PR #236, opened 2021-02-05
Issue 215 add a database access layer orm

FrancoisJ:


antarest/login/repository.py
```diff
@@ -1,36 +1,62 @@
...
+    finally:
+        sess.close()
 
 
 class UserRepository:
-    def __init__(self, config: Config) -> None:
-        self.users: Dict[int, User] = {
-            0: User(
-                id=0,
+    def __init__(self, config: Config, engine: Engine) -> None:
```
pl-buiquang, line 28: should be the session instead of the engine

pl-buiquang (approved):

antarest/login/repository.py
FrancoisJ: It should be a request-scoped session indeed. Today, we have choice between server-scoped session or atomic-scoped session. To avoid endlessly session, I code a atomic-scoped session.

PR #238, opened 2021-02-08
Issue 216 add storage permission schema

FrancoisJ:


antarest/login/model.py
```diff
@@ -67,3 +79,33 @@ def __eq__(self, o: Any) -> bool:
             and (o.name == self.name)
             and (o.role == self.role)
         )
+
+
+class Group(DTO, Base):  # type: ignore
+    __tablename__ = "groups"
+
+    id = Column(Integer, Sequence("user_id_seq"), primary_key=True)
```
pl-buiquang: create a new sequence table "group_id_seq"

pl-buiquang (changes requested):

antarest/login/model.py
FrancoisJ: done thanks

antarest/login/web.py
```diff
@@ -49,49 +49,81 @@ def auth() -> Any:
     def login() -> Any:
         return render_template("login.html")
 
-    @bp.route("/users", methods=["GET", "POST"], defaults={"id": None})
+    @bp.route("/users", methods=["GET"])
+    @jwt_required  # type: ignore
+    def users_get_all() -> Any:
+        if get_jwt_identity()["role"] != "ADMIN":
```
pl-buiquang, line 55: à remplacer par une méthode (isAdmin()) qui wrap ce code (comme ça on peut dire que "ADMIN" sera pas vraiment une magic string)

antarest/login/web.py
```diff
@@ -49,49 +49,81 @@ def auth() -> Any:
     def login() -> Any:
         return render_template("login.html")
 
-    @bp.route("/users", methods=["GET", "POST"], defaults={"id": None})
+    @bp.route("/users", methods=["GET"])
+    @jwt_required  # type: ignore
```
pl-buiquang, line 53: Wrap the decorator inside another decorator to disable it when login is disabled in config. eg. https://stackoverflow.com/questions/5952641/decorating-decorators-try-to-get-my-head-around-understanding-it

pl-buiquang (changes requested):
Protect storage module routes

antarest/login/web.py
FrancoisJ: yes, i will do that in other issue, when we refactor storage

PR #239, opened 2021-02-09
rename storage_api to storage

Hyralc:


Dockerfile
```diff
@@ -13,4 +13,4 @@ COPY ./static /static
 RUN pip3 install --upgrade pip \
     && pip3 install -r /conf/requirements.txt
 
-ENTRYPOINT gunicorn --config /conf/gunicorn.py antarest.storage_api.wsgi:app
\ No newline at end of file
+ENTRYPOINT gunicorn --config /conf/gunicorn.py antarest.storage.wsgi:app
```
pl-buiquang: just antarest:main

pl-buiquang (changes requested):

Dockerfile
pl-buiquang: also fix scripts

antarest/common/persistence.py
```diff
@@ -1,6 +1,9 @@
-from typing import Any
+from contextlib import contextmanager
```
pl-buiquang, line 1: rename db.py and add the create_engine() method used in the main

antarest/storage/repository/antares_io/reader/__init__.py
```diff
@@ -1,4 +1,4 @@
-from antarest.storage.antares_io.reader.ini_reader import (
+from antarest.storage.repository.antares_io.reader.ini_reader import (
```
pl-buiquang, line 1: why filesystem is not in "business" or "model" package instead ?

antarest/storage/service.py
```diff
@@ -0,0 +1,65 @@
+from io import BytesIO
+from pathlib import Path
+from typing import List, IO
+
+from antarest.common.custom_types import JSON
+from antarest.storage.web import RequestHandler
+from antarest.storage.web.request_handler import RequestHandlerParameters
+
+
+class StorageService:
```
pl-buiquang, line 57: should be the other way around. Business code should be in the service and request handler just calling the service methods

pl-buiquang (changes requested):

PR #249, opened 2021-02-16
Refresh

FrancoisJ:


tests/login/test_web.py
```diff
@@ -26,11 +28,11 @@ def create_client(service: Mock) -> Any:
...
+        token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJmcmVzaCI6ZmFsc2UsImlhdCI6MTYxMzQ2NzA3NSwianRpIjoiMTZlZDc3NmMtNmNlMS00MDdhLWFkMmItMzFiN2M0NjkyOTA5IiwibmJmIjoxNjEzNDY3MDc1LCJ0eXBlIjoiYWNjZXNzIiwic3ViIjp7ImlkIjoxLCJuYW1lIjoidXNlciIsInJvbGUiOiJVU0VSIn0sImNzcmYiOiI0ODIzYTlmYi1mNzJiLTQxZTAtODk1Mi1lYWU3MjlkNGUxZjQiLCJleHAiOjk5OTk5OTk5OTl9.QIcMKIEm_rE_0KrNgWMKE9ykAU9eHcbGckGFE8hhTk0"
     elif role == Role.ADMIN:
-        token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE2MTIxNzU4MDYsIm5iZiI6MTYxMjE3NTgwNiwianRpIjoiMzkzM2M0MTItZGI2NS00YjUwLTk5ZGMtYzJlYjgxMTBkNTlhIiwiZXhwIjo5OTk5OTk5OTk5LCJpZGVudGl0eSI6eyJpZCI6MCwibmFtZSI6ImFkbWluIiwicm9sZSI6IkFETUlOIn0sImZyZXNoIjpmYWxzZSwidHlwZSI6ImFjY2VzcyJ9.Gbxhmz6XSke0OZ2-fwRFKoh8LOrFj0pbJeR3Vfj536Q"
+        token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJmcmVzaCI6ZmFsc2UsImlhdCI6MTYxMzQ2NzA3NSwianRpIjoiMTZlZDc3NmMtNmNlMS00MDdhLWFkMmItMzFiN2M0NjkyOTA5IiwibmJmIjoxNjEzNDY3MDc1LCJ0eXBlIjoiYWNjZXNzIiwic3ViIjp7ImlkIjowLCJuYW1lIjoiYWRtaW4iLCJyb2xlIjoiQURNSU4ifSwiY3NyZiI6IjQ4MjNhOWZiLWY3MmItNDFlMC04OTUyLWVhZTcyOWQ0ZTFmNCIsImV4cCI6OTk5OTk5OTk5OX0.hvvGLnqbRzbuGyCmjuecX5-puMYDDOwsFXZHiUMJyFk"
```
pl-buiquang, line 35: You should generate the access_token. Also (for refresh test), it should be generated to be expired

antarest/login/web.py
```diff
@@ -41,13 +40,31 @@ def login() -> Any:
...
+    @bp.route("/refresh", methods=["POST"])
+    @auth.protected()
+    def refresh() -> Any:
+        identity = get_jwt_identity()
+        user = service.get_user(identity["id"])
+        if user:
+            access_token = create_access_token(identity=identity)
+            return jsonify(access_token=access_token)
+        else:
+            return "Token invalid", 403
```
pl-buiquang, line 66: Not sure if this is right (https://flask-jwt-extended.readthedocs.io/en/stable/refreshing_tokens/#explicit-refreshing-with-refresh-tokens)

tests/login/test_web.py
```diff
@@ -63,6 +65,22 @@ def test_auth_fail() -> None:
     assert res.status_code == 401
 
 
+@pytest.mark.unit_test
+def test_refresh() -> None:
```
pl-buiquang, line 69: this test should use the refresh token with a invalid access token

pl-buiquang (changes requested):

tests/login/test_web.py
FrancoisJ: Flask-JWT generates token only inside a request. Therefore we have choice between:
- hardcoded token like these with very long expiration delay (2 centuries...)
- request endpoint /login to generate token before each other endpoint tests. This solution avoid hardcoded token, but this is not simple unit test, since all endpoint need /login endpoint to success

Test expiration session with refresh token should be done, but inside integration or E2E test.

antarest/login/web.py
FrancoisJ: I really don't understand flask-jwt example. If server just create same access_token with new expiration date. Why we just don't expend token expiration date ? For me we use a short-time access_token with a long-time refresh token, to force user to sync frequently information in his access_token.

Each time user ask new access_token, we have to update its new groups,  permissions or blacklisted.

However, in the example, they use `@jwt_required(refresh=True)`, it's a better way than `@auth.protected()`. I  tried to implements it, but I can't find how to correctly send refresh_token during test.  We can come back to this issue later.

tests/login/test_web.py
FrancoisJ: Yes, I didn't find a way to send refresh token during test. So I pushed like that, and come back later.

antarest/login/web.py
pl-buiquang: In the example you have to replace the token in headers ('Bearer <token>') with the refresh_token instead of the access_token (it is documented at the last line of the linked section).

tests/login/test_web.py
pl-buiquang: Why can't you use create_access_token with a negative timedelta (for the expired token) ?

FrancoisJ: no it returns 401 unauthorize...., so i set expiration date at 99999999

antarest/login/web.py
FrancoisJ: yes, but it doesn't work... only for test. It test with curl, refresh endpoint work great, but with test it failed

PR #266, opened 2021-02-22
Add launcher service

pl-buiquang:


resources/application.yaml
```diff
@@ -12,4 +12,9 @@ db:
 storage:
   studies: examples/studies
 
+launcher:
+  type: local
+  binaries:
+    700: path/to/700
+
```
pl-buiquang: split config into launcher containing type and a new one specific for the local launcher config.
we may want to use different launcher
or something like : 
```
launcher:
   default: local
   local:
      binaries:
         700: ...
         640: ...
    slurm:
         url: ...
    worker:
        ...     
```

antarest/launcher/service.py
```diff
@@ -0,0 +1,51 @@
...
+        config: Config,
+        storage_service: StorageService,
+        repository: JobResultRepository,
+        factory_launcher: FactoryLauncher = FactoryLauncher(),
+    ) -> None:
+        self.config = config
+        self.storage_service = storage_service
+        self.repository = repository
+        self.launcher = factory_launcher.build_launcher(config)
+
```
pl-buiquang, line 30: add a threaded loop to watch for running job status by polling run_study.
sleep interval should be configurable (default 5s may be ok)
it will poll the launcher for study status and remove the watched study when completion occur or a configurable timeout (default 7 days). the status will be updated if different from the in memory cache list of running jobs.

PR #275, opened 2021-02-24
Add launcher UI

pl-buiquang:


tests/launcher/test_repository.py
```diff
@@ -36,32 +38,11 @@ def test_job_result() -> None:
...
+    d = repo.find_by_study(study_id)
+    assert len(d) == 1
+    assert a == d[0]
+
     repo.delete(a.id)
     assert repo.get(a.id) is None
 
-
-@pytest.mark.unit_test
-def test_save():
```
FrancoisJ, line 44: This test sound curious, but by default sqlalchemy raise error when we try to update an object existing in database (like b here). We fix it be using `Session.add` for new object and `Session.mege` for updating.

Can you keep this test, like a no regression test. Maybe rename it by `test_update` for more clarity

FrancoisJ (changes requested):

FrancoisJ (approved):

PR #277, opened 2021-02-25
Issue 265 add storage study permission check

FrancoisJ:


antarest/storage/business/storage_service_parameters.py
```diff
@@ -0,0 +1,11 @@
+from typing import Any, Optional
+
+from dataclasses import dataclass
+
+from antarest.login.model import User
+
+
+@dataclass
```
pl-buiquang, line 8: there is no dataclass in target python3.6

antarest/storage/service.py
```diff
@@ -1,328 +1,112 @@
...
-        url = "/".join(route_parts[1:])
-        study_path = self.path_to_studies / uuid
-
-        return uuid, url, study_path
+        return self.study_service.edit_study(route, new)
+
+    def _check_user_permission(self, user: User, uuid: str) -> None:
+        md = self.repository.get(uuid)
+        if not md:
+            raise StudyNotFoundError(uuid)
```
pl-buiquang: log a warning indicating that study metadata was not found and add a TODO for metadata autocreation from study files.
this will not raise any exception

antarest/storage/service.py
```diff
@@ -1,328 +1,112 @@
...
+        if user not in md.users:
```
pl-buiquang: and if user has not admin role

resources/application.yaml
```diff
@@ -10,7 +10,7 @@ db:
   url: "sqlite:///database.db"
 
 storage:
-  studies: examples/studies
+  studies: /tmp/Antares_Simulator_Examples
```
pl-buiquang: this file will be used to generate default config when launching application in local, le examples/studies be the default path for studies

antarest/storage/service.py
```diff
@@ -1,328 +1,112 @@
...
-                        k, str(self.__dict__[k]), type(self.__dict__[k])
-                    )
-                    for k in sorted(self.__dict__)
-                ]
-            ),
-        )
 
-    def __repr__(self) -> str:
-        return self.__str__()
+class UserHasNotPermissionError(Exception):
```
pl-buiquang: Make this extend werkzeug.exceptions.Forbidden or just use that when needed

antarest/storage/service.py
```diff
@@ -1,328 +1,184 @@
...
+            if user.role == Role.ADMIN:
+                return
 
-        output_name = f"{date}{mode}{name}"
-        path_output.rename(Path(path_output.parent, output_name))
+            md = self.repository.get(uuid)
+            if not md:
+                logging.warning(f"Study {uuid} not found in metadata db")
+                # TODO don't raise error. Study can be added manually by user
+                raise StudyNotFoundError(uuid)
```
pl-buiquang: so... it should return (to skip the user check) and not throw an exception.
Also the logging should be defined in the beginning of the file with logger = logging.get_logger(__name__)

pl-buiquang (changes requested):

antarest/storage/service.py
FrancoisJ: ok for logging. I prefere to raise an error now. When we have a clear view of files watcher and authorization logic, I will suppress it.

PR #279, opened 2021-02-26
Add logging conf

### LINKED ISSUE #278: Add logger config
Logging path and default level should be configured in the config application.yaml.
Default should be INFO and console output (when path is not configured).

pl-buiquang:
Also fixing the auth_need endpoint for web ui.

### DIFF
--- a/antarest/main.py
+++ b/antarest/main.py
@@ -1,4 +1,5 @@
 import argparse
+import logging
 import os
 import sys
 import webbrowser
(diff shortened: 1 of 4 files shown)


antarest/main.py
```diff
@@ -75,10 +76,28 @@ def get_local_path() -> Path:
         return Path(os.path.abspath(""))
 
 
+def configure_logger(config: ConfigYaml) -> None:
```
FrancoisJ: `ConfigYaml` is an implementation of `Config` class. Therefore we should use `Config` as type.

FrancoisJ (approved):

PR #284, opened 2021-03-01
Add user injection to launcher service

pl-buiquang:
Signed-off-by: Paul Bui-Quang <paul.buiquang@rte-france.com>

### DIFF
--- a/antarest/launcher/service.py
+++ b/antarest/launcher/service.py
@@ -2,6 +2,7 @@ from datetime import datetime
 from typing import List, Optional
 from uuid import UUID
 
+from antarest.common.auth import Auth
 from antarest.common.config import Config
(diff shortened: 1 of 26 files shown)


antarest/launcher/service.py
```diff
@@ -35,8 +37,7 @@ def update(self, job_result: JobResult) -> None:
         self.repository.save(job_result)
 
     def run_study(self, study_uuid: str) -> UUID:
-        # TODO use true parameters from user request
-        params = StorageServiceParameters()
+        params = StorageServiceParameters(user=Auth.get_current_user())
```
FrancoisJ: parameters were created here temporarily. Like for `storage`, parameters will be construct in `web` part. This will also avoid to use a patch in test.

PR #286, opened 2021-03-05
Extend tree

FrancoisJ:


antarest/storage/repository/filesystem/config/files.py
```diff
@@ -67,18 +78,23 @@ def parse_simulation(path: Path) -> "Simulation":
...
+            synthesis,
+        ) = ConfigPathBuilder._parse_outputs_parameters(path)
         return Simulation(
             date=regex.group(1),
             mode=modes[regex.group(2)],
             name=regex.group(3),
             nbyears=nbyears,
             by_year=by_year,
             synthesis=synthesis,
+            error=(path / "checkIntegrity.txt").exists(),
```
pl-buiquang: this cond should be inverted, error should be false when checkIntegrity exists

antarest/storage/repository/filesystem/config/json.py
```diff
@@ -69,10 +82,11 @@ def _parse_simulation(json: JSON) -> "Simulation":
             nbyears=nbyears,
             by_year=by_year,
             synthesis=synthesis,
+            error="checkIntegrity" in json,
```
pl-buiquang: invert cond (see previous comment)

antarest/storage/repository/filesystem/root/output/simulation/simulation.py
```diff
@@ -40,33 +40,39 @@ def build(self, config: StudyConfig) -> TREE:
...
                 config.next_file("simulation.log")
             ),
             "info": OutputSimulationInfoAntaresOutput(
                 config.next_file("info.antares-output")
             ),
         }
-        if self.simulation.mode == "economy":
-            children["economy"] = OutputSimulationMode(
-                config.next_file("economy"), self.simulation
+        if self.simulation.error:
```
pl-buiquang: invert condition (see prev comment)

pl-buiquang (changes requested):
This will not work when the mode is adequacyDraft (it doesn't output "checkIntegrity.txt')

PR #287, opened 2021-03-05
WIP Refactor validate

FrancoisJ:


antarest/storage/repository/filesystem/bucket_node.py
```diff
@@ -45,5 +45,7 @@ def build(self, config: StudyConfig) -> TREE:
         os.chdir(current_dir)
         return children
 
-    def validate(self, data: JSON) -> None:
-        pass  # no validation for bucket node
+    def validate(
```
pl-buiquang: even though, one can simply check if the return message array length is 0 to check the validity. It would be better to return a tuple (bool, List[str]) with validation boolean result and possibly just warning non blocking messages

pl-buiquang: also we could add an other optional input "raise_on_error" (default False) that would raise at first exception (if possible), in order not to bother parsing the whole broken simulation if needed.
that could be used to quickly invalidate studies before launch.

antarest/storage/repository/filesystem/folder_node.py
```diff
@@ -55,20 +54,29 @@ def save(self, data: JSON, url: Optional[List[str]] = None) -> None:
             (name,), sub_url = self.extract_child(children, url)
             return children[name].save(data, sub_url)
         else:
-            self.validate(data)
```
pl-buiquang, line 58: this method should be tranfered in StorageService::import_study and the result stored in the metadata study object

pl-buiquang (changes requested):

antarest/storage/repository/filesystem/bucket_node.py
FrancoisJ: with python a condition return false if a list is empty. Therefore we can rename method to check_errors to have some line like ) `if study.check_error()`. It will avoid to have a boolean

pl-buiquang (approved):

PR #290, opened 2021-03-12
Issue 274 workspaces

FrancoisJ:


antarest/storage/business/study_service.py
```diff
@@ -20,54 +22,51 @@
...
+    def assert_study_not_exist(self, metadata: Metadata) -> None:
+        if self.is_study_existing(metadata):
             raise StudyAlreadyExistError(
-                f"A study already exist with the uuid {uuid}."
+                f"A study already exist with the uuid {metadata.id}."
             )
 
-    def is_study_existing(self, uuid: str) -> bool:
-        return uuid in self.get_study_uuids()
+    def is_study_existing(self, metadata: Metadata) -> bool:
```
pl-buiquang: rename "study_exists" ("is study existing" a bit weird)

antarest/storage/business/study_service.py
```diff
@@ -20,54 +22,51 @@
...
+        return metadata.id in self.get_study_uuids(metadata.workspace)
```
pl-buiquang: you should just check if path exist and contain the study.antares, not checking if it is present among all existing studies.

antarest/storage/business/study_service.py
```diff
@@ -20,54 +22,51 @@
...
-            if (path / "study.antares").is_file()
+            path.name for path in folders if (path / "study.antares").is_file()
         ]
         # sorting needed for test
         return sorted(studies_list)
 
-    def get(self, route: str, depth: int) -> JSON:
-        uuid, url, study_path = self.extract_info_from_url(route)
-        self.check_study_exist(uuid)
+    def get(self, metadata: Metadata, url: str, depth: int) -> JSON:
```
pl-buiquang: make the url defaults to "/" or ""

antarest/storage/business/study_service.py
```diff
@@ -76,77 +75,70 @@ def get(self, route: str, depth: int) -> JSON:
...
-        path_study = self.get_study_path(uuid)
+        path_study = self.get_study_path(metadata)
         path_study.mkdir()
 
         with ZipFile(empty_study_zip) as zip_output:
             zip_output.extractall(path=path_study)
 
-        study_data = self.get(uuid, 10)
-        StorageServiceUtils.update_antares_info(study_name, study_data)
+        study_data = self.get(metadata, metadata.id, 10)
```
pl-buiquang: second arg should be the data route, so here the root "".
also the depth should be -1 since we want the whole study data to save

antarest/storage/service.py
```diff
@@ -92,97 +105,116 @@ def export_study(
...
+        if not user:
+            raise UserHasNotPermissionError()
+
+        md = self.repository.get(uuid)
+        if not md:
+            logger.warning(
+                f"Study %s not found in metadata db",
+                StorageServiceUtils.sanitize(uuid),
+            )
+            raise UserHasNotPermissionError()
```
pl-buiquang: now you can throw a StudyNotFound since every study should be in metadata table.

antarest/storage/service.py
```diff
@@ -92,97 +105,116 @@ def export_study(
...
-                raise UserHasNotPermissionError()
+        if owner:
+            metadata.owner = owner
+        if group:
+            metadata.groups = [group]
+        self.repository.save(metadata)
 
-            if user.role == Role.ADMIN:
-                return
+    def _get_metadata_according_permission(
```
pl-buiquang: why not have a method to retrieve metadata (raise StudyNotFound if not found) and an other one (the previously named "check_user_permission" with user and metadata in input ?

antarest/storage/service.py
```diff
@@ -38,52 +41,62 @@ def __init__(
...
-        self._check_user_permission(params.user, src_uuid)
-        uuid = self.study_service.copy_study(src_uuid, dest_study_name)
-        self._save_metadata(uuid, params.user)
+        src_md = self._get_metadata_according_permission(params.user, src_uuid)
+
+        dest_md = deepcopy(src_md)
+        dest_md.id = str(uuid4())
+        dest_md.name = dest_study_name
+
+        md = self.study_service.copy_study(src_md, dest_md)
```
pl-buiquang, line 99: why the creation of the metadata for the new study not done in copy_study (with second arg being the new name) ?

pl-buiquang: also copy_study should have access to metadata repository so it will also save the metadata which is an integral part of the study creation/copy process

antarest/storage/service.py
```diff
@@ -92,97 +105,116 @@ def export_study(
...
-            users=[user],
-        )
-        self.repository.save(meta)
+        info = self.study_service.get_study_information(metadata)["antares"]
+
+        metadata.name = info["caption"]
+        metadata.version = info["version"]
+        metadata.author = info["author"]
+        metadata.created_at = datetime.fromtimestamp(info["created"])
+        metadata.updated_at = datetime.fromtimestamp(info["lastsave"])
```
pl-buiquang, line 186: it should be the current datetime. and the underlying file should be updated. (i will create an issue for it)

antarest/storage/service.py
```diff
@@ -92,97 +105,116 @@ def export_study(
...
-                logger.warning(f"Study {uuid} not found in metadata db")
-                return
+        if not is_owner and not inside_group:
+            raise UserHasNotPermissionError()
 
-            if user not in md.users:
-                raise UserHasNotPermissionError()
+        return md
 
+    def _get_metadata_according_permission_no_raising(
```
pl-buiquang: why not have a flag in the previous method ?

antarest/storage/web/utils_blueprint.py
```diff
@@ -71,8 +71,8 @@ def get_file(path: str) -> Any:
         """
 
         try:
-            file_path = storage_service.study_service.path_to_studies / path
-            return send_file(file_path.absolute())
+            params = RequestParameters(user=Auth.get_current_user())
+            return storage_service.get_matrix(path, params)
```
pl-buiquang, line 75: why dropping the use of send_file ?

pl-buiquang (changes requested):

antarest/storage/service.py
FrancoisJ: Metadata object is managed by StorageService, Begin to create metadata elsewhere in code (and generate its id outside StorageService) create a less clear separate architecture.

antarest/storage/service.py
FrancoisJ: Flag change return expectation so method signature. When flag is off, method return a Metadata or send error. When flag is on, return can be None. Therefore I separate use in different method with their own return type

antarest/storage/service.py
```diff
@@ -182,22 +192,32 @@ def _save_metadata(
...
+        self,
+        user: Optional[User],
+        md: Optional[Metadata],
+        raising: bool = True,
+    ) -> bool:
+        if not user:
             raise UserHasNotPermissionError()
 
+        if not md:
+            raise ValueError("Metadata is None")
```
pl-buiquang, line 217: This should be a StudyNotFoundError

antarest/storage/service.py
```diff
@@ -182,22 +192,32 @@ def _save_metadata(
...
+            sanitized = StorageServiceUtils.sanitize(uuid)
             logger.warning(
                 f"Study %s not found in metadata db",
-                StorageServiceUtils.sanitize(uuid),
+                sanitized,
             )
+            raise StudyNotFoundError(uuid)
+        return md
+
+    def _assert_permission(
```
pl-buiquang, line 207: // why not _check_permission like before ?

antarest/storage/service.py
FrancoisJ: `StudyNotFoundError` is raising in `_get_metadata` because we search study in database. In this case, we don't not where come metadata. So if it none, we just said is none, we don't know why it is none....

antarest/storage/service.py
FrancoisJ: When code could raise an error during check I prefere to use `assert` key word. However, if an object method begin by `assert_*` there are troubles when this object will be mocked. But here we can because we prefix with `_`

antarest/storage/service.py
pl-buiquang: i had missed that, ok

pl-buiquang: i had missed that, ok

antarest/storage/service.py
pl-buiquang: "when code could raise an error", well it's pretty much everything.

PR #294, opened 2021-03-16
Add event bus + UI websocket support

pl-buiquang:
Also add some study management related events : create/edit/delete/run/complete


antarest/common/interfaces/eventbus.py
```diff
@@ -0,0 +1,67 @@
...
+    STUDY_JOB_COMPLETED = "STUDY_JOB_COMPLETED"
+
+
+@dataclass
+class Event:
+    type: str
+    payload: Any
+
+
+class IEventBus(ABC):
```
FrancoisJ, line 20: Do we really need event bus ? Currently it takes 1 000 line of code + 2 technos (websocket, redis) to update job status. Does a simple polling is not enough ?

antarest/eventbus/business/redis_eventbus.py
```diff
@@ -0,0 +1,57 @@
...
+class RedisConfig:
+    host: str
+    port: int = 6379
+
+
+def parse_config(redis_config: Any) -> RedisConfig:
+    return RedisConfig(**redis_config)
+
+
+class RedisEventBus(IEventBusBackend):
```
FrancoisJ, line 27: What is the purpose of redis ? Anyone seems to subscribe to redis, redis pubsub is just used to save events before launch in batch.

antarest/eventbus/service.py
```diff
@@ -0,0 +1,65 @@
...
+    ) -> None:
+        self.backend = backend
+        self.listeners: Dict[str, Callable[[Event], None]] = {}
+        self.lock = threading.Lock()
+        if autostart:
+            self.start()
+
+    def push(self, event: Event) -> None:
+        # TODO add arg permissions with group/role, user, public
+        self.backend.push_event(event)
```
FrancoisJ, line 25: Why events are not send directly to websocket ? What I understand, events are stored inside stack. And each 0.2s stack is clear and event are send to websocket.

PR #296, opened 2021-03-19
Issue 264 filewatcher

FrancoisJ:


antarest/storage/business/watcher.py
```diff
@@ -1,11 +1,66 @@
...
+
+
 class Watcher:
+    LOCK = Path("watcher")
+    DELAY = 1
+
     def __init__(self, config: Config, service: StorageService):
         self.service = service
         self.config = config
+        self._get_lock()
```
pl-buiquang: // why this call ? (it seem to be useful only when creating the thread)

antarest/storage/business/watcher.py
```diff
@@ -1,11 +1,66 @@
...
+    def _scan(self) -> None:
+        studies: List[StudyFolder] = list()
+        for name, workspace in self.config["storage.workspaces"].items():
+            path = Path(workspace["path"])
+            groups = [Group(id=g) for g in workspace.get("groups", [])]
+
+            for folder in path.iterdir():
+                if (folder / "study.antares").exists():
+                    logger.info(f"Study {folder.name} found in {name}")
+                    studies.append(StudyFolder(folder, name, groups))
```
pl-buiquang: this should be recursive

antarest/storage/business/watcher.py
```diff
@@ -1,11 +1,66 @@
...
+                return False
+
+    def _loop(self) -> None:
+        while True:
+            self._scan()
+            sleep(2)
+
+    def _scan(self) -> None:
+        studies: List[StudyFolder] = list()
+        for name, workspace in self.config["storage.workspaces"].items():
```
pl-buiquang, line 71: should skip default workspace

antarest/storage/business/watcher.py
```diff
@@ -1,11 +1,68 @@
...
+                int(f"0{Watcher.LOCK.read_text()}")
+                if Watcher.LOCK.exists()
+                else 0
+            )
+            now = int(time())
+            if now - start > Watcher.DELAY:
+                Watcher.LOCK.write_text(str(now))
+                logger.info("Watcher get lock")
+                return True
+            else:
```
pl-buiquang, line 48: i don't understand. if the second thread enter this lock within the watcher delay (1 second is large), it guess it will also start ?

antarest/storage/business/watcher.py
FrancoisJ: indeed

antarest/storage/business/watcher.py
FrancoisJ: indeed

antarest/storage/business/watcher.py
FrancoisJ: It starts only if date stored in file, is older than DELAY

pl-buiquang: oh yes, my bad

antarest/storage/service.py
```diff
@@ -91,6 +91,30 @@ def create_study(self, study_name: str, params: RequestParameters) -> str:
         )
         return str(md.id)
 
+    def sync_studies_on_disk(self, folders: List[StudyFolder]) -> None:
+
+        # delete orphan studies on database
+        paths = [str(f.path) for f in folders]
+        for md in self.repository.get_all():
+            if md.path not in paths:
+                self.repository.delete(md.id)
```
pl-buiquang, line 106: add some logs + push event (EventType.STUDY_DELETED)

antarest/storage/service.py
```diff
@@ -91,6 +91,30 @@ def create_study(self, study_name: str, params: RequestParameters) -> str:
...
+                    id=folder.path.name,
+                    name=folder.path.name,
+                    path=str(folder.path),
+                    workspace=folder.workspace,
+                    owner=User(id=0),
+                    groups=folder.groups,
+                )
+
+                md.content_status = self._analyse_study(md)
+                self.repository.save(md)
```
pl-buiquang, line 127: add some logs + push event (EventType.STUDY_CREATED)

PR #300, opened 2021-03-29
Registered config

pl-buiquang:


antarest/common/config.py
```diff
@@ -1,33 +1,50 @@
...
-from typing import Any, Optional
+from typing import Any, Optional, Dict, Type, TypeVar, Callable, cast
 
 import yaml
+from dataclasses_json import DataClassJsonMixin  # type: ignore
+from dataclasses_json.api import dataclass_json
 
 from antarest.common.custom_types import JSON
 
+config_definitions: Dict[str, Type[DataClassJsonMixin]] = {}
```
FrancoisJ, line 13: What is its purpose ?

antarest/common/config.py
```diff
@@ -1,33 +1,50 @@
...
-        self.data = data or dict()
+        self.raw_data = data or dict()
+        self.data: Dict[str, Any] = {}
+        self._init()
 
     def __getitem__(self, item: str) -> Any:
         return self._get(item)
 
+    def _init(self) -> None:
+        global config_definitions
```
FrancoisJ, line 26: can we do without global ?

antarest/common/logger.py
```diff
@@ -0,0 +1,27 @@
...
+from typing import Optional
+
+from dataclasses_json.api import dataclass_json
+
+from antarest.common.config import Config, register_config
+
+
+@dataclass_json
+@dataclass
+class LoggingConfig:
```
FrancoisJ, line 12: Why we separate config ? We don't want to centralize config inside one object ?

antarest/login/repository.py
```diff
@@ -44,7 +45,7 @@ def __init__(self, config: Config, session: Session) -> None:
                 User(
                     name="admin",
                     role=Role.ADMIN,
-                    password=Password(config["security.login.admin.pwd"]),
+                    password=Password(config.login.admin.pwd),
```
FrancoisJ, line 48: To resume, I match this target to avoid magic string and instead walk through object attributes. But I don't understand, why config is became so complexe just to achieve this target ?

FrancoisJ (changes requested):

antarest/common/config.py
pl-buiquang: this is the config module typed config definitions (they are module scoped)

antarest/common/config.py
pl-buiquang: as it is a idiom in python to have singleton as module, this is why this is a global (module scoped) variable

antarest/common/logger.py
pl-buiquang: at the very first I thought so, but as I like to think we could easily split the code in several micro services, I splitted the config in a plugin-like way

antarest/login/repository.py
pl-buiquang: I could have one single dataclass object definition in common but as I went the plugin way, it's that bit complex. Anyway, the end goal is not just removing magic strings (though its kind of linked), it is static type checking when accessing config and runtime config validation at startup.

PR #301, opened 2021-03-30
User permission

FrancoisJ:


antarest/common/jwt.py
```diff
@@ -0,0 +1,63 @@
...
+        return {
+            "id": self.id,
+            "name": self.name,
+            "groups": [g.to_dict() for g in self.groups],
+        }
+
+    def is_site_admin(self) -> bool:
+        return "admin" in [g.id for g in self.groups]
+
+    def is_group_admin(self, groups: Union[Group, List[Group]]) -> bool:
```
pl-buiquang, line 53: what is the use of this method when the arg is a list of group ? we won't know which group the user is admin of.

antarest/login/repository.py
```diff
@@ -43,16 +45,11 @@ def __init__(self, config: Config, session: Session) -> None:
...
                 )
             )
-            self.save(
-                User(name="alice", role=Role.USER, password=Password("alice"))
-            )
-            self.save(
-                User(name="bob", role=Role.USER, password=Password("bob"))
-            )
+            self.save(User(name="alice", password=Password("alice")))
+            self.save(User(name="bob", password=Password("bob")))
```
pl-buiquang: Remove this

antarest/storage/service.py
```diff
@@ -347,7 +345,7 @@ def _get_study(self, uuid: str) -> Study:
 
     def _assert_permission(
```
pl-buiquang, line 346: This should now take an additional parameter (action: read,run,edit,delete,manage_permissions) to enforce the permission matrix.

antarest/common/jwt.py
FrancoisJ: currently is used only for get_user. A group admin should access to user only if group admin belongs to user group.  So we fetch user groups and we check that if user logged as an admin belongs to one of these groups.

antarest/login/repository.py
FrancoisJ: indeed

antarest/storage/service.py
FrancoisJ: yes storage service permission is not yet implemented. I will do in other PR.

antarest/common/jwt.py
pl-buiquang: i see. maybe we should use any(is_group_admin(g) for g in groups) in the get_user method.
And is_group_admin should only take a single group.
Otherwise the result is ambiguous since we won't know if the user is admin of all groups or a single one which we won't know which.

PR #302, opened 2021-04-02
Add study permissions

pl-buiquang:


antarest/storage/service.py
```diff
@@ -306,6 +319,52 @@ def edit_study(
...
+        self.repository.save(study)
+
+    def add_group(
+        self, study_id: str, group_id: str, params: RequestParameters
+    ) -> None:
+        study = self._get_study(study_id)
+        self._assert_permission(
+            params.user, study, StudyPermissionType.MANAGE_PERMISSIONS
+        )
+        group = self.user_service.get_group(group_id, params)
```
FrancoisJ, line 340: This line create intermodule dependency. We could avoid by:
- if admin, we authorize any group without verify group exists...
- if group admin or owner, user can only add group where he is admin according to his jwt. 

Since we remain monolithic this line can remain too

antarest/storage/business/permissions.py
```diff
@@ -0,0 +1,72 @@
...
+def check_permission(
+    user: JWTUser, study: Study, permission: StudyPermissionType
+) -> bool:
+    if user.is_site_admin():
+        return True
+
+    if study.owner is not None and user.id == study.owner.id:
+        return True
+
+    global permission_matrix
```
FrancoisJ: Why this variable is global ? I delete this line and tests still passed.

PR #305, opened 2021-04-07
Storage fixes

pl-buiquang:


tests/storage/business/test_importer_service.py
```diff
@@ -40,16 +40,16 @@ def test_upload_matrix(tmp_path: Path, storage_service_builder) -> None:
         study_factory=Mock(),
     )
 
-    study_url = study_uuid + "/"
     matrix_path = "WRONG_MATRIX_PATH"
-    md = RawStudy(id=study_uuid, workspace=DEFAULT_WORKSPACE_NAME)
+    md = RawStudy(
+        id=study_uuid, workspace=DEFAULT_WORKSPACE_NAME, path=study_path
```
FrancoisJ, line 45: Do we really need path property ? In `StudyService` study path is build according to its id and workspace. `path` seems useless ?

pl-buiquang: No, not anymore. Since the watcher can pick up studies nested in directory. So the upload_matrix (that can be applied onto studies in watched workspaces) must know the path.

FrancoisJ: 👍

FrancoisJ (approved):

PR #312, opened 2021-05-06
Matrix parsing

FrancoisJ:


antarest/storage/repository/filesystem/matrix/output_series_matrix.py
```diff
@@ -0,0 +1,94 @@
...
+        df = pd.read_csv(
+            self.config.path, sep="\t", skiprows=4, na_values="N/A"
+        )
+
+        date, body = self.date_serializer.extract_date(df)
+
+        header = body.iloc[:2]
+        header.fillna("", inplace=True)
+        header = (
+            header.columns + "::" + header.iloc[0] + "::" + header.iloc[1]
```
pl-buiquang: declare magic string

antarest/storage/repository/filesystem/matrix/input_series_matrix.py
```diff
@@ -0,0 +1,63 @@
...
+            errors.append(
+                f"Input Series Matrix f{self.config.path} not exists"
+            )
+        if self.nb_columns and len(data) != self.nb_columns:
+            errors.append(
+                f"Data was wrong size. expected {self.nb_columns} get {len(data)}"
+            )
+        return errors
+
+    def _assert_url(self, url: Optional[List[str]] = None) -> None:
```
pl-buiquang: factor this code (used in RawNode and other MatrixNodes)

antarest/storage/repository/filesystem/matrix/date_serializer.py
```diff
@@ -0,0 +1,205 @@
...
+
+        return pd.concat([headers, matrix], axis=0)
+
+    def extract_date(self, df: pd.DataFrame) -> Tuple[pd.Index, pd.DataFrame]:
+        # Extract left part with date
+        date = df.loc[2:, ["Unnamed: 2", "Unnamed: 3", "Unnamed: 4"]]
+        date.columns = ["day", "month", "hour"]
+        date["month"] = date["month"].map(IDateMatrixSerializer._MONTHS)
+        date = (
+            date["hour"] + " " + date["day"] + "/" + date["month"].astype(str)
```
pl-buiquang: when using this as index, the sorting is not correct (as seen in the test resource file de_fr_values_hourly).
at least, switch the order (month > day > hour).
Yet there is another potential issue because sometime the series does not begin with the january (can start on july and end with june..)

pl-buiquang: the last issue, is real. There is (and it's the case most often than not) cases when the first date is something like "01	JUL	00:00" and the last "29	JUN	23:00".
So the index should be the very first column (integer from 1 to 8760 in a hourly scale) and add a new column containing the date

FrancoisJ: This was a tricky issue. Since we serialize Dataframe in dict with index as key, we don't have any control of order. I change the dict to format to `split` ([docs)](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_dict.html)), there are many improvment with this format: 
- index is a list, No issue any more with random order
- columns is a list, we can use tuple as column whereas before we can do only string column
- data size decrease.

pl-buiquang (approved):

PR #318, opened 2021-05-17
Add integration tests

pl-buiquang:


Hyralc (approved):

antarest/storage/service.py
```diff
@@ -388,8 +388,12 @@ def _save_study(
 
         if owner:
             study.owner = User(id=owner.impersonator)
-        if group:
-            study.groups = [group]
+            groups = [
```
FrancoisJ: Does it really what we want. I understand that if a user create a study, study belongs to all his groups automatically. It's like if a create a private repo on github, hadar and antares organization can see it.

tests/integration/test_integration.py
```diff
@@ -0,0 +1,199 @@
+import time
+
+from flask import Flask
+
+
+def test_main(app: Flask):
```
FrancoisJ, line 6: Perhaps split this test ?

pl-buiquang: perhaps later when this becomes really bulky. it will be cut into separate methods, but for now it's ok. It is to be a big integration test with dependent steps.

antarest/storage/service.py
pl-buiquang: indeed

FrancoisJ (approved):

PR #324, opened 2021-05-20
Fix optional detail file in output

pl-buiquang:
Signed-off-by: Paul Bui-Quang <paul.buiquang@rte-france.com>

### DIFF
--- a/antarest/storage/repository/filesystem/config/model.py
+++ b/antarest/storage/repository/filesystem/config/model.py
@@ -1,6 +1,6 @@
 from copy import deepcopy
 from pathlib import Path
-from typing import Optional, List, Dict
+from typing import Optional, List, Dict, Any
 
 from antarest.common.custom_types import JSON
 from antarest.common.persistence import DTO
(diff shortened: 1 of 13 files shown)


antarest/storage/repository/filesystem/config/model.py
```diff
@@ -29,7 +29,7 @@ class Area(DTO):
     def __init__(
         self,
         links: Dict[str, Link],
-        thermals: List[str],
+        thermals: Dict[str, Any],
```
FrancoisJ: I prefere to expand domain with new `Thermal` object. Like there are for `Link`. Today, there is only `enable` attribute in `Thermal` but it will be more easy to upgrade if we need to add other attributes.

FrancoisJ (changes requested):

antarest/storage/repository/filesystem/config/model.py
pl-buiquang: Indeed this would be cleaner

FrancoisJ (approved):

PR #340, opened 2021-05-28
add generic modal + groups settings

3lbanna:


FrancoisJ (approved):

pl-buiquang (changes requested):
We should prevent the creation of duplicate named groups

pl-buiquang (approved):

Issue #280, opened 2021-02-26
Add logs


pl-buiquang:
Add logs (debug or info) to every service method calls

PR #345, opened 2021-06-01
Logging #280

FrancoisJ:


pl-buiquang (commented):
Concerning the code smell (duplication). You could embed the string within the `StudyTypeUnsupported` class and parametrize its constructor with the id and type.

PR #351, opened 2021-06-04
fix settings and add list in group

3lbanna:


antarest/login/service.py
```diff
@@ -172,6 +175,19 @@ def save_bot(self, bot: BotCreateDTO, params: RequestParameters) -> Bot:
                     role_type = RoleType.from_dict(role_create.role)
                     if not (role and role.type.is_higher_or_equals(role_type)):
                         raise UserHasNotPermissionError()
+
+            if not bot.name.replace(" ", ""):
```
pl-buiquang: use `bot.name.strip()`

antarest/login/service.py
```diff
@@ -172,6 +175,19 @@ def save_bot(self, bot: BotCreateDTO, params: RequestParameters) -> Bot:
...
+                raise HTTPException(
+                    status_code=400, detail="Bot name is not valid"
```
pl-buiquang: you could precise :  "Bot name must not be empty"

antarest/login/web.py
```diff
@@ -172,11 +172,17 @@ def groups_get_all(
 
     @bp.get("/groups/{id}", tags=["User"])
     def groups_get_id(
-        id: str, current_user: JWTUser = Depends(auth.get_current_user)
+        id: str,
+        verbose: Optional[int] = None,
```
pl-buiquang: this is not "verbose", `list_user` or `details` could be better and with boolean type

pl-buiquang (approved):

PR #361, opened 2021-06-10
New UI for studies

3lbanna:


webapp/public/locales/en/singlestudy.json
```diff
@@ -0,0 +1,13 @@
+{
+    "informations": "Informations",
```
pl-buiquang: information is uncountable (there is no plural form)

webapp/public/locales/fr/singlestudy.json
```diff
@@ -0,0 +1,13 @@
+{
+    "informations": "Informations",
+    "treeView": "Vue détaillé",
```
pl-buiquang: détaillée

webapp/src/App/Pages/SingleStudy/InformationView.tsx
```diff
@@ -0,0 +1,232 @@
...
+const useStyles = makeStyles((theme: Theme) => createStyles({
+  root:{
+      flex: '0 0 30%',
+      minWidth: '320px',
+      minHeight: '250px',
+      height: '95%',
+      backgroundColor: 'white',
+      paddingBottom: theme.spacing(1),
+      margin: theme.spacing(1),
+      border: `1px solid ${theme.palette.primary.main}`,
```
pl-buiquang: no need for border. this is paper. so the surface-like will be sufficient to separate it.

webapp/src/App/Pages/SingleStudy/TaskView.tsx
```diff
@@ -0,0 +1,149 @@
...
+        catch(e)
+        {
+            enqueueSnackbar(t('singlestudy:failtoloadjobs'), { variant: 'error' });
+        }
+    }
+    init();
+  }, [t, enqueueSnackbar, studyId]);
+
+  return (
+    <Paper className={classes.root}>
```
pl-buiquang, line 98: Don't stack paper on paper. Here the border is ok but with a normal div.

pl-buiquang (approved):

PR #365, opened 2021-06-14
Issue 330 add file based matrix store api

FrancoisJ:


antarest/matrixstore/main.py
```diff
@@ -0,0 +1,51 @@
...
+        service = MatrixService(repo=repo, content=content)
+
+    @application.exception_handler(AuthJWTException)
+    def authjwt_exception_handler(
+        request: Request, exc: AuthJWTException
+    ) -> Any:
+        return JSONResponse(
+            status_code=HTTPStatus.UNAUTHORIZED,
+            content={"detail": exc.message},
+        )
```
pl-buiquang: to remove (exists already in login where it was needed to declare)

antarest/matrixstore/model.py
```diff
@@ -0,0 +1,63 @@
...
+            and self.type == other.type
+            and self.freq == other.freq
+            and self.created_at == other.created_at
+            and self.updated_at == other.updated_at
+        )
+        return res
+
+
+@dataclass
+class MatrixDTO(DataClassJsonMixin):  # type: ignore
```
pl-buiquang: For DTO that are used as a body in FastAPI enpoint, use pydantic BaseModel:  
```
class MatrixDTO(BaseModel)
```
(eg. UserCreateDTO)

antarest/matrixstore/model.py
```diff
@@ -0,0 +1,63 @@
...
+from dataclasses import dataclass
+from dataclasses_json import DataClassJsonMixin  # type: ignore
+from sqlalchemy import Column, String, Enum, DateTime  # type: ignore
+
+from antarest.common.persistence import Base
+
+
+class MatrixType(enum.IntEnum):
+    INPUT = 1
+    OUTPUT = 2
```
pl-buiquang: there will be no distinction between input and output (every matrix is an input matrix)

antarest/matrixstore/model.py
```diff
@@ -0,0 +1,63 @@
...
+    WEEKLY = 3
+    MONTHLY = 4
+    ANNUAL = 5
+
+
+class Matrix(Base):  # type: ignore
+    __tablename__ = "matrix"
+
+    id = Column(String(32), primary_key=True)
+    type = Column(Enum(MatrixType))
```
pl-buiquang: there will be no type

antarest/matrixstore/service.py
```diff
@@ -0,0 +1,88 @@
...
+            updated_at=datetime.fromtimestamp(dto.updated_at),
+        )
+
+        content = MatrixContent(
+            data=dto.data, index=dto.index, columns=dto.columns
+        )
+
+        return matrix, content
+
+    def save(self, data: MatrixDTO) -> str:
```
pl-buiquang: the name is ambiguous, it's only creation so it should be named create and check if it exist in database and throw an error if it is present.

antarest/matrixstore/model.py
```diff
@@ -0,0 +1,63 @@
...
+
+class MatrixFreq(enum.IntEnum):
+    HOURLY = 1
+    DAILY = 2
+    WEEKLY = 3
+    MONTHLY = 4
+    ANNUAL = 5
+
+
+class Matrix(Base):  # type: ignore
```
pl-buiquang, line 20: a name field would be nice so the user can find easily his matrix.
also we could add a metadata table so that user can add tags to matrix and query them to retrieve a list of matrix

pl-buiquang: for the metadata part, this will be in an other PR (issue : #370 )

antarest/matrixstore/main.py
FrancoisJ: ok

antarest/matrixstore/model.py
FrancoisJ: ok

antarest/matrixstore/model.py
FrancoisJ: ok

antarest/matrixstore/model.py
FrancoisJ: ok

antarest/matrixstore/service.py
FrancoisJ: Indeed matrix is immutable. So we can only create matrix. Ok for refactor. But since we share matrix, service doesn't care if matrix already exist or not

antarest/matrixstore/model.py
FrancoisJ: Matrix will be shared accros study, a name will be ambigus. Moreover matrix store is like a matrix bucket for storage module. It should be used by storage for its purpose but not by user as standalone

antarest/matrixstore/service.py
pl-buiquang: oh.. I thought the matrix would be mutable, which indeed would be better not. So you can remove the update_at field. 'save' is ok since there won't be an 'update'

antarest/matrixstore/model.py
pl-buiquang: no, actually it will be used by user. they will create matrix with name (like 'wind fc scenarioA', 'wind fc scenario B' and so on) so that they can find them easily later on and assign them to study inputs.

antarest/matrixstore/repository.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class MatrixContentRepository:
+    def __init__(self, config: Config) -> None:
+        self.bucket = config.matrixstore.bucket
+        self.bucket.mkdir(parents=True, exist_ok=True)
+
+    @staticmethod
+    def _compute_hash(data: str) -> str:
+        return hashlib.md5(data.encode()).hexdigest()
```
pl-buiquang: would be safer with sha256 or even 512 (a little afraid about potential collision, especially with md5)

antarest/matrixstore/model.py
pl-buiquang: actually (this wasn't specified in the specs) this lead me to think that matrix should have an owner. we could then add a unique (name x owner) unique constraint. (and group assignement but i'll create another issue for group later, since this will imply other management endpoints)

pl-buiquang (approved):

PR #394, opened 2021-06-23
UriResolver Issue 328

FrancoisJ:


antarest/storage/repository/filesystem/bucket_node.py
```diff
@@ -15,20 +15,28 @@ class BucketNode(FolderNode):
...
+                    value
+                )
 
     def build(self, config: StudyConfig) -> TREE:
         if not config.path.exists():
             return dict()
 
         children: TREE = {}
-        for item in config.path.iterdir():
+        for item in sorted(config.path.iterdir()):
```
pl-buiquang, line 31: why the sorted ?

antarest/storage/business/uri_resolver_service.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class UriResolverService:
+    def __init__(self, config: Config, matrix_service: MatrixService):
+        self.config = config
+        # StorageService must be injected after the class creation to avoid circular injection
+        self.storage_service = None  # type: ignore
+        self.matrix_service = matrix_service
+
+    def resolve(self, uri: str) -> Union[str, JSON]:
```
pl-buiquang: should take an additional optional parameter "parser" which is a function that can parse the file data (in case of matrix nodes which could hold a studyfile reference instead of a matrix).

antarest/storage/repository/filesystem/lazy_node.py
```diff
@@ -12,33 +13,42 @@ class LazyNode(INode, ABC, Generic[G, S, V]):  # type: ignore
...
+        if self.config.path.exists():
+            if expanded:
+                return self.context.resolver.build_studyfile_uri(
+                    self.config.path, self.config.study_id
+                )
+            else:
+                return self.load(url, depth, expanded)
+        else:
+            data = self.get_link_path().read_text()
+            return data if expanded else self.context.resolver.resolve(data)
```
pl-buiquang: you could pass the self.load method to the resolver, so that the raw file will be parsed if the link is a studyfile

antarest/storage/repository/filesystem/lazy_node.py
```diff
@@ -12,33 +13,42 @@ class LazyNode(INode, ABC, Generic[G, S, V]):  # type: ignore
...
-        if isinstance(data, str) and f"{self.url_prefix}://" in data:
-            src = Path(data[len(f"{self.url_prefix}://") :])
-            if src != self.config.path:
-                self.config.path.parent.mkdir(exist_ok=True, parents=True)
-                shutil.copyfile(src, self.config.path)
-            return None
+        self._assert_url_end(url)
 
+        if isinstance(data, str) and f"studyfile://" in data:
+            data = self.context.resolver.resolve(data)
```
pl-buiquang: so i guess there is an issue when it's a matrix reference since the data should be converted back to an antares file (if the study is on unsafe/unmanaged workspaces).

tests/conftest.py
```diff
@@ -42,16 +42,18 @@ def _assert_others(a: Any, b: Any) -> None:
...
-    elif (
-        isinstance(a, str)
-        and isinstance(b, str)
-        and ("file://" in a or "matrix://" in a)
-        and ("file://" in b or "matrix://" in b)
-    ):
-        _assert_pointer_path(a, b)
-    else:
-        _assert_others(a, b)
+    # TODO: REMOVE FUNCTION
```
pl-buiquang: why ?

antarest/storage/repository/filesystem/bucket_node.py
FrancoisJ: its for testing purpose. `iterdir` doesn't garanty order, whereas we need deterministic order for test.

antarest/storage/business/uri_resolver_service.py
FrancoisJ: done

antarest/storage/repository/filesystem/lazy_node.py
FrancoisJ: done

tests/conftest.py
FrancoisJ: mistake. We need yet assert_study because to compare pointer betwwen two studies. However new assert_study will be more robust that this one.

pl-buiquang (approved):

PR #479, opened 2021-09-01
Create variant tree view

3lbanna:


webapp/src/common/types.ts
```diff
@@ -21,9 +21,13 @@ export interface StudyMetadataOwner {
   name: string;
 }
 
+export type StudyType = 'study' | 'variantstudy' | 'rawstudy';
```
pl-buiquang: there should be no type "study"

pl-buiquang (approved):

PR #525, opened 2021-09-29
add command json importation and exportation

3lbanna:


antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -254,6 +254,21 @@ def remove_command(
                 command.index = idx
             self.repository.save(metadata=study, update_modification_date=True)
 
+    def remove_all_commands(
+        self, study_id: str, params: RequestParameters
+    ) -> None:
+        """
+        Remove command
```
pl-buiquang: remove all commands

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -254,6 +254,21 @@ def remove_command(
...
+        Args:
+            study_id: study id
+            command_id: command_id
```
pl-buiquang: arg doesn't exist

antarest/study/web/variant_blueprint.py
```diff
@@ -283,6 +283,23 @@ def remove_command(
             sanitized_uuid, sanitized_cid, params
         )
 
+    @bp.delete(
+        "/studies/{uuid}/all/commands",
+        tags=[APITag.study_variant_management],
+        summary="Remove a command",
```
pl-buiquang: Clear variant's commands

tests/integration/test_integration.py
```diff
@@ -638,6 +638,14 @@ def test_variant_manager(app: FastAPI):
...
     assert res.status_code == 404
 
+    res = client.delete(
+        f"/v1/studies/{variant_id}/all/commands",
+        headers={
+            "Authorization": f'Bearer {admin_credentials["access_token"]}'
+        },
+    )
+    assert res.status_code == 200
+
```
pl-buiquang, line 683: check by calling GET /studies/variant_id/commands that all commands has been deleted

antarest/study/web/variant_blueprint.py
```diff
@@ -283,6 +283,23 @@ def remove_command(
             sanitized_uuid, sanitized_cid, params
         )
 
+    @bp.delete(
+        "/studies/{uuid}/all/commands",
```
pl-buiquang: why not just :  `DELETE /studies/{uuid}/commands` ? adding "all" is weird, or it should be after "/commands"

webapp/src/components/Variants/Edition/index.tsx
```diff
@@ -121,6 +134,43 @@ const EditionView = (props: PropTypes) => {
...
+    exportJson(fromCommandItemToJsonCommand(commands), `${studyId}_commands.json`);
+  };
+
+  const onGlobalImport = async (json: object) => {
+    try {
+      const globalJson: Array<JsonCommandItem> = (json as Array<JsonCommandItem>);
+
+      await deleteAllCommands(studyId);
+
+      await appendCommands(studyId, globalJson);
```
pl-buiquang: as discussed, you have to combine those two calls in one.
We want to use the back to check the command validity.
Here, if I upload a proper json containing wrong commands (with wrong action names or args), it will delete all previous commands and apply the new wrong ones or just fail if the json doesn't contains commands.

webapp/src/components/Variants/Edition/DraggableCommands/CommandImportButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+    }
+  };
+
+  const handleChange = async (e: React.ChangeEvent<HTMLInputElement>) => {
+    e.preventDefault();
+    const reader = new FileReader();
+    reader.onload = async (ev: ProgressEvent<FileReader>) => {
+      if (ev.target) {
+        const text = (ev.target.result);
+        onImport(JSON.parse(text as string));
```
pl-buiquang: Separate json parsing from onImport calling and add a try catch surrounding JSON.parse since it may fail and throw if the json is not properly formated.
Display an error box if it happens.

webapp/src/components/Variants/Edition/index.tsx
pl-buiquang: or you could at least call `checkCommandValidity` on the globalJson to check if at least the content look like commands.

pl-buiquang: so, in appendCommand, appendCommands, updateCommand and a newly replaceCommands, it should check command validity by calling something like `self.command_factory.to_icommand()` which will try to read a real ICommand with validation from the command dtos.

webapp/src/components/Variants/Edition/DraggableCommands/CommandListItem.tsx
```diff
@@ -145,6 +150,15 @@ function Item({ provided, item, style, isDragging, index, onDelete, onArgsUpdate
     setJsonData(e.updated_src);
     onArgsUpdate(index, e.updated_src);
   };
+
+  const onImport = (json: object) => {
+    if (checkCommandValidity(json)) {
+      // eslint-disable-next-line dot-notation
+      setJsonData((json as any)['args']);
+      onCommandImport(index, json);
+    }
```
pl-buiquang: here it's just the args that is uploaded. Not the whole command.
Even if the code here prevents that, it could be weird that the user could upload a fully formed command with another action name that the one he is currently uploading the data into

webapp/src/components/Variants/Edition/DraggableCommands/CommandListItem.tsx
```diff
@@ -145,6 +150,15 @@ function Item({ provided, item, style, isDragging, index, onDelete, onArgsUpdate
     setJsonData(e.updated_src);
     onArgsUpdate(index, e.updated_src);
   };
+
+  const onImport = (json: object) => {
+    if (checkCommandValidity(json)) {
```
pl-buiquang: in the "else" condition, raise an error or display a error message box explaining that the json is not ok (though since it will be args in the json, there is, for now, no check to do here anymore.

webapp/src/components/Variants/Edition/DraggableCommands/CommandListItem.tsx
pl-buiquang: execute onSave within a trycatch right after. And, if the save fail (if the data is invalid, see previous comment on adding a check on command update in the back), revert the previous value.

webapp/src/components/Variants/Edition/index.tsx
```diff
@@ -121,6 +134,43 @@ const EditionView = (props: PropTypes) => {
...
+    enqueueSnackbar(t('variants:importSuccess'), { variant: 'success' });
+  };
+
+  const onCommandExport = (index: number) => {
+    const elm = commands[index];
+    exportJson({ action: elm.action, args: elm.args }, `${elm.id}_command.json`);
+  };
+
+  const onGlobalExport = () => {
+    exportJson(fromCommandItemToJsonCommand(commands), `${studyId}_commands.json`);
```
pl-buiquang: call the back api "GET /studies/{id}/commands", here you export the front content that could contains unsaved commands content. Anyway, the export should call the api, since it's the data truth source.

webapp/src/components/Variants/Edition/index.tsx
```diff
@@ -121,6 +134,43 @@ const EditionView = (props: PropTypes) => {
...
+    // eslint-disable-next-line dot-notation
+    tmpCommand[index].args = { ...((json as any)['args'] as object) };
+    tmpCommand[index].updated = true;
+    setCommands(tmpCommand);
+    enqueueSnackbar(t('variants:importSuccess'), { variant: 'success' });
+  };
+
+  const onCommandExport = (index: number) => {
+    const elm = commands[index];
+    exportJson({ action: elm.action, args: elm.args }, `${elm.id}_command.json`);
```
pl-buiquang: same here, call the api.

pl-buiquang: you can use the component DownloadLink

pl-buiquang (changes requested):

webapp/src/components/Variants/Edition/DraggableCommands/CommandListItem.tsx
3lbanna: The json we import should all have the same structure, so action + args.
On the other hand I find, on the contrary, that when one replaces a command at a given position, we should replace the complete command (therefore even the name of the action) and not only the args.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -155,6 +156,17 @@ def get_commands(
         study = self._get_variant_study(study_id, params)
         return [command.to_dto() for command in study.commands]
 
+    def _check_commands_validity(
+        self, study_id: str, commands: List[CommandDTO]
+    ) -> None:
+        for i, command in enumerate(commands):
+            try:
+                self.command_factory.to_icommand(command)
+            except Exception:
```
pl-buiquang: as a rule, when catching an exception, you should log the error, with exception detail:
```
except Exception as e:
  logger.error('Some message indicating what failed', exc_info=e)
```
it will be needed for debug purpose. also some information may want to be passed to the following rethrow later (in order for the user to understand better what went wrong)

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -285,6 +328,11 @@ def update_command(
             params: request parameters
         Returns: None
         """
+        try:
+            self.command_factory.to_icommand(command)
+        except Exception:
+            raise CommandNotValid(study_id)
+
```
pl-buiquang: reuse _check_command_validity with `[command]`

pl-buiquang (approved):

pl-buiquang (approved):

PR #535, opened 2021-10-04
Add variant interactive generation result view

3lbanna:


antarest/core/interfaces/eventbus.py
```diff
@@ -14,6 +14,10 @@ class EventType:
     STUDY_VARIANT_GENERATION_COMMAND_RESULT = (
         "STUDY_VARIANT_GENERATION_COMMAND_RESULT"
     )
+    TASK_STARTED = "TASK_STARTED"
```
pl-buiquang: actually this is a "task added", there is no guarantee (there is a threadpool with limited workers) that the task will run right after.

antarest/core/exceptions.py
```diff
@@ -29,6 +29,11 @@ def __init__(self, message: str) -> None:
         super().__init__(EXPECTATION_FAILED, message)
 
 
+class CommandUpdateAuthorizationError(HTTPException):
+    def __init__(self, message: str) -> None:
+        super().__init__(EXPECTATION_FAILED, message)
```
pl-buiquang: Use HttpStatus.LOCKED which is more semantically correct.
Also where EXPECTATION_FAILED is used, use the one from HttpStatus

antarest/core/tasks/model.py
```diff
@@ -34,6 +35,22 @@ class TaskResult(BaseModel):
...
 
+class TaskLogDTO(BaseModel):
+    id: str
+    message: str
+
+
+class TaskEventMessages(BaseModel):
+    start: Any
+    running: Any
+    end: Any
```
pl-buiquang: these are "str"

antarest/core/tasks/service.py
```diff
@@ -97,12 +104,27 @@ def add_task(
...
+
+        if event_messages is not None:
+            self.event_bus.push(
+                Event(
+                    EventType.TASK_STARTED,
+                    TaskEventPayload(
+                        id=task.id, message=event_messages.start
+                    ).dict(),
+                )
+            )
```
pl-buiquang: the event should be sent even without messages, with a default message "Task {task.id} added".
Same for other events (running, failed, completed).
In that case, the "event_messages" should be renamed "custom_event_messages" (same for the object model) to reflect the fact that this is only customization, but the event are sent with default messages.

antarest/core/tasks/service.py
```diff
@@ -154,33 +176,73 @@ def await_task(
...
 
-    def _run_task(self, callback: Task, task: TaskJob) -> None:
+    def _run_task(
+        self,
+        callback: Task,
+        task_id: str,
+        event_messages: Optional[TaskEventMessages] = None,
+    ) -> None:
+
+        if event_messages is not None:
```
pl-buiquang: see comment about start event

antarest/core/tasks/service.py
```diff
@@ -154,33 +176,73 @@ def await_task(
...
+                    task_id,
+                    TaskStatus.COMPLETED
+                    if result.success
+                    else TaskStatus.FAILED,
+                    result.success,
+                    result.message,
+                    result.return_value,
+                )
+
+                if event_messages is not None:
```
pl-buiquang: see comment about start event

antarest/core/tasks/service.py
```diff
@@ -154,33 +176,73 @@ def await_task(
...
+                        )
+                    )
+            except Exception as e:
+                logger.error(
+                    f"Exception when running task {task_id}", exc_info=e
+                )
+                self._update_task_status(
+                    task_id, TaskStatus.FAILED, False, str(e)
+                )
+                if event_messages is not None:
```
pl-buiquang: see comment about start event

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -243,7 +262,7 @@ def replace_commands(
         """
         study = self._get_variant_study(study_id, params)
         self._check_commands_validity(study_id, commands)
-
+        self._check_update_authorization(study)
```
pl-buiquang, line 268: check update authorization before validity

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -335,6 +360,8 @@ def update_command(
         """
         self._check_commands_validity(study_id, [command])
         study = self._get_variant_study(study_id, params)
+        self._check_update_authorization(study)
```
pl-buiquang, line 366: check update authorization before validity

pl-buiquang (changes requested):

pl-buiquang (approved):

PR #544, opened 2021-10-12
add link to root website

sgatto:


mkdocs.yml
```diff
@@ -29,6 +29,7 @@ theme:
 
 nav:
   - Home: index.md
+  - 'Antares ecosystem' : 'https://antares-doc.readthedocs.io'
```
pl-buiquang: There is already this link after the develop part.
I guess it's better to put it afterward because the user coming here wants foremost to see the antares web doc

PR #552, opened 2021-10-14
Issue 497 improve matrix storage

Hyralc:


antarest/matrixstore/repository.py
```diff
@@ -106,30 +108,81 @@ def __init__(self, config: Config) -> None:
...
+        file_path = self.bucket / f"{h}.tsv"
+        with open(file_path, "w", newline="") as fd:
+            tsv_output = csv.writer(fd, delimiter="\t")
+            tsv_output.writerows(data)
+
+    def _migration(self, h: str) -> bool:
+        # TODO:remove this method once the migration has been done
+        old_matrix_path = self.bucket / h
+        if old_matrix_path.exists():
+            data_json = json.load(open(old_matrix_path), newline="")
```
pl-buiquang: the newline doesn't exist in json.load, I guess you wanted it in open

antarest/matrixstore/repository.py
```diff
@@ -106,30 +108,81 @@ def __init__(self, config: Config) -> None:
...
         return hashlib.sha256(data.encode()).hexdigest()
 
+    def _write_matrix_data(self, h: str, data: List[List[MatrixData]]) -> None:
+        file_path = self.bucket / f"{h}.tsv"
+        with open(file_path, "w", newline="") as fd:
+            tsv_output = csv.writer(fd, delimiter="\t")
+            tsv_output.writerows(data)
+
+    def _migration(self, h: str) -> bool:
+        # TODO:remove this method once the migration has been done
```
pl-buiquang: the migration isn't executed on all matrices anywhere, so I guess, we won't be able to tell when the migration has been done.

pl-buiquang (changes requested):
As discussed, index and columns in MatrixContent are now irrelevant.
The code should be refactored so to get rid of it (keeping the model for parsing old matrices in migration). The service and repository should only create from and return List[List[float]].
Generating a MatrixContent (with index and column) at the very last moment when really needed.

pl-buiquang (approved):

PR #563, opened 2021-10-18
Add available study version list

### LINKED ISSUE #468: Add available study version list
Add an endpoint to indicate the list of available study version supported. 
GET /studies/_versions

UI should use this list to fill the version filter in the study listing component (or anywhere it is needed for that matter).

Hyralc:


pl-buiquang:
The available versions should not be driven by the local launcher version support.
There is currently multiple definition of study version availability:
- RawStudyService templates (l49 RawStudyService) : creating new study
- AVAILABLE_VERSIONS (CreateStudyForm.tsx) : creating new study
- versionList (l114 StudyManagement.tsx) : existing study
- local launcher config : launcher capability

The first three should be all coming from the first one (add missing study versions) and should be available through /studies/_versions
The local launcher config stand apart, as it is just the configuration of the launcher to indicate the compatibility with studies' version

pl-buiquang (changes requested):
See previous comment

pl-buiquang (approved):

PR #569, opened 2021-10-26
issue 521 add a matrix graph

### LINKED ISSUE #521: Add a matrix graph viewer component
The matrices should be viewable with a graph-like component (cf. imagrid).
That allow to select which version of the time series (column in the matrix) to display

Wintxer:


webapp/src/components/ui/MatrixView/MatrixGraphView.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  return (
+    <div className={classes.root}>
+      <div className={classes.container}>
+        <FormControl className={classes.form}>
+          <InputLabel className={classes.input} id="demo-multiple-chip-label">{t('data:graphSelector')}</InputLabel>
+          <Select
+            className={classes.select}
+            labelId="demo-multiple-chip-label"
+            id="demo-multiple-chip"
```
pl-buiquang: change the id to something meaningful (don't use the one from the examples in the documentation)

webapp/src/components/ui/MatrixView/MatrixGraphView.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  const unitChange = (tabBase: Array<number>) => {
+    const tab = [...tabBase];
+    for (let i = 0; i < tab.length; i += 1) {
+      if (i === 0) {
+        tab.splice(i, 1, 100 / tab.length);
+      } else {
+        tab.splice(i, 1, 100 / tab.length + tab[i - 1]);
+      }
+    }
```
pl-buiquang: Here, we compute "100/tab.length" tab.length times ! we should compute it once.
Also, this is an arithmetic progression 
```
U0 = a
Un = Un-1 + a
```
So we have the general term `Un = a(n+1)`

This could be then simplified with:
```
const stepLength = 100/tab.length;
return tabBase.map((el, index) => stepLength * (index + 1));
```

webapp/src/components/ui/MatrixView/MatrixGraphView.tsx
```diff
@@ -0,0 +1,152 @@
...
+          className={classes.monotonous}
+          control={(
+            <Checkbox
+              checked={monotonous}
+              onChange={monotonousChange}
+              name="checked"
+              color="primary"
+            />
+          )}
+          label="Affichage Monotone"
```
pl-buiquang: use translation (also just "Monotone" is fine, and "Monotonic" in english)

webapp/public/locales/en/data.json
```diff
@@ -30,5 +30,6 @@
     "analyzingmatrix": "Analyzing matrices",
     "copyid": "Copy the matrix id",
     "onMatrixIdCopySuccess": "Matrix id copied !",
-    "jsonFormat": "JSON Format"
+    "jsonFormat": "JSON Format",
+    "graphSelector":  "Seletc a column"
```
pl-buiquang: This is "Columns" (we can select multiple ones).
Also no need to add "select" as this is a select so the user already know that the action is a selection.

webapp/public/locales/fr/data.json
```diff
@@ -30,5 +30,6 @@
   "analyzingmatrix": "Analyse des matrices",
   "copyid": "Copie l'identifiant de la matrice",
   "onMatrixIdCopySuccess": "Identifiant de matrice copié !",
-  "jsonFormat": "Format JSON"
+  "jsonFormat": "Format JSON",
+  "graphSelector": "Sélectionnez une colonne"
```
pl-buiquang: This is "Colonnes" (we can select multiple ones)

webapp/public/locales/en/data.json
```diff
@@ -31,5 +31,6 @@
     "copyid": "Copy the matrix id",
     "onMatrixIdCopySuccess": "Matrix id copied !",
     "jsonFormat": "JSON Format",
-    "graphSelector":  "Seletc a column"
+    "graphSelector":  "Columns",
+    "monotonicView": "Monotone"
```
pl-buiquang: So, as the key hints it, the correct english traduction is Monotonic

webapp/src/components/ui/MatrixView/MatrixGraphView.tsx
```diff
@@ -63,41 +63,34 @@ export default function MatrixGraphView(props: PropTypes) {
...
       <div className={classes.container}>
         <FormControl className={classes.form}>
-          <InputLabel className={classes.input} id="demo-multiple-chip-label">{t('data:graphSelector')}</InputLabel>
+          <InputLabel className={classes.input} id="chip-label">{t('data:graphSelector')}</InputLabel>
           <Select
             className={classes.select}
-            labelId="demo-multiple-chip-label"
-            id="demo-multiple-chip"
+            labelId="chip-label"
+            id="chip"
```
pl-buiquang: prepend id with something specific like "matrix-graph" (in case it's in a page with another select)

pl-buiquang (approved):

PR #584, opened 2021-11-03
Allow users to stop a study launch

Hyralc:


antarest/launcher/web.py
```diff
@@ -120,4 +120,25 @@ def get_versions(
...
         return service.get_versions(params=params)
 
+    @bp.post(
+        "/launcher/jobs/{job_id}/kill",
+        tags=[APITag.launcher],
+        summary="Kill job",
+    )
+    def kill_job(
+        job_id: str,
+        engine: Optional[str] = None,
```
pl-buiquang: you don't need this, when a job is created, the launcher used is store in the job_result object

antarest/launcher/service.py
```diff
@@ -121,6 +126,40 @@ def run_study(
 
         return job_uuid
 
+    def kill_job(
+        self, job_id: str, params: RequestParameters, launcher: str
```
pl-buiquang: you don't need the launcher, it is already stored in the job_result object

pl-buiquang (changes requested):

pl-buiquang (approved):

PR #592, opened 2021-11-05
issue 528

Hyralc:


antarest/launcher/repository.py
```diff
@@ -40,3 +40,13 @@ def delete(self, id: str) -> None:
         g = db.session.query(JobResult).get(id)
         db.session.delete(g)
         db.session.commit()
+
+    def delete_by_study_id(self, study_id: str) -> None:
+        job_results_to_delete: List[JobResult] = (
```
pl-buiquang: add some log

antarest/launcher/repository.py
```diff
@@ -40,3 +40,13 @@ def delete(self, id: str) -> None:
...
+            db.session.query(JobResult)
+            .filter(JobResult.study_id == study_id)
+            .all()
+        )
+        for job_result in job_results_to_delete:
+            db.session.delete(job_result)
```
pl-buiquang: make this one query (replace all by delete) : https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.delete

antarest/study/service.py
```diff
@@ -96,6 +96,15 @@ def __init__(
         self.task_service = task_service
         self.areas = AreaManager(self.raw_study_service)
         self.config = config
+        self.callbacks: List[Callable[[str], None]] = []
```
pl-buiquang: this is a callback called only on study deletion, so this should be made clear. "on_delete_callabacks" or something like that

antarest/study/service.py
```diff
@@ -96,6 +96,15 @@ def __init__(
...
+
+    def add_callback(self, callback: Callable[[str], None]) -> None:
+        self.callbacks.append(callback)
+
+    def _callbacks(self, uuid: str) -> None:
```
pl-buiquang: this is a study delete "event", so this should be "on_study_delete"

antarest/launcher/repository.py
Hyralc: There are no logs in the whole class, i think the logs should either be put somewhere else, or should be added in all the methods of the class

pl-buiquang: yes, logs should be added in the service.
Yet, since this method is called directly from the on_delete callback, it should then be wrapped within a service method with a log.

pl-buiquang: also, it would be better to add the callback from the launcher service init constructor

pl-buiquang (approved):

PR #629, opened 2021-11-22
date conversion from local time to utc

### LINKED ISSUE #605: Parse utc date and display correct local date
Dates sent by the backend are utc dates.
We should use momentjs to parse the date and display a correctly formatted date corresponding to local timezone.

- [ ] dates in launcher jobs
- [ ] creation/modification dates of studies

Currently, modification/creation dates of studies are local dates... It should be changed to utc dates. So there is a need to modify these dates in a migration database script.

3lbanna:


antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -86,8 +86,8 @@ def update_from_raw_meta(
             if fallback_on_default is not None:
                 metadata.name = metadata.name or "unnamed"
                 metadata.version = metadata.version or 0
-                metadata.created_at = metadata.created_at or datetime.now()
-                metadata.updated_at = metadata.updated_at or datetime.now()
+                metadata.created_at = metadata.created_at or datetime.utcnow()
+                metadata.updated_at = metadata.updated_at or datetime.utcnow()
```
pl-buiquang, line 94: here, well l78-79 actually, we should convert the datetime.fromtimestamp into utc date

pl-buiquang (approved):

PR #658, opened 2021-12-09
Issue 654 add other searched fields in the study/data search tool

Wintxer:


webapp/src/components/Data/DataView.tsx
```diff
@@ -136,7 +136,7 @@ const DataView = (props: PropTypes) => {
     }
   };
 
-  const matchFilter = (input: string): boolean => input.search(filter) >= 0;
+  const matchFilter = (input: MatrixDataSetDTO): boolean => input.name.search(filter) >= 0;
```
pl-buiquang: where is the filter on the dataset.matrices ids ?

pl-buiquang (approved):

PR #677, opened 2021-12-13
Fix websocket channel subscription

3lbanna:


webapp/src/ducks/websockets.ts
```diff
@@ -21,17 +21,73 @@ export interface WebsocketState {
...
+  const { websockets } = getState();
+  if (websockets.socket && websockets.socket.readyState === WebSocket.OPEN) {
+    websockets.socket.send(JSON.stringify({
+      action: 'UNSUBSCRIBE',
+      payload: channel,
+    }));
+    dispatch({
+      type: 'WS/UNSUBSCRIBE',
+      payload: channel,
+    });
```
pl-buiquang: this should be outside the if, we want to remove the channel anyway

pl-buiquang (changes requested):

pl-buiquang (approved):

PR #769, opened 2022-03-09
Refacto JobService to add fallback output download + Delay study deletion and add rematch

### LINKED ISSUE #768: Add more resilience to watcher when a workspace is unavailable for some time
Currently when a workspace disappear (in case of a mounted network drive for instance), if the launcher loop over it, it will automatically delete the study reference in database.
It can happen that a network disk is temporary unavailable but will come back after some time.
For that we should, either : 
- delay the deletion and only mark studies in database as "unavailable"
or : 
- keep track of old tuple id / path so that when they are readded, we can use the old ids matching the path.

pl-buiquang:


antarest/launcher/service.py
```diff
@@ -61,37 +72,43 @@ def __init__(self, engine: str):
...
             config,
-            study_service,
             LauncherCallbacks(
                 update_status=lambda jobid, status, msg, output_id: self.update(
                     jobid, status, msg, output_id
                 ),
-                after_export_flat=lambda job_id, study_id, study_path, launcher_opts: self.after_export_flat_hooks(
-                    job_id, study_id, study_path, launcher_opts
+                export_study=lambda job_id, study_id, target_path, launcher_opts: self._export_study(
+                    job_id, study_id, target_path, launcher_opts
```
Hyralc: Does export_study = self._export_study  work ?

antarest/launcher/service.py
```diff
@@ -61,37 +72,43 @@ def __init__(self, engine: str):
...
                 append_before_log=lambda jobid, message: self.append_log(
                     jobid, message, JobLogType.BEFORE
                 ),
                 append_after_log=lambda jobid, message: self.append_log(
                     jobid, message, JobLogType.AFTER
                 ),
-                get_job_result=lambda jobid: self.job_result_repository.get(
-                    jobid
+                import_output=lambda jobid, output_path, additional_logs: self._import_output(
+                    jobid, output_path, additional_logs
```
Hyralc: import = self._export_study ?

antarest/launcher/service.py
```diff
@@ -281,10 +305,11 @@ def _filter_from_user_permission(
                 ):
                     allowed_job_results.append(job_result)
             except StudyNotFoundError:
-                logger.info(
-                    f"Removing job result {job_result.id} because of missing study"
-                )
-                self.job_result_repository.delete(job_result.id)
```
Hyralc, line 287: Why not keep it in a else statement ?

antarest/launcher/service.py
```diff
@@ -389,6 +430,132 @@ def get_log(
...
+            self.append_log(
+                job_id, f"Extracting study {study_id}", JobLogType.BEFORE
+            )
+            self.study_service.export_study_flat(
+                study_id,
+                RequestParameters(DEFAULT_ADMIN_USER),
+                target_path,
+                outputs=False,
+            )
+            self.append_log(job_id, "Study extracted", JobLogType.BEFORE)
```
Hyralc, line 444: JobLogType.AFTER

antarest/study/service.py
```diff
@@ -560,44 +567,72 @@ def sync_studies_on_disk(self, folders: List[StudyFolder]) -> None:
...
-                    study.id,
-                )
-                self.event_bus.push(
-                    Event(
-                        type=EventType.STUDY_DELETED,
-                        payload=study.to_json_summary(),
-                        permissions=create_permission_from_study(study),
+                if not study.missing:
+                    logger.info(
+                        "Study %s at %s is not present in disk and will be mark for deletion in %i days",
```
Hyralc: will be marked

antarest/study/service.py
```diff
@@ -560,44 +567,72 @@ def sync_studies_on_disk(self, folders: List[StudyFolder]) -> None:
...
-                logger.info(
-                    "Study=%s is not present in disk and will be deleted",
-                    study.id,
-                )
-                self.event_bus.push(
-                    Event(
-                        type=EventType.STUDY_DELETED,
-                        payload=study.to_json_summary(),
-                        permissions=create_permission_from_study(study),
+                if not study.missing:
```
Hyralc, line 570: This looks like a boolean. missing_since should be more appropriated
and change the condition to "if study.missing_since is not None" to be even more clear

Hyralc (approved):

antarest/launcher/service.py
pl-buiquang: nah, AFTER is for when the study is really launched, not only exported ^^'

antarest/launcher/service.py
pl-buiquang: It's intended, because if the user is not admin, it will remove the job result. But we don't want that. Admin should be able to view all job results and decide to delete them manually.

PR #774, opened 2022-03-10
Edit matrix with slices 

Hyralc:


antarest/matrixstore/business/matrix_editor.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @validator("column_to", always=True)
+    def set_column_to(cls, v: Optional[int], values: Any) -> int:
+        if v is None:
+            return cast(int, values["column_from"])
+        return v
+
+
+class Operation(BaseModel):
+    operation: str
```
pl-buiquang: could be an enum to enforce type (and you can add the _operation_to_operator into the Enum class)

antarest/matrixstore/business/matrix_editor.py
```diff
@@ -0,0 +1,77 @@
...
+    value: float
+
+
+class MatrixEditor:
```
pl-buiquang: could depend on an ISimpleMatrixService, so that you can do the full edit (createMatrix).
This way, you won't need to put the "assert type(matrix_service) is MatrixService" in MatrixManagement.

antarest/matrixstore/service.py
```diff
@@ -469,3 +474,14 @@ def download_matrix(
...
+    def update_matrix(
+        self,
+        matrix_data: List[List[MatrixData]],
+        slices: List[MatrixSlice],
+        operation: Operation,
+    ) -> str:
+        new_matrix_data = MatrixEditor.update_matrix_content_with_slices(
+            matrix_data=matrix_data, slices=slices, operation=operation
+        )
+        return self.create(new_matrix_data)
```
pl-buiquang: or you could put that in the MatrixManagement class, or in the interface/abstract class ISimpleMatrixService

pl-buiquang (approved):
Should add some unit test on the replace matrix function using all operator types

pl-buiquang (approved):

PR #791, opened 2022-03-22
Adequacy patch extension & improvements

pl-buiquang:


antarest/launcher/extensions/adequacy_patch/extension.py
```diff
@@ -30,26 +48,107 @@ def after_export_flat_hook(
...
+    ) -> Dict[str, bool]:
+        area_to_turn_on: List[str] = [
+            transform_name_to_id(area_id)
+            for area_id in adq_patch_config.get("areas", [])
+        ]
+        original_area_status: Dict[str, bool] = {}
+        original_link_status: Dict[str, bool] = {}
+        for area_id, area in study.config.areas.items():
+            # areas
+            original_area_status[area_id] = "hourly" in area.filters_year
```
Hyralc: The name of the variable suggests a status name such as a string, but not a boolean. It should be renamed for understanding purpose.

antarest/launcher/extensions/adequacy_patch/extension.py
```diff
@@ -30,26 +48,107 @@ def after_export_flat_hook(
...
+                            "input",
+                            "links",
+                            area_id,
+                            "properties",
+                            area_2,
+                            "filter-year-by-year",
+                        ],
+                    )
+
+        with FileLock(self.tmp_dir / "data.lock"):
```
Hyralc: In other files, when FileLock is used, it is used with a constant defined in the class and not directly a string, maybe do the same here ?

Hyralc (approved):

PR #800, opened 2022-03-31
Issue 710 improve packaging

Hyralc:


pl-buiquang (approved):
Before merge : 
- add a test for the new local launcher code
- remove the watcher code
- remove the matrices in resources/deploy/matrices (put a .placeholder file so that the directory is in the repository)

PR #900, opened 2022-05-10
Add managed studies virtual folder management

### LINKED ISSUE #780: Add managed studies virtual folder management
Managed studies (raw studies in default workspace or variant studies) does not have a folder property so they are all bunched together in the root folder.
We should be able to create new folders and "move" these studies into them so that they can be as organized as their unmanaged raw studies counterparts. 
Copied studies should retain their folder location.

pl-buiquang:


webapp/src/components/studies/MoveStudyDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { moveStudy } from "../../services/api/study";
+import { isStringEmpty } from "../../services/utils";
+import FormDialog, { SubmitHandlerData } from "../common/dialogs/FormDialog";
+
+interface Props {
+  study: StudyMetadata;
+  onClose: () => void;
+}
+
+function MoveStudyDialog(props: Props & DialogProps) {
```
skamril:
```typescript
interface Props extends FormDialog
```

webapp/src/components/studies/MoveStudyDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const { study, open, onClose } = props;
+  const [t] = useTranslation();
+  const mounted = usePromise();
+  const { enqueueSnackbar } = useSnackbar();
+  const enqueueErrorSnackbar = useEnqueueErrorSnackbar();
+
```
skamril, line 24: Block convention ^^ (Event Handlers and JSX here)

skamril (changes requested):

skamril (approved):

PR #917, opened 2022-05-18
Issue Update usePromise hook

skamril:


webapp/src/hooks/usePromise.ts
```diff
@@ -17,50 +28,46 @@ function usePromise<T>(
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
     [reloadCount, ...deps]
   );
 
-  return { data, isLoading, error, reload };
+  return {
+    data,
+    status,
+    isLoading:
+      status === PromiseStatus.Idle || status === PromiseStatus.Pending,
```
pl-buiquang, line 68: well, depending on the meaning, but when the status is Idle, it is not loading.

PR #922, opened 2022-05-24
Fix logmodal & subscription on websockets reconnection

pl-buiquang:


webapp/src/services/webSockets.ts
```diff
@@ -119,25 +111,32 @@ export function addWsMessageListener(listener: MessageListener): VoidFunction {
...
   function send(action: string): void {
-    const messagesToSend = RA.ensureArray(channels).map((chan) =>
+    channelSubscriptions = channelSubscriptions.filter(
+      (chan) => channelsList.indexOf(chan) === -1
+    );
+    if (action === "SUBSCRIBE") {
+      channelSubscriptions.push(...channelsList);
+    }
+
+    const messagesToSend = channelsList.map((chan) =>
```
skamril: Move inside the `if`

skamril (approved):

PR #947, opened 2022-06-14
Fixes : trad, variant cache bug, user config preservation on variant generation (xpansion) 

pl-buiquang:


MartinBelthle:
Hello @pl-buiquang, do you remember why you coded this: [Preserve unmanaged user config on variant generation](https://github.com/AntaresSimulatorTeam/AntaREST/pull/947/commits/6bd8942d0d2314603ad6a13be34938f25b87c06c) ?
It is preventing me to handle properly xpansion variants because it overwrites the parent configuration in the variant snapshot so i would like to remove this part of the code.
The problem is that i fail to understand why you coded it and so do other team member (also i do not find any doc on the subject) and i do not want to remove this part of the code carelessly.

PR #970, opened 2022-06-24
Link styles in map

Wintxer:


Wintxer:
TODO:
- Need an update link function
- Error 500 on getAllLinks when testing on a large study

webapp/src/components/App/Singlestudy/explore/Modelization/Map/index.tsx
```diff
@@ -300,19 +301,18 @@ function Map() {
...
-                    )
-                  ),
-                [] as Array<LinkProperties>
-              )
+              links.map((link) => {
+                return {
+                  source: link.area1,
+                  target: link.area2,
+                  color: link.ui
+                    ? `rgb(${link.ui?.color.split(",").join(", ")}`
```
skamril: `split` and `join` not necessary

webapp/src/services/api/studydata.ts
```diff
@@ -61,9 +61,12 @@ export const deleteLink = async (
 };
 
 export const getAllLinks = async (
-  uuid: string
+  uuid: string,
+  withUi?: boolean
 ): Promise<Array<LinkCreationInfo>> => {
```
skamril: Use conditional types return (cf. `getUsers` in "src/services/api/users.ts").
You need to have `LinkCreationInfoDTO` and `LinkCreationInfoWithUiDTO` that extends the previous one.

skamril (changes requested):

webapp/src/common/types.ts
```diff
@@ -563,9 +563,16 @@ export interface UpdateAreaUi {
   color_rgb: Array<number>;
 }
 
+export interface LinkCreationUI {
```
skamril: `DTO` suffix when it comes from server.
(Also for `LinkCreationInfo`)

skamril (changes requested):

webapp/src/components/App/Singlestudy/explore/Modelization/Map/index.tsx
```diff
@@ -300,19 +301,32 @@ function Map() {
...
+                  linecap = "round";
+                }
+                if (link.ui?.style === "dash") {
+                  style = [16, 8];
+                  linecap = "square";
+                }
+                if (link.ui?.style === "dotdash") {
+                  style = [10, 6, 1, 6];
+                  linecap = "square";
+                }
```
pl-buiquang: here you can use R.cond from Ramda (check examples in code) : https://ramdajs.com/docs/#cond

webapp/src/common/types.ts
```diff
@@ -563,11 +563,21 @@ export interface UpdateAreaUi {
   color_rgb: Array<number>;
 }
 
-export interface LinkCreationInfo {
+export interface LinkCreationUIDTO {
```
pl-buiquang: rename LinkUIInfoDTO

pl-buiquang (approved):

webapp/src/common/types.ts
skamril: `LinkInfoDTO` instead we have `LinkInfoDTOWithUI`

PR #1017, opened 2022-08-02
Refacto append commands + auto add matrix in store for managed studies

pl-buiquang:


pl-buiquang:
TODO: add automatic command transformation to put matrices in store and keep the id in args (for managed studies and variants)

PR #1119, opened 2022-10-06
fix(study): update study state after changes

hdinia:


webapp/src/components/App/Singlestudy/PropertiesDialog/index.tsx
```diff
@@ -88,6 +89,7 @@ function PropertiesDialog(props: Props) {
         );
       }
 
+      updateStudyData();
```
skamril, line 92:
```javascript
if (name || tags) {
  await updateStudyMetadata(studyId, {
    name: data.values.name,
    tags: data.values.tags,
  });
  
  if (name) {
    updateStudyData();
  }
}
```

skamril (changes requested):

skamril (approved):

PR #1160, opened 2022-11-02
refactor(general): update form

skamril:


antarest/study/business/general_management.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+DayNumberType = conint(ge=1, le=365)
+
+
+class GeneralFormFields(FormFieldsBaseModel):
+    mode: Optional[Mode]
+    first_day: Optional[DayNumberType]  # type: ignore
+    last_day: Optional[DayNumberType]  # type: ignore
+    horizon: Optional[StrictStr]
```
pl-buiquang: let loose with just "str" as sometimes this contains a casted int (like 2030)

pl-buiquang: in general, don't put StrictStr because the FileTree IniReader will always cast a string to a more specific type if possible

tests/storage/business/test_config_manager.py
```diff
@@ -1,113 +0,0 @@
...
-from antarest.study.storage.variantstudy.model.command_context import (
-    CommandContext,
-)
-from antarest.study.storage.variantstudy.model.dbmodel import VariantStudy
-from antarest.study.storage.variantstudy.variant_study_service import (
-    VariantStudyService,
-)
-
-
-def test_thematic_trimming_config():
```
pl-buiquang: keep this test

pl-buiquang (changes requested):

pl-buiquang (approved):

PR #1239, opened 2023-01-19
fix(ui-map): prevent duplicate layers and districts on create/update

hdinia:


laurent-laporte-pro:
Thanks for this implementation.

Rigorously, the detection of duplicates should be done on the identifiers of Layers and Districts and not on the names. For example, the districts "Foo" and "FOO" have different names but identical IDs, so there is a duplicate in this case. We have to consider the user's point of view: for him, the names "Foo" and "FOO" are equivalent.

The current implementation does not detect this case and does not match the requirements, sorry.

I think you should use the implementation that is done on the back office side because an error message is issued in case of duplicate = `HTTPException` with the code CONFLICT = 409, 'Conflict', 'Request conflict'.

hdinia:
> Thanks for this implementation.
> 
> Rigorously, the detection of duplicates should be done on the identifiers of Layers and Districts and not on the names. For example, the districts "Foo" and "FOO" have different names but identical IDs, so there is a duplicate in this case. We have to consider the user's point of view: for him, the names "Foo" and "FOO" are equivalent.
> 
> The current implementation does not detect this case and does not match the request, sorry.
> 
> I think you should use the implementation that is done on the back office side because an error message is issued in case of duplicate = `HTTPException` with the code CONFLICT = 409, 'Conflict', 'Request conflict'.

Thanks for the feedback, yes you are correct that comparing the names of layers and districts may not always accurately identify duplicates, as two layers or districts with different names but the same ID would be considered unique by the current implementation.
Implementing duplicate validation based on names on the front-end can be a useful first line of defense, but it is not a foolproof method.
It would be better to use the identifiers of layers and districts to detect duplicates on the front-end as well as back-end, as they are unique and guaranteed to be consistent across all instances of the layer or district.
However, if the front-end validation is in place as an additional check and not the only validation method(which is the case here), it could be considered a valid implementation. But it's always important to ensure that the back-end validation is also in place to ensure data integrity.

laurent-laporte-pro:
Currently, there is no error message displayed to the user when there is a duplicate (code 409). Can you add this?

laurent-laporte-pro:
On the other hand, when the form is displayed, it would be nice for the user if the cursor was automatically positioned on the first field of the form, i.e. on the Layer/District name.

hdinia:
> Currently, there is no error message displayed to the user when there is a duplicate (code 409). Can you add this?

indeed this error handling should be implemented

skamril (approved):

PR #1251, opened 2023-01-25
fix(ui-study): remove the create command button

hdinia:


webapp/src/components/App/Singlestudy/Commands/Edition/index.tsx
```diff
@@ -523,38 +497,12 @@ function EditionView(props: Props) {
...
-                    color="primary"
-                    variant="outlined"
-                    onClick={() => setOpenAddCommandDialog(true)}
-                  >
-                    {t("button.newCommand")}
-                  </Button>
-                }
-              />
-            </Box>
-          </Body>
```
pl-buiquang: this should stay, this is the placeholder view when there is no commands. Only the callToAction button should be removed

pl-buiquang (changes requested):

skamril (approved):

pl-buiquang (approved):

PR #1260, opened 2023-01-25
fix(storage/variant): create filestudy config with correct output path

### LINKED ISSUE #1259: Some variant modification generate in cache a study synthesis/config without the correct output path

pl-buiquang:


antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -884,6 +884,7 @@ def _generate(
             study = self.study_factory.create_from_fs(
                 self.get_study_path(variant_study),
                 study_id=variant_study.id,
+                output_path=Path(variant_study.path) / OUTPUT_RELATIVE_PATH,
```
laurent-laporte-pro, line 887: Sometimes we use `OUTPUT_RELATIVE_PATH`, and sometimes we use `"outpu"`.
What's the difference ? Why not always use `OUTPUT_RELATIVE_PATH` ?

laurent-laporte-pro (changes requested):

antarest/study/storage/variantstudy/variant_study_service.py
pl-buiquang: I don't know where we use "output" (not related to this PR), but yes we should use `OUTPUT_RELATIVE_PATH` instead of `"output"` in most cases I guess

laurent-laporte-pro (approved):

PR #1290, opened 2023-02-03
feat(ui-results): add download button on results matrix

### LINKED ISSUE #1225: Add download matrix button on matrix output
The issue is in the output matrices (Select a Study and open it, then select the "Results" tab to display the list of results, then select a specific result to display the matrix).

Image

For instance, we can add a "Download" button to the left of the mini-form "Monte-Carlo" / "Affichage" / "Temporalité'.

Take the example from the **Links** view (for example) which already has the button.

hdinia:


webapp/src/components/App/Singlestudy/explore/Results/ResultDetails/index.tsx
```diff
@@ -137,6 +138,16 @@ function ResultDetails() {
     setYear(year);
   };
 
+  const handleDownload = (data: MatrixType, filename: string): void => {
```
skamril: Create a matrix util

skamril (changes requested):

skamril (approved):

PR #1401, opened 2023-03-13
fix(package): do not create symlink for `AntaresWebServer` during package creation

flomnes:


laurent-laporte-pro (dismissed):

laurent-laporte-pro (changes requested):
Il faut supprimer le `else`

flomnes:
> Il faut supprimer le `else`

PR vide ?

laurent-laporte-pro:
Creating the symlink is not really necessary, as the user can navigate to the `AntaresWebServer` directory and run the application from there and/or create the symlink themselves. Currently there is a problem with GitHub when creating the ZIP package of the `dist` directory. The ZIP creation is done without preserving symbolic links; instead, the target directory is cloned.

laurent-laporte-pro (approved):

laurent-laporte-pro:
### Integration Test

On vérifie bien que le package Ubuntu ne contient plus le lien symbolique.

PR #1445, opened 2023-04-11
feat(api): spacial correlation matrix

### LINKED ISSUE #1443: [ANT-652] Add "Spatial correlation" API
## Problem Statement

Related to [[ANT-652] Add "Spatial correlation" form](https://github.com/AntaresSimulatorTeam/AntaREST/issues/1358)

The objective of this request is to develop the Swagger API necessary for the management of spatial correlations between the different hydraulic generators.

In the Antares Simulator GUI, the spatial correlations are represented as a large symmetrical matrix that links the zones to each other.

In Antares Web, the correlations will be presented as a form per zone. The form will display the current zone and the correlation coefficients with the other zones (except for zero coefficients).

We have the following properties:
- The values of this matrix are displayed in percentages, but values are stored as coefficients, where -1 <= coefficient <=1.
- By default, the correlation matrix is a unit matrix: the diagonal is 100%, the coefficients are 0 everywhere else.
- Coefficients can be negative or null, but the overall matrix is never null. Note that null values are not stored in the configuration files.
- For all zones, the correlation coefficient of a zone with itself is always 100%. This value cannot be changed. Note that this value is not stored in the configuration files.
- Symmetry property: if a zone *A* has a correlation value with a zone "B", then the correlation value of zone "B" with zone "A" is the same.
- If you delete the *A* zone, this means that the *A* column and the *A* row of the matrix are deleted: we must therefore update all the coefficients whose abscissa is *A* or whose ordinate is *A*.
- If we add a *B* zone, this is equivalent to adding a *B* column and a *B* row, where all the coefficients are null except for the coefficient of the cell (*B*, *B*) = 100%.

## Generalization to other correlation matrices

Correlation matrices can be found in the following files:

- `input/hydro/prepro/correlation.ini`: 🌧️ Hydraulic Generator
- `input/load/prepro/correlation.ini`: 🏋 Load Generator
- `input/solar/prepro/correlation.ini`: ☀️ Solar Generator
- `input/wind/prepro/correlation.ini`: 💨 Wind Turbine Generator

The vast majority of correlation matrices are "annual" mode matrices. Out of the 268 examples, there are 3 examples in "monthly":

* `Antares_Simulator_Tests_NR/short-tests/031 Wind Analysis`: 💨 `input/wind/prepro/correlation.ini` (empty matrices)
* `Antares_Simulator_Tests_NR/short-tests/031 Wind Analysis`: 🏋 `input/load/prepro/correlation.ini` (empty matrices)
* `Antares_Simulator_Tests_NR/medium-tests/032 Exploited Wind Fields`: 💨 [`input/wind/prepro/correlation.ini`](https://github.com/AntaresSimulatorTeam/Antares_Simulator_Tests_NR/blob/cf213310fab63db7c2ddb3f4f4cc7a5463eddf23/medium-tests/032%20Exploited%20Wind%20Fields/input/wind/prepro/correlation.ini)

### Regression matrices with negative values

* `input/hydro/prepro/correlation.ini`
  - 35 : False
  - 54 : True
* `input/load/prepro/correlation.ini`
  - 1  : True
  - 36 : False
* `input/solar/prepro/correlation.ini`
  - 1  : True
  - 34 : False
* `input/wind/prepro/correlation.ini`
  - 20 : True
  - 72 : False

### Null regression matrices

* `input/hydro/prepro/correlation.ini`
  - 89 : False
* `input/load/prepro/correlation.ini`
  - 37 : False
* `input/solar/prepro/correlation.ini`
  - 35 : False
* `input/wind/prepro/correlation.ini`
  - 92 : False

### Coefficient in range [-1; 1]

* `input/hydro/prepro/correlation.ini`
  - 89 : True
* `input/load/prepro/correlation.ini`
  - 37 : True
* `input/solar/prepro/correlation.ini`
  - 35 : True
* `input/wind/prepro/correlation.ini`
  - 92 : True

laurent-laporte-pro:
This PR implements the management of the hydro correlation matrices, but also the load, solar and wind correlations which work in the same way.

We have chosen to manage only the annual correlations because the monthly correlations are considered deprecated.

We have developed two pairs of end points:
- A GET/PUT pair to read and write a correlation matrix of a study,
- A GET/PUT pair to read and write the data from the correlation coefficient input form of a particular area.


antarest/study/business/correlation_management.py
```diff
@@ -0,0 +1,377 @@
...
+        if not correlation:
+            raise ValueError("correlation must not be empty")
+        counter = collections.Counter(field.area_id for field in correlation)
+        if duplicates := {id_ for id_, count in counter.items() if count > 1}:
+            raise ValueError(
+                f"correlation must not contain duplicate area IDs: {duplicates}"
+            )
+        # fmt: off
+        array = np.array([a.coefficient for a in correlation], dtype=np.float64)
+        if np.any((array < -100) | np.any(array > 100)):
```
hdinia, line 65: `if np.any((array < -1) | (array > 1)):
`
seems to be a logical error here

antarest/study/business/correlation_management.py
```diff
@@ -0,0 +1,377 @@
...
+        if invalid_ids := correlation_ids - set(area_ids):
+            # sort for deterministic error message and testing
+            raise AreaNotFound(*sorted(invalid_ids))
+
+        file_study = self.storage_service.get_storage(study).get_raw(study)
+        array = self._get_array(file_study, area_ids)
+        for field in data.correlation:
+            i = area_ids.index(field.area_id)
+            j = area_ids.index(area_id)
+            array[i][j] = field.coefficient / 100
```
hdinia, line 286: Optionnal: map area ids to their indexes before the loop, which will reduce the number of iterations

```
area_id_to_index = {id_: index for index, id_ in enumerate(area_ids)}

for field in data.correlation:
    i = area_id_to_index[field.area_id]
    j = area_id_to_index[area_id]
    array[i][j] = field.coefficient / 100

```

antarest/study/business/correlation_management.py
```diff
@@ -0,0 +1,377 @@
...
+        self._set_array(study, file_study, area_ids, array)
+
+        column = array[:, area_ids.index(area_id)] * 100
+        return CorrelationFormFields.construct(
+            correlation=[
+                CorrelationField.construct(area_id=a, coefficient=c)
+                for a, c in zip(area_ids, column)
+                if c
```
hdinia, line 294: Suggestion: change to `if c != 0` to explicitly states that you are filtering out the zero coefficients. `if c` can be hard to understand

antarest/study/business/correlation_management.py
```diff
@@ -0,0 +1,377 @@
...
+        area_ids = [area.id for area in all_areas]
+        array = self._get_array(file_study, area_ids)
+        column = array[:, area_ids.index(area_id)] * 100
+        return CorrelationFormFields.construct(
+            correlation=[
+                CorrelationField.construct(area_id=a, coefficient=c)
+                for a, c in zip(area_ids, column)
+                if c
+            ]
+        )
```
hdinia, line 251: Here you can remove the construct method since the `CorrelationField` and `CorrelationFormFields` are simple models and can be instantiated directly.

```
  correlation = [
        CorrelationField(area_id=a, coefficient=c)
        for a, c in zip(area_ids, column)
        if c
    ]

    return CorrelationFormFields(correlation=correlation)

```
note: same for `get_correlation_form_fields` etc

antarest/study/business/correlation_management.py
```diff
@@ -0,0 +1,377 @@
...
+    for i in range(count):
+        # not saved: values from the diagonal are always == 1.0
+        for j in range(i + 1, count):
+            coefficient = array[i][j]
+            if not coefficient:
+                # null values are not saved
+                continue
+            a1 = area_ids[i]
+            a2 = area_ids[j]
+            correlation_cfg[f"{a1}%{a2}"] = str(coefficient)
```
hdinia: is the `str(coefficient)` intentional ?

antarest/study/business/correlation_management.py
```diff
@@ -0,0 +1,377 @@
...
+    ) -> None:
+        self.storage_service = storage_service
+        if gen_cat not in {"hydro", "load", "solar", "wind"}:
+            raise ValueError(
+                f"The generator category '{gen_cat}' is not supported."
+                f" Today, only the 'hydro' category is fully supported,"
+                f" we also keep the possibility to manage the 'load',"
+                f" 'solar' and 'wind' categories but the usage is deprecated."
+            )
+        self.url = ["input", gen_cat, "prepro", "correlation", "annual"]
```
hdinia: `gen_cat` 
In my opinion it is better to remove this parameter because it will never be used and can be confusing, in addition to adding code

antarest/study/storage/variantstudy/model/command/remove_area.py
```diff
@@ -141,6 +141,34 @@ def _remove_area_from_hydro_allocation(
...
+        - ["input", "hydro", "prepro", "correlation"]
+        - ["input", "load", "prepro", "correlation"]
+        - ["input", "solar", "prepro", "correlation"]
+        - ["input", "wind", "prepro", "correlation"]
+
+        Args:
+            study_data:File Study to update.
+        """
+        # Update the correlation matrices
+        for generator in "hydro", "load", "solar", "wind":
```
hdinia: same here only `hydro` is needed

hdinia (dismissed):

antarest/study/business/correlation_management.py
laurent-laporte-pro: No, the form processes values between -100 and 100, while the matrix processes values between -1 and 1.

antarest/study/business/correlation_management.py
laurent-laporte-pro: We could do that but is is a micro-optimisation.

antarest/study/business/correlation_management.py
laurent-laporte-pro: We can also avoid it, but at least the conversion is done during serialization.

antarest/study/business/correlation_management.py
laurent-laporte-pro: When we use `construct`, the model validation is not done. This is intentional here to avoid it because it is not necessary.

antarest/study/business/correlation_management.py
laurent-laporte-pro: Actually, in Python, this is a bad practice.

antarest/study/business/correlation_management.py
laurent-laporte-pro: As verbally stated, I want to keep a trace in the code to notify the developer that a thorough analysis has been done for the other generator categories. If you prefer, I can use a constant and replace it with a comment, but I think that's a shame.

antarest/study/business/correlation_management.py
laurent-laporte-pro: conversion to `str` is not necessary

hdinia (dismissed):

antarest/study/business/correlation_management.py
hdinia: it's ok then

PR #1452, opened 2023-04-15
feat(matrix): improve matrix read/write using NumPy

### LINKED ISSUE #1451: Improve time series read/write performance
As a developer working on the Antares Web application you will be interested in the results of the performance tests that I performed on the NumPy read and write functions. The tests revealed that using the NumPy read/write functions is more efficient than using the Python `csv` library to read and
write large time series (8760 hours x 1000 years).

Here are the profiling I performed on my workstation:
- The `numpy.savetxt` function takes 2.6s to save a matrix, while it takes 4.5s for its equivalent with csv, which is 60% faster.
- The `numpy.loadtxt` function takes 2.1s to load a matrix, while it takes 3s for its equivalent with equivalent with `csv`, which is 70% faster.

These results are significant and suggest that it would be beneficial to review the implementation of matrix management using the NumPy functions.

Here is the source code:

```python
"""
Compare the processing time of numpy.savetxt and numpy.loadtxt functions
to read and write 2D matrices of floating point numbers
with the processing times of those using the `csv` python module.
"""
import csv
import tempfile
import time
from pathlib import Path

import numpy


def savetsv(tsv_file: Path, matrix: numpy.ndarray):
    with tsv_file.open(mode="w", newline="") as fd:
        writer = csv.writer(fd, delimiter="\t")
        writer.writerows(matrix)


def loadtsv(tsv_file: Path):
    with tsv_file.open(mode="r", newline="") as fd:
        reader = csv.reader(fd, delimiter="\t")
        return numpy.array(
            [[float(c) for c in line] for line in reader],
            dtype=numpy.float64,
        )


# noinspection PyTypeChecker
def profiling(profiling_dir: Path, iterations: int):
    tsv_file1 = profiling_dir.joinpath("numpy.tsv")
    tsv_file2 = profiling_dir.joinpath("csv.tsv")
    results = numpy.zeros((0, 4), dtype=float)
    for i in range(1, iterations + 1):
        print(f"Running iteration #{i}/{iterations}...")
        matrix = numpy.random.rand(8760, 1000)
        start = time.time()
        numpy.savetxt(tsv_file1, matrix, delimiter="\t", fmt="%.16g")
        save_numpy_time = time.time()
        matrix1 = numpy.loadtxt(tsv_file1, delimiter="\t", ndmin=2)
        load_numpy_time = time.time()
        savetsv(tsv_file2, matrix)
        save_csv_time = time.time()
        matrix2 = loadtsv(tsv_file2)
        load_csv_time = time.time()
        assert matrix1.all() == matrix.all()
        assert matrix2.all() == matrix.all()
        results = numpy.append(
            results,
            [
                [
                    save_numpy_time - start,
                    load_numpy_time - save_numpy_time,
                    save_csv_time - load_numpy_time,
                    load_csv_time - save_csv_time,
                ]
            ],
            axis=0,
        )
    print("Elapsed time for each function:")
    print(
        [
            "save_numpy_time",
            "load_numpy_time",
            "save_csv_time",
            "load_csv_time",
        ]
    )
    mean_time = numpy.mean(results, axis=0)
    print(mean_time)
    print()
    print("Performance of NumPy functions over `csv` module:")
    print(f"- numpy.savetxt / csv+write: {mean_time[0] / mean_time[2]:.1%}")
    print(f"- numpy.loadtxt / csv+read:  {mean_time[1] / mean_time[3]:.1%}")


if __name__ == "__main__":
    with tempfile.TemporaryDirectory(
        prefix="~profiling.", suffix=".tmp"
    ) as filepath:
        tmp_path = Path(filepath)
        profiling(tmp_path, 10)
```

laurent-laporte-pro:


antarest/matrixstore/repository.py
```diff
@@ -94,87 +90,110 @@ def save(self, matrix: Matrix) -> Matrix:
...
+        Args:
+            matrix_hash: SHA256 hash
+
+        Returns:
+            `True` if the matrix exist else `None`.
+        """
+        matrix_file = self.bucket_dir.joinpath(f"{matrix_hash}.tsv")
+        return matrix_file.exists()
 
     def save(self, content: List[List[MatrixData]]) -> str:
```
sylvlecl: In a future work, we should try to extend the use of numpy (or better pandas) to every parts of the code that uses arrays: the plain lists of floats will create inefficiencies everywhere (at least memory usage)

antarest/matrixstore/service.py
```diff
@@ -60,50 +48,42 @@ def create(self, data: List[List[MatrixData]]) -> str:
         raise NotImplementedError()
 
     @abstractmethod
-    def get(self, id: str) -> Optional[MatrixDTO]:
+    def get(self, matrix_hash: str) -> Optional[MatrixDTO]:
```
sylvlecl: Not in favor of using `matrix_hash` instead of `id` here, because the fact that it's a matrix hash is an implementation detail.

The user of the service should just use this as an identifier, without knowing that it's a matrix hash: actually, we should not rely anywhere outside of this class that it's a matrix hash, otherwise we will create logical dependencies between our components.

antarest/matrixstore/service.py
```diff
@@ -455,19 +503,21 @@ def export_task(notifier: TaskUpdateNotifier) -> TaskResult:
...
         """
         if not params.user:
             raise UserHasNotPermissionError()
-        matrix = self.get(matrix_id)
-        if matrix:
-            write_tsv_matrix_to_file(matrix, str(tmp_file))
+        if matrix := self.get(matrix_hash):
+            array = np.array(matrix.data, dtype=np.float64)
+            # noinspection PyTypeChecker
+            np.savetxt(filepath, array, delimiter="\t", fmt="%.18g")
```
sylvlecl: note that default encoding is latin1, not UTF-8, should not be an issue since we have only numbers here, though.

antarest/matrixstore/service.py
```diff
@@ -178,26 +168,41 @@ def create_by_importation(
...
         if is_json:
             return self.create(MatrixContent.parse_raw(file).data)
-        else:
-            data = parse_tsv_matrix(file)
-            return self.create(data)
+        # noinspection PyTypeChecker
+        matrix = np.loadtxt(
+            BytesIO(file), delimiter="\t", dtype=np.float64, ndmin=2
+        )
+        matrix_data = cast(List[List[float]], matrix.tolist())
```
sylvlecl: it's a pity to need to go through this conversion just to convert it back again to numpy when we save it later

antarest/matrixstore/service.py
```diff
@@ -324,23 +329,65 @@ def delete_dataset(self, id: str, params: RequestParameters) -> str:
...
+        """
+        matrix = self.repo.get(matrix_hash)
+        if matrix is None:
+            return None
+        content = self.matrix_content_repository.get(matrix_hash)
+        return MatrixDTO.construct(
+            id=matrix.id,
+            width=matrix.width,
+            height=matrix.height,
+            created_at=int(time.mktime(datetime.timetuple(matrix.created_at))),
```
sylvlecl: I guess you can make it simpler:
```python
int(matrix.created_at.timestamp())
```

antarest/matrixstore/repository.py
```diff
@@ -94,87 +90,110 @@ def save(self, matrix: Matrix) -> Matrix:
...
+        #    depending on the platform on which the calculations are performed,
+        # 2. The JSON conversion, necessarily involves a textual representation
+        #    of the floating point numbers which can introduce rounding errors.
+        # However, this method is still a good approach to calculate a hash value
+        # for a non-mutable NumPy Array.
+        matrix = np.array(content, dtype=np.float64)
+        matrix_hash = hashlib.sha256(matrix.data).hexdigest()
+        matrix_file = self.bucket_dir.joinpath(f"{matrix_hash}.tsv")
+        # noinspection PyTypeChecker
+        np.savetxt(matrix_file, matrix, delimiter="\t", fmt="%.18g")
```
sylvlecl: we should check if that matrix already exists and not save it again if it does (it's the whole point of using a hash)

sylvlecl: Also: 
it was not better before, but note that we have a race condition between threads and processes on this component.
If 2 threads want to write down the same matrix concurrently, it might end up in a "corrupted" file.

antarest/matrixstore/repository.py
```diff
@@ -94,87 +90,110 @@ def save(self, matrix: Matrix) -> Matrix:
...
+    def delete(self, matrix_hash: str) -> None:
+        """
+        Deletes the tsv file containing the content of a matrix with a given SHA256 hash.
 
-    def delete(self, id: str) -> None:
-        matrix_file = self.bucket / f"{id}.tsv"
-        if matrix_file.exists():
-            matrix_file.unlink()
+        Args:
+            matrix_hash: SHA256 hash
```
sylvlecl: We could add that this raises when the matrix does not exist: it's a change of behaviour compared to before.

antarest/matrixstore/service.py
```diff
@@ -153,17 +121,39 @@ def _from_dto(dto: MatrixDTO) -> Tuple[Matrix, MatrixContent]:
...
+                matrix = Matrix(
+                    id=matrix_hash,
+                    width=len(data[0]) if data else 0,
+                    height=len(data),
+                    created_at=datetime.now(timezone.utc),
                 )
+                self.repo.save(matrix)
+        except Exception:
+            # delete the file so as not to leave unreferenced files lying around
+            self.matrix_content_repository.delete(matrix_hash)
```
sylvlecl: we must not do this:
if the matrix already existed before (from a previous study for example), it will be deleted whereas it's still in use !

antarest/matrixstore/service.py
sylvlecl: I would be in favor of using `f` and not `g` in order to avoid scientific notation (also in other uses of savetxt in the repo)

In particular, it's not guaranteed that antares simulator does handle scientific notation in its input files.

tests/matrixstore/test_service.py
```diff
@@ -1,130 +1,144 @@
...
+    def test_create__side_effect(self, matrix_service: MatrixService):
+        """Creates a new matrix object with the specified data, but fail during saving."""
+        # if the matrix can't be created in the service
+        matrix_repo = matrix_service.repo
+        matrix_repo.save = Mock(side_effect=Exception("database error"))
+        with pytest.raises(Exception, match="database error"):
+            data = [[1, 2, 3], [4, 5, 6]]
+            matrix_service.create(data)
+
+        # The matrix is not saved in the content repository
```
sylvlecl: here the test is wrong : 
the file should still be here if the matrix existed before.

In the end, I think it's better to leave the garbage collector handle this, as it was done before.

sylvlecl (changes requested):
Mostly non blocking comments, but there are some design issues:
- we must take care to not delete files that are actually in use
- we should not write again matrices that are already present
- we must take care of interoperability of produced files (scientific notation?)

antarest/matrixstore/service.py
laurent-laporte-pro: > note that default encoding is latin1, not UTF-8, should not be an issue since we have only numbers here, though.

I know, actually this is platform-dependent. On Windows this is Latin-1, on Linux this is UTF8. In the previous implementation it wasn't specified, so it was UTF8 by default. I prefer to use UTF8, or best UTF8-sig 😁

laurent-laporte-pro: > note that default encoding is latin1, not UTF-8, should not be an issue since we have only numbers here, though.

I know, actually this is platform-dependent. On Windows this is Latin-1, on Linux this is UTF8. In the previous implementation it wasn't specified, so it was UTF8 by default. I prefer to use UTF8, or best UTF8-sig 😁

antarest/matrixstore/service.py
laurent-laporte-pro: You are right, I seen it, but…

Well, I decided to do the following:
- enhance the `MatrixContentRepository` class to support saving NumPy array directly;
- enhance the `ISimpleMatrixService` class and sub-classes to support creating from NumPy array;
- improve the importation function to handle NumPy arrays directly without the necessity of converting them.

antarest/matrixstore/repository.py
laurent-laporte-pro: I use a lock file to ensure exclusive access to the matrix file between multiple processes (or threads)

antarest/matrixstore/service.py
laurent-laporte-pro: Correct the `MatrixService.create` so that the matrix file is not deleted in case of exception.

tests/matrixstore/test_service.py
laurent-laporte-pro: corrected

antarest/matrixstore/service.py
```diff
@@ -324,23 +337,65 @@ def delete_dataset(self, id: str, params: RequestParameters) -> str:
...
+        # used by any study) and deletes them.
+        # So, we can ignore missing database records and/or missing files.
+        # Currently, the deletion is done matrix by matrix (eager deletion
+        # is not used, which is under-performing).
+        # In the case of a unitary deletion, it is preferable to use a transaction
+        # in order to have a rollback in case of failure, and to start with the
+        # database deletion and finish with the file deletion (considered as atomic).
+        with db():
+            self.repo.delete(matrix_id)
+            with contextlib.suppress(FileNotFoundError):
```
sylvlecl, line 397: Note that it would be more "gentle" with the database to not lock the session while we are deleting the file (which does not require the DB).
Probably does not make a significant diff here, but it's the kind of thing we must pay attention to, if we want the application to scale as best as possible.

antarest/matrixstore/service.py
```diff
@@ -375,7 +427,11 @@ def create_matrix_files(
                 mtx = self.get(mid)
                 if not mtx:
                     continue
-                write_tsv_matrix(mtx, tmpdir)
+                name = f"matrix-{mtx.id}.txt"
+                filepath = f"{tmpdir}/{name}"
+                array = np.array(mtx.data, dtype=np.float64)
+                # noinspection PyTypeChecker
+                np.savetxt(filepath, array, delimiter="\t", fmt="%.18f")
```
sylvlecl, line 434: We could have a small utility method so that we don't have to repeat the same arguments every time we write a matrix (and know them :) )

sylvlecl (dismissed):
A few more non-blocking comments that might be handled before merging.

antarest/matrixstore/service.py
laurent-laporte-pro: Sorry, no. I prefer to avoid one-line utility methods, because they're more readable in general.

hdinia (approved):

PR #1477, opened 2023-04-24
fix(api): fix uncaught exceptions in slurm launcher loop

### LINKED ISSUE #1476: Uncaught exceptions cause slurm launcher loop to stop
## Description

Seen in production: 
the loop which monitors studies state in slurm launcher stops, because of an uncaught exception in `_handle_success`

Stack:
```
  File "/antarest/launcher/adapters/slurm_launcher/slurm_launcher.py", line 430, in _handle_success
    output_id = self._import_study_output(
  File "/antarest/launcher/adapters/slurm_launcher/slurm_launcher.py", line 280, in _import_study_output
    return self.callbacks.import_output(
...
```

This leaves the worker in an invalid state where the progress of all tasks is not monitored anymore.

## Expected Behavior

The monitoring loop must never stops, exceptions need to be handled "gracefully" in that thread.

## Environment

- [X] Production
- [ ] Staging

### Add any other context about the problem here.

For an example of such an error in  RTE prod env: see task 721a1167-a2c5-41c7-bed2-b26635ac658e on 17/04 at 21:33.

sylvlecl:
**Description**

Fixes #1476.
In order to avoid the end of monitoring loopt:
- Exceptions are caught in result handling
- As a last resort, all exceptions are caught in the monitoring loop

### DIFF
--- a/antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
+++ b/antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
@@ -148,7 +148,13 @@ class SlurmLauncher(AbstractLauncher):
 
     def _loop(self) -> None:
         while self.check_state:
-            self._check_studies_state()
+            try:
+                self._check_studies_state()
+            except Exception:
+                logger.error(
+                    "An uncaught exception occurred in slurm_launcher loop",
+                    exc_info=True,
+                )
             time.sleep(2)
 
     def start(self) -> None:
@@ -406,7 +412,6 @@ class SlurmLauncher(AbstractLauncher):
                 study.name, JobStatus.FAILED, msg, None
             )
             logger.error(msg, exc_info=e)
-            raise
         else:
             msg = "Simulation failed (even if some output results may be available)"
             self.callbacks.append_after_log(study.name, msg)
@@ -444,7 +449,6 @@ class SlurmLauncher(AbstractLauncher):
                 study.name, JobStatus.FAILED, msg, None
             )
             logger.error(msg, exc_info=e)
-            raise
         else:
             self.callbacks.update_status(
                 study.name, JobStatus.SUCCESS, None, output_id


MartinBelthle (dismissed):

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -444,7 +449,6 @@ def _handle_success(self, study: StudyDTO) -> None:
                 study.name, JobStatus.FAILED, msg, None
             )
             logger.error(msg, exc_info=e)
-            raise
```
laurent-laporte-pro, line 447: Je me souviens avoir ajouté un `raise` explicitement pour voir les erreurs et les remonter au niveau de l’appelant. Si le on supprime, cela signifie que l’erreur est masquée.

Je suis d’accord pour supprimer le `raise` uniquement si j’ai une analyse claire des conséquences. À la suite de cela, il convient de documenter le code à l’aide d’un commentaire explicatif précis sans quoi, lors de la prochaine revue de code, je rajouterais probablement le `raise` 😜

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -406,7 +412,6 @@ def _handle_failure(self, study: StudyDTO) -> None:
...
```
laurent-laporte-pro, line 409: (même commentaire que ci-dessous)

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
sylvlecl: The "catch all" in the loop itself avoids to kill the slurm loop, which caused the progress of computations to not be monitored anymore (we had to restart the app to fix it).

However it's not enough to add that:
the individual result handling here must not raise, or else we will skip the handling of next results, and also the "cleaning" part.

To be honest we should have a unit test to check that behaviour, this really was a quick fix for deployment.

Where do you prefer to add comments? on this line, or on docstring?

PR #1488, opened 2023-04-28
feat(backend): Demonstrate metrics

sylvlecl:
**Description** 

This is a draft for defining metrics and exposing them to a prometheus instance, in order to improve the application observability.

The only metrics that are defined fro now are metrics about HTTP requests (count, duration).

Some technical details:
- metrics are exposed at the /metrics endpoint
- a "worker_id" label is used to enable the distinction between workers
- since our app works with multiple processes (workers), we need to aggregate the workers metrics. For that purpose we use the facility provided by prometheus client, which uses a temporary folder to store individual workers metrics and then aggregate them. This requires the definition of an additional env variable.

In the future, we could add metrics about various things: DB sessions, tasks durations, ...

In the long run, using [opentelemetry](https://opentelemetry.io/docs/instrumentation/python/) could be an alternative.

As a very basic illustration of grah in grafana:
image

### DIFF
--- a/antarest/core/config.py
+++ b/antarest/core/config.py
@@ -621,11 +621,7 @@ class MetricsConfig:
 
     @classmethod
     def from_dict(cls, data: JSON) -> "MetricsConfig":
-        return cls(
-            prometheus=PrometheusConfig.from_dict(data["prometheus"])
-            if "prometheus" in data
-            else None
-        )
+        return cls(prometheus=PrometheusConfig.from_dict(data["prometheus"]) if "prometheus" in data else None)
 
 
 @dataclass(frozen=True)
--- a/antarest/core/config.py
+++ b/antarest/core/config.py
@@ -599,6 +599,36 @@ class ServerConfig:
 
 
 @dataclass(frozen=True)
+class PrometheusConfig:
+    """
+    Sub config object dedicated to prometheus metrics
+    """
+
+    multiprocess: bool = False
+
+    @classmethod
+    def from_dict(cls, data: JSON) -> "PrometheusConfig":
+        return cls(multiprocess=bool(data["multiprocess"]))
+
+
+@dataclass(frozen=True)
+class MetricsConfig:
+    """
+    Sub config object dedicated to metrics
+    """
+
+    prometheus: Optional[PrometheusConfig] = None
+
+    @classmethod
+    def from_dict(cls, data: JSON) -> "MetricsConfig":
+        return cls(
+            prometheus=PrometheusConfig.from_dict(data["prometheus"])
+            if "prometheus" in data
+            else None
+        )
+
+
+@dataclass(frozen=True)
 class Config:
     """
     Root server config
@@ -616,6 +646,7 @@ class Config:
     eventbus: EventBusConfig = EventBusConfig()
     cache: CacheConfig = CacheConfig()
     tasks: TaskConfig = TaskConfig()
+    metrics: MetricsConfig = MetricsConfig()
     root_path: str = ""
     api_prefix: str = ""
     desktop_mode: bool = False
@@ -645,6 +676,7 @@ class Config:
             root_path=data.get("root_path", defaults.root_path),
             api_prefix=data.get("api_prefix", defaults.api_prefix),
             desktop_mode=desktop_mode,
+            metrics=MetricsConfig.from_dict(data["metrics"]) if "metrics" in data else MetricsConfig(),
         )
 
     @classmethod
--- a/antarest/core/exceptions.py
+++ b/antarest/core/exceptions.py
@@ -847,3 +847,10 @@ class OutputAggregationError(HTTPException):
     def __init__(self, output_id: str, message: str) -> None:
         msg = f"Could not aggregate output data for output '{output_id}' : {message}."
         super().__init__(HTTPStatus.UNPROCESSABLE_ENTITY, msg)
+
+
+class ConfigurationError(RuntimeError):
+    """
+    Raised when some configuration is invalid.
+    """
+    pass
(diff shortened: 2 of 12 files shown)


antarest/core/config.py
```diff
@@ -365,6 +365,36 @@ def from_dict(data: JSON) -> "ServerConfig":
...
+class PrometheusConfig:
+    """
+    Sub config object dedicated to prometheus metrics
+    """
+
+    multiprocess: bool = False
+
+    @staticmethod
+    def from_dict(data: JSON) -> "PrometheusConfig":
+        return PrometheusConfig(multiprocess=bool(data["multiprocess"]))
```
laurent-laporte-pro: The pythonic way to implement multiple constructors is to use `@classmethod`. See this article about ["Providing Multiple Constructors in Your Python Classes"](https://realpython.com/python-multiple-constructors/#providing-multiple-constructors-with-classmethod-in-python).

```python
@classmethod
def from_dict(cls, data: JSON) -> "PrometheusConfig":
    return cls(multiprocess=bool(data["multiprocess"]))
```

antarest/core/config.py
```diff
@@ -365,6 +365,36 @@ def from_dict(data: JSON) -> "ServerConfig":
...
+
+    prometheus: Optional[PrometheusConfig] = None
+
+    @staticmethod
+    def from_dict(data: JSON) -> "MetricsConfig":
+        return MetricsConfig(
+            prometheus=PrometheusConfig.from_dict(data["prometheus"])
+            if "prometheus" in data
+            else None
+        )
```
laurent-laporte-pro: use `@classmethod` instead

antarest/core/exceptions.py
```diff
@@ -224,3 +224,12 @@ def __init__(self) -> None:
...
         )
+
+
+class ConfigurationError(Exception):
+    """
+    Raised when some configuration is invalid.
+    """
+
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
```
laurent-laporte-pro, line 726: This method is not really useful unless you add some information, for instance, you can use the message template here (I mean "Environment variable {_PROMETHEUS_MULTIPROCESS_ENV_VAR} must be defined for use of prometheus in a multiprocess environment")

antarest/core/metrics.py
```diff
@@ -0,0 +1,92 @@
...
+from prometheus_client import multiprocess
+from starlette.requests import Request
+
+from antarest.core.config import Config
+from antarest.core.exceptions import ConfigurationError
+
+logger = logging.getLogger(__name__)
+
+
+_PROMETHEUS_MULTIPROCESS_ENV_VAR = "PROMETHEUS_MULTIPROC_DIR"
```
laurent-laporte-pro, line 42: Add a short comment to explain the purpose of this environment variable, for instance:

```python
# The `PROMETHEUS_MULTIPROC_DIR` environment variable is used by
# the Python Prometheus client library to configure process-level metrics
# when running in a multi-process environment.
```

antarest/core/metrics.py
```diff
@@ -0,0 +1,92 @@
...
+    Also registers metrics for HTTP requests.
+    """
+    prometheus_config = config.metrics.prometheus
+    if not prometheus_config:
+        return
+
+    if _PROMETHEUS_MULTIPROCESS_ENV_VAR not in os.environ:
+        raise ConfigurationError(
+            f"Environment variable {_PROMETHEUS_MULTIPROCESS_ENV_VAR} must be defined for use of prometheus in a multiprocess environment"
+        )
```
laurent-laporte-pro: We check that this environment variable exists but it is not used directly in the code (but only in the client). So, a little explanation is needed.
Moreover, we should also check that it is a directory that exists (with the right permissions).

requirements.txt
```diff
@@ -8,7 +8,7 @@ checksumdir~=1.2.0
 click~=8.0.3
 contextvars~=2.4
 fastapi-jwt-auth~=0.5.0
-fastapi[all]~=0.73.0
+fastapi[all]~=0.74.0
```
laurent-laporte-pro: FYI: A version upgrade would be welcome to take advantage of the latest developments that simplify the update of the Swagger API documentation. But, we need to analyse the impacts...

conf/gunicorn.py
```diff
@@ -28,3 +29,17 @@
 errorlog = "-"
 accesslog = "-"
 preload_app = False
+
+
+def post_fork(server, worker):
+    """
+    Put the worker_id into an env variable for further use within the app.
+    """
+    os.environ["APP_WORKER_ID"] = str(worker.age)
```
laurent-laporte-pro: If I had a choice, I would prefer `GUNICORN_WORKER_ID` for that.

Why do you use `worker.age` instead of `worker.id`?

`worker.age` is a number that indicates how many requests the worker process has handled since it was started.

laurent-laporte-pro (changes requested):

requirements.txt
sylvlecl: Ok! This was the minimum upgrade to get the "route" information from the request

conf/gunicorn.py
sylvlecl: From what I understood, `worker.age` is a counter incremented every time gunicorn creates a new worker (for ex. when one dies).
Using the age here allows to have an almost stable set of worker IDs between application runs (after a restart of the app, for example, we sill have IDs in the range [1-8] for 8 workers), to have mostly stable metrics labels.
I don't see any `worker.id`, we have the PID as an alternative, but it's not stable at all between runs.

laurent-laporte-pro:
@sylvlecl : do you know about that : https://autometrics.dev/

PR #1539, opened 2023-05-28
refactor(variant): enhance implementation of variant commands

### LINKED ISSUE #1525: [ANT-706] Add ST-storage commands for variants
## Problem Statement

Short-term storage is a new object similar to thermal clusters (and renewable clusters) that must be considered in the application.

We need to define new commands in the Study Variant Manager that allow the addition and removal of short-term storage.

There is no need to define a new command for updating short-term storage because we can already use the [`update_config`](https://antares-web.readthedocs.io/en/latest/user-guide/2-variant_manager/#command-list) command.

The short-term storage is an object that manages 5 time series. There is no need to develop commands for it because we can use the `replace_matrix` command.

## Proposed Solution

The commands to create are :

- `create_st_storage`: create a short term storage,
  
- `delete_st_storage`: delete a short term storage.

## Developer Notes

- The example of thermal clusters should be taken.

laurent-laporte-pro:
The aim of this PR is to refactor the code before implementing the news commands for short-term storage.

### DIFF
--- a/antarest/study/common/studystorage.py
+++ b/antarest/study/common/studystorage.py
@@ -248,7 +248,7 @@ class IStudyStorageService(ABC, Generic[T]):
     def export_study_flat(
         self,
         metadata: T,
-        dest: Path,
+        dst_path: Path,
         outputs: bool = True,
         output_list_filter: Optional[List[str]] = None,
         denormalize: bool = True,
@@ -257,13 +257,11 @@ class IStudyStorageService(ABC, Generic[T]):
         Export study to destination
 
         Args:
-            metadata: study
-            dest: destination path
-            outputs: list of outputs to keep
-            output_list_filter: list of outputs to keep (None indicate all outputs)
-            denormalize: denormalize the study (replace matrix links by real matrices)
-        Returns: None
-
+            metadata: study.
+            dst_path: destination path.
+            outputs: list of outputs to keep.
+            output_list_filter: list of outputs to keep (None indicate all outputs).
+            denormalize: denormalize the study (replace matrix links by real matrices).
         """
         raise NotImplementedError()
 
--- a/antarest/study/storage/variantstudy/command_factory.py
+++ b/antarest/study/storage/variantstudy/command_factory.py
@@ -1,7 +1,6 @@
 from typing import List
 
 from antarest.core.model import JSON
-from antarest.core.utils.utils import assert_this
 from antarest.matrixstore.service import ISimpleMatrixService
 from antarest.study.storage.patch_service import PatchService
 from antarest.study.storage.variantstudy.business.matrix_constants_generator import (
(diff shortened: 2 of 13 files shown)


antarest/study/storage/variantstudy/command_factory.py
```diff
@@ -96,134 +93,72 @@ def __init__(
...
-        elif action == CommandName.UPDATE_SCENARIO_BUILDER.value:
-            return UpdateScenarioBuilder(
-                **args, command_context=self.command_context
-            )
-        raise NotImplementedError()
-
-    def to_icommand(self, command_dto: CommandDTO) -> List[ICommand]:
+    def _to_single_command(self, action: str, args: JSON) -> ICommand:
+        """Convert a single CommandDTO to ICommand."""
+        command_mapping = {
```
hdinia, line 98: `action_map`

antarest/study/common/studystorage.py
```diff
@@ -248,7 +248,7 @@ def export_output(self, metadata: T, output_id: str, target: Path) -> None:
     def export_study_flat(
         self,
         metadata: T,
-        dest: Path,
+        dst_path: Path,
```
hdinia, line 251: I would have used a more descriptive name here like `target_path`, "dst" can be confusing 
note: same for other abbreviations elsewhere

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -852,34 +855,36 @@ def _generate(
...
-                        "Existing unamanaged user config. It will be overwritten."
+                        "Existing unmanaged user config. It will be overwritten."
                     )
-                    shutil.rmtree((dest_path / "user"))
-                shutil.copytree(unmanaged_user_config, dest_path / "user")
+                    shutil.rmtree((dst_path / "user"))
+                shutil.copytree(unmanaged_user_config, dst_path / "user")
             else:
                 logger.warning("Destination snapshot doesn't exist !")
             shutil.rmtree(unmanaged_user_config, ignore_errors=True)
```
hdinia, line 874:
```python
if not unmanaged_config:
    return

logger.info("Restoring previous unmanaged user configuration")

if not dest_path.exists():
    logger.warning("Destination snapshot doesn't exist!")
    return

user_config_path = dest_path / "user"
if user_config_path.exists():
    logger.warning("Existing unmanaged user configuration will be overwritten.")
    shutil.rmtree(user_config_path)

shutil.copytree(unmanaged_config, user_config_path)
shutil.rmtree(unmanaged_config, ignore_errors=True)
```
here I've used early returns to reduce the level of nesting, also created a `user_path` variable to avoid repetition. + `dst_path` > `dest_path`

hdinia (approved):

antarest/study/storage/variantstudy/variant_study_service.py
laurent-laporte-pro: Intéressant. Cela dit, je n'ai fait que renommer les variables. Je garde ça pour une prochaine fois.

PR #1568, opened 2023-06-07
fix(desktop): correct date parsing in localized environment

### LINKED ISSUE #1524: Standalone/desktop app can't be started
## Description
```bash
omnesflo@gm0winl370:~/Téléchargements/AntaresWeb-ubuntu-2.14.0$ ./AntaresWeb/AntaresWebServer 
plyer/platforms/linux/notification.py:65: UserWarning: The Python dbus package is not installed.
Try installing it with your distribution's package manager, it is usually called python-dbus or python3-dbus, but you might have to try dbus-python instead, e.g. when using pip.
libpng warning: iCCP: CRC error
Process Process-1:
  File "logging/config.py", line 542, in configure
  File "logging/config.py", line 674, in configure_formatter
  File "logging/config.py", line 89, in _resolve
ModuleNotFoundError: No module named 'pythonjsonlogger'

The above exception was the direct cause of the following exception:

  File "multiprocessing/process.py", line 315, in _bootstrap
  File "multiprocessing/process.py", line 108, in run
  File "antarest/gui.py", line 21, in run_server
  File "antarest/main.py", line 185, in fastapi_app
  File "antarest/core/logging/utils.py", line 160, in configure_logger
  File "logging/config.py", line 808, in dictConfig
  File "logging/config.py", line 545, in configure
ValueError: Unable to configure formatter 'json'
^CTraceback (most recent call last):
  File "antarest/gui.py", line 95, in <module>
  File "antarest/gui.py", line 89, in main
KeyboardInterrupt
[112379] Failed to execute script 'gui' due to unhandled exception!
```

1. Download https://github.com/AntaresSimulatorTeam/AntaREST/releases/download/v2.14.0/AntaresWeb-ubuntu-2.14.0.zip
2. Uncompress
3. Make the executable executable (chmod +x)
4. See error above

## Possible Solution
I tried to install pythonjsonlogger, but it doesn't fix the issue. Probably the environments are different.

Add [pythonjsonlogger](https://pypi.org/project/python-json-logger/) to `requirements.txt` ?

Looking at the diff since v2.13.0 (which was working), it seems that this bug is related to this PR "feat: standardize log messsage formats (#1431)"

```diff
@@ -41,11 +38,29 @@ def configure_logger(config: Config) -> None:
         "formatters": {
             "console": {
                 "class": "antarest.core.logging.utils.CustomDefaultFormatter",
-                "format": "%(asctime)s - %(trace_id)s - %(threadName)s - %(name)s - %(ip)s - %(user)s - %(pid)s - %(levelname)s - %(message)s",
+                "format": (
+                    "[%(asctime)s] [%(process)s] [%(name)s]"
+                    " - %(trace_id)s"
+                    " - %(threadName)s"
+                    " - %(ip)s"
+                    " - %(user)s"
+                    " - %(levelname)s"
+                    " - %(message)s"
+                ),
             },
             "json": {
-                "class": f"{JsonFormatter.__module__}.{JsonFormatter.__name__}",
-                "format": "%(asctime)s - %(trace_id)s - %(threadName)s - %(name)s - %(ip)s - %(user)s - %(pid)s - %(levelname)s - %(message)s",
+                "class": "pythonjsonlogger.jsonlogger.JsonFormatter",
+                "format": (
```

## Environment

- [X] Production (desktop)
- [x] Staging
### Add any other context about the problem here.

- If applicable, provide the study ID that this issue relates to:
- Any relevant logs or error messages: [e.g. console outputs, stack traces]

- The date and time when the issue occurred: 17/05/2023
- Browser and version: Firefox
- Operating System: Ubuntu 20.04

laurent-laporte-pro:
Date parsing with `strptime(starting_month, "%B")` and date formatting with `strftime("%A")` depend on the locale, but the "FR" locale is used on the server.

We need to use locale-independent functions, as the configuration is in English.

### DIFF
--- a/antarest/study/storage/utils.py
+++ b/antarest/study/storage/utils.py
@@ -265,7 +265,7 @@ def assert_permission(
 
 MATRIX_INPUT_DAYS_COUNT = 365
 
-MONTHS = [
+MONTHS = (
     "January",
     "February",
     "March",
@@ -278,9 +278,9 @@ MONTHS = [
     "October",
     "November",
     "December",
-]
+)
 
-DAY_NAMES = [
+DAY_NAMES = (
     "Monday",
     "Tuesday",
     "Wednesday",
@@ -288,7 +288,7 @@ DAY_NAMES = [
     "Friday",
     "Saturday",
     "Sunday",
-]
+)
 
 
 def get_start_date(
--- a/antarest/study/storage/utils.py
+++ b/antarest/study/storage/utils.py
@@ -265,6 +265,31 @@ def assert_permission(
 
 MATRIX_INPUT_DAYS_COUNT = 365
 
+ENGLISH_MONTHS = [
+    "January",
+    "February",
+    "March",
+    "April",
+    "May",
+    "June",
+    "July",
+    "August",
+    "September",
+    "October",
+    "November",
+    "December",
+]
+
+ENGLISH_DAY_NAMES = [
+    "Monday",
+    "Tuesday",
+    "Wednesday",
+    "Thursday",
+    "Friday",
+    "Saturday",
+    "Sunday",
+]
+
 
 def get_start_date(
     file_study: FileStudy,
@@ -273,6 +298,7 @@ def get_start_date(
 ) -> MatrixIndex:
     """
     Retrieve the index (start date and step count) for output or input matrices
+
     Args:
         file_study: Study data
         output_id: id of the output, if None, then it's the start date of the input matrices
@@ -287,28 +313,27 @@ def get_start_date(
     start_offset = cast(int, config.get("simulation.start"))
     end = cast(int, config.get("simulation.end"))
 
-    starting_month_index = strptime(starting_month, "%B").tm_mon
+    starting_month_index = ENGLISH_MONTHS.index(starting_month.title()) + 1
     target_year = 2000
     while True:
         if leapyear == calendar.isleap(target_year):
             first_day = datetime(target_year, starting_month_index, 1)
-            if first_day.strftime("%A") == starting_day:
+            first_day_name = ENGLISH_DAY_NAMES[first_day.weekday()]
+            if first_day_name == starting_day.title():
                 break
         target_year += 1
 
-    start_offset_days = (
-        timedelta(days=start_offset - 1)
-        if output_id is not None
-        else timedelta(days=0)
+    start_offset_days = timedelta(
+        days=(0 if output_id is None else start_offset - 1)
     )
     start_date = (
         datetime(target_year, starting_month_index, 1) + start_offset_days
     )
     # base case is DAILY
     steps = (
-        end - start_offset + 1
-        if output_id is not None
-        else MATRIX_INPUT_DAYS_COUNT
+        MATRIX_INPUT_DAYS_COUNT
+        if output_id is None
+        else end - start_offset + 1
     )
     if level == StudyDownloadLevelDTO.HOURLY:
         steps = steps * 24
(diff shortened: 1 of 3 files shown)


antarest/study/storage/utils.py
```diff
@@ -265,6 +265,31 @@ def assert_permission(
 
 MATRIX_INPUT_DAYS_COUNT = 365
 
+ENGLISH_MONTHS = [
```
sylvlecl: We could remove "English", it's quite obvious: simply MONTH_NAMES ?

antarest/study/storage/utils.py
```diff
@@ -287,28 +313,27 @@ def get_start_date(
     start_offset = cast(int, config.get("simulation.start"))
     end = cast(int, config.get("simulation.end"))
 
-    starting_month_index = strptime(starting_month, "%B").tm_mon
+    starting_month_index = ENGLISH_MONTHS.index(starting_month.title()) + 1
```
sylvlecl: Should we have an explicit dict for months indexes, instead of relying on linear search and +1?
(or a Month class?)

antarest/study/storage/utils.py
```diff
@@ -265,6 +265,31 @@ def assert_permission(
...
+    "June",
+    "July",
+    "August",
+    "September",
+    "October",
+    "November",
+    "December",
+]
+
+ENGLISH_DAY_NAMES = [
```
sylvlecl: Same about "english"

antarest/study/storage/utils.py
```diff
@@ -287,28 +313,27 @@ def get_start_date(
...
 
-    starting_month_index = strptime(starting_month, "%B").tm_mon
+    starting_month_index = ENGLISH_MONTHS.index(starting_month.title()) + 1
     target_year = 2000
     while True:
         if leapyear == calendar.isleap(target_year):
             first_day = datetime(target_year, starting_month_index, 1)
-            if first_day.strftime("%A") == starting_day:
+            first_day_name = ENGLISH_DAY_NAMES[first_day.weekday()]
+            if first_day_name == starting_day.title():
```
sylvlecl: It seems to me it would be clearer to convert `starting_day` to a weekday integer instead:
we can then simply have the comparison:
```python
if first_day.weekday() == starting_day
```

antarest/study/storage/utils.py
```diff
@@ -325,12 +350,11 @@ def get_start_date(
...
 
     first_week_offset = 0
-    while True:
-        if (start_date + timedelta(days=first_week_offset)).strftime(
-            "%A"
-        ) == first_week_day:
+    for first_week_offset in range(7):
+        first_day = start_date + timedelta(days=first_week_offset)
+        first_day_name = ENGLISH_DAY_NAMES[first_day.weekday()]
+        if first_day_name == first_week_day.title():
```
sylvlecl: Same as above about weekday comparison: would be clearer based on integer representation, in my opinion.
Actually here we could just substract the 2 weekdays to get the offset, I guess,with an additional `% 7`. Not sure it's clearer though. We could have a small named method to clarify.

antarest/study/storage/utils.py
sylvlecl: Note: I was wondering if antares simulator was based on locale or not.

It is not, it has the names hard coded like here, so this is the right solution.
Simulator reads months from case-insensitive month names.

sylvlecl (approved):
Ok for me, some minor suggestions for code clarity though.

sylvlecl:
Note for later work: those calendar operations should be isolated in a dedicated module, and tested separately.

antarest/study/storage/utils.py
laurent-laporte-pro: I remove "English"

antarest/study/storage/utils.py
laurent-laporte-pro: The algorithm is based on searching a mathing day name (not day offset).
What do you mean? Any suggestion?

antarest/study/storage/utils.py
laurent-laporte-pro: There is no performance issue for a list of size 12 (or 7 with days).
To me the code is clear enough. The difference : we may have `IndexError` instead of `KeyError`.

PR #1577, opened 2023-06-09
feature(api): add support for Microsoft Office documents in API

### LINKED ISSUE #1573: Add Support for Microsoft Office documents in Antares Web API
## Problem Statement

The Script team uses tools to generate studies in order to run simulations from Antares Web. Additional Microsoft Office Excel files are stored in the User directory. For example, an Excel sheet named "synthesys_windSolar.xlsx" is generated in the "/user/wind_solar" directory.

In order to standardize the development done in the R language, the Script team wants to use the Antares Web API instead of directly reading and writing files on the disk.

The current version of Antares Web allows reading and updating files using GET, POST, and PUT requests by specifying a path to the desired file. GET requests can also list the contents of directories, thus indicating the existence of files.

However, in the case of reading, only files in JSON format can be retrieved. In other words, it is not possible to read Microsoft Office files. Currently, the GET request `/v1/studies/{uuid}/raw` only provides JSON files.

This request proposes adding support for Microsoft Office formats when reading and writing documents in Antares Web.

Example URL: `/v1/studies/9cbd21b4-f9af-416f-8076-88173b1858b3/raw?path=%2Fuser%2Fwind_solar&depth=3&formatted=true`

image

## Proposed Solution

Detect the file's MIME type based on the path in the `raw_studies_blueprint.get_study` function.

The workaround is to use the JSON format instead of Excel.

laurent-laporte-pro:


laurent-laporte-pro:
Reprendre le message de commit pour le commit de PR : "fix(api): study data retrieval now considers document types, enabling retrieval of Microsoft Office documents. Binary data support has also been enhanced".

tests/integration/raw_studies_blueprint/assets/user/folder/example.txt
```diff
@@ -0,0 +1,24 @@
+Dans la douce clarté du matin enchanté,
```
MartinBelthle, line 1: We should probably write examples in English don't you think ? This also applies to other examples in the same folder

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -59,26 +104,42 @@ def get_study(
...
-            separators=(",", ":"),
-        ).encode("utf-8")
-        return Response(content=json_response, media_type="application/json")
+            resource_path = pathlib.Path(path)
+            suffix = resource_path.suffix.lower()
+            content_type, encoding = CONTENT_TYPES.get(suffix, (None, None))
+            if content_type == "application/json":
+                # Use `JSONResponse` to ensure to return a valid JSON response
+                # that checks `NaN` and `Infinity` values.
+                output = json.loads(output)
```
MartinBelthle: To be sure I've understood, if there's a NaN in the JSON, the back end used to allow it and the check was done on the front end. But with this development, the front end will always receive a valid JSON or an exception. But is this exception caught in the front end ?

tests/integration/raw_studies_blueprint/assets/user/folder/example.txt
laurent-laporte-pro: No, I would like to check accentuated characters.

antarest/study/web/raw_studies_blueprint.py
laurent-laporte-pro: In case of error, we get a `ValueError`, for instance:

```python
>>> import json
>>> import math
>>> data = {"foo": math.nan}
>>> json.dumps(data, allow_nan=False)
  ...
ValueError: Out of range float values are not JSON compliant
```

This exception will raise a HTTP 500 error. But you're right, it might be better to raise a 422 error to indicate that the configuration data is invalid.

MartinBelthle: I let you do that and i will approve + merge the PR when done &#x1F642;

MartinBelthle (approved):

PR #1630, opened 2023-06-28
feat(commands): add ST-Storage commands

### LINKED ISSUE #1629: [ANT-706] Add ST-storage (variant) commands
## Problem statement

The purpose of this request is to add the `CreateSTStorage` and `RemoveSTStorage` commands for creating and deleting short-term storage. These commands are needed to implement the entry points for this new object (#1522). They are used both for managing a study variant and for a RAW study.

## Proposed solution

Create a new `CreateSTStorage` class which inherits from the `ICommand` class.

The attributes are :
- area_id: the unique identifier of the area containing this new short-term storage,
- storage_name: the name of the short-term storage,
- parameters: short-term storage configuration parameters (INI file),
- pmax_injection: the Pmax injection matrix (time series),
- pmax_withdrawal: Pmax withdrawal matrix (time series),
- lower_rule_curve: the min curve matrix (time series),
- upper_rule_curve: max curve matrix (time series),
- inflows: the inflows matrix (time series).

Create a new `RemoveSTStorage` class which inherits from the `ICommand` class.

Attributes are :
- area_id: the unique identifier of the area containing this new short-term storage,
- storage_id: the unique identifier of the short-term storage to be removed.

laurent-laporte-pro:
L'objectif de cette PR est d'ajouter les commandes `CreateSTStorage` et `RemoveSTStorage` nécessaires à la gestion des stockages court terme.

La gestion des matrices a été modifié pour ajouter les séries temporelles nécessaires aux stockages court terme.

L'implémentation de ces commandes nécessite un refactoring d'autres commandes comme la creation de zones, les cluster thermiques et renouvelable. Le refactoring vise principalement à améliorer l'implémentation ou le style de programmation afin d'en réduire la complexité (surconception).

Une confusion a été repérée entre les noms et les IDs des clusters thermiques et renouvelable. Ces anomalies sont signalées avec des `# fixme` facilement repérables dans le code. Elles ne sont pas corrigées, cela ferra l'objet d'une issue/PR.

Cette PR met aussi à jour la liste des commandes dans la documentation. À ce niveau, il y a encore du travail à faire, mais cela ferra l'objet d'une autre User Story.

On a aussi ajouté de la documentation dans le code source (doccstrings) afin d'améliorer la maintenabilité et la compréhension du code pour les développeurs.

### DIFF
--- a/antarest/study/storage/variantstudy/model/command/create_st_storage.py
+++ b/antarest/study/storage/variantstudy/model/command/create_st_storage.py
@@ -1,0 +1,482 @@
+import json
+from typing import Any, Dict, List, Optional, Tuple, Union, cast
+
+import numpy as np
+from antarest.core.model import JSON
+from antarest.matrixstore.model import MatrixData
+from antarest.study.business.enum_ignore_case import EnumIgnoreCase
+from antarest.study.storage.rawstudy.model.filesystem.config.model import (
+    Area,
+    FileStudyTreeConfig,
+    STStorage,
+    transform_name_to_id,
+)
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.variantstudy.business.matrix_constants_generator import (
+    GeneratorMatrixConstants,
+)
+from antarest.study.storage.variantstudy.business.utils import (
+    strip_matrix_protocol,
+    validate_matrix,
+)
+from antarest.study.storage.variantstudy.model.command.common import (
+    CommandName,
+    CommandOutput,
+)
+from antarest.study.storage.variantstudy.model.command.icommand import (
+    MATCH_SIGNATURE_SEPARATOR,
+    ICommand,
+)
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+from pydantic import BaseModel, Extra, Field, root_validator, validator
+from pydantic.fields import ModelField
+
+# minimum required version.
+REQUIRED_VERSION = 860
+
+MatrixType = List[List[MatrixData]]
+
+
+class STStorageGroup(EnumIgnoreCase):
+    """
+    This class defines the specific energy storage systems.
+
+    Enum values:
+        PSP_OPEN: Represents an open pumped storage plant.
+        PSP_CLOSED: Represents a closed pumped storage plant.
+        PONDAGE: Represents a pondage storage system (reservoir storage system).
+        BATTERY: Represents a battery storage system.
+        OTHER: Represents other energy storage systems.
+    """
+
+    PSP_OPEN = "PSP_open"
+    PSP_CLOSED = "PSP_closed"
+    PONDAGE = "Pondage"
+    BATTERY = "Battery"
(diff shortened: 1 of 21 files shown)


laurent-laporte-pro:
I've just switched my feature branch to the development branch to take into account the update regarding the use of Antares Solver 8.6.1.

antarest/study/storage/variantstudy/model/command/create_st_storage.py
```diff
@@ -0,0 +1,381 @@
...
+        """The normalized version of the storage's name used as the ID."""
+        return self.parameters.id
+
+    @property
+    def storage_name(self) -> str:
+        """The label representing the name of the storage for the user."""
+        return self.parameters.name
+
+    @validator(*_MATRIX_NAMES, always=True)
+    def register_matrix(
```
sylvlecl, line 101: I think it's a design issue that storing the matrix in the matrix service is part of the validation process.

It makes this process hidden inside the validation, which should be more focused and short-duration.

In particular, if for any reason we encounter an exception while writing the matrix to store, there is chances that the user receives a 4** error, whereas it's not an input problem.

Also, this mechanism does not allow us to type the model as strongly as possible:
from what I understand, after validation, we will only have matrix IDs, not matrix data, which does not appear into the model typing.

antarest/study/storage/variantstudy/model/command/create_st_storage.py
```diff
@@ -0,0 +1,381 @@
...
+    # =====================
+
+    command_name = CommandName.CREATE_ST_STORAGE
+    version = 1
+
+    # Command parameters
+    # ==================
+
+    area_id: str = Field(description="Area ID", regex=r"[a-z0-9_(),& -]+")
+    parameters: STStorageConfig
```
sylvlecl, line 68: Although it's important to introduce a stronger model than in existing commands, it will be a design issue in the long run to use the same model all the way from file storage to API.

We will need to separate layers at some point. We should probably have at least 2 layers, maybe 3:
- one for the file format
- one for the business model (in-memory representation)
- possibly one more for the API (DTOs)

This separation will:
- be needed when we want to have another storage strategy than the current file format (database, other format ...).
- allow us to have a better user facing model, without being constrained by the current file format.

sylvlecl (approved):
Just raising some design question for future improvements.

antarest/study/storage/variantstudy/model/command/create_st_storage.py
laurent-laporte-pro: You are right. Matrices are optionals.

antarest/study/storage/variantstudy/model/command/create_st_storage.py
laurent-laporte-pro: The separation between the file format and the business model is a hard work, but we (with Florian) have some good ideas for the future.

An effort is done here to separate the File format and the DTO. But again, this is a work in progress…

PR #1647, opened 2023-07-13
refactor: refactoring of unit tests and Model improvement

laurent-laporte-pro:
This pull request proposes several changes aimed at improving code quality and readability.

- **Unit Test Correction (`test_exporter.py`):** I simplified the unit tests to make the code more concise and easier to maintain.

- **Conftest.py Enhancements:** I performed refactorings and added type annotations to improve code readability.
  Additionally, I updated these files to enhance the calculation of the project's root path.

- **Model Modifications (`tasks`):** I made changes to the existing PyDantic models to prevent inadvertent attribute creation.

- **Service Refactoring:** I made a minor modification to improve code readability in the relevant service.

Goal: improve code quality and maintainability.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Any, Callable, Dict, List, cast
 
 import numpy as np
+import numpy.typing as npt
 import pytest
 from antarest.core.model import SUB_JSON
 from antarest.core.utils.fastapi_sqlalchemy import DBSessionMiddleware, db
@@ -22,9 +23,9 @@ def project_path() -> Path:
     return PROJECT_DIR
 
 
-def with_db_context(f):
+def with_db_context(f: Callable[..., Any]) -> Callable[..., Any]:
     @wraps(f)
-    def wrapper(*args, **kwargs):
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
         engine = create_engine("sqlite:///:memory:", echo=False)
         Base.metadata.create_all(engine)
         # noinspection SpellCheckingInspection
@@ -39,7 +40,7 @@ def with_db_context(f):
     return wrapper
 
 
-def _assert_dict(a: dict, b: dict) -> None:
+def _assert_dict(a: Dict[str, Any], b: Dict[str, Any]) -> None:
     if a.keys() != b.keys():
         raise AssertionError(
             f"study level has not the same keys {a.keys()} != {b.keys()}"
@@ -48,7 +49,7 @@ def _assert_dict(a: dict, b: dict) -> None:
         assert_study(v, b[k])
 
 
-def _assert_list(a: list, b: list) -> None:
+def _assert_list(a: List[Any], b: List[Any]) -> None:
     for i, j in zip(a, b):
         assert_study(i, j)
 
@@ -65,7 +66,10 @@ def _assert_others(a: Any, b: Any) -> None:
         raise AssertionError(f"element in study not the same {a} != {b}")
 
 
-def _assert_array(a: np.ndarray, b: np.ndarray) -> None:
+def _assert_array(
+    a: npt.NDArray[np.float64],
+    b: npt.NDArray[np.float64],
+) -> None:
     if not (a == b).all():
         raise AssertionError(f"element in study not the same {a} != {b}")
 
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -1,31 +1,31 @@
...
-
 from antarest.core.model import SUB_JSON
 from antarest.core.utils.fastapi_sqlalchemy import DBSessionMiddleware, db
 from antarest.dbmodel import Base
+from sqlalchemy import create_engine  # type: ignore
 
-project_dir: Path = Path(__file__).parent.parent
-sys.path.insert(0, str(project_dir))
+# fmt: off
+HERE = Path(__file__).parent.resolve()
```
MartinBelthle, line 16: Not sure why we change the old behavior, i find the new way of setting the PROJECT_DIR more complicated.

tests/conftest.py
```diff
@@ -66,7 +66,10 @@ def _assert_others(a: Any, b: Any) -> None:
         raise AssertionError(f"element in study not the same {a} != {b}")
 
 
-def _assert_array(a: np.ndarray, b: np.ndarray) -> None:
+def _assert_array(
+    a: npt.NDArray[np.float64],
```
MartinBelthle, line 70: When calling _assert_array line 92, we do this : ` isinstance(a, np.ndarray) ` so i do not understand why we expect this type `npt.NDArray[np.float64]`. We should probably harmonize the 2.

tests/storage/conftest.py
```diff
@@ -313,7 +294,7 @@ def add_task(
         custom_event_messages: Optional[CustomTaskEventMessages],
         request_params: RequestParameters,
     ) -> str:
-        action(self._create_notifier())
+        action(lambda message: None)
```
MartinBelthle, line 297: I understand why you removed the useless method but i fail to understand what this line achieved in the first place

tests/storage/integration/test_exporter.py
```diff
@@ -82,72 +79,64 @@ def assert_url_content(
...
-    [
-        (True, None, True),
-        (True, [], False),
-        (True, ["20201014-1427eco"], False),
-        (False, ["20201014-1427eco"], False),
-    ],
-)
+@pytest.mark.parametrize("outputs", [True, False, "prout"])
+@pytest.mark.parametrize("output_list", [None, [], ["20201014-1427eco"]])
+@pytest.mark.parametrize("denormalize", [True, False])
```
MartinBelthle, line 104: Why did you remove some cases ? The case `(True, ["20231014-1427-eco"], False)` seemed relevant to me

tests/conftest.py
laurent-laporte-pro: The method used to calculate the project's root directory is more intricate but offers greater reliability in the event of refactoring. For example, if you move the test file to a different directory or if a developer copies and pastes this code to another location, no modifications will be required. This approach is highly recommended for improved maintainability.

tests/conftest.py
laurent-laporte-pro: The change made in the function signature involves specifying the type of a parameter that expects a NumPy array.

`npt` is an abbreviation for the NumPy typing module, which provides enhanced type hints specifically for NumPy arrays. `npt.NDArray` represents a generic NumPy array, and `[np.float64]` specifies the element type of the array.

tests/storage/conftest.py
laurent-laporte-pro: Actually, `_create_notifier` is a function which returns a function that does nothing. This is the same as `lambda message: None` above. So, no need to use boilerplates, here.

tests/storage/integration/test_exporter.py
laurent-laporte-pro: Actually, I haven't removed the test cases, but rather added some. When you use several `@pytest.mark.parametrize` decorators in a row, they combine (it's a Cartesian product of each list with the others). All in all, we have 3 × 3 × 2 = 18 tests!

tests/conftest.py
MartinBelthle: I understand but isn't it weird that we can enter the function only checking if the parameter is a `np.ndarray` and not checking if `isinstance(a, npt.NDArray[np.float64])` ?

tests/storage/conftest.py
MartinBelthle: So why don't we delete this line `action(lambda message: None)` as the lambda function does nothing ?

tests/storage/integration/test_exporter.py
MartinBelthle: Oh okay i didn't know that they combined, nice  &#x1F603;

MartinBelthle (approved):

PR #1660, opened 2023-07-20
Fix login submit issue

skamril:
Impossible to click on the submit button after an error of login (the button remains disabled).

### DIFF
--- a/webapp/src/components/common/Form/index.tsx
+++ b/webapp/src/components/common/Form/index.tsx
@@ -30,9 +30,15 @@ import clsx from "clsx";
 import { LoadingButton, LoadingButtonProps } from "@mui/lab";
 import UndoIcon from "@mui/icons-material/Undo";
 import RedoIcon from "@mui/icons-material/Redo";
+import axios from "axios";
 import useEnqueueErrorSnackbar from "../../../hooks/useEnqueueErrorSnackbar";
 import useDebounce from "../../../hooks/useDebounce";
-import { getDirtyValues, stringToPath, toAutoSubmitConfig } from "./utils";
+import {
+  ROOT_ERROR_KEY,
+  getDirtyValues,
+  stringToPath,
+  toAutoSubmitConfig,
+} from "./utils";
 import useDebouncedState from "../../../hooks/useDebouncedState";
 import usePrompt from "../../../hooks/usePrompt";
 import { mergeSxProp } from "../../../utils/muiUtils";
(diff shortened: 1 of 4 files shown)


webapp/src/components/common/Form/index.tsx
```diff
@@ -304,8 +328,14 @@ function Form<TFieldValues extends FieldValues, TContext>(
           <FormProvider {...formApiPlus}>{children}</FormProvider>
         )}
       </FormContext.Provider>
+      {rootError && (
+        <Box color="error.main" sx={{ fontSize: "0.9rem", mb: 2 }}>
+          {rootError.type && `${rootError.type}: `}
```
hdinia: in my opinion it's useless to display the error type here, unless requested by users, as the error notification displays it already

webapp/src/components/common/Form/index.tsx
```diff
@@ -314,7 +344,13 @@ function Form<TFieldValues extends FieldValues, TContext>(
                 disabled={!isSubmitAllowed}
                 loading={isSubmitting}
                 loadingPosition="start"
-                startIcon={<SaveIcon />}
+                startIcon={
+                  RA.isNotUndefined(submitButtonIcon) ? (
+                    submitButtonIcon
+                  ) : (
```
hdinia, line 345: the `startIcon` prop is expected to be a React node or undefined, no ? 

```javascript
startIcon={
  submitButtonIcon || <SaveIcon />
}

```

webapp/src/components/wrappers/LoginWrapper.tsx
```diff
@@ -62,18 +60,8 @@ function LoginWrapper(props: Props) {
...
-    try {
-      await dispatch(login(values)).unwrap();
-    } catch (err) {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      setLoginError((err as any).data?.message || t("login.error"));
-      throw err;
-    }
+  const handleSubmit = (data: SubmitHandlerPlus<FormValues>) => {
+    return dispatch(login(data.values)).unwrap();
   };
```
hdinia:
```javascript
  const handleSubmit = ({ values }: SubmitHandlerPlus<FormValues>) => {
    return dispatch(login(values)).unwrap();
  };
```

webapp/src/components/common/Form/index.tsx
```diff
@@ -123,14 +131,17 @@ function Form<TFieldValues extends FieldValues, TContext>(
...
   // * /!\ `formState` is a proxy
-  const { isSubmitting, isSubmitSuccessful, isDirty, dirtyFields } = formState;
+  const { isSubmitting, isSubmitSuccessful, isDirty, dirtyFields, errors } =
+    formState;
   // Don't add `isValid` because we need to trigger fields validation.
   // In case we have invalid default value for example.
   const isSubmitAllowed = isDirty && !isSubmitting;
   const showSubmitButton = !hideSubmitButton && !autoSubmitConfig.enable;
   const showFooter = showSubmitButton || enableUndoRedo;
+  const rootError = errors.root?.[ROOT_ERROR_KEY];
```
hdinia, line 144: maybe you can explain a bit more about the usage of `default` key. If I understand correctly, this allows errors to be persisted and keep any errors that have not yet been corrected across submits?

hdinia (dismissed):

webapp/src/components/common/Form/index.tsx
skamril: No because `submitButtonIcon` is optional, so if it's not defined, its value will be `undefined` and we display the default icon. To hide the icon `null` must be used.

webapp/src/components/common/Form/index.tsx
skamril: I will go into more detail

hdinia (approved):

PR #1668, opened 2023-07-25
feat(watcher): scanning a managed workspace only logs a warning msg

### LINKED ISSUE #1667: Users shouldn't be able to scan a managed workspace in the front-end
## Description

Scanning a workspace allows AntaresWeb to retrieve studies that were not already in its database
So it does not make sense to scan any managed folder.

However, it is possible using the front-end (little "target" icon at the very right of the screenshot below):
scanner_default

It is also possible to do so using the API.
If this happens, the back-end currently raises an Exception and the task fails.

**2 things need to be implemented:**
- The back office should not raise an Exception but simply log the fact that this doesn't make sense
-  The front office should not display the little "target" icon when the folder to scan is a child of "default" workspace.

The back-end part is quite easy and already implemented in the linked PR, so it solves the API issue.

The front-office however could be more tricky as every information is gathered with the endpoint GET `/studies` when the user clicks on the Studies page and no other API call is raised when the user clicks on any folder.

MartinBelthle:


hdinia (changes requested):
As seen with @MartinBelthle I'll work on the front side for this issue

tests/integration/test_integration_watcher.py
```diff
@@ -1,20 +1,32 @@
 from fastapi import FastAPI
 from starlette.testclient import TestClient
 
+from antarest.core.tasks.model import TaskStatus
```
laurent-laporte-pro: Can you follow the [Integration Testing Guide with Pytest](https://github.com/AntaresSimulatorTeam/AntaREST/discussions/1712) to reorganize your test scenarios.

tests/integration/test_integration_watcher.py
```diff
@@ -1,20 +1,32 @@
...
 
-def test_integration_xpansion(app: FastAPI, tmp_path: str):
-    client = TestClient(app, raise_server_exceptions=False)
+
+def test_integration_watcher(app: FastAPI, tmp_path: str):
```
laurent-laporte-pro: The `tmp_path` parameter is not used.

tests/integration/test_integration_watcher.py
```diff
@@ -1,20 +1,32 @@
...
+    client = TestClient(app)
     res = client.post(
```
laurent-laporte-pro: You can consider using the `admin_access_token` fixture to retrive admin credentials.

tests/integration/test_integration_watcher.py
```diff
@@ -1,20 +1,32 @@
...
-        f"/v1/watcher/_scan?path=/tmp",
+    res = client.post("/v1/watcher/_scan", headers=headers)
+    assert res.status_code == 422
+    assert res.json() == {
+        "description": "field required",
+        "exception": "RequestValidationError",
+        "body": None,
+    }
+
+    task_id = client.post(
```
laurent-laporte-pro: Please, can you add a nominal cas on another workspace?

tests/integration/test_integration_watcher.py
```diff
@@ -1,20 +1,32 @@
...
+        "/v1/watcher/_scan?path=/default",
         headers=headers,
-    )
+    ).json()
+
+    # asserts that scanning default workspace doesn't fail
+    res = client.get(
+        f"v1/tasks/{task_id}?wait_for_completion=true", headers=headers
+    ).json()
+    assert res["status"] == TaskStatus.COMPLETED.value
```
laurent-laporte-pro: OK, but the scanning of the `default` is prohibited, so we need to check that the task failed.

laurent-laporte-pro: Although comparing the status string is acceptable, I believe it's preferable to create a `TaskResult` object for comparison. This way, you can evaluate both the status and the message simultaneously.

antarest/study/storage/rawstudy/watcher.py
```diff
@@ -179,7 +179,10 @@ def oneshot_scan(
         """
 
         def scan_task(notifier: TaskUpdateNotifier) -> TaskResult:
-            self.scan(workspace, path)
+            try:
+                self.scan(workspace, path)
+            except CannotScanInternalWorkspace as e:
+                logger.warning(e)
```
laurent-laporte-pro, line 188: Why not returning a failed status with a message:

```python
return TaskResult(success=False, message=(str(e) or "Scan failed"))
```

laurent-laporte-pro (changes requested):
I believe that if the scan operation fails, it should be reported to the user in a clear manner, ensuring they are aware of the error. Therefore, I would prefer the task status to be set to 'failed' (accompanied by an error message) in case of failure instead of handling errors silently. If this cannot be achieved, it should be well-documented in the source code with a comment (in a verbose manner).

I'm not entirely convinced that the `CannotScanInternalWorkspace` class should be an HTTPException; to me, it seems more like a standard exception that should inherit from `Exception`. In this case, it would be best to declare the exception where it is used.

Regarding integration tests, some refactoring is needed. I've recently created a handy guide for this purpose. It aims to enhance fixture management and add a 'normal case' test.

PR #1671, opened 2023-07-26
fix(launcher): fixing launcher versions display and API endpoint for solver versions

### LINKED ISSUE #1672: The list of solver available versions in the launcher dialog box is wrong
## Description

In the launcher dialog box, you can choose a solver version to launch your study, example on recette environment : 

liste_versions

This list is returned by the endpoint `GET /studies/_versions` but it has nothing to do with the list of available solvers which can be retrieved via the endpoint `GET /launcher/_versions`

## Expected Behavior

The front end should use the 2nd API to show the available launcher versions : `GET /launcher/_versions`
The API returns a dict object such as : 

new_model_response

The key `default` is always present and its value will be a dict object with a key (slurm or local) and the list of solver versions. This is the list the front should display (we do not care about the key `others` here).

## Environment

- [x] Production
- [x] Staging

## Developer note

You should do your devlopment on the linked PR because it implements the new reponse model of the endpoint described right above.

MartinBelthle:
Description:
This pull request addresses the issue with the display of Launcher versions by examining the available versions in the application's configuration.

Currently, when launching a simulation study, the list of default Solver versions is displayed correctly.

image

Regarding the API, the `/v1/launcher/versions` endpoint is used to retrieve the default versions in the form of a JSON list, for example: `["800", "810", "820"]`. The response status will be 200 OK.

The endpoint includes an optional parameter named `solver`, which allows specifying the desired Solver for obtaining versions. The valid values for this parameter are as follows: "default" (default Solver based on the application's configuration), "slurm" (Solver installed on the computation server), or "local" (Solver installed locally). If an unknown Solver name is provided, the API will return a 422 error with a JSON-formatted error message, for example: `{"detail": "Unknown solver configuration: 'remote'"}`.

image

Changes proposed in this pull request:
- Correct the display of Launcher versions when launching a simulation study.
- Implement the `/v1/launcher/versions` endpoint to retrieve default Solver versions.
- Introduce the `solver` parameter in the endpoint to specify desired Solver versions.
- Handle the case of an unknown Solver name and return an appropriate error response.

These changes aim to improve the overall functionality and user experience of the application's Launcher and Solver version management.

### DIFF
--- a/antarest/launcher/service.py
+++ b/antarest/launcher/service.py
@@ -1,14 +1,12 @@
 import logging
 import os
 import shutil
-from datetime import datetime, timedelta
-from functools import reduce
+from datetime import datetime, timedelta, timezone
+import functools
 from http import HTTPStatus
 from pathlib import Path
 from typing import Dict, List, Optional, cast
 from uuid import UUID, uuid4
-
-from fastapi import HTTPException
 
 from antarest.core.config import Config
 from antarest.core.exceptions import StudyNotFoundError
@@ -57,6 +55,7 @@ from antarest.study.storage.utils import (
     extract_output_name,
     find_single_output_path,
 )
+from fastapi import HTTPException
 
 logger = logging.getLogger(__name__)
 
@@ -186,7 +185,7 @@ class LauncherService:
                 job_result.output_id = output_id
                 final_status = status in [JobStatus.SUCCESS, JobStatus.FAILED]
                 if final_status:
-                    job_result.completion_date = datetime.utcnow()
+                    job_result.completion_date = datetime.now(timezone.utc)
                 self.job_result_repository.save(job_result)
                 self.event_bus.push(
                     Event(
@@ -434,7 +433,7 @@ class LauncherService:
                     or ""
                 )
             if log_type == LogType.STDOUT:
-                app_logs: Dict[JobLogType, List[str]] = reduce(
+                app_logs: Dict[JobLogType, List[str]] = functools.reduce(
                     lambda logs, log: LauncherService.sort_log(log, logs),
                     job_result.logs or [],
                     {JobLogType.BEFORE: [], JobLogType.AFTER: []},
@@ -707,7 +706,7 @@ class LauncherService:
         if self.config.launcher.slurm:
             if from_cluster:
                 raise NotImplementedError
-            slurm_used_cpus = reduce(
+            slurm_used_cpus = functools.reduce(
                 lambda count, j: count
                 + (
                     LauncherParametersDTO.parse_raw(
@@ -722,7 +721,7 @@ class LauncherService:
                 float(slurm_used_cpus) / self.config.launcher.slurm.max_cores
             )
         if self.config.launcher.local:
-            local_used_cpus = reduce(
+            local_used_cpus = functools.reduce(
                 lambda count, j: count
                 + (
                     LauncherParametersDTO.parse_raw(
@@ -736,19 +735,28 @@ class LauncherService:
             load["local"] = float(local_used_cpus) / (os.cpu_count() or 1)
         return load
 
-    def get_versions(self, params: RequestParameters) -> Dict[str, List[str]]:
-        version_dict = {}
-        if self.config.launcher.local:
-            version_dict["local"] = list(
-                self.config.launcher.local.binaries.keys()
-            )
-
-        if self.config.launcher.slurm:
-            version_dict[
-                "slurm"
-            ] = self.config.launcher.slurm.antares_versions_on_remote_server
-
-        return version_dict
+    def get_versions(self) -> Dict[str, Dict[str, List[str]]]:
+        local_config = self.config.launcher.local
+        local_list = sorted(local_config.binaries) if local_config else []
+        slurm_config = self.config.launcher.slurm
+        slurm_list = (
+            sorted(slurm_config.antares_versions_on_remote_server)
+            if slurm_config
+            else []
+        )
+        default_launcher = self.config.launcher.default
+        if default_launcher == "local":
+            return {
+                "default": {"local": local_list},
+                "others": {"slurm": slurm_list},
+            }
+        elif default_launcher == "slurm":
+            return {
+                "default": {"slurm": slurm_list},
+                "others": {"local": local_list},
+            }
+        else:  # pragma: no cover
+            raise NotImplementedError(default_launcher)
 
     def get_launch_progress(
         self, job_id: str, params: RequestParameters
--- a/antarest/launcher/service.py
+++ b/antarest/launcher/service.py
@@ -748,7 +748,7 @@ class LauncherService:
             This list is empty if the configuration is not available.
 
         Raises:
-            ValueError: if the configuration is not "default", "slurm" or "local".
+            KeyError: if the configuration is not "default", "slurm" or "local".
         """
         local_config = self.config.launcher.local
         slurm_config = self.config.launcher.slurm
@@ -760,12 +760,7 @@ class LauncherService:
             else [],
         }
         versions_map["default"] = versions_map[default_config]
-        try:
-            return versions_map[solver]
-        except KeyError:
-            raise ValueError(
-                f"Unknown solver configuration: '{solver}'"
-            ) from None
+        return versions_map[solver]
 
     def get_launch_progress(
         self, job_id: str, params: RequestParameters
--- a/antarest/launcher/service.py
+++ b/antarest/launcher/service.py
@@ -1,12 +1,14 @@
+import functools
 import logging
 import os
 import shutil
 from datetime import datetime, timedelta, timezone
-import functools
 from http import HTTPStatus
 from pathlib import Path
 from typing import Dict, List, Optional, cast
 from uuid import UUID, uuid4
+
+from fastapi import HTTPException
 
 from antarest.core.config import Config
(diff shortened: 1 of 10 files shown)


laurent-laporte-pro:
@MartinBelthle, je ne vois pas quel avantage apporte la nouvelle implémentation par rapport à l'ancienne.

La seule chose qui semble manquer dans l'ancienne implémentation est un _flag_ indiquant quelle version est la version par défaut utilisée lors du lancement de la simulation.

Par exemple :

```json
{
    "slurm": ["800", "810", "820"],
    "local": ["820", "830"],
    "default": "local"
}
```

Cette approche permettait de maintenir une meilleure compatibilité avec le format existant.

laurent-laporte-pro (changes requested):
Je pense que ça serait plus simple d'ajouter un _flag_ "default" = "sulrm" ou "local".

MartinBelthle:
Cette approche me va aussi:
```python
{
  "slurm": ["800", "810", "820"],
  "local": ["820", "830"],
  "default": "local"
}
```
Je me rappelle que je voulais ajouter la clé `default`, mais pourquoi j'ai choisi un autre ResponseModel, je ne sais plus.
Tu peux effectuer ce changement :+1: 

Il faudra aussi mettre à jour le ticke car j'ai mis un screen du ReponseModel

laurent-laporte-pro:
Finalement, il est préférable d'avoir un liste à plat `["800", "810", "820"]` (comme pour les versions d'études).

MartinBelthle:
En fait je voulais garder les 2 pour pouvoir récupérer l'informations des binaires sur les 2 environnements si jamais je souhaite modifier le launcher par défaut. Mais ça n'a pas de sens car on ne peut pas faire ça avec l'API, on doit passer par le application.yaml et on y voit la liste.
Donc effectivement, on peut renvoyer juste une liste

laurent-laporte-pro:
En cas d'erreur, on aura :

HTTP 422 :

```json
{
  "description": "Unknown solver configuration: 'remote'",
  "exception": "HTTPException"
}
```

En cas de succès, on aura :

HTTP 200 :

```json
[ "700", "720", "800", "810", "820", "830", "840", "850", "860" ]
```

antarest/launcher/web.py
```diff
@@ -203,14 +205,44 @@ def get_load(
     @bp.get(
         "/launcher/_versions",
```
hdinia: Endpoints that follow RESTful naming conventions typically don't have underscores unless specific actions, here we `GET` launcher versions which is a common action, not needed and confusing. it's old code, but we should change it as it is not yet used front side

antarest/launcher/web.py
```diff
@@ -203,14 +205,44 @@ def get_load(
...
-        logger.info(f"Fetching version list")
-        return service.get_versions(params=params)
+    def get_solver_versions(
+        solver: Optional[str] = Query(
+            None,
+            examples={
+                "default solver": {
+                    "description": "Get the solver versions of the default configuration",
+                    "value": "",
+                },
```
hdinia:
```python
"default solver": {
                    "description": "Get the solver versions of the default configuration",
                    "value": "default",
                },
```
I think you missed the value

antarest/launcher/web.py
```diff
@@ -203,14 +205,44 @@ def get_load(
...
+            },
+        )
+    ) -> List[str]:
+        """
+        Get list of supported solver versions defined in the configuration.
+
+        Args:
+        - `solver`: name of the configuration to read: "default", "slurm" or "local".
+        """
+        solver = solver or "local"
```
hdinia: If no solver is not provided you define it to `local` is it intentional? shouldn't it be `default` instead?

antarest/launcher/web.py
```diff
@@ -203,14 +205,44 @@ def get_load(
     @bp.get(
         "/launcher/_versions",
         tags=[APITag.launcher],
-        summary="Get list of supported study version for all configured launchers",
-        response_model=Dict[str, List[str]],
+        summary="Get list of supported solver versions",
+        response_model=List[str],
```
hdinia, line 222: a detail but you can add the `status_code = HHTPStatus.OK`

antarest/launcher/service.py
```diff
@@ -736,19 +736,36 @@ def get_load(self, from_cluster: bool = False) -> Dict[str, float]:
...
+            if slurm_config
+            else [],
+        }
+        versions_map["default"] = versions_map[default_config]
+        try:
+            return versions_map[solver]
+        except KeyError:
+            raise ValueError(
+                f"Unknown solver configuration: '{solver}'"
+            ) from None
```
hdinia: we can use a helper function to reduce duplication and simplify:
(errors catched by the route handler, the try/except is no longer needed)

```python
 def get_sorted_versions(solver_config, attribute):
            return sorted(getattr(solver_config, attribute)) if solver_config else []

        solvers = {
            "local": get_sorted_versions(self.config.launcher.local, "binaries"),
            "slurm": get_sorted_versions(self.config.launcher.slurm, "antares_versions_on_remote_server"),
        }

        solvers["default"] = solvers[self.config.launcher.default]

       return solvers.get(solver)
        ```

**Review by hdinia — CHANGES_REQUESTED** [review]

File: antarest/launcher/web.py [hunk 6]
```diff
@@ -203,14 +205,44 @@ def get_load(
...
+        logger.info(
+            f"Fetching the list of solver versions for the '{solver}' configuration"
+        )
+        try:
+            return service.get_solver_versions(solver)
+        except ValueError as e:
+            raise HTTPException(
+                status_code=http.HTTPStatus.UNPROCESSABLE_ENTITY,
+                detail=str(e),
+            )
```
hdinia: redundant error check with the service function `get_solver_versions`,we can move the check for an invalid solver configuration from the service to the route handler. The route handler is the appropriate place to validate user input, declaring an exception with appropriate message:

```python
class InvalidSolverConfig(HTTPException):
    def __init__(self) -> None:
        super().__init__(
            HTTPStatus.UNPROCESSABLE_ENTITY,
            "Invalid solver configuration",
        )
```

we can also be more specific and check against the params allowed only:

```python
 if solver not in ["default", "slurm", "local"]:
            raise InvalidSolverConfig
```

we would have this:

```python
      solver = solver or "default"

        if solver not in ["default", "slurm", "local"]:
            raise InvalidSolverConfig

        logger.info(
            f"Fetching the list of solver versions for the '{solver}' configuration"
        )

        return service.get_solver_versions(solver)
```

antarest/launcher/web.py
laurent-laporte-pro: OK to remove the underscore.

antarest/launcher/service.py
laurent-laporte-pro: Well, I prefer to keep the code simple.

antarest/launcher/web.py
laurent-laporte-pro: Not needed, because [200 is the default status code](https://fastapi.tiangolo.com/tutorial/response-status-code/#about-http-status-codes).

> the more you have lines, the more you have bugs 😋

antarest/launcher/web.py
laurent-laporte-pro: You right, could you add it?

antarest/launcher/web.py
laurent-laporte-pro: You're right, could you set `default`?

laurent-laporte-pro (changes requested):

antarest/launcher/web.py
laurent-laporte-pro: I think an exception called `UnknownSolverConfig` is more appropriate, because the error message should be something like that: "Unknown solver configuration: '{solver}'". Do not forget to pass the `solver` value as a parameter to the exception.

I suggest this:

```python
class UnknownSolverConfig(HTTPException):
    """
    Exception raised during solver versions retrieval when
    the name of the launcher is not "default", "slurm" or "local".
    """

    def __init__(self, solver: str) -> None:
        super().__init__(
            http.HTTPStatus.UNPROCESSABLE_ENTITY,
            f"Unknown solver configuration: '{solver}'",
        )

solver = solver or "local"
logger.info(
    f"Fetching the list of solver versions for the '{solver}' configuration"
)
if solver not in {"default", "slurm", "local"}:
    raise UnknownSolverConfig(solver)
return service.get_solver_versions(solver)
```

laurent-laporte-pro:
```python
def get_solver_versions(self, solver: str) -> List[str]:
    """
    Fetch the list of solver versions from the configuration.

    Args:
        solver: name of the configuration to read: "default", "slurm" or "local".

    Returns:
        The list of solver versions.
        This list is empty if the configuration is not available.

    Raises:
        KeyError: if the configuration is not "default", "slurm" or "local".
    """
    local_config = self.config.launcher.local
    slurm_config = self.config.launcher.slurm
    default_config = self.config.launcher.default
    versions_map = {
        "local": sorted(local_config.binaries) if local_config else [],
        "slurm": sorted(slurm_config.antares_versions_on_remote_server)
        if slurm_config
        else [],
    }
    versions_map["default"] = versions_map[default_config]
    return versions_map[solver]
```

laurent-laporte-pro (approved):

skamril (approved):

hdinia (approved):

PR #1674, opened 2023-07-26
fix(table-mode): fixes the reading of UI information in the case where the study has only one area 

### LINKED ISSUE #1623: Table mode : wrong area names
## Description
Open table mode

Areas should appear in row labels

Instead, we get **ui**, **optimization** and **adequacy_patch**. Note that each area directory contains those three INI files.

image

- If applicable, provide the study ID that this issue relates to: f079dd4f-f478-4f92-a6c6-ff2165aba28c

- Any relevant library versions: [e.g. React 17.0.1]
- The date and time when the issue occurred: 27/06/2023
- Browser and version: Firefox 107.0
- Operating System: Ubuntu 20.04

MartinBelthle:
The problem lies in the folder_node.py in the method _forward_get. But this is called in so many places in the code i did a little trick elsewhere to bypass the issue. 

Not sure you will like it @laurent-laporte-pro but I did not find any satisfying solution :disappointed:

### DIFF
--- a/antarest/study/business/area_management.py
+++ b/antarest/study/business/area_management.py
@@ -1,17 +1,17 @@
 import logging
 import re
 from enum import Enum
-from typing import Optional, Dict, List, Tuple, Any
+from typing import Any, Dict, List, Optional, Sequence, Tuple
 
 from pydantic import BaseModel
 
-from antarest.core.exceptions import LayerNotFound, LayerNotAllowedToBeDeleted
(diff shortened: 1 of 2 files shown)


laurent-laporte-pro (approved):

laurent-laporte-pro:
@MartinBelthle, you wasn't wrong 😅. This situation is similar to the one already identified in the `get_layers` method. Therefore, I refactored the code to have only one function to read the UI information of the areas.

PR #1678, opened 2023-07-27
fix(ui-common): prevent null values for string cells in HandsonTable component

### LINKED ISSUE #1663: [ANT-804] Saving modifications impossible on table mode
## Description

When we make some modification on the table mode of this study ID **0b25cf01-d6a8-4917-a88a-2dc030342e50** (on production), it's not saving it. This issue isn't present on other studies.

hdinia:
fix #1663  

-This Pull Request addresses an issue in the TableMode, when you press the delete button, the default behavior is to set the cell value to `null`. This behavior is now modified so that the cell's value is set to an empty string (`""`) instead.

Note: only **string** inputs are concerned

### DIFF
--- a/webapp/src/components/common/Handsontable.tsx
+++ b/webapp/src/components/common/Handsontable.tsx
@@ -2,6 +2,7 @@ import HotTable, { HotTableProps } from "@handsontable/react";
 import { styled } from "@mui/material";
 import { registerAllModules } from "handsontable/registry";
 import { forwardRef } from "react";
+import * as RA from "ramda-adjunct";
 import { SECONDARY_MAIN_COLOR } from "../../theme";
 
 // Register Handsontable's modules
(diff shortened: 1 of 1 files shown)


webapp/src/components/common/FormTable/Table.tsx
```diff
@@ -48,7 +48,7 @@ function Table(props: TableProps) {
   const handleAfterChange: HandsontableProps["afterChange"] =
     function afterChange(this: unknown, changes, ...rest): void {
       changes?.forEach(([row, column, _, nextValue]) => {
-        setValue(`${data[row].id}.${column}`, nextValue);
+        setValue(`${data[row].id}.${column}`, nextValue ?? "");
```
skamril: No. Like I said, `nextValue` is not necessary a string.

skamril (changes requested):

webapp/src/components/common/FormTable/Table.tsx
hdinia: correct me if I am wrong but `null` value is possible only on strings. when you press delete on booleans `true` or `false` only is possible. on numbers inputs this behavior is disabled.

So I dont see in what case we can have an other type than string that is `null` ?

hdinia: We can maybe check for the type of `nextValue` ? 
```javascript
if(typeof nextValue === 'string' && nextValue === null) {
...
}
```

hdinia: We can check the type of nextValue and add a specific case for strings only ? 

```javascript
changes?.forEach(([row, column, _, nextValue]) => {
  if (typeof nextValue === 'string') {
    setValue(`${data[row].id}.${column}`, nextValue ?? "");
  } else {
    setValue(`${data[row].id}.${column}`, nextValue);
  }
});
```

hdinia: EDIT:

`afterChange` event in Handsontable provides the previous value of the cell `prevValue`, in addition to `nextValue`. We can leverage this to determine if the cell was originally a `string`. If it was, and if `nextValue` is `null`, we can then replace it with an empty string. This ensures that we only replace null with "" for cells that were originally strings.

```javascript
  const handleAfterChange: HandsontableProps["afterChange"] =
    function afterChange(this: unknown, changes, ...rest): void {
      changes?.forEach(([row, column, prevValue, nextValue]) => {
        if (typeof prevValue === "string" && nextValue === null) {
          setValue(`${data[row].id}.${column}`, "");
        } else {
          setValue(`${data[row].id}.${column}`, nextValue);
        }
      });
      restProps.afterChange?.call(this, changes, ...rest);
    };
```
Tested, it works as expected, I dont see a better solution if this does not suits you I let you provide a better one

skamril: Use `beforeChange` instead, because is called before re-render.

```typescript
const handleBeforeChange: HandsontableProps["beforeChange"] =
  function beforeChange(this: unknown, changes, ...rest): void {
    changes.forEach(([row, prop, oldValue, newValue]) => {
      if (restProps.allowEmpty && typeof oldValue === "string" && newValue === null) {
        // You must mutate the value
        changes[row][3] = '';
      }
    });
    restProps.beforeChange?.call(this, changes, ...rest);
  }
```

Add it to the HandsonTable component.

skamril (dismissed):

skamril (approved):

PR #1687, opened 2023-08-02
feat(ui-study): enhance study upgrade dialog

### LINKED ISSUE #1508: Enhance Upgrade Dialog to allow selective version upgrade
## Problem Statement

Currently, when a user clicks the "Upgrade" button on the study's web interface, the upgrade dialog is always displayed. This can be confusing for the user, especially when there are no available versions to upgrade to. Therefore, we would like to propose an enhancement to the upgrade dialog to only display it when a superior version is available.

image

In addition, we understand that some users may want to upgrade to multiple versions later, rather than upgrading sequentially one version at a time. Therefore, we propose enhancing the upgrade dialog to allow for selective version upgrades. The new behavior should be as follows:

## Proposed Solution

1. The upgrade dialog should only be displayed when a superior version is available.
2. The upgrade dialog should display a list of available versions in reverse chronological order (from newest to oldest). The list should only show versions that are higher than the current version of the study.
3. The user should be able to select one or more versions to upgrade to. The selected versions should be upgraded starting from the current version of the study, and the user should be able to skip versions in between if desired.
4. The upgrade dialog should provide an option to cancel the upgrade process at any time.

### TODO

- [ ] Create UpgradeStudyDialog component

hdinia:
fix #1508 

# Add Selective Study Upgrade and Refactor NavHeader Component

## Description

This PR introduces a new feature that allows selective study upgrades. In addition, it includes a significant refactor of the `NavHeader` component. The refactor simplifies the component's structure, improves naming conventions, and removes unnecessary styles and conditions. The changes are aimed at enhancing maintainability and scalability without altering the existing functionality.

Key changes include:
- The `NavHeader` component has been broken down into smaller, more manageable components. This makes the code more readable and easier to maintain.
- The naming conventions have been improved for better clarity and consistency.
- Unnecessary styles and conditions have been removed to reduce complexity.
- Vertical spaces have been reduced as per user requests to improve the overall user experience.

## Impact

These changes should improve both the developer experience and the user experience. Developers will find the code easier to understand and modify, which can speed up future development. Users will benefit from the new selective study upgrade feature and the improved layout.


laurent-laporte-pro (changes requested):
For an improved user experience, it would be desirable to have the latest version selected by default. The list of versions should be in reverse chronological order: with the most recent one displayed first. Since the action associated with the 'submit' button is a version update, it would be preferable to label it as 'Upgrade' (or 'Update').

hdinia:
> For an improved user experience, it would be desirable to have the latest version selected by default. The list of versions should be in reverse chronological order: with the most recent one displayed first. Since the action associated with the 'submit' button is a version update, it would be preferable to label it as 'Upgrade' (or 'Update').
@laurent-laporte-pro 
- Not sure about the version selected by default.. (I think it can confuse the user)  ❌
- I will add the reversed order. 👍

> Edit: seen with @skamril, we can change the submit button text easily with a dedicated prop `submitButtonText`, I'll add it 👍

webapp/src/components/App/Singlestudy/NavHeader/NavHeaderCommands.tsx
```diff
@@ -0,0 +1,131 @@
...
+interface Props {
+  study: StudyMetadata | undefined;
+  isExplorer: boolean | undefined;
+  onCopyId: () => Promise<void>;
+  onUnarchive: () => Promise<void>;
+  onLaunch: VoidFunction;
+  onOpenCommands: VoidFunction;
+}
+
+function NavHeaderCommands({
```
skamril: Rename all components with format NavHeaderX.tsx -> X.tsx, beceause they already are in the NavHeader folder.

webapp/src/components/App/Singlestudy/NavHeader/NavHeaderCommands.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { useTranslation } from "react-i18next";
+import { StudyMetadata, StudyType } from "../../../../common/types";
+import { toggleFavorite } from "../../../../redux/ducks/studies";
+import StarToggle from "../../../common/StarToggle";
+import useAppDispatch from "../../../../redux/hooks/useAppDispatch";
+import useAppSelector from "../../../../redux/hooks/useAppSelector";
+import { isCurrentStudyFavorite } from "../../../../redux/selectors";
+
+interface Props {
+  study: StudyMetadata | undefined;
```
skamril:
```javascript
study?: StudyMetadata;
isExplorer?: boolean;

```

webapp/src/components/App/Singlestudy/NavHeader/NavHeaderCommands.tsx
```diff
@@ -0,0 +1,131 @@
...
+  isExplorer,
+  onCopyId,
+  onUnarchive,
+  onLaunch,
+  onOpenCommands,
+}: Props) {
+  const [t] = useTranslation();
+  const dispatch = useAppDispatch();
+  const isStudyFavorite = useAppSelector(isCurrentStudyFavorite);
+  const isManaged = study?.managed;
```
skamril: `!!` to have a boolean in all case. Because `isX` refers to a boolean.

webapp/src/services/api/study.ts
```diff
@@ -166,8 +166,15 @@ export const unarchiveStudy = async (sid: string): Promise<void> => {
...
 };
 
-export const upgradeStudy = async (sid: string): Promise<void> => {
-  await client.put(`/v1/studies/${sid}/upgrade`);
+export const upgradeStudy = async (
+  studyId: string,
+  targetVersion: string
+): Promise<void> => {
+  await client.put(
+    `/v1/studies/${studyId}/upgrade?target_version=${encodeURIComponent(
```
skamril, line 174: `encodeURIComponent` is necessary for version?

webapp/src/components/App/Singlestudy/UpgradeDialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+        value: id,
+        label: name,
+      }));
+  }, [versions, study.version]);
+
+  ////////////////////////////////////////////////////////////////
+  // Event handlers
+  ////////////////////////////////////////////////////////////////
+
+  const handleSubmit = async (
```
skamril, line 40: TODO @skamril

webapp/src/components/App/Singlestudy/NavHeader/index.tsx
```diff
@@ -0,0 +1,339 @@
...
+              width: 1,
+              height: "auto",
+              gap: 1,
+            }}
+          >
+            <Typography>{t("studies.question.delete")}</Typography>
+            <CheckBoxFE
+              value={deleteChildren}
+              label={t("studies.deleteSubvariants")}
+              onChange={(_e, checked) => setDeleteChildren(checked)}
```
skamril: `_` or `e`

webapp/src/components/App/Singlestudy/NavHeader/NavHeaderMenu.tsx
```diff
@@ -0,0 +1,60 @@
...
+  action: () => void;
+  condition?: boolean;
+  color?: string;
+}
+
+interface Props {
+  anchorEl: HTMLElement | null;
+  openMenu: string;
+  onClose: () => void;
+  menuItems: NavMenuItem[];
```
skamril: `items` is sufficient

webapp/src/components/App/Singlestudy/NavHeader/NavHeaderCommands.tsx
```diff
@@ -0,0 +1,131 @@
...
+        alignItems: "center",
+        boxSizing: "border-box",
+      }}
+    >
+      <Tooltip title={study.folder} placement="bottom-start">
+        <Typography
+          variant="h6"
+          noWrap
+          sx={{
+            width: 1,
```
skamril: `flex: 1` instead, and remove `minWidth` for unarchive button.

skamril (changes requested):

webapp/src/services/api/study.ts
hdinia: you are right, it might not be strictly necessary, it's often a good idea to encode URI params to be safe. As you like I can remove it

webapp/src/components/App/Singlestudy/NavHeader/index.tsx
hdinia: old code didn't catch it when refactored, my bad

webapp/src/components/App/Singlestudy/NavHeader/NavHeaderCommands.tsx
hdinia: I will keep the explicit `undefined` because in this case, the `| undefined`  signals that the prop can be `undefined` due to the parent's logic, but it's not optional in the sense that the parent should always pass it (even if it's undefined).

webapp/src/components/App/Singlestudy/NavHeader/NavHeaderCommands.tsx
hdinia: didn't think about it, thanks

webapp/src/components/App/Singlestudy/NavHeader/NavHeaderCommands.tsx
hdinia: Not sure about this.. Its true that NavHeader/ X.tsx gives context , but I find having very generic names like Info.tsx can lead to confusion. (Not a common component)

laurent-laporte-pro (dismissed):

webapp/src/services/api/study.ts
laurent-laporte-pro: I agree that it is not necessary to validate the values since they are provided by the back office and guaranteed to be numbers (in other words, the values do not come from user input). However, it is important to note that the value can be generated from an R script using the API. Therefore, my question is whether you ensure that the value matches a certain regex, for instance, `"\d+"`?

hdinia: I am not sure to understand, this is not validation, just encoding for URI parameters when we send the update request:
this ensure to have proper encoding for URI components (encode spaces or specials chars etc..), also useful for security aspects as it prevent potential injections by manipulating the payload request, see:
[encodeURIComponent](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)

webapp/src/components/App/Singlestudy/NavHeader/NavHeaderCommands.tsx
skamril: Ok but `isExplorer` must be a mandatory boolean so.

webapp/src/components/App/Singlestudy/NavHeader/NavHeaderCommands.tsx
skamril: Not a probleme because the component is inside the folder NavHeader. Long names can be avoided here. 
All component folders have an index.tsx for example, and there is no confusion. When we search a file with VS Code, the folder name is include (search: `navheaderinfo` => result: `NavHeader/info.tsx`)

skamril (changes requested):

webapp/src/components/App/Singlestudy/NavHeader/Actions.tsx
```diff
@@ -0,0 +1,131 @@
...
+        />
+      ) : (
+        <Chip
+          label={study.workspace}
+          variant="filled"
+          size="small"
+          sx={{ mr: 1 }}
+        />
+      )}
+      {(study.tags || []).map((tag) => (
```
skamril: `study.tags?.map`

webapp/src/components/App/Singlestudy/NavHeader/Actions.tsx
```diff
@@ -0,0 +1,131 @@
...
+          color="info"
+          size="small"
+          sx={{ mr: 1 }}
+        />
+      ) : (
+        <Chip
+          label={study.workspace}
+          variant="filled"
+          size="small"
+          sx={{ mr: 1 }}
```
skamril: Remove all margin left and right on buttons. Replace it by adding a gap in the box container.

webapp/src/components/App/Singlestudy/NavHeader/NavHeaderCommands.tsx
hdinia: it can be optional on the parent..

skamril (approved):

PR #1698, opened 2023-08-11
feat(ui): remove patch number in version list when it's equal to 0

### LINKED ISSUE #1204: In version list, drop "patch" version number
### Current
- 8.0.0
- 8.1.0
- 8.2.0
- 8.3.0
- etc.

### Proposed
- 8.0
- 8.1
- 8.2
- 8.3
- etc.

### Why
Currently, AntaresWeb lies to the user. For exemple, it doens't launch antares-solver 8.2.0, but rather 8.2.2.

### Screenshot

image

skamril:


webapp/src/services/utils/index.ts
```diff
@@ -148,8 +148,14 @@ export const exportText = (fileData: string, filename: string): void => {
...
-export const displayVersionName = (version: string): string =>
-  version.split("").join(".");
+export const displayVersionName = (version: string): string => {
+  const [major, minor, patch] = version.split("");
+  // cf. https://github.com/AntaresSimulatorTeam/AntaREST/issues/1204
+  if (patch === "0") {
+    return `${major}.${minor}`;
+  }
+  return `${major}.${minor}.${patch}`;
+};
```
hdinia: you dont want to change this one has you will affect the study versions displayed on the entire application by removing the last digit.

instead define a new util function to format solver versions:

```javascript
const displaySolverVersion = (v: string): string => `${v[0]}.${v[1]}`;
```
and keep the original one by renaming properly to: `displayStudyVersions` or `formattedStudyVersions` as you like
that will keep the same format for study versions:

```javascript
const displayStudyVersion = (v: string): string => `${v[0]}.${v[1]}.${v[2]}`;
```

hdinia (changes requested):

webapp/src/services/utils/index.ts
hdinia: > Edit: Seen with @skamril @laurent-laporte-pro,  in fact we want to change all displayed versions and exclude the patch digit, that most likely will never be used.

```javascript
const displayVersion = (v: string): string => `${v[0]}.${v[1]}`;
```

hdinia (approved):

Issue #1212, opened 2023-01-03
When upgrading a study, upgrade all its children


pl-buiquang:
# Requirements

When a study is updated, all its descendant variants must be updated.

This is easy, as it is sufficient to clear the snapshot/invalidate cache and update its version number in the database, thus avoiding any failure.

## Context

A study can be derived from a main study that serves as the basis for the calculations. The derived study is considered to be a child of the main study: it is a variant that contains an ordered sequence of commands to modify the main study. Any derived study can also have one or may children.

The upgrade of a study consists of applying data migrations to move to a higher version of the data format. There are two types of studies: raw studies (stored on disk) and variants (studies with a command history).

# Acceptance Criteria

- [ ] After an upgrade, the version number of the upgraded study must be increased and all descendant studies must have the same version number as the upgraded study.
- [ ] After an upgrade, the version number of the upgraded study must be greater than the version number of all its ancestors.

laurent-laporte-pro:
@pl-buiquang, What happens if the operator wants to upgrade a main study in version 8.2 that has a child study in 8.3?
My feeling is that the main study is upgraded to 8.3 and the child study is unchanged. Am I right?

pl-buiquang:
The children variant are always in a higher or equal version

laurent-laporte-pro:
It is necessary to call `clear_snapshot` to remove the snapshot from the working directory. This function must also be called for each descendant.

MartinBelthle:
Indeed the snapshot should be cleaned but 2 things need to be done.
1) Be sure that every command behave correctly according to the study version
   - I think the only commands concerned by the update are : create and remove for areas and links. These commands are versioned so IMO this point is okay.
2) Choose what to do with the very specific case : the parameter "custom-ts-numbers" in settings/generaldata.ini. This parameter disappears with the version 8.0. Therefore, if a command used to modify this parameter, the command will fail after the upgrade. It is such a weird case, I propose we do not handle this and the user will simply have to remove this command by himself. Another solution could be to remove the parameter ourselves but we should notify the user if we do so. What do you think @laurent-laporte-pro ?

PR #1703, opened 2023-08-23
feat(upgrade): upgrading a study also upgrades all of its children

MartinBelthle:
Linked to this issue : https://github.com/AntaresSimulatorTeam/AntaREST/issues/1212 but i cannot link them i do not know why

### DIFF
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -229,16 +229,28 @@ class StudyUpgraderTask:
                 file_study.tree.denormalize()
             try:
                 # sourcery skip: extract-method
-                if is_variant:
-                    self.storage_service.variant_study_service.clear_snapshot(
-                        study_to_upgrade
-                    )
-                else:
+                if not is_variant:
                     study_path = Path(study_to_upgrade.path)
                     upgrade_study(study_path, target_version)
-                remove_from_cache(self.cache_service, study_to_upgrade.id)
-                study_to_upgrade.version = target_version
-                self.repository.save(study_to_upgrade)
+                    remove_from_cache(self.cache_service, study_to_upgrade.id)
+                    study_to_upgrade.version = target_version
+                    self.repository.save(study_to_upgrade)
+
+                variants = self.storage_service.variant_study_service.repository.get_children(
+                    study_id
+                )
+                while variants:
+                    for variant in variants:
+                        self.storage_service.variant_study_service.clear_snapshot(
+                            variant
+                        )
+                        variant.version = target_version
+                        self.repository.save(variant)
+                        variants.remove(variant)
+                        variants += self.storage_service.variant_study_service.repository.get_children(
+                            variant.id
+                        )
+
                 self.event_bus.push(
                     Event(
                         type=EventType.STUDY_EDITED,
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -229,12 +229,15 @@ class StudyUpgraderTask:
                 file_study.tree.denormalize()
             try:
                 # sourcery skip: extract-method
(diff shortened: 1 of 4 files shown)


ghost:
### Features

*  **upgrade:** upgrading a study also upgrades all of its children ([605897d](https://github.com/AntaresSimulatorTeam/AntaREST/commit/605897d6dfaa7d5c7596d279b4c52ac1347ff560))


### Bug Fixes

*  **tests:** rollback test_variant_model.py ([96ab67d](https://github.com/AntaresSimulatorTeam/AntaREST/commit/96ab67d6167d08ed2ce6babd34812afd0b4a79e7))
*  **upgrade:** clear snapshot when upgrading a variant as it used to be done before ([15ddaa4](https://github.com/AntaresSimulatorTeam/AntaREST/commit/15ddaa42fa319028b500b5d43063735efffb3525))
*  **imports:** remove unused import ([0fb1469](https://github.com/AntaresSimulatorTeam/AntaREST/commit/0fb1469dc61e5f7d77da3ece6a23015c530374e0))


### Contributors

<a href="https://github.com/MartinBelthle">MartinBelthle</a>

<details>
<summary>Commit-Lint commands</summary>
<br />

You can trigger Commit-Lint actions by commenting on this PR:
- `@Commit-Lint merge patch` will merge dependabot PR on "patch" versions (X.X.Y - Y change)
- `@Commit-Lint merge minor` will merge dependabot PR on "minor" versions (X.Y.Y - Y change)
- `@Commit-Lint merge major` will merge dependabot PR on "major" versions (Y.Y.Y - Y change)
- `@Commit-Lint merge disable` will desactivate merge dependabot PR
- `@Commit-Lint review` will approve dependabot PR
- `@Commit-Lint stop review` will stop approve dependabot PR



</details>

antarest/study/service.py
```diff
@@ -234,11 +234,26 @@ def _upgrade_study(self) -> None:
...
                 else:
-                    study_path = Path(study_to_upgrade.path)
-                    upgrade_study(study_path, target_version)
-                remove_from_cache(self.cache_service, study_to_upgrade.id)
+                    upgrade_study(Path(study_to_upgrade.path), target_version)
+                    remove_from_cache(self.cache_service, study_to_upgrade.id)
                 study_to_upgrade.version = target_version
                 self.repository.save(study_to_upgrade)
+
+                variants = self.storage_service.variant_study_service.repository.get_children(
```
laurent-laporte-pro, line 242: To simplify the algorithm, you should consider a recursive approach. Here, we are changing the study version, but it's also necessary to upgrade the commands in the study history.

laurent-laporte-pro (changes requested):
Upgrading a study involves not only changing the version of a study but also modifying its properties, and in some cases, the matrices of the study. For a RAW study, it's sufficient to modify the files on disk, but for study variants, it's more complex: we need to consider upgrading some commands in the study history. It's possible that commands may operate on older properties or matrices, and they may also need modification to accommodate the format change.

**Martin**:
Indeed, you are right. I only considered commands  `create` and `remove` for areas and links that are versioned so I thought it was okay. But modifying some matrix in a link, raises a command `replace_matrix` which does not work when we upgrade a study from 8.1 to 8.2. Moreover, as the matrix is splitted in 3 different matrices, the command should become 3 distinct ones :confused: . So i think handling such a case is really complicated and takes a lot of time to develop. Plus it is not really important as users use almost only v8.2 and newer versions.

MartinBelthle:
Le problème est plus compliqué que prévu et cette PR ne le résolvera pas. Je la close car elle ne fix pas le souci

PR #1714, opened 2023-08-31
fix(raw): impossible to see matrix containing NaN values

### LINKED ISSUE #1659: Impossible to see some matrix file in output
## Description

After run a simulation in 1 MC year and synthesis and year-by-year output, I can't see in results view this matrix : 

- id-<period>.txt
- values-<period>.txt

I got the message below and associates files are present in the output folder

image

When associates files are not presents I got the right message "no data available"
Also see in study version 8.4.2 .

MartinBelthle:
In the file `raw_studies_blueprint.py`,   i "rollbacked" the behaviour as it was before this PR:  https://github.com/AntaresSimulatorTeam/AntaREST/pull/1577 to allow NaN values in matrices.

I added an integration test to ensure this worked. 

I've also cleaned a bit the `test_integration.py` file:
- I used the fixture `admin_access_token` when needed
- I added a line `admin_headers = {"Authorization": f"Bearer {admin_access_token}"}` at the beginning of each test to use this value everywhere and by doing so, I lightened the code

### DIFF
--- a/antarest/study/web/raw_studies_blueprint.py
+++ b/antarest/study/web/raw_studies_blueprint.py
@@ -136,6 +136,10 @@ def create_raw_study_routes(
                 # because it's better to avoid raising an exception.
                 return Response(content=output, media_type="application/octet-stream")
 
+        # We want to allow `NaN`, `+Infinity`, and `-Infinity` values in the JSON response
+        # even though they are not standard JSON values because they are supported in JavaScript.
+        # Additionally, we cannot use `orjson` because, despite its superior performance, it converts
+        # `NaN` and other values to `null`, even when using a custom encoder.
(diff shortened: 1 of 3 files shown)


MartinBelthle:
I tried to use `ORJSONResponse` but that converts `NaN` to `None` and we do not want that. Therefore, i keep the code as it is

laurent-laporte-pro:
> I tried to use `ORJSONResponse` but that converts `NaN` to `None` and we do not want that. Therefore, i keep the code as it is

You are right: it is written in the [documentation](https://github.com/ijl/orjson#float).
However, we could use orjson for its performance, but it doesn't work for those values 👎🏻

laurent-laporte-pro (approved):

PR #1726, opened 2023-09-04
fix(api): correct API endpoints and CLI tools

### LINKED ISSUE #1724: Upgrade version of FastAPI and Pydantic
## Problem Statement

The version of FastAPI we are currently using is 0.73, released on 23 January 2022. Although this version is relatively recent, it should be noted that the library is undergoing rapid development within the open-source community. 

This library has dependencies such as Pydantic, Starlette and Requests (used by Starlette). These dependencies also need to be updated to take advantage of new features, bug fixes and security patches.

However, we will not be able to upgrade to version 2 of Pydantic as this would require substantial changes to our code.

Since version 0.21 of Starlette, this library no longer has a dependency on Requests, having replaced it with Httpx, a more modern library. We need to adapt the source code of our application to take account of this change. What's more, given that we use Requests in a very limited and basic way, we can also abandon Requests in favour of Httpx without requiring any considerable effort.

We want to target the following versions:
- FastAPI: 0.103.1
- Httpx: 0.24.1
- Pydantic: 1.10.12
- Starlette: 0.27.0

and drop Requests.

laurent-laporte-pro:
This PR includes updates to the API endpoints and CLI tools as per the provided change log.

## Summary

- Updated the construction of `UploadFile` objects in unit tests to accommodate changes in Starlette's API (>= 0.24).
- Updated stream file download tests for compatibility with both Requests and Httpx.
- Prepared for migration from `Requests` to `httpx` by adjusting imports and usage in LDAP, command line tools, and GUI code.
- Corrected and updated integration tests for Starlette >= 0.21, which introduced changes to the `TestClient` implementation.
- Corrected unit tests for Xpansion-related endpoints to use `io.BytesIO` for file uploads instead of `io.StringIO`.
- Corrected the response model for the `update_area_ui` endpoint to return `None`.
- Corrected JSON payloads for endpoints related to "table mode" to properly convert enumerated values into strings.
- Corrected the endpoint for `HTTPStatus.NO_CONTENT (204)` and updated the response model to `None`.

## Related Issues

Fixes #1724

### DIFF
--- a/antarest/study/web/studies_blueprint.py
+++ b/antarest/study/web/studies_blueprint.py
@@ -79,9 +79,10 @@ def create_study_routes(study_service: StudyService, ftm: FileTransferManager, c
 
     @bp.put(
         "/studies/{uuid}/comments",
-        status_code=HTTPStatus.NO_CONTENT.value,
+        status_code=HTTPStatus.NO_CONTENT,
         tags=[APITag.study_raw_data],
         summary="Update comments",
+        response_model=None,
     )
     def edit_comments(
         uuid: str,
--- a/tests/matrixstore/test_service.py
+++ b/tests/matrixstore/test_service.py
@@ -1,12 +1,14 @@
 import datetime
 import io
 import time
+import typing as t
 from unittest.mock import ANY, Mock
 from zipfile import ZIP_DEFLATED, ZipFile
 
 import numpy as np
 import pytest
 from fastapi import UploadFile
+from starlette.datastructures import Headers
 
 from antarest.core.jwt import JWTGroup, JWTUser
 from antarest.core.requests import RequestParameters, UserHasNotPermissionError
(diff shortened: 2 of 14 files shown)


tests/matrixstore/test_service.py
```diff
@@ -345,6 +347,17 @@ def test_dataset_lifecycle():
     dataset_repo.delete.assert_called_once()
 
 
+def _create_upload_file(filename: str, file: t.IO = None, content_type: str = "") -> UploadFile:
```
TLAIDI, line 350: Need the docstring for this function

laurent-laporte-pro: It doesn't truly justify documenting a unit test function since it's an internal implementation detail. Furthermore, the code is self-explanatory and straightforward. I intend to remove this function in my upcoming PR related to the version upgrade as it will no longer serve a purpose. Thank you nonetheless.

antarest/study/web/studies_blueprint.py
```diff
@@ -79,9 +79,10 @@ def get_comments(
 
     @bp.put(
         "/studies/{uuid}/comments",
-        status_code=HTTPStatus.NO_CONTENT.value,
+        status_code=HTTPStatus.NO_CONTENT,
         tags=[APITag.study_raw_data],
         summary="Update comments",
+        response_model=None,
```
hdinia, line 85: If you have no particular reason to explicitly set `response_model=None` it can be removed instead

hdinia (approved):

TLAIDI (approved):

antarest/study/web/studies_blueprint.py
laurent-laporte-pro: En fait, les versions récentes de FastAPI peuvent se baser sur la signature de la fonction au lieu de `response_model`. Je peux voir ça dans ma future PR.

PR #1728, opened 2023-09-06
build(python): bump project dependencies 

### LINKED ISSUE #1724: Upgrade version of FastAPI and Pydantic

laurent-laporte-pro:
**Martin:**

This PR aims at bumping several project dependencies:
- `pydantic` from 1.9.0 to 2.8.2 :warning: **HUGE** breaking change on their side
- `fastapi` from 0.73.0 to 0.110.3
- `uvicorn` from to 0.15.0 to 0.30.6
- `mypy` from 1.4.1 to 1.11.1

It also brings few changes inside dependencies
- Drop `requests` in favor of `httpx`
- Drop `fastapi-jwt-auth` as they do not and will not support pydantic v2. We've decided to copy their code and adapt it as it's really light (see new folder `/antarest/fastapi_jwt_auth`)

These changes also induced other minor dependencies bump: `jinja2`, `typing_extensions`, `PyJWT`, `python-multipart`


flomnes (commented):
Any documentation about migration ? What deps are you upgrading ? Only FastAPI ?

MartinBelthle:
`fastapi-jwt-auth` doesn't and will not support Pydantic v2 so we have to find another library to do the work ...

MartinBelthle:
Still 85 test failures after commit `use own all optional constructor`

MartinBelthle:
**TODO:** 
After commit `try to fix mpy`:
- Still 45 test failures
- Still 99 mypy issues

MartinBelthle:
**TODO:** 

After commit `24febd1`:
- Still 30 test failures (all inside `integration` folder)
- Still 98 mypy issues

MartinBelthle:
Sonar raises a Security issue concerning the code `Client(verify=False)`.
But this code is only used by the `cli` so I believe we should suppress it.
I'll let you decide @sylvlecl.

sylvlecl (approved):

sylvlecl:
> Sonar raises a Security issue concerning the code `Client(verify=False)`. But this code is only used by the `cli` so I believe we should suppress it. I'll let you decide @sylvlecl.

For this, although it's in the CLI tool, I preferred to really solve the issue by letting the command line user take the responsibility of disabling SSL certification, through a new argument `--no-verify`.

PR #1740, opened 2023-09-21
chore: correct handling of base class with no `__annotations__` in `AllOptionalMetaclass`

laurent-laporte-pro:
This correction addresses an issue where the `AllOptionalMetaclass` metaclass was not handling cases where the base class has no `__annotations__` attribute properly. In such situations, this fix ensures that an empty annotation dictionary is assumed for correct behavior.


laurent-laporte-pro:
Solved in #1742

PR #1742, opened 2023-09-21
fix(thematic-trimming): correct the name of the "Profit by plant" variable

### LINKED ISSUE #1741: [ANT-857] Invalid Thematic Trimming variable name "Profit" must be replaced with "Profit by plant"
## Description

The variable used to display or hide power plant profits in the simulation output is named "Profit,"
which should be renamed to "Profit by plant" for accurate representation.

## Proposed Solution

Correct this variable in both the back end and front end by renaming it to "Profit by plant".

The unit tests should be updated to account for this change and test scenarios when the "Profit by plant"
feature is enabled or disabled.

## Steps to Reproduce

1. Choose a study in version 8.3 or above
2. Select the "CONFIGURATION" tab,
3. In the "Output profile" field set, choose "Thematic Trimming" and select "Custom",
4. Click on the gear button "⚙ Parameters" to customize the "Thematic Trimming",
5. In the "Thematic Trimming" dialog box, you can see the "Profit" variable.

## Environment

- Application Version: v2.14.5
- [x] Staging
- [x] Production

## Screenshots

thematic_trimming_profit_by_plant

thematic_trimming_profit_by_plant-2

laurent-laporte-pro:
This modification addresses the variable name used to enable or disable the calculation of power plant profits. It replaces the name "Profit" with "Profit by plant." On the back-office implementation side, changes have been made to simplify maintenance, and tests have been added to ensure proper handling of this variable. On the front end, the variable has been renamed to `profitByPlant` to align with TypeScript/ReactJS naming conventions.

image

### DIFF
--- a/antarest/study/business/thematic_trimming_management.py
+++ b/antarest/study/business/thematic_trimming_management.py
@@ -1,82 +1,92 @@
-from typing import Any, Dict, List, Optional, cast
+import typing as t
 
-from pydantic.types import StrictBool
-
-from antarest.study.business.utils import GENERAL_DATA_PATH, FieldInfo, FormFieldsBaseModel, execute_or_add_commands
+from antarest.study.business.utils import (
+    GENERAL_DATA_PATH,
+    AllOptionalMetaclass,
+    FieldInfo,
+    FormFieldsBaseModel,
+    execute_or_add_commands,
+)
 from antarest.study.model import Study
 from antarest.study.storage.storage_service import StudyStorageService
 from antarest.study.storage.variantstudy.model.command.update_config import UpdateConfig
 
 
-class ThematicTrimmingFormFields(FormFieldsBaseModel):
-    ov_cost: Optional[StrictBool]
-    op_cost: Optional[StrictBool]
-    mrg_price: Optional[StrictBool]
-    co2_emis: Optional[StrictBool]
-    dtg_by_plant: Optional[StrictBool]
-    balance: Optional[StrictBool]
-    row_bal: Optional[StrictBool]
-    psp: Optional[StrictBool]
-    misc_ndg: Optional[StrictBool]
-    load: Optional[StrictBool]
-    h_ror: Optional[StrictBool]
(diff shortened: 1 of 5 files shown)


laurent-laporte-pro:
Depends on #1740

antarest/study/business/thematic_trimming_management.py
```diff
@@ -1,82 +1,92 @@
-from typing import Any, Dict, List, Optional, cast
-
-from pydantic.types import StrictBool
-
-from antarest.study.business.utils import GENERAL_DATA_PATH, FieldInfo, FormFieldsBaseModel, execute_or_add_commands
+import typing as t
```
skamril, line 1: Why do you import typing like this?

antarest/study/business/thematic_trimming_management.py
```diff
@@ -1,82 +1,92 @@
...
+    solar_rooft: bool
+    renw_1: bool
+    renw_2: bool
+    renw_3: bool
+    renw_4: bool
     # For study versions >= 830
-    dens: Optional[StrictBool]
-    profit: Optional[StrictBool]
+    dens: bool
+    profit_by_plant: bool
```
skamril, line 86: Indicate the version for this new field in comment, and keep `profit` field in the list.
You cannot remove a field in production like this.

skamril (changes requested):

skamril (dismissed):

antarest/study/business/thematic_trimming_management.py
laurent-laporte-pro: Using the `import typing as t` form allows for the introduction of a dedicated namespace for typing. I find this approach more elegant because:
- The namespace isolates typing-related classes from other classes.
- It results in a concise import statement rather than a long list of classes (Dict, List, Mapping, Optional, cast, Union, etc.).
So, I find it clearer and thus more maintainable. It's a matter of preference.

skamril (approved):

PR #1733, opened 2023-09-13
feat(thermal): refactor thermal view

hdinia:
fix #1641 

## Description
This pull request introduces several enhancements and refactors to the Thermal component, resulting in improved performance, readability, and user experience. The changes made are in line with the ongoing effort to enhance the front-end of our application.

## Changes Made
- Usage of the GroupedDataTable to summarize important information
- New Columns and Calculations: Added new columns and calculations to display essential information about thermal clusters, including their group, enabled status, unit count, nominal capacity, installed capacity, and market bid cost.
- Event Handling: Implemented event handlers for creating and deleting thermal clusters. Users can now create new clusters and delete selected clusters with ease, enhancing the manageability of thermal cluster data.
- Total Summaries: Included total summaries for unit count, installed capacity, and enabled capacity, making it easier for users to understand the aggregate values of thermal clusters at a glance.
- Endpoints Added:
GET `/studies/{uuid}/areas/{area_id}/clusters/thermal`: Retrieves clusters for a given area.
GET `/studies/{uuid}/areas/{area_id}/clusters/thermal/{cluster_id}`: Retrieves thermal configuration for a given cluster.
POST `/studies/{uuid}/areas/{area_id}/clusters/thermal`: Creates a new thermal cluster for a given area and returns the created cluster.
PATCH `/studies/{uuid}/areas/{area_id}/clusters/thermal/{cluster_id}`: Updates a thermal cluster for a given area and returns the updated cluster.
DELETE `/studies/{uuid}/areas/{area_id}/clusters/thermal`: Deletes thermal clusters for a given area.

## Purpose
Main goal is better user experience and maintainability. By displaying crucial information more clearly, introducing interactive features, and enhancing the overall visual appeal, we aim to make working with thermal cluster data more efficient and user-friendly.

### Screenshots
image

Checklist

- [x] The feature works as expected, and has been tested 
- [ ] Code has been reviewed and approved by team members.
- [ ] Unit tests have been updated or added where necessary.
- [x] Documentation has been updated to reflect the changes, if applicable.
- [ ] All tests pass successfully, and there are no new linting errors.
- [x] No regression issues have been introduced.

PR #1764, opened 2023-10-11
fix(model): error 500 in cluster list when area change

### LINKED ISSUE #1686: [ANT-798] Exception thrown while viewing thermal clusters
## Description
When navigating in a study (no matter which one), an exception is systematically raised.

- Navigate with the mouse to Modeling / Area / Thermal clusters
- Open view of any cluster, for example CCGT-old-2 from GR area
- Return directly to another area, for example BG

Then we get this exception:
image

## Expected Behavior

When consulting a thermal cluster, if you click on a area, you must obtain the view of the list of thermal clusters of the associated area. No exception should be thrown.

## Environment

- [X] Production
- [X] Staging

skamril:


hdinia (changes requested):
We are planning to deprecate and no longer use the ClusterRoot component in upcoming updates. As a result, addressing this issue here would be redundant.

We'll handle this particular concern in a separate PR focusing on the new components that will replace ClusterRoot.
Such as: #1733 (+Storages, Renewables)

skamril:
Cherry picked in 2.15.3 by @laurent-laporte-pro

PR #1766, opened 2023-10-12
fix(api-raw): correct the `/studies/{uuid}/raw` endpoint to return "text/plain" content for matrices

### LINKED ISSUE #1765: [ANT-866] Problem with "raw" content (`formatted=False`) of the GET `/v1/studies/{study_id}/raw` endpoint
## Description

When users attempt to retrieve the "raw" (unprocessed) content of a file from the endpoint GET `/v1/studies/{study_id}/raw` with the `formatted=False` option, we observe unexpected behavior.
The issue lies in determining the content-type of the response, especially when retrieving matrices, which should be in TSV (tab-separated values) format.
Currently, the response can either be a binary file (content-type "application/octet-stream") or a JSON file (content-type "application/json").

## Details

As a reminder, a convention has been chosen to omit the file extension when providing the `path` parameter to the endpoint.
When the requested path does not contain a file extension, the `get_study` function that implements the endpoint cannot correctly deduce the content type of the file being read.
Two cases are possible:

- In the case where the read data is of type `bytes`, the response is assigned the content-type "application/octet-stream".
- In the case where the read data is of type `str`, the response is assigned the content-type "application/json".

## Proposed Fix

The recommended fix is to consistently use the content-type "text/tab-separated-values" in the response for the `formatted=False` option.
This content-type should be validated during integration tests with the Script R team to ensure its compatibility with scripts using the API.
It is also important to ensure that there is no regression in the Debug view (the risk is minimal because we use `formatted=True` in this case).

## Tasks

- [ ] Implement the modification to use "text/tab-separated-values" as the content-type for matrices in the `formatted=False` case.

## Tests

- [ ] Validate the content-type "text/tab-separated-values" with the Script R team.
- [ ] Perform integration tests to ensure compatibility.

## Resources

- Link to the relevant source code: [get_study](https://github.com/AntaresSimulatorTeam/AntaREST/blob/master/antarest/study/web/raw_studies_blueprint.py#L101)
- Example where a response of "application/octet-stream" is obtained: **NT2030** `/v1/studies/1c1441c7-0d15-4ae0-88f3-e310ecb60c67/raw?path=/input/thermal/series/es00/nuclear_nuclear/series&formatted=false`.
- Example where a response of "application/json" is obtained: **test_storage_cbe** `/v1/studies/45873be8-ddc8-484b-8943-91f693aea3ff/raw?path=/input/thermal/series/it/it_clust_test/series&formatted=false`.

## Environment

- Antares Web Version: v2.15.1

For more information, contact @jacquesaguilera

laurent-laporte-pro:
Correct the `/studies/{uuid}/raw` to return "text/plain" content for matrices

### DIFF
--- a/antarest/study/web/raw_studies_blueprint.py
+++ b/antarest/study/web/raw_studies_blueprint.py
@@ -96,11 +96,16 @@ def create_raw_study_routes(
             extra={"user": current_user.id},
         )
         parameters = RequestParameters(user=current_user)
-        output = study_service.get(uuid, path, depth, formatted, parameters)
+
+        resource_path = pathlib.PurePosixPath(path)
+        output = study_service.get(uuid, str(resource_path), depth=depth, formatted=formatted, params=parameters)
 
         if isinstance(output, bytes):
-            resource_path = pathlib.Path(path)
-            suffix = resource_path.suffix.lower()
+            # Guess the suffix form the target data
+            parent_cfg = study_service.get(uuid, str(resource_path.parent), depth=2, formatted=True, params=parameters)
+            child = parent_cfg[resource_path.name]
+            suffix = pathlib.PurePosixPath(child).suffix
+
             content_type, encoding = CONTENT_TYPES.get(suffix, (None, None))
             if content_type == "application/json":
                 # Use `JSONResponse` to ensure to return a valid JSON response
--- a/tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
+++ b/tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
@@ -2,7 +2,7 @@ import http
 import json
 import pathlib
 import shutil
-from urllib.parse import urlencode
+from unittest.mock import ANY
 
 import numpy as np
 import pytest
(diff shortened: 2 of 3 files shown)


antarest/study/web/raw_studies_blueprint.py
```diff
@@ -96,11 +96,16 @@ def get_study(
             extra={"user": current_user.id},
         )
         parameters = RequestParameters(user=current_user)
-        output = study_service.get(uuid, path, depth, formatted, parameters)
+
+        resource_path = pathlib.PurePosixPath(path)
```
MartinBelthle, line 100: You could maybe put this variable in the `if` condition as it was done before.
That would also allow you to use `path` instead of `str(resource_path)` in the next line.

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -96,11 +96,16 @@ def get_study(
...
+        output = study_service.get(uuid, str(resource_path), depth=depth, formatted=formatted, params=parameters)
 
         if isinstance(output, bytes):
-            resource_path = pathlib.Path(path)
-            suffix = resource_path.suffix.lower()
+            # Guess the suffix form the target data
+            parent_cfg = study_service.get(uuid, str(resource_path.parent), depth=2, formatted=True, params=parameters)
```
MartinBelthle, line 105: I do not understand why you use depth=2, can you explain it to me by responding to this comment ?

laurent-laporte-pro: I only need to get the properties' value.

antarest/study/web/raw_studies_blueprint.py
laurent-laporte-pro: Discutable, mais j'ai pas trop le temps => avec un nouveau push, on perd au moins 20 min.

PR #1769, opened 2023-10-12
fix(raw): fix HTTP exception when going on debug view

### LINKED ISSUE #1768: Clicking on Debug View raises an HTTP 404 Exception
## Description

Everything is in the title.

1. Create a study
2. Go on the debug view
3. See the error

## Possible Solution

This is due to a very recent change.
We convert a string to a path and reconvert it to a str at some point.
But if the string is an empty one, when doing this operation, it becomes "."

## Environment

- [x] Production
- [x] Staging

MartinBelthle:


laurent-laporte-pro (dismissed):
Can you move your UT in `integration.raw_studies_blueprint.test_fetch_raw_data.TestFetchRawData.test_get_study` instead.

**Martin response:** Done

laurent-laporte-pro (approved):
J'ai effectué un petit changement pour d'augmenter le couverture de test.

PR #1774, opened 2023-10-16
feat(service): support import for outputs generated by the solver with the `-z` option

### LINKED ISSUE #1738: [ANT-583] Improve retrieval of simulation results (option `-z`)
## Problem Statement

Currently, Antares Web is able to handle scenarios where Antares Launcher provides the entire study archive, including the results in the study's `output` folder.
Antares Web is able to take into account the case where results are stored in a sub-folder, e.g.
`output/20230914-1918eco`, but also the case where results are provided in a ZIP archive, e.g.
`output/20230914-1918eco.zip`.

However, Antares Web does not have the capacity to handle the situation where the archive does not contain the entire study, but only the simulation results in the form of a compressed archive.
This is particularly the case when using the `-z` option in Antares Solver.

## Proposed Solution

The aim of this feature request is to modify Antares Web to take account of this new scenario in addition to the existing scenarios.

Specifically, the `SlurmLauncher._import_study_output` method (which is an implementation detail) needs to be modified to take into account the case where the directory path `self.local_workspace / STUDIES_OUTPUT_DIR_NAME / job_id` represents a compressed archive of results instead of a entire study.

The old behavior should also be retained, i.e.
where the directory path `self.local_workspace / STUDIES_OUTPUT_DIR_NAME / job_id` is a complete study containing an `output` sub-folder.

## Developer Notes

- In order to test this evolution, we need to use a version of the `launchAntares.sh` script that exports only the results, not the entire study.

MartinBelthle:
This PR is linked to the MR `feature/use-simulator-zip`  on `antares-cd`.  

In order to validate this PR we should launch studies with and without:
- `xpress`
- `auto_unzip` option
- `xpansion`
- `xpansion` with `sensitivity`
- `study.version` higher and lower than `v8.4`

Also, we should check that the local launcher works.

**I've performed all these tests on my local environment and they passed.**


MartinBelthle:
Needs to be deployed with this branch of antares-launcher https://github.com/AntaresSimulatorTeam/antares-launcher/pull/67

laurent-laporte-pro:
This PR requires Antares-Launcher version 1.4 or above.

❗⚠❗ This feature must be tested on integration server

PR #1786, opened 2023-10-26
fix(tasks): resolve incorrect UTC timezone usage for task completion dates in list view

### LINKED ISSUE #1785: Invalid completion date in task view
## Description

In the task view, we have observed that the completion date of the task (`completion_date`) is ahead by two hours,
Whereas the task creation date (`creation_date`) is correctly displayed in local time.

## Expected Behavior

We expect the task completion date (`completion_date`) to be displayed in local time, in line with the behavior of the
task creation date (`creation_date`).

## Steps to Reproduce the Issue

1. Open a study (reputed fast to simulate).
2. Launch the study simulation and wait for the result.
3. Open the task view.
4. Observe the task completion date compared to the task creation date.

## Current Behavior

The task completion date (`completion_date`) is ahead by two hours, displayed in the local timezone.

## Screenshot

task_view-wrong_completion_date

## Environment

- Antares Web Version: v2.15.4 and previous versions

laurent-laporte-pro:
This PR correct the way `completion_date` is updated at the end of task processing.

wrong-completion-date-in-task-view

The task list in the study main view is also fixed:

study_view-completion_date-corrected

### DIFF
--- a/tests/core/tasks/test_task_job_service.py
+++ b/tests/core/tasks/test_task_job_service.py
@@ -1,0 +1,29 @@
+import datetime
+
+from sqlalchemy.orm import Session  # type: ignore
+
+from antarest.core.tasks.model import TaskJob
+
+
+def test_database_date_utc(db_session: Session) -> None:
+    now = datetime.datetime.utcnow()
+    later = now + datetime.timedelta(seconds=1)
+
+    with db_session:
+        task_job = TaskJob(name="foo")
+        db_session.add(task_job)
+        db_session.commit()
+
+    with db_session:
+        task_job = db_session.query(TaskJob).filter(TaskJob.name == "foo").one()
+        assert now <= task_job.creation_date <= later
+        assert task_job.completion_date is None
+
+    with db_session:
+        task_job = db_session.query(TaskJob).filter(TaskJob.name == "foo").one()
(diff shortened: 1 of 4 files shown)


tests/core/tasks/test_task_job_service.py
```diff
@@ -0,0 +1,29 @@
...
+        task_job = TaskJob(name="foo")
+        db_session.add(task_job)
+        db_session.commit()
+
+    with db_session:
+        task_job = db_session.query(TaskJob).filter(TaskJob.name == "foo").one()
+        assert now <= task_job.creation_date <= later
+        assert task_job.completion_date is None
+
+    with db_session:
```
MartinBelthle, line 22: Apart from the fact that these lines test interactions with the db, I do not see their real utility.

MartinBelthle (approved):

tests/core/tasks/test_task_job_service.py
laurent-laporte-pro: These lines ensure that the creation date and time `creation_date`, which is updated by the database during INSERT, corresponds to a UTC date in the `now`..`later` interval. Actually, it's difficult to check whether the UTC date and the local date coincide. We may need to check the timezone.

PR #1797, opened 2023-11-07
test(apidoc): add unit test for apidoc page

MartinBelthle:


tests/integration/test_apidoc.py
```diff
@@ -0,0 +1,19 @@
...
+logger = logging.getLogger(__name__)
+
+
+def test_apidoc(client: TestClient) -> None:
+    # Asserts that the apidoc can be loaded
+    flat_models = get_flat_models_from_routes(client.app.routes)
+    model_name_map = get_model_name_map(flat_models)
+    try:
+        get_model_definitions(flat_models=flat_models, model_name_map=model_name_map)
+    except Exception as e:
```
laurent-laporte-pro: Actually, the raised exception is `ValidationError`.
It's better not to catch the exception at all, and let the test raises.

tests/integration/test_apidoc.py
```diff
@@ -0,0 +1,19 @@
...
+        logger.error("A pydantic model is invalid. Therefore the APIdoc page cannot be loaded.", exc_info=e)
```
laurent-laporte-pro: It doesn't really make sense to use a logger in unit tests.

tests/integration/test_apidoc.py
laurent-laporte-pro: Here is what we get when the validation error is raised:
```python
tests/integration/test_apidoc.py:11: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
venv/lib/python3.8/site-packages/fastapi/utils.py:24: in get_model_definitions
    m_schema, m_definitions, m_nested_models = model_process_schema(
pydantic/schema.py:602: in pydantic.schema.model_process_schema
    ???
antarest/study/business/areas/st_storage_management.py:48: in schema_extra
    schema["example"] = STStorageInput(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   pydantic.error_wrappers.ValidationError: 1 validation error for STStorageInput
E   foo
E     extra fields not permitted (type=value_error.extra)

pydantic/main.py:341: ValidationError
```

laurent-laporte-pro (changes requested):
Could you simplify the unit test to:
- let the exception raised => it is managed by pytest anyway,
- do not use logger.

tests/integration/test_apidoc.py
MartinBelthle: Indeed

tests/integration/test_apidoc.py
MartinBelthle: Indeed, it's way better

laurent-laporte-pro (approved):

PR #1808, opened 2023-11-14
feat(launcher): add information about which user launched a study

laurent-laporte-pro:
feat(launcher): add information about which user launched a study (#1761)


laurent-laporte-pro:
Checked in integration server ✅ (API).

PR #1816, opened 2023-11-16
fix(ui-tasks): filter deleted studies jobs

### LINKED ISSUE #1470: Presence of tasks from a deleted study in the task view.
## Description
When you delete a study that has tasks and these are still present in the application's tasks page, the name of the study no longer appears in the list. Instead we have the Unknown name followed by the study ID.

And when you click on it to be redirected to this study that has been deleted, we have an empty page that grinds and returns nothing.

## Possible Solution
As a solution, after deleting the study, remove all of its tasks from the tasks page. or again, if one clicks on a task called Unknown (study id) the user receives a message informing them that the study no longer exists (because it has been deleted).
And when you click on it to be redirected to this study that has been deleted, we have a black page that grinds and returns nothing.

## Screenshot
image

## Environment

- [x] Production
- [x] Staging
- [x] Desktop

hdinia:
fixes #1470

### DIFF
--- a/webapp/src/services/api/study.ts
+++ b/webapp/src/services/api/study.ts
@@ -1,4 +1,5 @@
 import { AxiosRequestConfig } from "axios";
+import * as R from "ramda";
 import { isBoolean, trimCharsStart } from "ramda-adjunct";
 import client from "./client";
 import {
@@ -329,19 +330,21 @@ export const mapLaunchJobDTO = (j: LaunchJobDTO): LaunchJob => ({
 });
 
 export const getStudyJobs = async (
-  sid?: string,
+  studyId?: string,
   filterOrphans = true,
   latest = false,
 ): Promise<LaunchJob[]> => {
-  let query = sid
-    ? `?study=${sid}&filter_orphans=${filterOrphans}`
-    : `?filter_orphans=${filterOrphans}`;
-  if (latest) {
-    query += "&latest=100";
-  }
-  const res = await client.get(`/v1/launcher/jobs${query}`);
-  const data = await res.data;
-  return data.map(mapLaunchJobDTO);
+  const queryParams = new URLSearchParams(
+    R.filter(Boolean, {
+      filter_orphans: filterOrphans,
+      study: studyId,
+      latest: latest && 100,
+    }).toString(),
+  );
+
+  return client
+    .get(`/v1/launcher/jobs?${queryParams}`)
+    .then(({ data }) => data.map(mapLaunchJobDTO));
 };
 
 export const getStudyJobLog = async (
--- a/webapp/src/components/App/Tasks/index.tsx
+++ b/webapp/src/components/App/Tasks/index.tsx
@@ -97,7 +97,7 @@ function JobsListing() {
       if (studies.length === 0) {
         await dispatch(fetchStudies()).unwrap();
       }
-      const allJobs = await getStudyJobs(undefined, false, fetchOnlyLatest);
+      const allJobs = await getStudyJobs(undefined, true, fetchOnlyLatest);
       setJobs(allJobs);
       const dlList = await getDownloadsList();
       setDownloads(dlList);


webapp/src/services/api/study.ts
```diff
@@ -329,17 +329,18 @@ export const mapLaunchJobDTO = (j: LaunchJobDTO): LaunchJob => ({
...
   latest = false,
 ): Promise<LaunchJob[]> => {
-  let query = sid
-    ? `?study=${sid}&filter_orphans=${filterOrphans}`
-    : `?filter_orphans=${filterOrphans}`;
-  if (latest) {
-    query += "&latest=100";
-  }
-  const res = await client.get(`/v1/launcher/jobs${query}`);
+  const queryParams = new URLSearchParams({
```
skamril:
```javascript
const queryParams = new URLSearchParams([
  ["filter_orphans", filterOrphans],
  studyId && ["study", studyId],
  latest && ["latest", 100],
].filter(Boolean));
```

skamril (changes requested):

webapp/src/services/api/study.ts
hdinia: Thanks but it causes TS issues, I let you see

hdinia: I think it would be this instead: (which is the same as my solution)
```javascript
const queryParams = new URLSearchParams(
  [
    ["filter_orphans", String(filterOrphans)],
    ...(studyId ? [["study", studyId]] : []),
    ...(latest ? [["latest", "100"]] : []),
  ]
);
```

skamril:
```javascript
const queryParams = new URLSearchParams(
    R.filter(Boolean, {
      filter_orphans: filterOrphans,
      study: studyId,
      latest: latest && 100,
    }),
  );
```

webapp/src/services/api/study.ts
```diff
@@ -329,19 +330,21 @@ export const mapLaunchJobDTO = (j: LaunchJobDTO): LaunchJob => ({
...
-    : `?filter_orphans=${filterOrphans}`;
-  if (latest) {
-    query += "&latest=100";
-  }
-  const res = await client.get(`/v1/launcher/jobs${query}`);
-  const data = await res.data;
-  return data.map(mapLaunchJobDTO);
+  const queryParams = new URLSearchParams(
+    R.filter(Boolean, {
+      filter_orphans: filterOrphans.toString(),
```
skamril:
```javascript
filter_orphans: filterOrphans,
study: studyId,
latest: latest && 100
```

`R.filter` will remove false key.

webapp/src/services/api/study.ts
```diff
@@ -329,19 +330,21 @@ export const mapLaunchJobDTO = (j: LaunchJobDTO): LaunchJob => ({
...
+      study: studyId || "",
+      latest: latest ? "100" : "",
+    }),
+  ).toString();
```
skamril: `toString()` useless.

webapp/src/services/api/study.ts
```diff
@@ -329,19 +330,21 @@ export const mapLaunchJobDTO = (j: LaunchJobDTO): LaunchJob => ({
...
+
+  return (await client.get(`/v1/launcher/jobs?${queryParams}`)).data.map(
```
skamril: `client.get` can throw an error.

```javascript
return client.get(`/v1/launcher/jobs?${queryParams}`).then(({ data }) => data.map(mapLaunchJobDTO));
```

skamril (changes requested):

webapp/src/services/api/study.ts
hdinia: no queryParams must be a string

skamril: yes but you include it in a string, so automatically `toString()` will be called

skamril (dismissed):

skamril (approved):

PR #1817, opened 2023-11-17
feat(api-raw): add the `create_missing` flag to allow creating missing files when uploading a file

### LINKED ISSUE #1573: Add Support for Microsoft Office documents in Antares Web API

laurent-laporte-pro:
This PR modifies the `/v1/studies/{uuid}/raw` endpoint to add a `create_missing` flag authorizing creation of the directories and file targeted by the `path` if they don't exist.

This creation of files and folders is only possible for the `user` directory, to guard against possible data corruption. The `user/expansion` directory is also excluded to prevent modification of the Xpansion configuration.

Note that if the file targeted by the `path` does exist, there is no restriction and the user can modify it.
Note: setting up a control here could be very complicated.

Before this modification, if the `path` didn't exist, we'd get a 404 error.

Remember that the aim is to enable the user to store Microsoft Office documents, in particular Excel sheets, in the "user" directory.

The documentation of RAW Study endpoints is also improved.

Fix #1573

### DIFF
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -5,7 +5,7 @@ import logging
 import os
 from datetime import datetime, timedelta
 from http import HTTPStatus
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from time import time
 from typing import Any, BinaryIO, Callable, Dict, List, Optional, Sequence, Tuple, Union, cast
 from uuid import uuid4
@@ -534,15 +534,15 @@ class StudyService:
             self._assert_study_unarchived(study)
             study_settings = self.storage_service.get_storage(study).get(study, study_settings_url)
             study_settings["horizon"] = metadata_patch.horizon
-
             self._edit_study_using_command(study=study, url=study_settings_url, data=study_settings)
+
         if metadata_patch.author:
             study_antares_url = "study/antares"
             self._assert_study_unarchived(study)
             study_antares = self.storage_service.get_storage(study).get(study, study_antares_url)
             study_antares["author"] = metadata_patch.author
-
             self._edit_study_using_command(study=study, url=study_antares_url, data=study_antares)
+
         study.additional_data = study.additional_data or StudyAdditionalData()
         if metadata_patch.name:
             study.name = metadata_patch.name
@@ -1411,20 +1411,50 @@ class StudyService:
                 )
         raise NotImplementedError()
 
-    def _edit_study_using_command(self, study: Study, url: str, data: SUB_JSON) -> ICommand:
+    def _edit_study_using_command(
+        self,
+        study: Study,
+        url: str,
+        data: SUB_JSON,
+        *,
+        create_missing: bool = False,
+    ) -> ICommand:
         """
-        Replace data on disk with new, using ICommand
+        Replace data on disk with new, using variant commands.
+
+        In addition to regular configuration changes, this function also allows the end user
+        to store files on disk, in the "user" directory of the study (without using variant commands).
+
         Args:
             study: study
             url: data path to reach
             data: new data to replace
+            create_missing: Flag to indicate whether to create file or parent directories if missing.
         """
-
         study_service = self.storage_service.get_storage(study)
         file_study = study_service.get_raw(metadata=study)
-        tree_node = file_study.tree.get_node(url.split("/"))
+
+        file_relpath = PurePosixPath(url.strip().strip("/"))
+        file_path = study_service.get_study_path(study).joinpath(file_relpath)
+        create_missing &= not file_path.exists()
+        if create_missing:
+            # IMPORTANT: We prohibit deep file system changes in private directories.
+            # - File and directory creation is only possible for the "user" directory,
+            #   because the "input" and "output" directories are managed by Antares.
+            # - We also prohibit writing files in the "user/expansion" folder which currently
+            #   contains the Xpansion tool configuration.
+            #   This configuration should be moved to the "input/expansion" directory in the future.
+            if file_relpath and file_relpath.parts[0] == "user" and file_relpath.parts[1] != "expansion":
+                # In the case of variants, we must write the file directly in the study's snapshot folder,
+                # because the "user" folder is not managed by the command mechanism.
+                file_path.parent.mkdir(parents=True, exist_ok=True)
+                file_path.touch()
+
+        # A 404 Not Found error is raised if the file does not exist.
+        tree_node = file_study.tree.get_node(file_relpath.parts)  # type: ignore
 
         command = self._create_edit_study_command(tree_node=tree_node, url=url, data=data)
+
         if isinstance(study_service, RawStudyService):
             res = command.apply(study_data=file_study)
             if not is_managed(study):
@@ -1432,11 +1462,12 @@ class StudyService:
             if not res.status:
                 raise CommandApplicationError(res.message)
 
-            lastsave_url = "study/antares/lastsave"
-            lastsave_node = file_study.tree.get_node(lastsave_url.split("/"))
-            self._create_edit_study_command(tree_node=lastsave_node, url=lastsave_url, data=int(time())).apply(
-                file_study
-            )
+            # noinspection SpellCheckingInspection
+            url = "study/antares/lastsave"
+            last_save_node = file_study.tree.get_node(url.split("/"))
+            cmd = self._create_edit_study_command(tree_node=last_save_node, url=url, data=int(time()))
+            cmd.apply(file_study)
+
             self.storage_service.variant_study_service.invalidate_cache(study)
 
         elif isinstance(study_service, VariantStudyService):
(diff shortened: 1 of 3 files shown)


ghost:
### Tests

*  **api-raw:** check the case where the required path points to an empty matrix ([fefa868](https://github.com/AntaresSimulatorTeam/AntaREST/commit/fefa868d0c1e189b22a6ce461aae31033f767390))
*  **api-raw:** add a UT to check resource update with `create_missing=True` ([c14aef6](https://github.com/AntaresSimulatorTeam/AntaREST/commit/c14aef61a0661c861d27e62b2ac59e0403d143c5))


### Documentation

*  **api-raw:** improve the documentation and the typing of the `/studies/{uuid}/raw` endpoints ([9fb8e06](https://github.com/AntaresSimulatorTeam/AntaREST/commit/9fb8e06a16c9d5545330b16f378786b17464bd99))


### Features

*  **api-raw:** add the `create_missing` flag to allow creating missing files when uploading a file. ([78b10ba](https://github.com/AntaresSimulatorTeam/AntaREST/commit/78b10bad439467ed7a1654c369bc179f567fdb63))


### Chore

*  **api-raw:** fix typo in unit tests ([8446734](https://github.com/AntaresSimulatorTeam/AntaREST/commit/84467347997af48b0b066fe9b3f841273525ca4e))


### Contributors

<a href="https://github.com/laurent-laporte-pro">laurent-laporte-pro</a>

<details>
<summary>Commit-Lint commands</summary>
<br />

You can trigger Commit-Lint actions by commenting on this PR:
- `@Commit-Lint merge patch` will merge dependabot PR on "patch" versions (X.X.Y - Y change)
- `@Commit-Lint merge minor` will merge dependabot PR on "minor" versions (X.Y.Y - Y change)
- `@Commit-Lint merge major` will merge dependabot PR on "major" versions (Y.Y.Y - Y change)
- `@Commit-Lint merge disable` will desactivate merge dependabot PR
- `@Commit-Lint review` will approve dependabot PR
- `@Commit-Lint stop review` will stop approve dependabot PR



</details>

antarest/study/service.py
```diff
@@ -1411,32 +1411,63 @@ def _create_edit_study_command(
...
+            create_missing: Flag to indicate whether to create file or parent directories if missing.
         """
-
         study_service = self.storage_service.get_storage(study)
         file_study = study_service.get_raw(metadata=study)
-        tree_node = file_study.tree.get_node(url.split("/"))
+
+        file_relpath = PurePosixPath(url.strip().strip("/"))
+        file_path = study_service.get_study_path(study).joinpath(file_relpath)
+        create_missing &= not file_path.exists()
```
MartinBelthle, line 1439: This comment is only here to be sure we're on the same page.

You wrote this in the PR : `Note that if the file targeted by the path does exist, there is no restriction and the user can modify it.`
But, this line of code does not allow the user to modify an existing file.
If an user wants to modify an existing file, he should let the flag to `False` as it was done before.

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
```diff
@@ -120,25 +166,43 @@ def test_get_study(
...
         )
-        res.raise_for_status()
+        assert res.status_code == 200, res.json()
         actual = res.text
         actual_lines = actual.splitlines()
         first_row = [float(x) for x in actual_lines[0].split("\t")]
-        assert first_row == [100000, 100000, 0.010000, 0.010000, 0, 0, 0, 0]
+        assert first_row == [100000, 100000, 0.01, 0.01, 0, 0, 0, 0]
+
+        # If wa ask for and empty matrix, we should have an empty binary content
```
MartinBelthle: **typo**: `wa` -> `we` and `and` -> `an`

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
```diff
@@ -86,19 +87,64 @@ def test_get_study(
...
+
+        # You can check that the resource has been created or updated.
+        res = client.get(
+            f"/v1/studies/{study_id}/raw",
+            params={"path": "user/somewhere/something.txt"},
+            headers=headers,
+        )
+        assert res.status_code == 200, res.json()
+        assert res.content == b"Goodbye Cruel World!"
+
```
MartinBelthle, line 149: You could probably add a test with the flag to `True` and a path leading to an existing file.

antarest/study/service.py
```diff
@@ -1411,32 +1411,63 @@ def _create_edit_study_command(
...
+        if create_missing:
+            # IMPORTANT: We prohibit deep file system changes in private directories.
+            # - File and directory creation is only possible for the "user" directory,
+            #   because the "input" and "output" directories are managed by Antares.
+            # - We also prohibit writing files in the "user/expansion" folder which currently
+            #   contains the Xpansion tool configuration.
+            #   This configuration should be moved to the "input/expansion" directory in the future.
+            if file_relpath and file_relpath.parts[0] == "user" and file_relpath.parts[1] != "expansion":
+                # In the case of variants, we must write the file directly in the study's snapshot folder,
+                # because the "user" folder is not managed by the command mechanism.
```
MartinBelthle, line 1449: Indeed, so it's temporary and not really satisfying.
The command `update_raw_file` could do what we want but I'm not sure if it works with unexisting files :/

MartinBelthle (changes requested):

antarest/study/service.py
laurent-laporte-pro: I think you misunderstood the logic of this `&=` operator. Here is the truth table:

- MISS <=> `create_missing`
- EXIST <=> `file_path.exists()`
- MISS and not EXIST <=> `create_missing &= not file_path.exists()`

| MISS  | EXIST | MISS and not EXIST | Comment                                                  |
|-------|-------|--------------------|----------------------------------------------------------|
| False | False | False              | Don't create missing and file is missing => don't create |
| False | True  | False              | Don't create missing and file exists => don't create     |
| True  | False | True               | Create missing and file is missing => create             |
| True  | True  | False              | Create missing but file exists => don't create           |

So, if the user wants to allow missing file creation and the target file is actually missing, the function will create it with a `touch()`.

But, there is no restriction if the file already exit. The user can update it. This is what is done in line 1454.

antarest/study/service.py
laurent-laporte-pro: What do you mean by "temporary"?

The user can upload a file to create the target file if he set `create_missing=True`, so it works for a missing file in that case only.

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
laurent-laporte-pro: This is exactly what is it done in the previous test, because the default value of `create_missing` is `False`:

```python
# If you want to update an existing resource, you can use PUT method.
# But, if the resource doesn't exist, you should have a 404 Not Found error.
res = client.put(
    f"/v1/studies/{study_id}/raw",
    params={"path": "user/somewhere/something.txt"},
    headers=headers,
    files={"file": io.BytesIO(b"Goodbye World!")},
)
assert res.status_code == 404, res.json()
assert res.json() == {
    "description": "'somewhere' not a child of User",
    "exception": "ChildNotFoundError",
}
```

laurent-laporte-pro: Done

antarest/study/service.py
MartinBelthle: Indeed, you're right I got confused

antarest/study/service.py
MartinBelthle: What I mean is that you will directly write in the snapshot which is by desing temporary.
The right way to do it would be to create a command because when the snapshot will be cleaned, this file will disappear forever.
But as I wrote in my previous comment, i think the `update_raw_file` command does not work for unexisting files.
So this solution is not optimal but i do not see any workaround

MartinBelthle (approved):

PR #1828, opened 2023-11-23
perf(variants-generation): correct slow variant study snapshot generation

laurent-laporte-pro:
## Description

This pull request tackles a speed issue where users face delays when creating a study variant.
The problem arises from unnecessary copying of outputs from the parent study (or ancestor) during the snapshot generation.
This extra copy process could take a long time and cause timeouts for users.

In this pull request, we've fixed the problem by eliminating the need to copy outputs from the reference study
during variant snapshot generation.

## Changes Made

1. Adjusted the snapshot generation process to avoid unnecessary copying of outputs from the parent study or ancestor.
2. Ensured that the variant snapshot is now generated efficiently without the previously slow copy operation.

## Testing

- Thoroughly tested the modified snapshot generation process to ensure efficiency.
- Verified that removing redundant copying doesn't impact the accuracy of the generated variant snapshot.

## Instructions for Reviewers

- [ ] Confirm the removal of unnecessary output copying as described.
- [ ] Check the improved speed by assessing the time taken for generating study variant snapshots.
- [ ] Ensure that the changes don't introduce new issues or compromise the accuracy of the snapshots.

## Additional Information

- This pull request addresses [issue number].
- Local testing has been completed, considering various scenarios.

### DIFF
--- a/antarest/study/storage/utils.py
+++ b/antarest/study/storage/utils.py
@@ -390,12 +390,8 @@ def export_study_flat(
 
     stop_time = time.time()
     duration = "{:.3f}".format(stop_time - start_time)
-    logger.info(
-        {
-            False: f"Study '{study_dir}' exported (flat mode) in {duration}s",
-            True: f"Study '{study_dir}' exported with outputs (flat mode) in {duration}s",
-        }[bool(outputs)]
-    )
+    with_outputs = "with outputs" if outputs else "without outputs"
+    logger.info(f"Study '{study_dir}' exported ({with_outputs}, flat mode) in {duration}s")
     study = study_factory.create_from_fs(dest, "", use_cache=False)
     if denormalize:
         study.tree.denormalize()


antarest/study/storage/utils.py
```diff
@@ -390,9 +390,14 @@ def export_study_flat(
 
     stop_time = time.time()
     duration = "{:.3f}".format(stop_time - start_time)
-    logger.info(f"Study {path_study} exported (flat mode) in {duration}s")
+    logger.info(
+        {
+            False: f"Study '{study_dir}' exported (flat mode) in {duration}s",
+            True: f"Study '{study_dir}' exported with outputs (flat mode) in {duration}s",
+        }[bool(outputs)]
```
sylvlecl: Do we really need to cast ?
Why not ternary syntax instead ? :) " ... if outputs else ..."

sylvlecl (dismissed):

antarest/study/storage/utils.py
laurent-laporte-pro: Better: add "with outputs" or "without outputs" according to the boolean value

PR #1847, opened 2023-11-30
Demonstrate N+1 queries problem

sylvlecl:
**Description**

This is not an actual work to be merged, just a test for demonstration purpose.

The test queries the study service for the list of studies. The log show that multiple queries are executed for each study present in the database (N+1 queries problem).

### DIFF
--- a/tests/storage/test_service.py
+++ b/tests/storage/test_service.py
@@ -54,7 +54,7 @@ from antarest.study.storage.rawstudy.model.filesystem.config.model import (
     Link,
     Simulation,
 )
-from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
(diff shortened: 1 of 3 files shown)


tests/storage/test_service.py
```diff
@@ -1822,3 +1822,82 @@ def test_upgrade_study__raw_study__failed(upgrade_study_mock: Mock, tmp_path: Pa
...
+    db.session.add(bob)
+    db.session.add(alice)
+    db.session.commit()
+
+    import logging
+    logging.getLogger('sqlalchemy.engine').setLevel(logging.DEBUG)
+    logging.root.setLevel(logging.DEBUG)
+
+    # noinspection PyArgumentList
+    studies = study_service.get_studies_information(
```
sylvlecl, line 1892: If you look at the queries logs, you'll see queries for study A, then queries for study B, etc.

PR #1854, opened 2023-12-08
perf(variant): avoid snapshot generation if the snapshot directory already exists

laurent-laporte-pro:
Cette PR permet de corriger (en partie) le problème de lenteur de la génération des variants en évitant de générer le snapshot s'il existe déjà.

En effet, lorsqu'un variant est modifié, son snapshot n'est plus à jour et il suffit d'appliquer les nouvelles commandes pour le mettre à jour.

Dans de rares cas, par exemple si l'utilisateur modifie l'historique des commandes, le snapshot ne sera pas mis à jour correctement. Cette situation n'est pour l'instant pas gérée automatiquement.

### DIFF
--- a/antarest/study/storage/variantstudy/snapshot_generator.py
+++ b/antarest/study/storage/variantstudy/snapshot_generator.py
@@ -4,7 +4,6 @@ This module dedicated to variant snapshot generation.
 import datetime
 import logging
 import shutil
-import tempfile
 import typing as t
 from pathlib import Path
 
@@ -51,8 +50,6 @@ class SnapshotGenerator:
         self.study_factory = study_factory
         self.patch_service = patch_service
         self.repository = repository
-        # Temporary directory used to generate the snapshot
-        self._tmp_dir: Path = Path()
 
     def generate_snapshot(
         self,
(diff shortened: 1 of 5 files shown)


antarest/study/storage/variantstudy/snapshot_generator.py
```diff
@@ -77,22 +77,39 @@ def generate_snapshot(
...
-        # at the end of the process. This prevents incomplete snapshots in case of error.
-
-        # Get snapshot directory and prepare a temporary directory next to it.
+        # Get snapshot directory
         variant_study = descendants[-1]
         snapshot_dir = variant_study.snapshot_dir
-        snapshot_dir.parent.mkdir(parents=True, exist_ok=True)
-        self._tmp_dir = Path(tempfile.mkdtemp(dir=snapshot_dir.parent, prefix=f"~{snapshot_dir.name}", suffix=".tmp"))
+
+        if snapshot_dir.exists():
```
sylvlecl: I think this is not a sufficient condition, there are many conditions under which this snapshot dir will not be consistent with the cmd_block we get from `search_ref_study`:
- if we generate from scratch, we must not re-use this previously generated snapshots, or we will apply all commands on something where other commands have already been applied
- if snapshots are identified as "old" in search_ref_study, we will be in the same situation
- if we fall in the "unlikely" case where last_exec_cmd is None, same thing

We need to think of a more robust approach!
Side note:
I am not convinced it's worth keeping the temp directory approach, since we will not be able to rely on it anyway

sylvlecl: Also I think we have an issue in `search_ref_study` at lines 281-284:
```python
command_ids = [c.id for c in ref_study.commands]
last_exec_index = command_ids.index(last_exec_cmd)
cmd_blocks = ref_study.commands[last_exec_index + 1 :]
```
If the last executed command is not anymore in the list of commands (it has been deleted), there will probably be an exception or at least an unexpected behaviour.

sylvlecl (changes requested):
See comments: I think this implementation will introduce some bugs

antarest/study/storage/variantstudy/snapshot_generator.py
laurent-laporte-pro: I changed the `search_ref_study` so that it returns a `force_regenerate` flag to force the snapshot regeneration.
This flag should fix the issues.
But, we need to discuss about "if snapshots are identified as "old" in search_ref_study, we will be in the same situation"...

PR #1870, opened 2023-12-19
fix(services): rollback the loop for services

mabw-rte:
Context: 
- In previous PR #1837 we found no direct usage of SingletonServices but in the `main` program. So we decided to remove it and to include all of its functionalities in a function `start_all_services`.
- When we included all the functionalities in `start_all_services`, we noticed a loop that does no thing but running for eternity 
- From there we decided to remove that loop

Issue:
- That loop was essential to keep the program and thus the services running when we only want build containers for services without the application 

Solution:
- re-insert the loop in start_all_services

### DIFF
--- a/antarest/singleton_services.py
+++ b/antarest/singleton_services.py
@@ -1,3 +1,4 @@
+import time
 from pathlib import Path
 from typing import Dict, List, cast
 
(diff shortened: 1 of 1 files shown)


laurent-laporte-pro (approved):

sylvlecl:
I just thought that the more clear/standard way of doing this would probably simply be to join the services threads (Thread.join method).

PR #1915, opened 2024-01-30
fix(disk-usage): fix bug on variants

MartinBelthle:
If the study is a variant, it's possible that it only exists in db and not on disk.
If so, we used to throw a HTTP 500 Exception. Now, we return 0.

### DIFF
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -2373,5 +2373,6 @@ class StudyService:
         """
         study = self.get_study(uuid=uuid)
         assert_permission(params.user, study, StudyPermissionType.READ)
-        path = str(self.storage_service.get_storage(study).get_study_path(study))
-        return get_disk_usage(path=path)
+        study_path = self.storage_service.get_storage(study).get_study_path(study)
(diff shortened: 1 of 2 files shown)


antarest/study/service.py
```diff
@@ -2373,5 +2373,6 @@ def get_disk_usage(self, uuid: str, params: RequestParameters) -> int:
         """
         study = self.get_study(uuid=uuid)
         assert_permission(params.user, study, StudyPermissionType.READ)
-        path = str(self.storage_service.get_storage(study).get_study_path(study))
-        return get_disk_usage(path=path)
+        study_path = self.storage_service.get_storage(study).get_study_path(study)
+        # If the study is a variant, it's possible that it only exists in db and not on disk. If so, we return 0.
+        return get_disk_usage(path=str(study_path)) if study_path.exists() else 0
```
laurent-laporte-pro, line 2378: 👍🏻

laurent-laporte-pro (changes requested):
Unfortunately, the calculation of the size occupied by a variant when the variant has a snapshot does not work. This needs to be diagnosed and corrected.

laurent-laporte-pro (approved):

PR #1934, opened 2024-02-14
fix(tags-db): correct `tag` and `study_tag` migration script

laurent-laporte-pro:
Avoid bulk insertion if the list of values to insert is empty.

fix #1929


mabw-rte (changes requested):
Il faut utiliser une liste.

mabw-rte (approved):

PR #1945, opened 2024-02-23
fix(ui): UI fixes

### LINKED ISSUE #1409: Add timestamp in results
## Problem Statement

In the "results" tab, it's currently impossible to determine what timestamp corresponds to a given entry.

image

## Proposed Solution

Add a special column that contains the timestamp (hour, day, week, etc.)

image

## Developer Notes

- Any potential roadblocks or concerns: [e.g. dependencies on other features, limitations in technology]

hdinia:


webapp/src/components/App/Singlestudy/explore/Results/ResultDetails/index.tsx
```diff
@@ -148,15 +149,27 @@ function ResultDetails() {
     },
   );
 
+  // !NOTE: Workaround to display the date in the correct format, to be replaced by a proper solution.
+  const dateTimeFromIndex = useMemo(() => {
```
skamril, line 153: Add it in `usePromise` instead of creating a `useMemo`.

webapp/src/components/common/EditableMatrix/style.ts
```diff
@@ -6,7 +6,7 @@ export const Root = styled(Box)(({ theme }) => ({
   display: "flex",
   flexDirection: "column",
   alignItems: "center",
-  overflow: "auto",
+  overflow: "hidden",
```
skamril, line 9: Are you sure you don't break anything? Test with big matrix.

skamril (changes requested):

webapp/src/components/App/Singlestudy/explore/Results/ResultDetails/index.tsx
hdinia: It's better to not mix it with the matrix promise and keep clear separation of concerns. For the `useMemo`, the `index` array in `matrixRes.data` is large, making date formatting expensive. If we trigger a filter for example it will be re-calculated.
don't you think we should keep it?

webapp/src/components/common/EditableMatrix/style.ts
hdinia: Handsontable docs recommands the parent container overflow is set to "hidden", normally it doesn't break anything i've tested

webapp/src/components/App/Singlestudy/explore/Results/ResultDetails/index.tsx
skamril: Because it's workaround before the backend fix it, I think it's correct to place it in the usePromise.
As you want.

webapp/src/components/common/EditableMatrix/style.ts
skamril: Doc recommands `auto` or `hidden`. If you're sure ok.

skamril (dismissed):

skamril (approved):

PR #1952, opened 2024-02-27
fix(ui-tablemode): 'co2' column not working

skamril:
ANT-1270


webapp/src/services/api/studies/tableMode/index.ts
```diff
@@ -15,7 +15,7 @@ export async function getTableMode<T extends TableModeType>(
   const res = await client.get(url, {
     params: {
       table_type: type,
-      columns: columns.map(snakeCase).join(","),
+      columns: columns.map(toColumnApiName).join(","),
```
hdinia, line 18:
```javascript
columns: columns
        .map((column) => (column !== "co2" ? snakeCase(column) : column))
        .join(","),
```

hdinia (approved):

PR #1958, opened 2024-02-29
fix(outputs): build outputs config even when using cache

### LINKED ISSUE #1936: [AntaresWeb] Output disappeared
## Description
I launched simulations on 3 studies which are variants of the same initial study. All three simulations went well and the results were initially accessible.
When trying to access them several hours later, the results had disappeared: they can't be seen in the "Results" tab or in the debug view. However, they are still downloadable from the "Tasks" view.

The outputs of 2 studies were accessible again several hours later, but they are still invisible for one of the studies.
The size of this study is identical to the size of a study with outputs, suggesting that the outputs are there.

## Steps to Reproduce
Initial study ID: 390e2a89-2abe-4707-bf59-bd1877d8429e
Variant study ID which output has disappeared: 7aa4f373-5e3e-40ff-be02-88c76c160eca
Antares version of the study: V8.0

MartinBelthle:
Resolves [ANT-1314]

### DIFF
--- a/antarest/study/storage/rawstudy/model/filesystem/config/files.py
+++ b/antarest/study/storage/rawstudy/model/filesystem/config/files.py
@@ -74,7 +74,7 @@ def build(study_path: Path, study_id: str, output_path: t.Optional[Path] = None)
         version=_parse_version(study_path),
         areas=_parse_areas(study_path),
         sets=_parse_sets(study_path),
-        outputs=_parse_outputs(outputs_dir),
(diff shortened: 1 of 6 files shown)


MartinBelthle:
The test is good and the code works but it's more a workaround than a solution.
Because, i did not investigate why we lose the information at some point :/

laurent-laporte-pro (dismissed):

laurent-laporte-pro:
Good job! I only improve the unit tests to use fixtures.

PR #1969, opened 2024-03-04
fix(desktop): correct configuration and launcher load indicator

laurent-laporte-pro:
- Correct the configuration file for the Desktop version: add missing configuration for watcher
- Correct the launcher load indicator to be in range (0, 100)
- Update the packaging script to use Antares Solver v8.8.3 (new bug fix version)

### DIFF
--- a/scripts/package_antares_web.sh
+++ b/scripts/package_antares_web.sh
@@ -9,7 +9,7 @@
 set -e
 
 ANTARES_SOLVER_VERSION="8.8"
-ANTARES_SOLVER_FULL_VERSION="8.8.2"
+ANTARES_SOLVER_FULL_VERSION="8.8.3"
 ANTARES_SOLVER_VERSION_INT="880"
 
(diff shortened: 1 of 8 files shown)


scripts/package_antares_web.sh
```diff
@@ -31,6 +31,7 @@ echo "INFO: Preparing the Git Commit ID..."
 git log -1 HEAD --format=%H > ${RESOURCES_DIR}/commit_id
 
 echo "INFO: Generating the Desktop version of the Web Application..."
+rm -rf ${DIST_DIR}
```
MartinBelthle, line 35: Why do you remove this ? Because you use this path later in the script, ex : 
```pyinstaller --distpath ${DIST_DIR} AntaresWebWin.spec```

MartinBelthle (dismissed):

scripts/package_antares_web.sh
laurent-laporte-pro: This directory removal is required only in scenarios where the script is executed twice. This situation occurs in development mode.

MartinBelthle (approved):

Issue #1783, opened 2023-10-26
Rework LauncherDialog for Improved User and Technical Experience


hdinia:
## Problem Statement

🔍 Overview
The `LauncherDialog` currently holds legacy code and design patterns that contribute to a subpar user experience and various technical inefficiencies. To rectify this, a full-scale redesign and refactoring are necessary to modernize the component, improve performance, and enhance usability.

Problems:
- Outdated state management leading to performance issues.
- No default values nor validation of inputs (that can lead to major bugs)
- Lecagy form fields that needs the support of react-hook-form to centralize the state
- Multiple endpoints for data fetching, causing inefficiencies.
- Cluttered and complicated user interface.

## Proposed Solution

🛠 Redesign and Refactoring Goals:
- Modernize state management to improve performance and maintainability.
- Refactor the form 
- Include validators and default values constraints 
- Consolidate multiple endpoints into a single, dedicated endpoint.
- Revamp the user interface for a cleaner and more intuitive experience:
 Apply modern UX/UI principles to make the interface more user-friendly, intuitive, and less cluttered, enhancing the overall user experience and interaction

## Developer Notes

Tasks:

- [ ]  Conduct a detailed analysis and planning of the redesign and refactoring.
- [ ]  Execute the redesign, focusing on state management, endpoints consolidation, and UI improvements.
- [ ]  Conduct rigorous testing to ensure performance and usability enhancements are realized.

PR #1976, opened 2024-03-11
fix(launcher): adds default time limit submission

hdinia:
Related to: https://gopro-tickets.rte-france.com/browse/ANT-1271

### Description:

This pull request fixes an issue in the `LauncherDialog` component where the `time_limit` state was not correctly set to a default value if the user modified fields other than the `time limit` itself before launching a simulation. 
Despite the UI correctly displaying the default time limit (converted from seconds to hours for user readability), the internal state `time_limit` remained unset if the user did not explicitly change the time limit field. 
This could lead to unintended behavior when submitting the form.

### Key Changes:

Initialization of `time_limit` State:
Ensured the time_limit state is initialized with DEFAULT_TIME_LIMIT in the component's initial state. This guarantees that the state correctly reflects the displayed default time limit even if the user does not interact with the time limit field.

### DIFF
--- a/antarest/launcher/model.py
+++ b/antarest/launcher/model.py
@@ -25,7 +25,7 @@ class LauncherParametersDTO(BaseModel):
     adequacy_patch: t.Optional[t.Dict[str, t.Any]] = None
     nb_cpu: t.Optional[int] = None
     post_processing: bool = False
-    time_limit: t.Optional[int] = 240 * 3600  # Default value set to 240 hours (in seconds)
+    time_limit: int = 240 * 3600  # Default value set to 240 hours (in seconds)
     xpansion: t.Union[XpansionParametersDTO, bool, None] = None
     xpansion_r_version: bool = False
     archive_output: bool = True
--- a/antarest/launcher/model.py
+++ b/antarest/launcher/model.py
@@ -25,13 +25,14 @@ class LauncherParametersDTO(BaseModel):
     adequacy_patch: t.Optional[t.Dict[str, t.Any]] = None
     nb_cpu: t.Optional[int] = None
     post_processing: bool = False
-    time_limit: t.Optional[int] = None  # 3600 ≤ time_limit < 864000 (10 days)
+    time_limit: t.Optional[int] = 240 * 3600  # Default value set to 240 hours (in seconds)
     xpansion: t.Union[XpansionParametersDTO, bool, None] = None
     xpansion_r_version: bool = False
     archive_output: bool = True
     auto_unzip: bool = True
     output_suffix: t.Optional[str] = None
     other_options: t.Optional[str] = None
+
     # add extensions field here
 
     @classmethod
--- a/webapp/src/components/App/Studies/LauncherDialog.tsx
+++ b/webapp/src/components/App/Studies/LauncherDialog.tsx
@@ -38,6 +38,7 @@ import CheckBoxFE from "../../common/fieldEditors/CheckBoxFE";
 import { convertVersions } from "../../../services/utils";
 import UsePromiseCond from "../../common/utils/UsePromiseCond";
 import SwitchFE from "../../common/fieldEditors/SwitchFE";
+import moment from "moment";
 
 const DEFAULT_NB_CPU = 22;
 const DEFAULT_TIME_LIMIT = 240 * 3600; // 240 hours in seconds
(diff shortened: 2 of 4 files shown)


antarest/launcher/model.py
```diff
@@ -23,15 +23,16 @@ class LauncherParametersDTO(BaseModel):
     # The reason is that it's stored in json format in database and deserialized using the latest class version
     # If compatibility is to be broken, an (alembic) data migration script should be added
     adequacy_patch: t.Optional[t.Dict[str, t.Any]] = None
-    nb_cpu: t.Optional[int] = None
+    nb_cpu: t.Optional[int] = 22  # Default value set to 22 cores
```
laurent-laporte-pro: Cette valeur par défaut est acceptable pour la production mais fausse en recette :

```yaml
nb_cores_min: 1
nb_cores_default: 34
nb_cores_max: 36
```

laurent-laporte-pro: Il faut utiliser le endpoint `GET /v1/launcher/nbcores` pour obtenir les valeurs par défaut.

Exemple de réponse :
```json
{
  "min": 1,
  "defaultValue": 34,
  "max": 36
}
```

hdinia: ok merci, du coup tu est d'avis de retirer cette valeur, vu que c'est faux pour la recette ?
je pense qu'il faut une valeur de fallback quand même, si l'appel `/v1/launcher/nbcores`  échoue on donne 22, il n'y aurait pas une valeur correcte pour la prod + recette ?

hdinia: > Il faut utiliser le endpoint `GET /v1/launcher/nbcores` pour obtenir les valeurs par défaut.
> 
> Exemple de réponse :
> 
> ```json
> {
>   "min": 1,
>   "defaultValue": 34,
>   "max": 36
> }
> ```

C'est ce que l'on fait déjà. Donc la valeur par défaut en dur ne sert à rien ?

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -53,8 +54,9 @@ function LauncherDialog(props: Props) {
   const { enqueueSnackbar } = useSnackbar();
   const enqueueErrorSnackbar = useEnqueueErrorSnackbar();
   const [options, setOptions] = useState<LaunchOptions>({
-    nb_cpu: LAUNCH_LOAD_DEFAULT,
+    nb_cpu: DEFAULT_NB_CPU,
     auto_unzip: true,
+    time_limit: DEFAULT_TIME_LIMIT,
```
sylvlecl, line 60: We should do it for all other options or we'll have the same issue some day with those options.
In the type `LaunchOptions`, nothing should be optional, to ensure we always send all displayed values.

sylvlecl (changes requested):

webapp/src/components/App/Studies/LauncherDialog.tsx
hdinia: Thank you for your feedback. If I'm right the backend already assigns default values to these options in the `LauncherParametersDTO` class, which mitigates potential issues. 

The one missing were `nb_cpu` and `time_limit` that I added, and according to @laurent-laporte-pro  the `nb_cpu` can be tricky to handle as default because it's not the same for all environments, I need some more precisions about that.

Also please note that we're dealing with legacy code, and many issues comes from that as outlined in this issue: 
[ Rework LauncherDialog for Improved User and Technical Experience #1783 ](https://github.com/AntaresSimulatorTeam/AntaREST/issues/1783)

When the form will be refactored, all values will be validated against constraints(min,max, valid input etc) and default values will be explicitly set in the inital state of the from.

I'm leaning towards not duplicating defaults in the frontend, considering the backend already handles them and our current legacy code scenario. But I'm open to other perspectives. What’s your take? Looking forward on this.

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -170,12 +172,17 @@ function LauncherDialog(props: Props) {
...
-    }
+  /**
+   * Parses an hour value from a string and converts it to seconds.
+   * If the input is invalid, returns a default value (240 hours).
+   *
+   * @param hourString - A string representing the number of hours.
+   * @returns The equivalent number of seconds, or a default value for invalid inputs.
+   */
+  const parseHoursToSeconds = (hourString: string): number => {
+    const hours = parseInt(hourString, 10);
```
skamril:
```typescript
const seconds = moment.duration(hourString, 'hours').asSeconds();
return seconds > 0 ? seconds : DEFAULT_TIME_LIMIT;
```

hdinia: I think its overkill to use moment for this, but thanks for the alternative

skamril: Is already imported, so no problem

hdinia: Not only about bundle size, JS native parse is faster than moment 🐢  and I don't see the point using an external lib for such simple thing. But let's add it for readability if you prefer this, it's true that it's a bit simpler

laurent-laporte-pro (approved):

PR #1982, opened 2024-03-18
feat(bc): add constraints groups endpoints and constraints service enhancements

MartinBelthle:
1- Addition of validation endpoint
2- Add response model to PUT and POST endpoints
3- PUT endpoint does not crash anymore to let the front-end update bc groups.

### DIFF
--- a/antarest/study/business/binding_constraint_management.py
+++ b/antarest/study/business/binding_constraint_management.py
@@ -1,7 +1,7 @@
 import collections
 import itertools
 import logging
-from typing import Any, Dict, List, Mapping, MutableSequence, Optional, Sequence, Union
+from typing import Any, Dict, List, Mapping, MutableSequence, Optional, Sequence, Type, Union
 
 import numpy as np
 from pydantic import BaseModel, Field, root_validator, validator
@@ -351,40 +351,35 @@ class BindingConstraintManager:
         self.storage_service = storage_service
 
     @staticmethod
-    def parse_constraint(key: str, value: str, char: str, new_config: ConstraintOutput) -> bool:
-        split = key.split(char)
-        if len(split) == 2:
-            value1 = split[0]
-            value2 = split[1]
-            weight = 0.0
-            offset = None
-            try:
-                weight = float(value)
-            except ValueError:
-                weight_and_offset = value.split("%")
-                if len(weight_and_offset) == 2:
-                    weight = float(weight_and_offset[0])
-                    offset = float(weight_and_offset[1])
-            if new_config.terms is None:
-                new_config.terms = []
-            new_config.terms.append(
-                ConstraintTerm(
-                    id=key,
-                    weight=weight,
-                    offset=offset if offset is not None else None,
-                    data=LinkTerm(
-                        area1=value1,
-                        area2=value2,
+    def parse_and_add_terms(
+        key: str, value: Any, adapted_constraint: Union[ConstraintOutputBase, ConstraintOutput870]
+    ) -> None:
+        """Parse a single term from the constraint dictionary and add it to the adapted_constraint model."""
+        if "%" in key or "." in key:
+            separator = "%" if "%" in key else "."
+            term_data = key.split(separator)
+            weight, offset = (value, None) if isinstance(value, (float, int)) else map(float, value.split("%"))
+
+            if separator == "%":
+                # Link term
+                adapted_constraint.terms.append(
+                    ConstraintTerm(
+                        id=key,
+                        weight=weight,
+                        offset=offset,
+                        data={
+                            "area1": term_data[0],
+                            "area2": term_data[1],
+                        },
                     )
-                    if char == "%"
-                    else ClusterTerm(
-                        area=value1,
-                        cluster=value2,
-                    ),
                 )
-            )
-            return True
-        return False
+            # Cluster term
+            else:
+                adapted_constraint.terms.append(
+                    ConstraintTerm(
+                        id=key, weight=weight, offset=offset, data={"area": term_data[0], "cluster": term_data[1]}
+                    )
+                )
 
     @staticmethod
     def constraint_model_adapter(constraint: Mapping[str, Any], version: int) -> ConstraintOutput:
@@ -409,6 +404,10 @@ class BindingConstraintManager:
         ensures data integrity when storing or retrieving constraint configurations from the database.
         """
 
+        ConstraintModel: Type[Union[ConstraintOutputBase, ConstraintOutput870]] = (
+            ConstraintOutput870 if version >= 870 else ConstraintOutputBase
+        )
+
         constraint_output = {
             "id": constraint["id"],
             "name": constraint["name"],
@@ -416,22 +415,28 @@ class BindingConstraintManager:
             "time_step": constraint.get("type", BindingConstraintFrequency.HOURLY),
             "operator": constraint.get("operator", BindingConstraintOperator.EQUAL),
             "comments": constraint.get("comments", ""),
-            "filter_year_by_year": constraint.get("filter_year_by_year") or constraint.get("filter-year-by-year"),
-            "filter_synthesis": constraint.get("filter_synthesis") or constraint.get("filter-synthesis"),
             "terms": constraint.get("terms", []),
         }
 
-        if version < 870:
-            adapted_constraint = ConstraintOutputBase(**constraint_output)
-        else:
-            constraint_output["group"] = constraint.get("group", DEFAULT_GROUP)
-            adapted_constraint = ConstraintOutput870(**constraint_output)
+        if version >= 840:
+            constraint_output["filter_year_by_year"] = constraint.get("filter_year_by_year", "") or constraint.get(
+                "filter-year-by-year", ""
+            )
+            constraint_output["filter_synthesis"] = constraint.get("filter_synthesis", "") or constraint.get(
+                "filter-synthesis", ""
+            )
 
-        for key, value in constraint.items():
-            if BindingConstraintManager.parse_constraint(key, value, "%", adapted_constraint):
-                continue
-            if BindingConstraintManager.parse_constraint(key, value, ".", adapted_constraint):
-                continue
+        if version >= 870:
+            constraint_output["group"] = constraint.get("group", DEFAULT_GROUP)
+
+        adapted_constraint = ConstraintModel(**constraint_output)
+
+        # If 'terms' were not directly provided in the input, parse and add terms dynamically
+        if not constraint.get("terms"):
+            for key, value in constraint.items():
+                if key not in constraint_output:  # Avoid re-processing keys already included
+                    BindingConstraintManager.parse_and_add_terms(key, value, adapted_constraint)
+
         return adapted_constraint
 
     @staticmethod
--- a/antarest/study/business/binding_constraint_management.py
+++ b/antarest/study/business/binding_constraint_management.py
@@ -178,36 +178,32 @@ class ConstraintFilters(BaseModel, frozen=True, extra="forbid"):
         if self.time_step is not None and self.time_step != constraint.time_step:
             return False
 
-        # Filter on terms
         terms = constraint.terms or []
 
         if self.area_name:
             matching_terms = []
-
             for term in terms:
                 if term.data:
                     if isinstance(term.data, LinkTerm):
-                        # Check if either area in the link matches the specified area_name
+                        # Check if either area in the link term matches the specified area_name.
                         if self.area_name.upper() in (term.data.area1.upper(), term.data.area2.upper()):
                             matching_terms.append(term)
                     elif isinstance(term.data, ClusterTerm):
-                        # Check if the cluster's area matches the specified area_name
+                        # Check if the area matches the specified area_name for a cluster term.
                         if self.area_name.upper() == term.data.area.upper():
                             matching_terms.append(term)
-
-            # If no terms match, the constraint should not pass
             if not matching_terms:
                 return False
 
         if self.cluster_name:
-            all_clusters = []
+            matching_terms = []
             for term in terms:
                 if term.data is None:
                     continue
-                if isinstance(term.data, ClusterTerm):
-                    all_clusters.append(term.data.cluster)
-            upper_cluster_name = self.cluster_name.upper()
-            if all_clusters and not any(upper_cluster_name in cluster.upper() for cluster in all_clusters):
+                if term.data and isinstance(term.data, ClusterTerm):
+                    if self.cluster_name.upper() == term.data.cluster.upper():
+                        matching_terms.append(term)
+            if not matching_terms:
                 return False
 
         if self.link_id:
@@ -448,14 +444,12 @@ class BindingConstraintManager:
         :return: A dictionary of term IDs mapped to a list of their coefficients.
         """
         coeffs = {}
-
         if terms is not None:
             for term in terms:
                 if term.id and term.weight is not None:
                     coeffs[term.id] = [term.weight]
                     if term.offset is not None:
                         coeffs[term.id].append(term.offset)
-
             return coeffs
 
     def get_binding_constraint(
--- a/antarest/study/business/binding_constraint_management.py
+++ b/antarest/study/business/binding_constraint_management.py
@@ -105,7 +105,7 @@ class ConstraintTerm(BaseModel):
 
     id: Optional[str]
     weight: Optional[float]
-    offset: Optional[float]
+    offset: Optional[int]
     data: Optional[Union[LinkTerm, ClusterTerm]]
 
     @validator("id")
@@ -362,7 +362,12 @@ class BindingConstraintManager:
         if "%" in key or "." in key:
             separator = "%" if "%" in key else "."
             term_data = key.split(separator)
-            weight, offset = (value, None) if isinstance(value, (float, int)) else map(float, value.split("%"))
+            if isinstance(value, (float, int)):
+                weight, offset = (float(value), None)
+            else:
+                _parts = value.partition("%")
+                weight = float(_parts[0])
+                offset = int(_parts[2]) if _parts[2] else None
 
             if separator == "%":
                 # Link term
@@ -455,7 +460,7 @@ class BindingConstraintManager:
             for term in terms:
                 if term.id and term.weight is not None:
                     coeffs[term.id] = [term.weight]
-                    if term.offset is not None:
+                    if term.offset:
                         coeffs[term.id].append(term.offset)
             return coeffs
 
@@ -831,7 +836,7 @@ class BindingConstraintManager:
 
         for term in constraint_terms:
             coeffs[term.id] = [term.weight]
-            if term.offset is not None:
+            if term.offset:
                 coeffs[term.id].append(term.offset)
 
         command = UpdateBindingConstraint(
--- a/antarest/study/business/binding_constraint_management.py
+++ b/antarest/study/business/binding_constraint_management.py
@@ -391,7 +391,7 @@ class BindingConstraintManager:
         return False
 
     @staticmethod
-    def constraint_model_adapter(constraint: ConstraintInput, version: int) -> ConstraintOutput:
+    def constraint_model_adapter(constraint: Mapping[str, Any], version: int) -> ConstraintOutput:
         """
         Adapts a constraint configuration to the appropriate version-specific format.
 
@@ -479,10 +479,10 @@ class BindingConstraintManager:
 
         # If a specific constraint ID is provided, we return that constraint
         if filters.bc_id:
-            return filtered_constraints.get(filters.bc_id)
+            return filtered_constraints.get(filters.bc_id) # type: ignore
 
         # Else we return all the matching constraints, based on the given filters
-        return list(filtered_constraints.values())
+        return list(filtered_constraints.values())  
 
     def get_grouped_constraints(self, study: Study) -> Mapping[str, Sequence[ConstraintOutput]]:
         """
@@ -677,7 +677,7 @@ class BindingConstraintManager:
         }
 
         if study_version >= 870:
-            updated_constraint["group"] = data.group or existing_constraint.group
+            updated_constraint["group"] = data.group or existing_constraint.group  # type: ignore
 
         args = {
             **updated_constraint,
@@ -836,7 +836,7 @@ class BindingConstraintManager:
         binding_constraint_id: str,
         term_id: str,
     ) -> None:
-        return self.update_constraint_term(study, binding_constraint_id, term_id)
+        return self.update_constraint_term(study, binding_constraint_id, term_id)  # type: ignore
 
 
 def _replace_matrices_according_to_frequency_and_version(
--- a/antarest/study/business/binding_constraint_management.py
+++ b/antarest/study/business/binding_constraint_management.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional, Union, cast
+from typing import Any, Dict, List, Optional, Union
 
 from pydantic import BaseModel, validator
 
@@ -32,6 +32,7 @@ from antarest.study.storage.variantstudy.business.matrix_constants.binding_const
 from antarest.study.storage.variantstudy.business.matrix_constants.binding_constraint.series_before_v87 import (
     default_bc_weekly_daily as default_bc_weekly_daily_86,
 )
+from antarest.study.storage.variantstudy.model.command.common import BindingConstraintOperator
 from antarest.study.storage.variantstudy.model.command.create_binding_constraint import (
     BindingConstraintMatrices,
     BindingConstraintProperties,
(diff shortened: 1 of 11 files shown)


antarest/study/business/binding_constraint_management.py
```diff
@@ -111,9 +112,19 @@ def generate_id(self) -> str:
         return self.data.generate_id()
 
 
-class UpdateBindingConstProps(BaseModel):
-    key: str
-    value: Any
+class BindingConstraintForm(BaseModel, metaclass=AllOptionalMetaclass):
```
hdinia: suffix `Form` can be omitted.
in RESTful API design, models often serve as representations of resources rather than forms, a REST resource model typically goes beyond that by acting as a data structure for API responses and requests

antarest/study/business/binding_constraint_management.py
```diff
@@ -239,11 +250,26 @@ def get_binding_constraint(
             binding_constraint.append(new_config)
         return binding_constraint
 
+    def validate_binding_constraint(self, study: Study, constraint_id: str) -> None:
+        if int(study.version) < 870:
+            return  # There's nothing to check for constraints before v8.7
+        file_study = self.storage_service.get_storage(study).get_raw(study)
+        config = file_study.tree.get(["input", "bindingconstraints", "bindingconstraints"])
+        group = next((value["group"] for value in config.values() if value["id"] == constraint_id), None)
```
hdinia: if the group has no value this will raise a key error before your `BindingConstraintNotFoundError` making it useless, also consider a more appropriate exception like `InvalidConstraint` or anything else that suits you

To make your code more robust you can use `.get()`:
```python
 group = next((value.get("group") for value in config.values() if value.get("id") == constraint_id), None)
```

antarest/study/web/study_data_blueprint.py
```diff
@@ -937,15 +937,39 @@ def update_binding_constraint(
...
+    ) -> Any:
+        """
+        Validates the binding constraint configuration.
+
+        Parameters:
+        - `uuid`: The study UUID.
+        - `binding_constraint_id`: The binding constraint id to validate
+
+        For studies version prior to v8.7, does nothing.
+        Else, it checks the coherence between the group and the size of its columns.
```
hdinia: Its unclear to me, you mean the size of the matrices columns ?

antarest/study/web/study_data_blueprint.py
```diff
@@ -937,15 +937,39 @@ def update_binding_constraint(
...
 
-    @bp.post(
-        "/studies/{uuid}/bindingconstraints",
+    @bp.get(
+        "/studies/{uuid}/bindingconstraints/{binding_constraint_id}/validate",
         tags=[APITag.study_data],
-        summary="Create a binding constraint",
+        summary="Validate binding constraint configuration",
         response_model=None,
     )
```
hdinia: you can make the endpoint more clearly aligned with RESTful principles and to avoid confusion, you might consider:

If it's a read operation that returns validation results, for example, you could use a path like:
`/studies/{uuid}/bindingconstraints/{binding_constraint_id}/validation-status` 
or `/studies/{uuid}/bindingconstraints/{binding_constraint_id}/check`
 this suggests that you're retrieving the status of a validation check, rather than performing an action.

antarest/study/web/study_data_blueprint.py
```diff
@@ -937,15 +937,39 @@ def update_binding_constraint(
...
         return study_service.binding_constraint_manager.update_binding_constraint(study, binding_constraint_id, data)
 
-    @bp.post(
-        "/studies/{uuid}/bindingconstraints",
+    @bp.get(
+        "/studies/{uuid}/bindingconstraints/{binding_constraint_id}/validate",
         tags=[APITag.study_data],
-        summary="Create a binding constraint",
+        summary="Validate binding constraint configuration",
         response_model=None,
```
hdinia, line 1038: This status code can be used to indicate that the validation passed without errors but there is no content to send in the response (204)

```python
status_code=HTTPStatus.NO_CONTENT
```
Tip: In addition it's generally a good practice to inform your client about the result when performing such operations, you can add for example a json response containing a status and the requested object.

```json
{
"status": "A status message",
"constraint": { ... }
}
```
In this case the status code will be a 200

antarest/study/business/binding_constraint_management.py
```diff
@@ -111,9 +112,19 @@ def generate_id(self) -> str:
...
+    time_step: BindingConstraintFrequency
+    operator: BindingConstraintOperator
+    filter_year_by_year: str
+    filter_synthesis: str
+    comments: str
+    coeffs: Dict[str, List[float]]
+
+
+class BindingConstraintEdition(BindingConstraintMatrices, BindingConstraintForm):
+    pass
```
hdinia: you can simplify to:

```python
class BindingConstraint(BindingConstraintConfig, metaclass=AllOptionalMetaclass):
    group: str
    coeffs: Dict[str, List[float]]

class BindingConstraintEdition(BindingConstraintMatrices, BindingConstraint):
    pass
```

hdinia (changes requested):

antarest/study/business/binding_constraint_management.py
MartinBelthle: `BindingConstraintConfig` contains `id`, `name` and `constraints` so i don't think we should do that.

antarest/study/web/study_data_blueprint.py
MartinBelthle: Yes i don't mind. I just used this name because we already created a validation endpoint for st storages and it finishes with `/validate`

antarest/study/business/binding_constraint_management.py
MartinBelthle: Okay. I've renamed it `BindingConstraintEditionModel`. Don't know if it's better but I'm not really inspired

antarest/study/business/binding_constraint_management.py
MartinBelthle: It's impossible for a bc after v8.7 to not have a group so this case will never happen.
Therefore, it the group is None, it means the bc does not exist, that's why i raised such an Exception

antarest/study/business/binding_constraint_management.py
hdinia: mybad I mean `BindingConstraintProperties` not Config you're right it contains the id and name that should not be updated

antarest/study/business/binding_constraint_management.py
hdinia: In fact it's possible if you update a BC via the `/raw` endpoint you can set the group to `None`. Which in this case raise a key error instead of the one you want. (Maybe it's a edge case, just to let you know about that)

MartinBelthle: You're right but it would be a non-sense.

antarest/study/business/binding_constraint_management.py
MartinBelthle: Ah okay. Why not, i'll let Laurent change this if he wants.

antarest/study/business/binding_constraint_management.py
hdinia: No problem, thanks for the change I would called this one just `BindingConstraintInput` as you already defined `BindingConstraintEdition` for the update. If it's ok for you I will update it or Laurent.

Maybe we should use `Input` and `Output` suffix to stay consistent with the rest, but it's a detail

laurent-laporte-pro: We also need to add `use_none=True` or else default values are used:

```python
class BindingConstraintEditionModel(BaseModel, metaclass=AllOptionalMetaclass, use_none=True):
    ...
```

antarest/study/business/binding_constraint_management.py
```diff
@@ -239,11 +250,26 @@ def get_binding_constraint(
...
 
+    def validate_binding_constraint(self, study: Study, constraint_id: str) -> None:
+        if int(study.version) < 870:
+            return  # There's nothing to check for constraints before v8.7
+        file_study = self.storage_service.get_storage(study).get_raw(study)
+        config = file_study.tree.get(["input", "bindingconstraints", "bindingconstraints"])
+        group = next((value["group"] for value in config.values() if value["id"] == constraint_id), None)
+        if not group:
+            raise BindingConstraintNotFoundError(study.id)
+        matrix_terms = {
```
laurent-laporte-pro: A note on vocabulary: the word "term" introduces confusion (which already exists elsewhere in the code) between the term of a binding constraint (list of terms) and matrices. Matrices concern the so-called second member.

antarest/study/business/binding_constraint_management.py
```diff
@@ -239,11 +250,26 @@ def get_binding_constraint(
...
+            "eq": get_matrix_data(file_study, constraint_id, "eq"),
```
laurent-laporte-pro: Actually, it's not necessary to check all the matrices: for the sake of efficiency, we could compare only the matrices corresponding to the type of binding constraint:

- if `operator == "equal"`, we check the "eq" matrix,
- if `operator == "greater"`, we check the "gt" matrix,
- if `operator == "less"`, we check the "lt" matrix,
- if `operator == "both"`, we check the "lt" and "gt" matrices.

antarest/study/web/study_data_blueprint.py
laurent-laporte-pro: Could you rephrase:

For studies with versions prior to v8.7, no validation is performed.
For studies with version 8.7 or later, the endpoint checks if the dimensions
of the right-hand side matrices are consistent with the dimensions of the
binding constraint matrices within the same group.

laurent-laporte-pro (changes requested):

antarest/study/web/study_data_blueprint.py
laurent-laporte-pro: For Short-term storage validation (see `/v1/studies/{uuid}/areas/{area_id}/storages/{storage_id}/validate`), we used a boolean. Actually, I have no preferences.

antarest/study/business/binding_constraint_management.py
laurent-laporte-pro: We can also add camelCase aliases

antarest/study/business/binding_constraint_management.py
```diff
@@ -111,9 +112,19 @@ def generate_id(self) -> str:
...
-    value: Any
+class BindingConstraintEditionModel(BaseModel, metaclass=AllOptionalMetaclass):
+    group: str
+    enabled: bool
+    time_step: BindingConstraintFrequency
+    operator: BindingConstraintOperator
+    filter_year_by_year: str
+    filter_synthesis: str
+    comments: str
+    coeffs: Dict[str, List[float]]
```
hdinia: `coeffs` can be removed from the Input model, as it represents the constraint terms. And the constraints terms are handled separately to the constraint properties. 

(We can note that the client sends/expects `constraints` of type `List[ConstraintTermDTO]` for the constraint terms.)

I think that I may asked you for this, but it's not necessary actually, sorry about that. (I will update it)
Note: we also should update `process_contraint` to not send back `constraints` List

antarest/study/business/binding_constraint_management.py
```diff
@@ -253,52 +279,55 @@ def create_binding_constraint(
...
+            }
+            check_matrices_coherence(file_study, data.group, bc_id, matrix_terms_list)
+
+        args = {
+            "name": data.name,
+            "enabled": data.enabled,
+            "time_step": data.time_step,
+            "operator": data.operator,
+            "coeffs": data.coeffs,
+            "values": data.values,
```
hdinia, line 650: what is values ? If it represents the matrix for < v8.7 we should consider a more obvious name

MartinBelthle: Indeed it represents matrices before v8.7. Even though I agree it's not an obvious name, we kept it to ensure backward compatibility (the commands create and update both have this field so it would break scripts and DB).

antarest/study/business/binding_constraint_management.py
MartinBelthle: Same thing as the comment below, the commands create and update have the field `coeffs`.

Also, we currently use `process_contraint` to send the information about  `constraints` with the GET, PUT, POST endpoints. So I'm not sure why you want to remove this.

hdinia: The input you pass to commands is not the same as the model represented by Pydantic, you can have a representation of a given model that you want to validate, and construct the data you want to persist differently.

Pydantic models are primarily used for data validation and serialization, there might be scenarios where the structure of data coming from or going to the database (or commands here) needs to be transformed or adapted for various reasons.

Given that the `coeffs` field (`Dict[str, List[float]]`)  you pass to commands is not like `constraints` (`List[ConstraintTermDTO]`) that comes and goes to the client. So you have at least to change it from coeffs to constraints (only in the model, of course you keep it coeffs for the commands)

hdinia: And the reason why I suggested to remove `constraints` from the Input model, is cause it represents the constraint terms. And the constraints terms are handled separately to the constraints properties. (they are never required both at the same time) so we can easily dissociate them.

But you're right it's maybe a bad idea, we never know what it could breaks.

antarest/study/business/binding_constraint_management.py
hdinia: Ok it make sense thanks for the explanation 👍🏼

antarest/study/web/study_data_blueprint.py
laurent-laporte-pro: In case of failure, we raise an HTTP 422 exception with the following body:

```json
{
   "msg": "Matrix shapes mismatch in binding constraints group. Expected shape: (8784, 3)",
   "invalid_constraints": {"third bc": ["'third bc_lt' (8784, 4)"]}
}
```

- `msg`: is a error message
- `invalid_constraints` is a mapping between BC IDs and invalid matrix IDs / shapes

antarest/study/web/study_data_blueprint.py
laurent-laporte-pro: New endpoints to manage groups are:

- `GET /v1/studies/{uuid}/constraint-groups`
- `GET /v1/studies/{uuid}/constraint-groups/validate-all`
- `GET /v1/studies/{uuid}/constraint-groups/{group}`
- `GET /v1/studies/{uuid}/constraint-groups/{group}/validate`

In the future, we also need:
- `DELETE /v1/studies/{uuid}/constraint-groups/{group}`: to delete all BC of a given group

hdinia:
## Breaking Change in API for Binding Constraints

We are committed to continually improving our API's usability and alignment with domain-specific concepts. As part of this commitment, we have made a significant update to the format used for specifying binding constraints. This message outlines the nature of the change, the rationale behind it, and provides a detailed migration guide.

### Changes mades

We have clarified the terminology and format used for defining constraint terms in our API. Previously, constraint definitions could be referred to by either `coeffs` or `constraints`, which was confusing and inconsistent. 
Now, we have standardized on using `terms`, enhancing clarity and consistency. This change affects how constraints are specified in API requests and responses, except in the configuration where coeffs remains unchanged.

#### Previous Format: `coeffs` or `constraints`

Type: `Optional[List[ConstraintTerm]]`

Usage: Encoded constraint terms with the relationship indicated by a % separator for links (e.g., area1%area2) or a . separator for clusters (e.g., area.cluster), where the key is a string representing the term and the value is a float representing the weight.

Example:

```json
"constraints": [
  {
    "id": "z3%z4",
    "weight": 0.0,
    "offset": 5.0,
    "data": {
      "area1": "z3",
      "area2": "z4"
    }
  }
]
```
could also be:

```json
"coeffs": [
  {
    "id": "z1.cluster1",
    "weight": 0.0,
    "offset": 5.0,
    "data": {
      "area": "z1",
      "cluster": "cluster1"
    }
  }
]
```

#### New Format: terms

Type: `Optional[List[ConstraintTerm]]`

Usage: Each item in the list is a ConstraintTerm object that explicitly defines the term's components, including id, weight, offset, and a data object for additional metadata.

Example:

```json

    "terms": [
      {
        "id": "z1.cluster1",
        "weight": 0.0,
        "offset": null,
        "data": {
          "area": "z1",
          "cluster": "cluster1"
        }
      }
    ]
```

### Rationale for the Change

This update is aimed at enhancing the clarity, consistency, and expressiveness of our API by:

- Reducing Confusion: By eliminating the use of two terms (coeffs and constraints) for similar concepts, we reduce confusion and make the API more intuitive.
- Improving Readability and Flexibility: The structured format of terms allows for clearer inputs and outputs, and easier future enhancements.
- Aligning with Domain Language: The new format more accurately reflects the terminology and concepts used in the field.

### Migration Guide

To adapt to this change, please update your application's API interactions as follows:

- Understand the Terminology Update: Recognize that what was previously referred to as coeffs or constraints for defining constraint terms is now uniformly referred to as terms in API interactions, with the exception of the configuration context where coeffs remains unchanged.
- Adjust Data Structures: Replace any structures for constraint definitions with the new terms format where applicable.
- Update Request Construction and Response Handling: Ensure your application uses the terms list format for creating or updating constraints and parses this format in API responses.

Example Conversion from constraints to terms:

Before:

```json

"constraints": [
  {
    "id": "z3%z4",
    "weight": 0.0,
    "offset": 5.0,
    "data": {
      "area1": "z3",
      "area2": "z4"
    }
  }
]
```

After:

```json

{
  "terms": [
    {
      "id": "z3%z4",
      "weight": 0.0,
      "offset": 5.0,
      "data": {
        "area1": "z3",
        "area2": "z4"
      }
    }
  ]
}
```

Support and Feedback

We understand that changes like this can require adjustments in how you interact with our API. The AntaresWeb team is here to assist with any questions or challenges you may encounter during this transition.

hdinia:
@laurent-laporte-pro  @MartinBelthle 

Overview of Recent Refactor and Improvements:: [edc37fb](https://github.com/AntaresSimulatorTeam/AntaREST/pull/1982/commits/edc37fb9d7f258085101dcb8bc42ecf0255b6043)

This commit introduces method renamings, model consolidations, bug fixes, and significant improvements in our handling of binding constraints. 

Below is a detailed breakdown of the changes:

#### Method and Model Renaming
To improve code readability and consistency, we've renamed several methods and Pydantic models. This includes aligning naming conventions across our codebase and ensuring that names accurately reflect their functionality.

> Note: To not extend this PR too much we will address the rest of the refactoring,(such as renaming endpoints etc) in a dedicated one.

#### Model Consolidation
We've merged redundant models to streamline our data structures and reduce complexity. 

#### Bug Fixes
- Filter Functionality: Addressed an issue where added filters (e.g., area_name and cluster_name) were not functioning as expected. The fix ensures that constraints results are accurately retrieved based on the provided filters.
- Response Model Types: We've added return types and response model types to route handlers to enforce type safety and clarity in our API responses.
- Get Response Accuracy: Corrected a bug where get responses would incorrectly list terms as an empty array even when existing terms were present in a constraint. Now, responses accurately reflect the constraint's content.
- Update Behavior: Fixed an issue where updating constraints could unintentionally mutate existing data, particularly with terms. Previously, not passing terms was treated as a request to delete all terms, which has been corrected to prevent unintended data loss.
- Enabled Field Update: Resolved a problem where updating the enabled field to false was not possible due to it reverting to its default value.

#### Additional Improvements
- Missing Exception Handling: Added a TODO to address a missing exception for invalid IDs in the GET /v1/{uuid}/bindingconstraints route. This future enhancement will further improve error handling and user feedback.

Feel free to review & give feedbakc on those changes, thanks ✌🏼

MartinBelthle:
I'll wait until you fix the tests you broke before reading what you've done :smile:

antarest/study/business/binding_constraint_management.py
```diff
@@ -111,27 +123,228 @@ def generate_id(self) -> str:
...
+                            matching_terms.append(term)
+
+            # If no terms match, the constraint should not pass
+            if not matching_terms:
+                return False
+
+        if self.cluster_name:
+            all_clusters = []
+            for term in terms:
+                if term.data is None:
```
laurent-laporte-pro: This condition is useless, because we have `isinstance(term.data, ClusterTerm):`

antarest/study/business/binding_constraint_management.py
```diff
@@ -111,27 +123,228 @@ def generate_id(self) -> str:
...
+                        # Check if the cluster's area matches the specified area_name
+                        if self.area_name.upper() == term.data.area.upper():
+                            matching_terms.append(term)
+
+            # If no terms match, the constraint should not pass
+            if not matching_terms:
+                return False
+
+        if self.cluster_name:
+            all_clusters = []
```
laurent-laporte-pro: Can be rewrite:

```python
all_clusters = [term.data.cluster for term in terms if isinstance(term.data, ClusterTerm)]
```

antarest/study/business/binding_constraint_management.py
```diff
@@ -178,167 +391,329 @@ def parse_constraint(key: str, value: str, char: str, new_config: BindingConstra
...
-            "id": constraint_value["id"],
-            "name": constraint_value["name"],
-            "enabled": constraint_value["enabled"],
-            "time_step": constraint_value["type"],
-            "operator": constraint_value["operator"],
-            "comments": constraint_value.get("comments", None),
-            "filter_year_by_year": constraint_value.get("filter-year-by-year", ""),
-            "filter_synthesis": constraint_value.get("filter-synthesis", ""),
-            "constraints": None,
+    def constraint_model_adapter(constraint: ConstraintInput, version: int) -> ConstraintOutput:
```
laurent-laporte-pro: Wrong typing for `constraint`, should be:

```python
constraint: t.Mapping[str, t.Any]]
```

antarest/study/business/binding_constraint_management.py
```diff
@@ -178,167 +391,329 @@ def parse_constraint(key: str, value: str, char: str, new_config: BindingConstra
...
+        """
+
+        constraint_output = {
+            "id": constraint["id"],
+            "name": constraint["name"],
+            "enabled": constraint.get("enabled", True),
+            "time_step": constraint.get("type", BindingConstraintFrequency.HOURLY),
+            "operator": constraint.get("operator", BindingConstraintOperator.EQUAL),
+            "comments": constraint.get("comments", ""),
+            "filter_year_by_year": constraint.get("filter_year_by_year", ""),
```
laurent-laporte-pro: Are you aware that `filter_year_by_year` and `filter_synthesis` are only available for study v8.3 or above?

See File format changes [v8.3.2 — Marginal price of a binding constraint](https://antares-simulator.readthedocs.io/en/latest/reference-guide/13-file-format/#marginal-price-of-a-binding-constraint)

laurent-laporte-pro (changes requested):

antarest/study/business/binding_constraint_management.py
hdinia: Yes their some typing issues, I am fixing this as well as tests

antarest/study/business/binding_constraint_management.py
hdinia: This is handled on the front end, but if it causes problems for the simulator by adding data through the API directly(scripts), we should handle it (we didn't before).

hdinia:
@laurent-laporte-pro  please consider keeping `feat(bc)!` in the title `!` is for a breaking change, usually its more used for commits titles, but here we use this convention to name PR's

hdinia (approved):

antarest/study/business/binding_constraint_management.py
```diff
@@ -658,34 +677,31 @@ def update_binding_constraint(
...
-        ):
-            raise BindingConstraintNotFoundError(study.id)
-
         check_attributes_coherence(data, study_version)
 
         # Because the update_binding_constraint command requires every attribute we have to fill them all.
         # This creates a `big` command even though we only updated one field.
         # fixme : Change the architecture to avoid this type of misconception
-        updated_constraint = {
+        upd_constraint = {
```
hdinia, line 692: in my opinion, it's better not to use this kind of abbreviation, as it's less comprehensible. It's better to be more explicit, even if it means taking up a little more space.

hdinia: (it as meaning for you in the given moment but another developer might not get it later on)

antarest/study/business/binding_constraint_management.py
```diff
@@ -181,39 +184,41 @@ def match_filters(self, constraint: "ConstraintOutput") -> bool:
...
+                if (
+                    isinstance(data, LinkTerm)
+                    and (area_name_upper in data.area1.upper() or area_name_upper in data.area2.upper())
+                ) or (
+                    isinstance(data, ClusterTerm)
+                    and area_name_upper in data.area.upper()
+                ):
+                    break
+                # fmt: on
+            else:
```
hdinia, line 199: the filters works as expected now 👍🏼

hdinia:
@MartinBelthle @laurent-laporte-pro  thanks for this teamwork i've tested the api integration it works well i think we can merge safely 👍🏼

laurent-laporte-pro (approved):

PR #1984, opened 2024-03-25
feat(clusters-ui): implement new duplication endpoint and optimistic update

skamril:
ANT-1357


webapp/public/locales/en/main.json
```diff
@@ -483,8 +487,8 @@
   "study.modelization.clusters.matrix.timeSeries": "Time-Series",
   "study.modelization.clusters.backClusterList": "Back to cluster list",
   "study.modelization.clusters.tsInterpretation": "TS interpretation",
-  "study.modelization.clusters.group": "Group",
-  "studies.modelization.clusters.question.delete": "Are you sure you want to delete this cluster?",
+  "studies.modelization.clusters.question.delete_one": "Are you sure you want to delete this cluster?",
+  "studies.modelization.clusters.question.delete_other": "Are you sure you want to delete these {{count}} clusters?",
```
hdinia, line 494: use camel case instead (same for FR keys)
`"studies.modelization.clusters.question.deleteOne":`

`"studies.modelization.clusters.question.deleteMany":`

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/index.tsx
```diff
@@ -1,198 +1,158 @@
...
-    id,
-    installedCapacity,
-    enabledCapacity,
-    ...cluster
-  }: RenewableClusterWithCapacity) => {
-    return createRenewableCluster(study.id, areaId, cluster);
+  const handleCreate = async (values: TRow<RenewableGroup>) => {
+    const cluster = await createRenewableCluster(study.id, areaId, values);
+    return addCapacity(cluster);
+  };
```
hdinia: optional: you could name the cluster values parameter more explicitly
```typescript
 const handleCreate = async (cluster: TRow<RenewableGroup>) => {
    const createdCluster = await createRenewableCluster(study.id, areaId, cluster);
    return addCapacity(createdCluster);
  };
```
(if you do it don't forget Thermals and Storages, to not repeat comments)

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/index.tsx
```diff
@@ -1,198 +1,158 @@
...
+  ) => {
+    const cluster = await duplicateRenewableCluster(
+      study.id,
+      areaId,
+      row.id,
+      newName,
+    );
+
+    return { ...row, ...cluster };
   };
```
hdinia: optional: row > cluster
```typescript
  const handleDuplicate = async (
    cluster: RenewableClusterWithCapacity,
    newName: string,
  ) => {
    const duplicatedCluster = await duplicateRenewableCluster(
      study.id,
      areaId,
      cluster.id,
      newName,
    );

    return { ...row, ...duplicatedCluster };
  };
```
(if you do it don't forget Thermals and Storages, to not repeat comments)

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/index.tsx
```diff
@@ -1,198 +1,158 @@
...
-  const handleDeleteSelection = (ids: string[]) => {
+  const handleDelete = (rows: RenewableClusterWithCapacity[]) => {
+    const ids = rows.map((row) => row.id);
     return deleteRenewableClusters(study.id, areaId, ids);
   };
 
+  const handleNameClick = (row: RenewableClusterWithCapacity) => {
+    navigate(`${location.pathname}/${row.id}`);
+  };
+
```
hdinia, line 140: same as 2 previous comments you can use "cluster(s)" for more clarity if you want

(if you do it don't forget Thermals and Storages, to not repeat comments)

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/utils.ts
```diff
@@ -115,18 +118,32 @@ export async function updateRenewableCluster(
...
     "post",
     getClustersUrl(studyId, areaId),
     data,
   );
 }
 
+export function duplicateRenewableCluster(
+  studyId: StudyMetadata["id"],
+  areaId: Area["name"],
+  clusterIdSource: RenewableCluster["id"],
```
hdinia: `sourceClusterId`

(also in thermals, storages)

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/utils.ts
```diff
@@ -115,18 +118,32 @@ export async function updateRenewableCluster(
...
+  newName: RenewableCluster["name"],
+) {
+  return makeRequest<RenewableCluster>(
+    "post",
+    `/v1/studies/${studyId}/areas/${areaId}/renewables/${clusterIdSource}`,
+    null,
```
hdinia, line 134: as the payload is optional you can remove `null` and pass an empty body object instead `{}`
don't forget to update the `makeRequest` data type

(to no repeat comments you can do the same for thermals and storages)

webapp/public/locales/en/main.json
```diff
@@ -78,6 +79,8 @@
   "global.error.failedtoretrievejobs": "Failed to retrieve job information",
   "global.error.failedtoretrievelogs": "Failed to retrieve job logs",
   "global.error.failedtoretrievedownloads": "Failed to retrieve downloads list",
+  "global.error.createFailed": "Unable to create",
+  "global.error.deleteFailed": "Unable to delete",
```
hdinia:
```json
    "global.error.create": "Creation failed. Please try again."
    "global.error.delete": "Deletion failed. Please try again."
```

webapp/public/locales/fr/main.json
```diff
@@ -78,6 +79,8 @@
   "global.error.failedtoretrievejobs": "Échec de la récupération des tâches",
   "global.error.failedtoretrievelogs": "Échec de la récupération des logs",
   "global.error.failedtoretrievedownloads": "Échec de la récupération des exports",
+  "global.error.createFailed": "Impossible d'effectuer la création",
+  "global.error.deleteFailed": "Impossible d'effectuer la suppression",
```
hdinia:
```json
    "global.error.create": "La création a échoué. Veuillez réessayer."
    "global.error.delete": "La suppression a échoué. Veuillez réessayer."
```

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/index.tsx
```diff
@@ -1,198 +1,158 @@
...
-        size: 200,
-        Cell: ({ cell }) => Math.floor(cell.getValue<number>()),
-      },
-      {
-        accessorKey: "installedCapacity",
+        size: 220,
+        Cell: ({ cell }) => Math.floor(cell.getValue()),
+      }),
+      columnHelper.accessor("installedCapacity", {
         header: "Enabled / Installed (MW)",
```
hdinia, line 86: you have refactored the name and group columns to be common it's a nice addition 👍🏼 
but the i18n labels are not added in Renewables, so their is a mismatch between translations, consider adding i18n keys here

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Thermal/utils.ts
```diff
@@ -143,18 +146,32 @@ export async function updateThermalCluster(
...
 }
 
-export async function createThermalCluster(
+export function createThermalCluster(
   studyId: StudyMetadata["id"],
   areaId: Area["name"],
-  data: Partial<ThermalCluster>,
-): Promise<ThermalClusterWithCapacity> {
-  return makeRequest<ThermalClusterWithCapacity>(
+  data: PartialExceptFor<ThermalCluster, "name">,
```
hdinia, line 147: the group is also mandatory

webapp/src/components/common/GroupedDataTable/index.tsx
```diff
@@ -1,69 +1,299 @@
...
+  deleteConfirmationMessage?: string | ((count: number) => string);
+  fillPendingRow?: (
+    pendingRow: TRow<TGroups[number]>,
+  ) => TRow<TGroups[number]> & Partial<TData>;
 }
 
-function GroupedDataTable<TData extends TRow>({
+// Use ids to identify default columns (instead of `accessorKey`),
+// to have a unique identifier. It is more likely to have a duplicate
+// `accessorKey` with `columns` prop.
```
hdinia, line 53: use a multi line comment

webapp/src/components/common/GroupedDataTable/index.tsx
```diff
@@ -1,69 +1,299 @@
...
-  );
-
-  const selectedRow = useMemo(() => {
-    if (isOneRowSelected) {
-      const selectedIndex = Object.keys(rowSelection).find(
-        (key) => rowSelection[key],
-      );
-      return selectedIndex && tableData[+selectedIndex];
+  // Update once `data` only if previous value was empty.
+  // It allows to handle loading data.
```
hdinia: `// single line comment`
```javascript
  /**
   * Multi
   * line comment
   */
```

webapp/src/components/common/GroupedDataTable/index.tsx
```diff
@@ -1,69 +1,299 @@
...
+    pendingRow: TRow<TGroups[number]>,
+  ) => TRow<TGroups[number]> & Partial<TData>;
 }
 
-function GroupedDataTable<TData extends TRow>({
+// Use ids to identify default columns (instead of `accessorKey`),
+// to have a unique identifier. It is more likely to have a duplicate
+// `accessorKey` with `columns` prop.
+const GROUP_COLUMN_ID = "_group";
+const NAME_COLUMN_ID = "_name";
```
hdinia, line 55: is this an advice from the docs, or a "hack" ?

webapp/src/components/common/GroupedDataTable/index.tsx
```diff
@@ -1,69 +1,299 @@
...
+  fillPendingRow,
+}: GroupedDataTableProps<TGroups, TData>) {
   const { t } = useTranslation();
   const [openDialog, setOpenDialog] = useState<
     "add" | "duplicate" | "delete" | ""
   >("");
   const [tableData, setTableData] = useState(data);
   const [rowSelection, setRowSelection] = useState<MRT_RowSelectionState>({});
+  const enqueueErrorSnackbar = useEnqueueErrorSnackbar();
+  const callbacksRef = useAutoUpdateRef({ onNameClick });
```
hdinia, line 81: maybe you can comment on this, it's an advanced pattern not easy to understand at first look. You are ensuring callback functions that may change over time have the latest version ?

webapp/src/components/common/GroupedDataTable/index.tsx
```diff
@@ -1,69 +1,299 @@
...
+
+    // Type can be asserted as `TData` because the row will be checked in cell renders
+    // and `fillPendingRow` allows to add needed data
+    setTableData((prev) => [...prev, pendingRow as TData]);
+
+    return function removePendingRow() {
+      pendingRows.current = pendingRows.current.filter((r) => r !== pendingRow);
+      setTableData((prev) => prev.filter((r) => r !== pendingRow));
+    };
+  };
```
hdinia: it's very confusing why the `addPendingRow` is also responsible for deleting pending rows, you should split this, also it can potentially lead to stale state, If the state changes before the operation completes, calling this cleanup function could inadvertently revert some of those changes.

maybe you can define 1 for adding, 1 for replacing 1 for removing

```typescript
  const addPendingRow = (pendingRow: TRow<TGroups[number]>) => {
    pendingRows.current.push(pendingRow);
    setTableData((prev) => [...prev, pendingRow as TData]);
  };

  const replacePendingRow = (
    pendingRow: TRow<TGroups[number]>,
    newRow: TData,
  ) => {
    setTableData((prev) => {
      const index = prev.findIndex((row) => row === pendingRow);

      if (index > -1) {
        return [...prev.slice(0, index), newRow, ...prev.slice(index + 1)];
      }

      return prev;
    });
  };

  const removePendingRow = (pendingRow: TRow<TGroups[number]>) => {
    pendingRows.current = pendingRows.current.filter(
      (row) => row !== pendingRow,
    );

    setTableData((prev) => prev.filter((row) => row !== pendingRow));
  };
  ```
  and update with them:
  
  ```typescript
  const handleCreate = async (values: TRow<TGroups[number]>) => {
    closeDialog();

    if (!onCreate) {
      return;
    }

    const pendingRow = fillPendingRow?.(values) || values;
    createOps.increment();
    addPendingRow(pendingRow); // Optimistically add row

    try {
      const newRow = await onCreate(values);
      replacePendingRow(pendingRow, newRow); // Replace with actual data on success
    } catch (error) {
      removePendingRow(pendingRow); // Remove if operation fails
      enqueueErrorSnackbar(t("global.error.createFailed"), toError(error));
    } finally {
      createOps.decrement();
    }
  };
  ```

File: webapp/src/components/common/GroupedDataTable/index.tsx [hunk 16]
```diff
@@ -74,51 +304,80 @@ function GroupedDataTable<TData extends TRow>({
...
+    try {
+      const newRow = await onDuplicate(selectedRow, newName);
+      setTableData((prev) => [...prev, newRow]);
+    } catch (error) {
+      enqueueErrorSnackbar(t("global.error.createFailed"), toError(error));
+    }
+
+    removePendingRow();
+    createOps.decrement();
+  };
```
hdinia: you have omitted part of the optimistic updates for the duplication:

```typescript
  const handleDuplicate = async (newName: string) => {
    closeDialog();

    if (!onDuplicate || !selectedRow) {
      return;
    }

    setRowSelection({});

    const duplicatedRow = {
      ...selectedRow,
      name: newName,
    };
    // optimistically add row
    addPendingRow(duplicatedRow);

    createOps.increment();

    try {
      const newRow = await onDuplicate(selectedRow, newName);
      // replace the optimistic duplicate with the real one
      replacePendingRow(duplicatedRow, newRow);
    } catch (error) {
      // operation fails, remove the optimistic duplicate and show an error
      removePendingRow(duplicatedRow);
      enqueueErrorSnackbar(t("global.error.createFailed"), toError(error));
    } finally {
      createOps.decrement();
    }
  };
```

webapp/src/components/common/GroupedDataTable/index.tsx
```diff
@@ -74,51 +304,80 @@ function GroupedDataTable<TData extends TRow>({
...
+  const handleDelete = async () => {
+    closeDialog();
+
+    if (!onDelete) {
+      return;
+    }
+
+    setRowSelection({});
+
+    const rowsToDelete = selectedRows;
```
hdinia, line 369: (optional) you can use a snapshot of all the table to avoid side effects when restoring data, if the order may change

`  const snapshot = [...tableData]; `

webapp/src/utils/tsUtils.ts
```diff
@@ -1,3 +1,16 @@
...
+/**
+ * Allow to use `any` with `Promise` type without disabling ESLint rule.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type PromiseAny = Promise<any>;
+
+/**
+ * Make all properties in T optional, except for those specified by K.
+ */
+export type PartialExceptFor<T, K extends keyof T> = O.Required<Partial<T>, K>;
```
hdinia, line 12: can this allow multiple keys to be required ?

webapp/src/utils/validationUtils.ts
```diff
@@ -99,7 +99,7 @@ export function validateString(
 
   // Check for duplication against existing values.
   if (existingValues.map(normalize).includes(comparisonValue)) {
-    return t("form.field.duplicate", { 0: value });
```
hdinia, line 102: their is case-sensitive comparisons involved, and you don't always have the visibility on the existing values, so I think you should revert that change to inform the user about the duplicated value

webapp/src/hooks/useOperationInProgressCount.ts
hdinia, line 1: we should avoid such manual approach for this kind of things, it can easily introduce hard to find bugs like:

- race conditions: using simple state updates for async CRUD operations can lead to race conditions, where the operation counts might not reflect the actual number of active operations due to the asynchronous nature of state updates

- state desynchronization: without centralized control and automated rollback mechanisms, manual error handling can cause the UI to show an optimistic state that doesn't match the backend, especially if an operation fails

- error handling and consistency: incrementing and decrementing counts without considering all possible state transitions can lead to inconsistent states

Instead you could centralize the logic into a generic hook that handles state and data updates at the same time, here's a quick draft (needs adjustments):

```typescript
import { useCallback, useState } from "react";

type OperationTypes = "create" | "read" | "update" | "delete";
type OperationCounts = {
  [key in OperationTypes]: number;
};

type OperationFunction<T = any> = () => Promise<T>;
type CompletionCallback<T = any> = (result: T) => void;
type ErrorCallback = (error: any) => void;
type UpdateStateCallback<T = any> = (result: T) => void;
type RevertStateCallback = () => void;

interface PerformOperationArgs<T = any> {
  operationType: OperationTypes;
  operation: OperationFunction<T>;
  optimisticResult: T;
  onCompletion: CompletionCallback<T>;
  onError: ErrorCallback;
  onUpdateState: UpdateStateCallback<T>;
  onRevertState: RevertStateCallback;
}

function useOptimistic() {
  const [operationCounts, setOperationCounts] = useState<OperationCounts>({
    create: 0,
    read: 0,
    update: 0,
    delete: 0,
  });

  const updateOperationCount = useCallback(
    (operationType: OperationTypes, delta: number) => {
      setOperationCounts((prevCounts) => ({
        ...prevCounts,
        [operationType]: Math.max(prevCounts[operationType] + delta, 0),
      }));
    },
    [],
  );

  const performOperation = useCallback(
    <T>({
      operationType,
      operation,
      optimisticResult,
      onUpdateState,
      onCompletion,
      onError,
      onRevertState,
    }: PerformOperationArgs<T>) => {
      // optimistic update
      if (onUpdateState) {
        onUpdateState(optimisticResult);
      }

      updateOperationCount(operationType, 1); // increment operation count

      operation()
        .then((result) => {
          // handle success
          if (onCompletion) {
            onCompletion(result);
          }

          updateOperationCount(operationType, -1); // decrement operation count
        })
        .catch((error) => {
          // operation fails revert optimistic update and handle error
          if (onRevertState) {
            onRevertState();
          }

          if (onError) {
            onError(error);
          }
          updateOperationCount(operationType, -1); // decrement happens on error too
        });
    },
    [updateOperationCount],
  );

  return {
    operationCounts,
    performOperation,
  };
}

export default useOptimistic;
```
Usage:
```typescript
const [items, setItems] = useState<ItemData[]>([]);
const { performOperation } = useOptimistic();

const handleCreate = (itemData: ItemData) => {
  performOperation<ItemData>({
    operationType: "create",
    operation: () => api.createItem(itemData),
    optimisticResult: itemData,
    onUpdateState: (newItem) =>
      setItems((prevItems) => [...prevItems, newItem]),
    onCompletion: (result) => console.log("Item created successfully:", result),
    onRevertState: () =>
      setItems((prevItems) =>
        prevItems.filter((item) => item.id !== itemData.id),
      ),
    onError: (error) => console.error("Failed to create item:", error),
  });
};
```

webapp/src/components/common/GroupedDataTable/index.tsx
```diff
@@ -1,69 +1,299 @@
...
-      );
-      return selectedIndex && tableData[+selectedIndex];
+  // Update once `data` only if previous value was empty.
+  // It allows to handle loading data.
+  useUpdateEffectOnce(() => {
+    if (prevData && prevData.length === 0) {
+      setTableData(data);
     }
-  }, [isOneRowSelected, rowSelection, tableData]);
+  }, [data.length]);
```
hdinia: I dont know what was your intent here but this is causing a major state bug, the tableData is never updated, so at each re-render even if we change the area the state remains the same. here's my analysis:

- initial fetch: when data changes from empty to filled, `prevData` is empty `([])`, and data is now filled `([...items])`. The effect runs, setting tableData with data. This is the first and only time this effect runs.

- data changes: if data changes again (like we change the area), useUpdateEffectOnce will not execute its effect function again because `hasUpdated.current` is permanently set to `true` after the first execution. no updates to `tableData` will occur no matter how data changes.

hdinia (changes requested):

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/index.tsx
hdinia: I just noticed the same on thermals too 😮

webapp/public/locales/en/main.json
skamril: `_one` and `_other` are internally used by i18n => https://www.i18next.com/translation-function/plurals

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/index.tsx
skamril: At this moment it's not a cluster already, but the values with give to create a new one.
Give a cluster to create a cluster it's strange.

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/index.tsx
skamril: I prefer `row` (it's really a row) to understand the origin of the data (the selected row in the table).

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/index.tsx
skamril: Same response.

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/utils.ts
skamril: Can you explain why?

webapp/public/locales/en/main.json
skamril: Ok, but without 'Please try again.'

webapp/public/locales/fr/main.json
skamril: Ok, but without 'Veuillez réessayer.'

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/index.tsx
skamril: Sorry can you give more details.

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Thermal/utils.ts
skamril: No, they give a default group.

hdinia: Ok so it's required on the UI but not for the api ? It's ok if so

webapp/src/components/common/GroupedDataTable/index.tsx
hdinia: you can ignore this comment

webapp/public/locales/en/main.json
hdinia: ok didn't know about that I will check it thanks

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/utils.ts
hdinia: sorry if my comment is not clear to explain in more details:
`data?: Partial<ThermalCluster> | { data: Array<Cluster["id"]> } | null` data is optional in the type signature, so you don't necessarily have to define null and instead you can just pass an empty body (i'ts the same in fine), and the signature will be `data?: Partial<ThermalCluster> | { data: Array<Cluster["id"]> },`

```typescript
 return makeRequest<ThermalCluster>(
    "post",
    `/v1/studies/${studyId}/areas/${areaId}/thermals/${clusterIdSource}`,
    {},
    { newName },
  );
```

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Thermal/utils.ts
skamril: Yes this is it

webapp/src/components/common/GroupedDataTable/index.tsx
skamril: I wrote a comment on top. It is for prevent duplicate columns (for  `grouping` and `columnPinning`), it is most probable to have a column in props named 'group' than '_group'.

When an id is defined, it take ascendant over the column name.

webapp/src/components/common/GroupedDataTable/index.tsx
skamril: Ok, you can read this to understand https://overreacted.io/making-setinterval-declarative-with-react-hooks/

webapp/src/components/common/GroupedDataTable/index.tsx
skamril: Adding `replacePendingRow` complicate the code for nothing.

>  it can potentially lead to stale state, If the state changes before the operation completes, calling this cleanup function could inadvertently revert some of those changes.

No because I use a callback in `setTableData`, `prev` is the last value when the function is called.
And `pendingRows` is a ref.

I can have something like this if you prefer:

```javascript
const pendingRow = addPendingRow(values);
...
removePendingRow(pendingRow);
```

Like `setTimeout` mechanism.

webapp/src/utils/validationUtils.ts
skamril: The `value` is actually visible in the invalid field, there is not reason to display it in the error message.
Tell me if I don't understand the issue.

webapp/src/hooks/useOperationInProgressCount.ts
skamril: Actually, I wanted to make an easy hook for my needs, but the goal is to used the new `useOptimistic` hook from React when it would be release (https://fr.react.dev/reference/react/useOptimistic), instead of creating a complicated one that will be replaced. I tested it with the canary version, but I wasn't able to create a polyfill.

webapp/src/utils/tsUtils.ts
skamril: I you  mean like this: `PartialExceptFor<MyType, "name" | "id">`, the answer is yes.

webapp/src/components/common/GroupedDataTable/index.tsx
skamril: Thanks for testing, I will investigate.

webapp/src/components/common/GroupedDataTable/index.tsx
hdinia: thanks

webapp/src/hooks/useOperationInProgressCount.ts
hdinia: If it releases soon I agree it's maybe not ideal to implement it. but stable React 19 can be long to release, and keep in mind this view will evolve soon.
We can also consider Tan stack built-in optimistic updates and state management: https://tanstack.com/query/v3/docs/framework/react/guides/optimistic-updates (maybe on another issue, but it would be nice to have a solid base for when we'll update the view soon

webapp/src/utils/validationUtils.ts
hdinia: ok if you think it's good, it's a detail anyway (my point was just about case-sensitive differences like Area vs area, if you don't see the existing values)

webapp/src/hooks/useOperationInProgressCount.ts
skamril: Yes I saw this too, it's preferable to create a new ticket to implement this.

webapp/src/components/common/GroupedDataTable/index.tsx
hdinia: I personally prefer to separate concerns this way you know exactly at first look what each function does, but as you prefer. Yes it's better  like in the example 👍🏼

webapp/src/components/common/GroupedDataTable/index.tsx
skamril: It is more complicated to implement, because during the deletion the state may have changed.

hdinia: Ok, yes if the other operations are not blocked during deletion it can complicate

webapp/src/components/common/GroupedDataTable/index.tsx
skamril: Yes I know lol, but I use this format only for JSDoc or for top file comment (IDE syntax color is not the same for both). It is very common to see single line comment for long sentences like this (in fact I always seen this for internal function comment).

(cf. https://github.com/facebook/react/blob/main/packages/react-dom/src/client/ReactDOMFB.js)

webapp/src/components/common/GroupedDataTable/index.tsx
skamril: https://github.com/AntaresSimulatorTeam/AntaREST/pull/1984#discussion_r1561342680

webapp/src/components/common/GroupedDataTable/index.tsx
skamril: Ok finally I prefer separate too.

webapp/src/components/common/GroupedDataTable/index.tsx
skamril: Cause:  before `GroupedDataTable` was unmounted when an area changed by `UsePromiseCond` (cf. `ifPending` and `ifResolved` props), but I remove it, so I need to add a `key` in `GroupedDataTable`.

hdinia: Ok so finally it's not because of the `useEffectOnce`  mybad 🤔  thanks for the return I'll take a look 👍🏼

webapp/src/components/common/GroupedDataTable/index.tsx
hdinia: I agree that multi-line comments should be avoided inside code blocks to simplify commenting out sections as needed.But for documenting functions, hooks or constants at the start of blocks (like you did in this case), multi-line comments are suitable as they're typically used for more detailed explanations. Many well-regarded libraries also follow this practice. If your preference for single-line comments is just personal, that's completely fine I'm not advocating for overly strict rules just aiming for clarity and consistency where it might enhance understanding.

hdinia: My initial comment was maybe too short I could've explain the rationale behind

skamril: > But for documenting functions, hooks or constants at the start of blocks (like you did in this case)

Yes but here I comment the **call** of `useUpdateEffectOnce` function hook inside `GroupedDataTable` function component, and not the function itself. I explain why I call it in this context.
I used multi line comment for the JSDoc of `useUpdateEffectOnce`.

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/utils.ts
skamril: Done

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Renewables/utils.ts
skamril: I removed `makeRequest` abstraction, it's useless,the difference with the use of `client` directly is insignificant.

hdinia (approved):

PR #1988, opened 2024-03-26
fix(study-search): correct pagination SQL query

mabw-rte:


antarest/study/repository.py
```diff
@@ -257,9 +257,17 @@ def get_all(
...
+            offset = pagination.page_nb * pagination.page_size
+            end = offset + limit
+            if sort_by is None:
+                q = q.order_by(entity.name.asc())
+            if study_filter.groups or study_filter.tags:
+                studies: t.Sequence[Study] = q.all()[offset:end]
+                return studies
+            q = q.offset(offset).limit(limit)
+
+        studies = q.all()
```
hdinia, line 270: You can apply eager loading for related entities to avoid N+1 query problem:

```python

    if sort_by is None:
        q = q.order_by(entity.name.asc())

    q = q.options(selectinload(entity.owner), selectinload(entity.groups))

    if pagination.page_size:
        offset = pagination.page_nb * pagination.page_size
        q = q.offset(offset).limit(pagination.page_size)

    return q.all()
```

laurent-laporte-pro (approved):

PR #1996, opened 2024-04-02
feat(common-ui,results-ui): update matrix buttons (import & export)

skamril:
ANT-1052


webapp/src/components/common/DownloadRawButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+      downloadFile(
+        res,
+        `matrix_${studyId}_${path.replace("/", "_")}.${format}`,
+      );
+    } catch (err) {
+      enqueueErrorSnackbar(t("global.download.error"), String(err));
+    }
+
+    setIsDownloading(false);
+  };
```
hdinia, line 61: this a common component, but you are coupling the matrix logic into it therefore making it unusable for other use-cases
if this is intended, consider moving it to the parent folder of the matrix component , and give it a more descriptive name.
Or if you want to keep it common, move the business logic back into the matrix component

webapp/src/components/common/DownloadRawButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+const EXPORT_OPTIONS = [
+  { label: "TSV", value: "tsv" },
+  { label: "Excel", value: "xlsx" },
+] as const;
+
+type ExportFormat = (typeof EXPORT_OPTIONS)[number]["value"];
+
+function DownloadRawButton(props: DownloadRawButtonProps) {
```
hdinia: just `DownloadButton` is fine (optional)

webapp/src/components/common/DownloadRawButton.tsx
hdinia: this a common component, but you are coupling the matrix logic into it therefore making it unusable for other use-cases
if this is intended, consider moving it to the parent folder of the matrix component , and give it a more descriptive name.
Or if you want to keep it common, move the business logic back into the matrix component, for example you can extract it into a `onDownload` prop

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { useRef, useState } from "react";
+import LoadingButton, { type LoadingButtonProps } from "@mui/lab/LoadingButton";
+
+interface OptionObj<TOptionValue extends string = string> {
+  value: TOptionValue;
+  label?: string;
+  disabled?: boolean;
+}
+
+export interface SlitButtonProps<TOptionValue extends string = string>
```
hdinia: typo here `SplitButtonProps`

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+  extends Omit<ButtonGroupProps, "onClick"> {
+  options: Array<TOptionValue | OptionObj<TOptionValue>>;
+  dropdownActionMode?: "change" | "trigger";
+  onClick?: (optionValue: TOptionValue, optionIndex: number) => void;
+  ButtonProps?: Omit<LoadingButtonProps, "onClick">;
+}
```
hdinia: you can refine the interfaces

```typescript
interface Option<Value extends string = string> {
  value: Value;
  label?: string;
  disabled?: boolean;
}

interface SplitButtonProps<Value extends string>
  extends Omit<ButtonGroupProps, "onClick"> {
  options: Array<OptionOrString<Value>>;
  dropdownActionMode?: "change" | "trigger";
  onClick?: (value: Value, index: number) => void;
  loading?: boolean;
  ButtonProps?: Omit<LoadingButtonProps, "onClick">;
}
```

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+function formatOptions<TOptionValue extends string>(
+  options: SlitButtonProps<TOptionValue>["options"],
+) {
+  return options.map((option) =>
+    RA.isString(option) ? { value: option } : option,
+  );
+}
```
hdinia: (optional) you can use a type guard

```typescript
type OptionOrString<Value extends string> = Value | Option<Value>;

function isOption<Value extends string>(
  option: OptionOrString<Value>,
): option is Option<Value> {
  return typeof option === "object" && "value" in option;
}
```

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+    dropdownActionMode = "trigger",
+    onClick,
+    children,
+    ButtonProps: loadingButtonProps,
+    ...buttonGroupProps
+  } = props;
+  const [open, setOpen] = useState(false);
+  const [selectedIndex, setSelectedIndex] = useState(0);
+  const anchorRef = useRef<HTMLDivElement>(null);
+  const formattedOptions = formatOptions(options);
```
hdinia: (optional) remove `formatOptions()` and define options directly:
```typescript
  const formattedOptions = options.map((option) =>
    isOption(option) ? option : { value: option },
  );
```
you can also memoize it but it seem not necessary

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+interface OptionObj<TOptionValue extends string = string> {
+  value: TOptionValue;
+  label?: string;
+  disabled?: boolean;
+}
+
+export interface SlitButtonProps<TOptionValue extends string = string>
+  extends Omit<ButtonGroupProps, "onClick"> {
+  options: Array<TOptionValue | OptionObj<TOptionValue>>;
+  dropdownActionMode?: "change" | "trigger";
```
hdinia, line 22: (optional) you can simplify to `action`

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+  props: SlitButtonProps<TOptionValue>,
+) {
+  const {
+    options,
+    dropdownActionMode = "trigger",
+    onClick,
+    children,
+    ButtonProps: loadingButtonProps,
+    ...buttonGroupProps
+  } = props;
```
hdinia, line 37: (optional) destructure in the function definition

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+      return;
+    }
+
+    if (isChangeMode) {
+      const { value } = formattedOptions[selectedIndex];
+      onClick(value, selectedIndex);
+      return;
+    }
+
+    onClick(formattedOptions[0].value, 0);
```
hdinia:
```typescript
const index = isChangeMode ? selectedIndex : 0;
    onClick(formattedOptions[index].value, index);
```

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+        transition
+        disablePortal
+      >
+        {({ TransitionProps, placement }) => (
+          <Grow
+            {...TransitionProps}
+            style={{
+              transformOrigin:
+                placement === "bottom" ? "center top" : "center bottom",
+            }}
```
hdinia: I don't know your intent but it seem not necessary to have `placement` 

```typescript
transformOrigin: "center top",
```

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+  value: TOptionValue;
+  label?: string;
+  disabled?: boolean;
+}
+
+export interface SlitButtonProps<TOptionValue extends string = string>
+  extends Omit<ButtonGroupProps, "onClick"> {
+  options: Array<TOptionValue | OptionObj<TOptionValue>>;
+  dropdownActionMode?: "change" | "trigger";
+  onClick?: (optionValue: TOptionValue, optionIndex: number) => void;
```
hdinia: why `onClick` is optional ? consider making it mandatory and so you can simplify the checks inside the event handlers

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+    children,
+    ButtonProps: loadingButtonProps,
+    ...buttonGroupProps
+  } = props;
+  const [open, setOpen] = useState(false);
+  const [selectedIndex, setSelectedIndex] = useState(0);
+  const anchorRef = useRef<HTMLDivElement>(null);
+  const formattedOptions = formatOptions(options);
+  const isChangeMode = dropdownActionMode === "change";
+  const isTriggerMode = dropdownActionMode === "trigger";
```
hdinia, line 42: (optional) used only 1 time you can use it directly

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+  };
+
+  const handleMenuItemClick = (index: number) => {
+    setSelectedIndex(index);
+    setOpen(false);
+
+    if (isTriggerMode) {
+      const { value } = formattedOptions[index];
+      onClick?.(value, index);
+    }
```
hdinia:
```typescript
  dropdownActionMode === "trigger" &&
      onClick(formattedOptions[index].value, index);
```
or 
```typescript
if (dropdownActionMode === "trigger") {
      onClick(formattedOptions[index].value, index);
    }
```

webapp/src/components/common/buttons/SplitButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+            }}
+          >
+            <Paper>
+              <ClickAwayListener onClickAway={handleClose}>
+                <MenuList autoFocusItem>
+                  {formattedOptions.map((option, index) => (
+                    <MenuItem
+                      key={option.value}
+                      disabled={option.disabled}
+                      selected={isChangeMode && index === selectedIndex}
```
hdinia, line 130: normally it's used to indicate the currently selected item, it seem useless

webapp/src/services/api/studies/raw/types.ts
```diff
@@ -0,0 +1,9 @@
+import type { StudyMetadata } from "../../../../common/types";
+
+export interface RawDownloadParams {
+  studyId: StudyMetadata["id"];
+  path: string;
+  format?: "tsv" | "xlsx";
+  header?: boolean;
+  index?: boolean;
+}
```
hdinia: their is no need to define an interface for this you can remove the file

webapp/src/services/api/studies/raw/index.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export async function downloadRaw(params: RawDownloadParams) {
+  const { studyId, ...rest } = params;
+  const url = `${formatText(RAW_API_URL, { studyId })}/download`;
+  const res = await client.get<Blob>(url, {
+    params: rest,
+    responseType: "blob",
+  });
+  return res.data;
+}
```
hdinia: at first look it's difficult to read and understand, please simplify to:

```typescript
import client from "../../client";
import { StudyMetadata } from "../../../../common/types";

export async function downloadFile(
  studyId: StudyMetadata["id"],
  path: string,
  format: "tsv" | "xlsx",
) {
  const params = {
    path,
    format,
    ...(format === "xlsx" && { header: true, index: true })
  };

  const res = await client.get<Blob>(`v1/studies/${studyId}/raw/download`, {
    params,
    responseType: "blob",
  });

  return res.data;
}
```
rename `downloadFile` and move it the utils file of the `DownloadButton` component, then adapt usage:

```typescript
try {
      const file = await downloadFile(studyId, path, format);

      const fileName = `matrix_${studyId}_${path.replace(
        /\//g,
        "_",
      )}.${format}`;

      processDownload(file, fileName);
    } catch (err) {
      enqueueErrorSnackbar(t("global.download.error"), String(err));
    }
```

webapp/src/utils/fileUtils.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Triggers the download of a file with the given data and name.
+ *
+ * @param fileData - The data of the file to be downloaded.
+ * @param fileName - The name of the file to be downloaded.
+ */
+export function downloadFile(fileData: BlobPart, fileName: string) {
```
hdinia, line 7: rename `processDownload` or something else and move it to the utils file of `DownloadButton`

webapp/src/utils/fileUtils.ts
```diff
@@ -0,0 +1,13 @@
...
+  const link = document.createElement("a");
+  link.href = URL.createObjectURL(new Blob([fileData]));
```
hdinia, line 9: you can add the mimeType `{ type: "application/vnd. ms-excel }` (excel files only may be adjusted if the processDownload is intended for a global usage)

webapp/src/services/api/constants.ts
```diff
@@ -2,3 +2,4 @@ const API_URL_BASE = "v1";
 const STUDIES_API_URL = `${API_URL_BASE}/studies/{studyId}`;
 
 export const TABLE_MODE_API_URL = `${STUDIES_API_URL}/tablemode`;
+export const RAW_API_URL = `${STUDIES_API_URL}/raw`;
```
hdinia: in my opinion you can remove it it's not worth it to over optimize just to replace: `v1/studies/${studyId}/raw/download`

hdinia (changes requested):

webapp/src/components/common/DownloadRawButton.tsx
skamril: No, because the component use the raw API, it's not a simple UI button.

webapp/src/components/common/DownloadRawButton.tsx
skamril: The raw API in not specific for the download of matrices. This is why I named the component with 'Raw'. When you want to download a button that use this API, you just use it, instead of repeat the same logic every times.

webapp/src/components/common/buttons/SplitButton.tsx
skamril: Done

webapp/src/components/common/buttons/SplitButton.tsx
skamril: Where does `OptionOrString` come from? And what the the purpose of this abstraction here?
`loading` is including in `ButtonProps`.

PS: I just saw the comment with `OptionOrString`.

webapp/src/components/common/buttons/SplitButton.tsx
skamril: Type guard  is not necessary to differentiate a string and a plain object, TS detect the appropriate type automatically after `RA.isString(option)`.

webapp/src/components/common/buttons/SplitButton.tsx
skamril: With `action` only, we can think of the button click, but here it is the action for the dropdown click only.

webapp/src/components/common/buttons/SplitButton.tsx
skamril: Done

webapp/src/components/common/buttons/SplitButton.tsx
skamril: I know that you like it, but when it's too long I prefer make it inside the function.

webapp/src/components/common/buttons/SplitButton.tsx
skamril: It is useless for trigger mode (action direct on dropdown click), for change mode (change the button main action but need to click to the button to run the action) it allow to show the selected item.

webapp/src/components/common/buttons/SplitButton.tsx
skamril: lol really?

webapp/src/services/api/constants.ts
skamril: The goal is to used it for the other Raw API (I don't make it in this PR) instead of repeat it.

laurent-laporte-pro (dismissed):

hdinia (approved):

PR #1997, opened 2024-04-03
fix(ui-api-bc): v8.7 binding constraints fix

hdinia:
- Fix a UI issue when no binding constraint data exists.
- Fix default values for optional strings to prevent null default values that triggers a UI error.
- Group default value is now displayed only for v8.7 without causing an error for versions below.
- Add broader special chars set

### DIFF
--- a/antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
+++ b/antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
@@ -75,13 +75,13 @@ class BindingConstraintProperties(BaseModel, extra=Extra.forbid, allow_populatio
     enabled: bool = True
     time_step: BindingConstraintFrequency = BindingConstraintFrequency.HOURLY
     operator: BindingConstraintOperator = BindingConstraintOperator.EQUAL
-    comments: t.Optional[str] = None
-    filter_year_by_year: t.Optional[str] = None
-    filter_synthesis: t.Optional[str] = None
+    comments: t.Optional[str] = ""
+    filter_year_by_year: t.Optional[str] = ""
+    filter_synthesis: t.Optional[str] = ""
 
 
 class BindingConstraintProperties870(BindingConstraintProperties):
-    group: t.Optional[str] = None
+    group: t.Optional[str] = ""
 
 
 class BindingConstraintMatrices(BaseModel, extra=Extra.forbid, allow_population_by_field_name=True):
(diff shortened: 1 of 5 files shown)


antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
```diff
@@ -75,13 +75,13 @@ class BindingConstraintProperties(BaseModel, extra=Extra.forbid, allow_populatio
...
-    filter_year_by_year: t.Optional[str] = None
-    filter_synthesis: t.Optional[str] = None
+    comments: t.Optional[str] = ""
+    filter_year_by_year: t.Optional[str] = ""
+    filter_synthesis: t.Optional[str] = ""
 
 
 class BindingConstraintProperties870(BindingConstraintProperties):
-    group: t.Optional[str] = None
+    group: t.Optional[str] = ""
```
MartinBelthle, line 84: I don't really like these changes on default values. But I'll let Laurent choose.

MartinBelthle: Other than that I like the fact that we kept back-end validation on the "group" field.

hdinia: can you explain a bit more, you think it can lead to potential issues ? In fact it would make more sense to set it to `"default"` because it's not send for versions below 8.7 so it will not cause issues on the UI

hdinia: indeed it's better for the direct users of the API like the scripts to be informed about the error yes 👍🏼

MartinBelthle: I prefer to use "default" as a default value yes.

hdinia: I agree it makes more sense, i will update it. I need to check it's not sent for versions below 8.7 before

hdinia:
`default` instead of `""` for group default value breaks tests, the tests don't take versions into account I guess what do you think about this change @laurent-laporte-pro

laurent-laporte-pro (approved):
Nous avons observé avec Martin des incohérences entre la commande de création et la commande de mise à jour, car, cette dernière devrait avoir tous ses paramètres optionnels (avec `None`). Nous ferrons une nouvelle PR pour corriger ces incohérences, qui sera indépendante de ta correction.

PR #2003, opened 2024-04-05
feat(ui-thermals): add v8.7 thermal cluster fields

hdinia:
ANT-1463


webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Thermal/Matrix.tsx
```diff
@@ -19,91 +19,87 @@ interface Props {
...
-              <MatrixInput
-                study={study}
-                url={`input/thermal/series/${areaId}/${clusterId}/series`}
-                computStats={MatrixStats.NOCOL}
-                title={t("study.modelization.clusters.matrix.timeSeries")}
-              />
-            ),
-          ],
-        ])()}
+        {value === 0 && (
```
skamril, line 61:
```jsx
const matrixData = [
  { url: "modulation",  titleKey: "common", columns:  COMMON_MATRIX_COLS },
  ...
];

...

{matrixData.map(({ url, titleKey, columns }) => (
  <MatrixInput
    key={url}
    study={study}
    computStats={MatrixStats.NOCOL}
    url={`input/thermal/prepro/${areaId}/${clusterId}/${url}`}
    title={t(`study.modelization.clusters.matrix.${titleKey}`)}
    columnsNames={columns}
  />
))
}
```

webapp/src/utils/validationUtils.ts
```diff
@@ -150,6 +150,37 @@ export function validatePassword(password: string): string | true {
...
+ * @param options - Configuration options for validation including min and max values. (Optional)
+ * @param [options.min=Number.MIN_SAFE_INTEGER] - Minimum allowed value for the number.
+ * @param [options.max=Number.MAX_SAFE_INTEGER] - Maximum allowed value for the number.
+ * @returns True if validation is successful, or a localized error message if it fails.
+ */
+export function validateNumber(
+  value: number,
+  options?: ValidationOptions,
+): string | true {
+  if (typeof value !== "number" || isNaN(value) || !isFinite(value)) {
```
skamril, line 166: `if (!Number.isFinite(value))` is sufficient

webapp/src/utils/validationUtils.ts
```diff
@@ -150,6 +150,37 @@ export function validatePassword(password: string): string | true {
...
+/**
+ * Validates a number against specified numerical limits.
+ *
+ * @param value - The number to validate.
+ * @param options - Configuration options for validation including min and max values. (Optional)
+ * @param [options.min=Number.MIN_SAFE_INTEGER] - Minimum allowed value for the number.
+ * @param [options.max=Number.MAX_SAFE_INTEGER] - Maximum allowed value for the number.
+ * @returns True if validation is successful, or a localized error message if it fails.
+ */
+export function validateNumber(
```
skamril, line 162: Update `validateString` and `validatePassword` to use this function

webapp/public/locales/fr/main.json
```diff
@@ -467,11 +468,14 @@
...
-  "study.modelization.clusters.marginalCost": "Coûts marginaux (€/MWh)",
-  "study.modelization.clusters.fixedCost": "Coûts fixes (€/h)",
+  "study.modelization.clusters.marginalCost": "Coût marginal (€/MWh)",
+  "study.modelization.clusters.fixedCost": "Coûts fixes O&M (€/h)",
   "study.modelization.clusters.startupCost": "Coûts de démarrage (€)",
   "study.modelization.clusters.marketBidCost": "Offre de marché (€/MWh)",
-  "study.modelization.clusters.spreadCost": "Spread (€/MWh)",
+  "study.modelization.clusters.spreadCost": "Random Spread (€/MWh)",
+  "study.modelization.clusters.variableOMCost": "Coût variable O&M (€/MWh)",
+  "study.modelization.clusters.costGeneration": "TS Cost",
```
skamril, line 496: Missing translations for `spreadCost` and `costGeneration`

webapp/public/locales/fr/main.json
```diff
@@ -481,6 +485,8 @@
   "study.modelization.clusters.matrix.common": "Common",
   "study.modelization.clusters.matrix.tsGen": "TS generator",
   "study.modelization.clusters.matrix.timeSeries": "Séries temporelles",
+  "study.modelization.clusters.matrix.fuelCost": "Fuel Cost",
```
skamril, line 507: Not translated

skamril (changes requested):

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Thermal/Matrix.tsx
hdinia: thanks I thought about it too, but I don't think it's necessary to optimize here

webapp/src/utils/validationUtils.ts
hdinia: I dont understand what you mean here ? `validateString` validate string types only

webapp/public/locales/fr/main.json
hdinia: for `spreadCost` it's not missing the field is the same it's intended. For `costGeneration` it has to be confirmed and added for fr key if needed (not provided)

webapp/public/locales/fr/main.json
hdinia: same it's not provided if a translation is needed we will add it later

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Thermal/Matrix.tsx
skamril: It's not an optimization for the speed, but for the code, to remove duplicate JSX

webapp/src/utils/validationUtils.ts
skamril: Yes but `validateNumber` can be used inside:

```typescript
export function validateString(
  value: string,
  options?: ValidationOptions,
): string | true {

  ...

  //if (trimmedValue.length < min) {
  //  return t("form.field.minValue", { 0: min });
  //}

  //if (trimmedValue.length > max) {
  //  return t("form.field.maxValue", { 0: max });
  //}
  
  const lengthValidation = validateNumber(trimmedValue.length, { min, max });
  if (lengthValidation !== true) {
    return lengthValidation;
  }
  
  ...
 
}
```

Same logic for `validatePassword`.

hdinia: Thanks for the explanation I understand, but be careful about over optimization in some cases repetition is fine, here I dont think it's good:
- using `validateNumber` to check string lengths mixes responsibilities, it should only manage numbers type.
- integrating one validator within another leads to tight coupling so it decrease comprehension and risk introducing hidden bugs if numeric validation logic impacts string length checks in a future change.
- it makes futures changes harder if we want to add a special case for strings or numbers length
etc, a more appropriate approach would be to define a helper function that is shared between validators but in my opinion it's not necessary

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Thermal/Matrix.tsx
hdinia: I know, I dont think it's necessary but your point is valid I will update this

webapp/src/utils/validationUtils.ts
skamril: I think you complicate the things. You are doing the same operation 3 times, `validateNumber` is neutral and very simple, its operation is not likely to change. The function validating any number (so length too). It's normal to compose with function to not repeat itself, especially with utilities (in this file all functions are about validation of values). It is a basic concept in development, it is not an over optimization.

I would understand if the function were `validateAreasNumber` for example, even if the content would be the same, it could probably change because it is context specific.

But if you keep considering `validateNumber` not appropriate, create a function `validateStringLength` (with new new i18n keys) (can be added later). It's better to create little functions, instead of duplication.

PS: in fact the "form.field.minValue" and "form.field.maxValue" i18n keys used in `validateString` and `validatePassword` are not appropriate for lengths.

PR #2004, opened 2024-04-08
fix(bc): better handling of default values

MartinBelthle:
This pull request significantly enhances the management of default values when using the `create_binding_constraint` and `update_binding_constraint` commands. It also addresses an issue with updating terms of a binding constraint using the `update_binding_constraint` command.

Corrections have also been made to the `CommandExtractor` class, which extracts commands from a base study and a modified study through comparison, and to the `CommandReverter` class, which reverses commands based on a command history.

The generation of a study from a list of commands has been corrected: now, the generation stops as soon as an error is encountered, making it easier to identify the failing command, as it is necessarily the last one executed.


antarest/study/business/binding_constraint_management.py
```diff
@@ -642,38 +650,36 @@ def create_binding_constraint(
...
-            "values": data.values,
-            "less_term_matrix": data.less_term_matrix,
-            "equal_term_matrix": data.equal_term_matrix,
-            "greater_term_matrix": data.greater_term_matrix,
-            "filter_year_by_year": data.filter_year_by_year,
-            "filter_synthesis": data.filter_synthesis,
-            "comments": data.comments or "",
-        }
+        new_constraint = {"name": data.name}
+        properties = list(data.dict(exclude={"terms", "name", "operator", "time_step"}).keys())
```
MartinBelthle: It seems that this code is duplicated between creation and update

antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
```diff
@@ -206,17 +216,81 @@ def get_corresponding_matrices(
...
-            self.less_term_matrix = self.get_corresponding_matrices(self.less_term_matrix, version, create)
-            self.greater_term_matrix = self.get_corresponding_matrices(self.greater_term_matrix, version, create)
-            self.equal_term_matrix = self.get_corresponding_matrices(self.equal_term_matrix, version, create)
+            self.less_term_matrix = self.get_corresponding_matrices(self.less_term_matrix, time_step, version, create)
+            self.greater_term_matrix = self.get_corresponding_matrices(
+                self.greater_term_matrix, time_step, version, create
+            )
+            self.equal_term_matrix = self.get_corresponding_matrices(self.equal_term_matrix, time_step, version, create)
+
+    def apply_binding_constraint(
```
MartinBelthle, line 290: Some parts of this code could proably be merged into AbstractCommand as now both classes are optional

antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
```diff
@@ -73,15 +74,22 @@ def check_matrix_values(time_step: BindingConstraintFrequency, values: MatrixTyp
...
+class BindingConstraintProperties830(BindingConstraintProperties):
+    filter_year_by_year: str = Field("", alias="filter-year-by-year")
+    filter_synthesis: str = Field("", alias="filter-synthesis")
+
+
+class BindingConstraintProperties870(BindingConstraintProperties830):
+    group: str = "default"
+
+
+class _OptionalProperties(BindingConstraintProperties870, metaclass=AllOptionalMetaclass, use_none=True):
```
laurent-laporte-pro: The `_OptionalProperties` should be public (w/o underscore) since it is used elsewhere.

laurent-laporte-pro: Could you, please, remove the `__all__` constant. Actually, it is noisy and useless.

antarest/study/business/binding_constraint_management.py
laurent-laporte-pro: Well, we could try to avoid that…

laurent-laporte-pro (changes requested):

laurent-laporte-pro (changes requested):
I don't know where those `__getattribute__` comes from, but it is a bad practice. We should use `getattr()` function instead.

antarest/study/business/binding_constraint_management.py
laurent-laporte-pro: I think we can simplify:

```python
new_constraint = {"name": data.name, **json.loads(data.json(exclude={"terms", "name"}, exclude_none=True))}
```

MartinBelthle:
modify the constraint term to remove optional values

antarest/study/business/binding_constraint_management.py
MartinBelthle: It works, really nice :)

MartinBelthle:
There's an issue with the removal of constraint terms.

MartinBelthle:
@laurent-laporte-pro I've checked your commit and I think you should revert it because you missed this part of the code that makes your test fail inside the old `match` function in `create_binding_constraint`: 

        simple_match = self.name == other.name
        if not equal:
            return simple_match

So IMO we should not touch at this code as it would mean a big refactoring for other commands and we do not want that in this PR

laurent-laporte-pro (dismissed):

laurent-laporte-pro:
> @laurent-laporte-pro I've checked your commit and I think you should revert it because you missed this part of the code that makes your test fail inside the old `match` function in `create_binding_constraint`:
> 
> ```
>     simple_match = self.name == other.name
>     if not equal:
>         return simple_match
> ```
> 
> So IMO we should not touch at this code as it would mean a big refactoring for other commands and we do not want that in this PR

I've done a rollback + refactoring. It's OK now.

laurent-laporte-pro (approved):

PR #2006, opened 2024-04-10
feat(simulator): API change to add support for study version 8.8

MartinBelthle:
**Only supports the back end part.**
1- Study upgrader
2- New field for st storage
3- New field for thematic trimming (already existed but not in the right version)


antarest/study/storage/study_upgrader/__init__.py
```diff
@@ -196,6 +197,9 @@ def _copies_only_necessary_files(files_to_upgrade: t.List[Path], study_path: Pat
     files_to_retrieve = []
     for path in files_to_copy:
         entire_path = study_path / path
+        if not entire_path.exists():
```
laurent-laporte-pro, line 204: Je ne comprends pas bien dans quel cas cela se produit.

laurent-laporte-pro (approved):

antarest/study/storage/study_upgrader/__init__.py
MartinBelthle: I made a little explanation in the upgrader 8.8 file. I believe it can happen when upgrading a study without st storages to v8.8.

PR #2012, opened 2024-04-18
feat(launcher): add new API endpoint `/v1/launcher/time-limit` and update `LauncherDialog`

laurent-laporte-pro:
**Description:**

This pull request introduces a new API endpoint, `/v1/launcher/time-limit`, which retrieves the time limit for a job (in seconds) for a given launcher: "local" or "slurm". The value is read from the `application.yaml` configuration file.

In addition, the `LauncherDialog` has been updated to use this endpoint to initialize the default and maximum values for the `timeLimit` field. The minimum value is set to 3600 seconds. The time limit is displayed in hours, with a step of 1 hour.

This enhancement provides a more dynamic and configurable approach to setting the time limit for jobs, allowing for greater flexibility and control in job execution.

### DIFF
--- a/antarest/core/config.py
+++ b/antarest/core/config.py
@@ -391,8 +391,8 @@ class LauncherConfig:
         launcher_config = config_map.get(launcher)
         if launcher_config is None:
             raise InvalidConfigurationError(launcher)
-        # The default time limit is not available for the local launcher
-        return getattr(launcher_config, "default_time_limit", 3600)
+        # Default to 172800 (48 hours) for local launcher.
+        return getattr(launcher_config, "default_time_limit", 172800)
 
 
 @dataclass(frozen=True)
--- a/antarest/core/config.py
+++ b/antarest/core/config.py
@@ -198,12 +198,47 @@ class NbCoresConfig:
 
 
 @dataclass(frozen=True)
+class TimeLimitConfig:
+    """
+    The TimeLimitConfig class is designed to manage the configuration of the time limit for a job.
+
+    Attributes:
+        min: int: minimum allowed value for the time limit (in hours).
+        default: int: default value for the time limit (in hours).
+        max: int: maximum allowed value for the time limit (in hours).
+    """
+
+    min: int = 1
+    default: int = 48
+    max: int = 48
+
+    def to_json(self) -> Dict[str, int]:
+        """
+        Retrieves the time limit parameters, returning a dictionary containing the values "min"
+        (minimum allowed value), "defaultValue" (default value), and "max" (maximum allowed value)
+
+        Returns:
+            A dictionary: `{"min": min, "defaultValue": default, "max": max}`.
+            Because ReactJs Material UI expects "min", "defaultValue" and "max" keys.
+        """
+        return {"min": self.min, "defaultValue": self.default, "max": self.max}
+
+    def __post_init__(self) -> None:
+        """validation of CPU configuration"""
+        if 1 <= self.min <= self.default <= self.max:
+            return
+        msg = f"Invalid configuration: 1 <= {self.min=} <= {self.default=} <= {self.max=}"
+        raise ValueError(msg)
+
+
+@dataclass(frozen=True)
 class LocalConfig:
     """Sub config object dedicated to launcher module (local)"""
 
     binaries: Dict[str, Path] = field(default_factory=dict)
     enable_nb_cores_detection: bool = True
     nb_cores: NbCoresConfig = NbCoresConfig()
+    time_limit: TimeLimitConfig = TimeLimitConfig()
 
     @classmethod
     def from_dict(cls, data: JSON) -> "LocalConfig":
@@ -251,7 +286,7 @@ class SlurmConfig:
     key_password: str = ""
     password: str = ""
     default_wait_time: int = 0
-    default_time_limit: int = 0
+    time_limit: TimeLimitConfig = TimeLimitConfig()
     default_json_db_name: str = ""
     slurm_script_path: str = ""
     partition: str = ""
@@ -279,6 +314,9 @@ class SlurmConfig:
             nb_cores["max"] = max(nb_cores["max"], nb_cores["default"])
         if enable_nb_cores_detection:
             nb_cores.update(cls._autodetect_nb_cores())
+        # In the configuration file, the default time limit is in seconds, so we convert it to hours
+        max_time_limit = data.get("default_time_limit", defaults.time_limit.max * 3600) // 3600
+        time_limit = TimeLimitConfig(min=1, default=max_time_limit, max=max_time_limit)
         return cls(
             local_workspace=Path(data.get("local_workspace", defaults.local_workspace)),
             username=data.get("username", defaults.username),
@@ -288,7 +326,7 @@ class SlurmConfig:
             key_password=data.get("key_password", defaults.key_password),
             password=data.get("password", defaults.password),
             default_wait_time=data.get("default_wait_time", defaults.default_wait_time),
-            default_time_limit=data.get("default_time_limit", defaults.default_time_limit),
+            time_limit=time_limit,
             default_json_db_name=data.get("default_json_db_name", defaults.default_json_db_name),
             slurm_script_path=data.get("slurm_script_path", defaults.slurm_script_path),
             partition=data.get("partition", defaults.partition),
@@ -371,7 +409,7 @@ class LauncherConfig:
             raise InvalidConfigurationError(launcher)
         return launcher_config.nb_cores
 
-    def get_time_limit(self, launcher: str) -> int:
+    def get_time_limit(self, launcher: str) -> TimeLimitConfig:
         """
         Retrieve the time limit for a job of the given launcher: "local" or "slurm".
         If "default" is specified, retrieve the configuration of the default launcher.
@@ -380,7 +418,7 @@ class LauncherConfig:
             launcher: type of launcher "local", "slurm" or "default".
 
         Returns:
-            Time limit for a job of the given launcher.
+            Time limit for a job of the given launcher (in seconds).
 
         Raises:
             InvalidConfigurationError: Exception raised when an attempt is made to retrieve
@@ -391,8 +429,7 @@ class LauncherConfig:
         launcher_config = config_map.get(launcher)
         if launcher_config is None:
             raise InvalidConfigurationError(launcher)
-        # Default to 172800 (48 hours) for local launcher.
-        return getattr(launcher_config, "default_time_limit", 172800)
+        return launcher_config.time_limit
 
 
 @dataclass(frozen=True)
--- a/antarest/core/config.py
+++ b/antarest/core/config.py
@@ -308,7 +308,7 @@ class SlurmConfig:
 
 class InvalidConfigurationError(Exception):
     """
-    Exception raised when an attempt is made to retrieve the number of cores
+    Exception raised when an attempt is made to retrieve a property
     of a launcher that doesn't exist in the configuration.
     """
 
@@ -370,6 +370,29 @@ class LauncherConfig:
         if launcher_config is None:
             raise InvalidConfigurationError(launcher)
         return launcher_config.nb_cores
+
+    def get_time_limit(self, launcher: str) -> int:
+        """
+        Retrieve the time limit for a job of the given launcher: "local" or "slurm".
+        If "default" is specified, retrieve the configuration of the default launcher.
+
+        Args:
+            launcher: type of launcher "local", "slurm" or "default".
+
+        Returns:
+            Time limit for a job of the given launcher.
+
+        Raises:
+            InvalidConfigurationError: Exception raised when an attempt is made to retrieve
+                a property of a launcher that doesn't exist in the configuration.
+        """
+        config_map = {"local": self.local, "slurm": self.slurm}
+        config_map["default"] = config_map[self.default]
+        launcher_config = config_map.get(launcher)
+        if launcher_config is None:
+            raise InvalidConfigurationError(launcher)
+        # The default time limit is not available for the local launcher
+        return getattr(launcher_config, "default_time_limit", 3600)
 
 
 @dataclass(frozen=True)
--- a/webapp/src/components/App/Studies/LauncherDialog.tsx
+++ b/webapp/src/components/App/Studies/LauncherDialog.tsx
@@ -39,11 +39,6 @@ import CheckBoxFE from "../../common/fieldEditors/CheckBoxFE";
 import { convertVersions } from "../../../services/utils";
 import UsePromiseCond from "../../common/utils/UsePromiseCond";
 import SwitchFE from "../../common/fieldEditors/SwitchFE";
-import moment from "moment";
-
-const DEFAULT_NB_CPU = 22;
-const MIN_TIME_LIMIT = 1 * 3600; // 1 hour in seconds.
-const MAX_TIME_LIMIT = 240 * 3600; // 240 hours in seconds.
 
 interface Props {
   open: boolean;
@@ -57,9 +52,7 @@ function LauncherDialog(props: Props) {
   const { enqueueSnackbar } = useSnackbar();
   const enqueueErrorSnackbar = useEnqueueErrorSnackbar();
   const [options, setOptions] = useState<LaunchOptions>({
-    nb_cpu: DEFAULT_NB_CPU,
     auto_unzip: true,
-    time_limit: MIN_TIME_LIMIT,
   });
   const [solverVersion, setSolverVersion] = useState<string>();
   const [isLaunching, setIsLaunching] = useState(false);
@@ -92,7 +85,7 @@ function LauncherDialog(props: Props) {
         setOptions((prevOptions) => {
           return {
             ...prevOptions,
-            time_limit: timeLimit,
+            time_limit: timeLimit.defaultValue * 3600,
           };
         });
         return timeLimit;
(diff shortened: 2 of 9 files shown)


webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -273,16 +286,21 @@ function LauncherDialog(props: Props) {
             type="number"
             variant="filled"
             // Convert from seconds to hours the displayed value
-            value={(options.time_limit ?? DEFAULT_TIME_LIMIT) / 3600}
+            value={
```
skamril: Wait for the server response before enabling the field. Because if the server takes time, `DEFAULT_TIME_LIMIT` could be used when submitting.

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -49,15 +50,15 @@ interface Props {
   onClose: () => void;
 }
 
-function LauncherDialog(props: Props) {
+function LauncherDialog(props: Readonly<Props>) {
```
skamril: Remove `Readonly`

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -49,15 +50,15 @@ interface Props {
...
   const { studyIds, open, onClose } = props;
   const [t] = useTranslation();
   const { enqueueSnackbar } = useSnackbar();
   const enqueueErrorSnackbar = useEnqueueErrorSnackbar();
   const [options, setOptions] = useState<LaunchOptions>({
     nb_cpu: DEFAULT_NB_CPU,
     auto_unzip: true,
-    time_limit: DEFAULT_TIME_LIMIT,
+    time_limit: undefined,
```
skamril: Best to initialize with `null`

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -83,6 +84,18 @@ function LauncherDialog(props: Props) {
     },
   );
 
+  const { data: launcherTimeLimit } = usePromiseWithSnackbarError(
+    async () => {
```
skamril: async/await is useless

webapp/src/services/api/study.ts
```diff
@@ -226,7 +226,7 @@ export const importStudy = async (
   if (onProgress) {
     options.onUploadProgress = (progressEvent): void => {
       const percentCompleted = Math.round(
-        (progressEvent.loaded * 100) / (progressEvent.total || 1),
+        (progressEvent.loaded * 100) / (progressEvent.total ?? 1),
```
skamril: `progressEvent.total ?? 1` => if `progressEvent.total` equals `0` we will have an infinite value:
`0 ?? 1` => 0
`0 || 1` => 1

So put back `||`.

skamril (changes requested):

webapp/src/services/api/study.ts
```diff
@@ -253,7 +253,7 @@ export const importFile = async (
...
```
skamril: https://github.com/AntaresSimulatorTeam/AntaREST/pull/2012/files#r1582813726

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -83,6 +84,18 @@ function LauncherDialog(props: Props) {
...
+      return await getLauncherTimeLimit();
+    },
+    {
+      errorMessage: t("study.error.launcherTimeLimit"),
+    },
+  );
+
+  const minSeconds = 3600;
```
skamril, line 96: Create outside the component `const SECOND = 3600;` and replace all occurence of `3600` with it

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -458,7 +476,7 @@ function LauncherDialog(props: Props) {
                   name: o.name,
                 }))}
                 disabled={!!options.xpansion_r_version || !options.xpansion}
-                data={options.xpansion?.output_id || ""}
+                data={options.xpansion?.output_id ?? ""}
```
skamril: Change not necessary

webapp/src/services/api/study.ts
```diff
@@ -206,7 +206,7 @@ export const exportStudy = async (
 
 export const getExportUrl = (sid: string, skipOutputs = false): string =>
   `${
-    getConfig().downloadHostUrl ||
+    getConfig().downloadHostUrl ??
```
skamril: Keep `||`:
`"" ?? "foo"` => ""
`"" || "foo"` => "foo"

skamril (changes requested):

MartinBelthle:
I think the back-end should call this new endpoint inside the `/run` one to avoid any issue. Because this development isn't enough to avoid issues with scripts.

hdinia:
As stated in this issue: https://github.com/AntaresSimulatorTeam/AntaREST/issues/1783

The API endpoints related to this form should be consolidated into one, that returns a structured json with all the necessary data like launcher versions, nb cpu, time limit, outputs list etc.
It could be named for example `GET /v1/launcher/metrics`

webapp/src/components/App/Studies/LauncherDialog.tsx
hdinia: it's a `number`  type, would make more sense to set the initial state to 0.

also we can consider adding configuration files or using `.env` to handle env values instead of hardcoded values:
env =>
```typescript
const DEFAULT_TIME_LIMIT = process.env.REACT_APP_DEFAULT_TIME_LIMIT || 864000; // 240 hours in seconds as fallback

const [options, setOptions] = useState<LaunchOptions>({
    nb_cpu: DEFAULT_NB_CPU,
    auto_unzip: true,
    time_limit: DEFAULT_TIME_LIMIT,
  });
```
or config files =>
```typescript
// config.development.ts
export default {
  timeLimit: 864000, // 240 hours in seconds
};

// config.production.ts
export default {
  timeLimit: 172800, // 48 hours in seconds
};

// Usage
import config from './config';

const [options, setOptions] = useState<LaunchOptions>({
    nb_cpu: DEFAULT_NB_CPU,
    auto_unzip: true,
    time_limit: config.timeLimit,
  });
```

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -83,6 +84,18 @@ function LauncherDialog(props: Props) {
...
+    async () => {
+      return await getLauncherTimeLimit();
+    },
+    {
+      errorMessage: t("study.error.launcherTimeLimit"),
+    },
+  );
+
+  const minSeconds = 3600;
+  const maxSeconds = launcherTimeLimit ?? DEFAULT_TIME_LIMIT;
```
hdinia: this could lead to unexpected behaviour => you are using `const DEFAULT_TIME_LIMIT = 240 * 3600;`
to calculate the max seconds value, but the default can be different for each env, so your fallback value is potentially wrong (`?? DEFAULT_TIME_LIMIT` => always equals 240), consider using `.env` or a config file instead.(explained in the previous comment)

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -273,16 +286,21 @@ function LauncherDialog(props: Props) {
             type="number"
             variant="filled"
             // Convert from seconds to hours the displayed value
-            value={(options.time_limit ?? DEFAULT_TIME_LIMIT) / 3600}
+            value={
+              options.time_limit === undefined
+                ? maxSeconds / 3600
+                : options.time_limit / 3600
+            }
```
hdinia: revert that and set the initial state to a number (preferably using env variables in hours)

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -273,16 +286,21 @@ function LauncherDialog(props: Props) {
...
               handleChange("time_limit", parseHoursToSeconds(e.target.value))
             }
             InputLabelProps={{
               shrink: true,
             }}
             inputProps={{
-              min: 1,
-              max: 240,
+              min: minSeconds / 3600,
+              max: maxSeconds / 3600,
```
hdinia: it's confusing, 3600/3600 to set a min value to 1. Please revert that change you don't need constants for this and the validation should be separate from the initial state of the form / api return values

webapp/src/components/App/Studies/LauncherDialog.tsx
hdinia: you a are referring to hours for `minSeconds`, it should be 1 (sec) , or `minHours` (3600 sec = 1h)

also its value is set to 3600 and never changed, it is a fixed value defining the minimum seconds so it should be a constant => `const MIN_SECONDS = 1` (it's just for information we don't need these 2 variables, please consider removing them)

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -182,7 +195,7 @@ function LauncherDialog(props: Props) {
    */
   const parseHoursToSeconds = (hourString: string): number => {
     const seconds = moment.duration(hourString, "hours").asSeconds();
-    return seconds > 0 ? seconds : DEFAULT_TIME_LIMIT;
+    return Math.max(minSeconds, Math.min(seconds, maxSeconds));
```
hdinia: please revert that change, this function is not meant to enforce min/max validation ( +it's a bad practice, validation in a React form is not handeled like this usually)
some ressources:
https://ibaslogic.com/simple-guide-to-react-form/
https://stackoverflow.com/questions/72344280/how-can-i-validate-min-max-setting-dynamic-value
https://daily.dev/blog/form-on-react-best-practices

hdinia: ⚠️ ⚠️ `return Math.max(minSeconds, Math.min(seconds, maxSeconds));` => this will always return 3600, and so entirely block this field + send 3600 as value

hdinia (changes requested):

webapp/src/components/App/Studies/LauncherDialog.tsx
laurent-laporte-pro: Non, je ne crois pas que ce soit la bonne façon de faire. Je préfère éviter d’intégrer ce type de fichier de configuration du côté frontend, surtout quand un mécanisme existe déjà côté backend. Par ailleurs, il est important de considérer la version Desktop pour laquelle le time_limit n’est pas pertinent : cette valeur est exclusivement utile pour le SLURM et non pour les exécutions locales.

En réalité, c’est la responsabilité du backend de définir les valeurs par défaut, ce qui est le cas ici : le nouveau endpoint récupère la valeur en lisant le fichier `config.yaml`, qui inclut le paramètre `default_time_limit`. Il est impératif que le frontend adopte cette valeur comme limite maximale.

En fait, la valeur fixe `DEFAULT_TIME_LIMIT` dans le frontend s’avère utile uniquement si le endpoint plante (erreur 500). À la limite, on pourrait envisager une valeur nulle. Cependant, si le endpoint est hors service, il y a de fortes chances que l’application soit inutilisable. Cette valeur pourrait également être pertinente lors des tests du frontend en l’absence du backend.

hdinia: Si tu considère que le lancement doit se faire uniquement si le ce endpoint est en service dans ce cas effectivement on peut s'en passer, tu peux initialiser le state pour `time_limit` a 0 et laisser l'api renvoyer les valeurs par defaut selon l'environnement.

webapp/src/components/App/Studies/LauncherDialog.tsx
laurent-laporte-pro: Non, je n'ai pas constaté ce problème lorsque j'ai testé en intégration. Tu dois avoir un autre problème ailleurs, non  ?

Par exemple, en Python, ça fonctionne très bien :

```python
minSeconds = 3600
maxSeconds = 240 * 3600


def parse_hours_to_seconds(hour_string: str) -> int:
    seconds = int(hour_string) * 3600
    return max(minSeconds, min(seconds, maxSeconds))


assert parse_hours_to_seconds("0") == minSeconds
assert parse_hours_to_seconds("1") == minSeconds
assert parse_hours_to_seconds("123") == 123 * 3600
assert parse_hours_to_seconds("240") == maxSeconds
assert parse_hours_to_seconds("241") == maxSeconds
```

webapp/src/components/App/Studies/LauncherDialog.tsx
laurent-laporte-pro: La valeur min doit être 1h, donc on devrait bien avoir `MIN_SECONDS = 3600`, ou, si tu préfère, `MIN_HOURS = 1`. Mais certainement pas `MIN_SECONDS = 1`.

webapp/src/components/App/Studies/LauncherDialog.tsx
laurent-laporte-pro: En fait, le plus simple serait sûrement d'utiliser des heures à la place des secondes. Cela rendrait le code bien plus clair. La conversion en secondes ne se ferrait que lors des GET et POST.

webapp/src/components/App/Studies/LauncherDialog.tsx
hdinia: Oui en fait cela vient de l'api qui retourne 3600 pour le launcher local, mis a part cela la synthaxe est bonne mais niveau implementation c'est quelque chose a eviter de forcer la validation comme cela, surtout sur un ecouteur d'evenements

webapp/src/components/App/Studies/LauncherDialog.tsx
hdinia: oui erreur de ma part ce serait `MIN_HOURS = 1` mais c'etait plus pour donner un exemple de definition de constante en JS (pour info 😃 )

webapp/src/services/api/study.ts
```diff
@@ -304,6 +304,17 @@ export const getLauncherCores = async (): Promise<Record<string, number>> => {
   return res.data;
 };
 
+/**
+ * Time limit for SLURM jobs.
+ * If a jobs exceed this time limit, SLURM kills the job and it is considered failed.
+ *
+ * @returns The time limit in seconds, Often used value: 172800 (48 hours).
+ */
+export const getLauncherTimeLimit = async (): Promise<number> => {
```
laurent-laporte-pro: On devrait renommer cette fonction  `getLauncherMaxTimeLimit`, car il s'agit bien d'une valeur maximale.

laurent-laporte-pro: Si c'est plus simple, on pourrait envisager une API similaire à nb_cores, c'est à dire avec un `min`, un `defaultValue` et un `max`.

webapp/src/services/api/study.ts
```diff
@@ -304,6 +304,17 @@ export const getLauncherCores = async (): Promise<Record<string, number>> => {
   return res.data;
 };
 
+/**
+ * Time limit for SLURM jobs.
+ * If a jobs exceed this time limit, SLURM kills the job and it is considered failed.
+ *
+ * @returns The time limit in seconds, Often used value: 172800 (48 hours).
```
skamril: Fix typo `seconds. Often`

Remove `: Promise<number>`

Add `client.get<number>`

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -182,7 +201,7 @@ function LauncherDialog(props: Props) {
    */
   const parseHoursToSeconds = (hourString: string): number => {
     const seconds = moment.duration(hourString, "hours").asSeconds();
-    return seconds > 0 ? seconds : DEFAULT_TIME_LIMIT;
+    return Math.max(MIN_TIME_LIMIT, Math.min(seconds, MAX_TIME_LIMIT));
```
skamril: Use the timeLimit from server instead of `MAX_TIME_LIMIT`

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -267,35 +289,48 @@ function LauncherDialog(props: Props) {
...
+                    "time_limit",
+                    parseHoursToSeconds(e.target.value),
+                  );
+                }}
+                InputLabelProps={{
+                  shrink: true,
+                }}
+                inputProps={{
+                  min: MIN_TIME_LIMIT,
+                  max: MAX_TIME_LIMIT,
```
skamril: Use the timeLimit from server for `max`

skamril (changes requested):

antarest/core/config.py
```diff
@@ -371,6 +371,29 @@ def get_nb_cores(self, launcher: str) -> "NbCoresConfig":
             raise InvalidConfigurationError(launcher)
         return launcher_config.nb_cores
 
+    def get_time_limit(self, launcher: str) -> int:
```
skamril: You should return the minimum value also (so `get_time_limits`)

skamril (changes requested):

webapp/src/components/App/Studies/LauncherDialog.tsx
```diff
@@ -267,35 +289,48 @@ function LauncherDialog(props: Props) {
...
-              max: 240,
-            }}
-            sx={{
-              minWidth: "125px",
-            }}
+
+          <UsePromiseCond
+            response={launcherTimeLimit}
+            ifResolved={(timeLimit) => (
+              <TextField
```
skamril, line 273: Step buttons are not working on the UI field

skamril (changes requested):

webapp/src/services/api/study.ts
laurent-laporte-pro: Maintenant, le endpoint retourne un objet :
```json
{
  "min": 1,
  "defaultValue": 48,
  "max": 48
}
```

skamril (approved):

PR #2013, opened 2024-04-18
feat(archive-apis)!: use `.7z` format to archive studies

mabw-rte:
Context: 
After a benchmark comparing `.zip` vs `.7z` in order to archive studies, the conclusions drawn from the experiments showed that we may gain better efficiency in both time and memory if we use `.7z` format. 

Goal of PR:
Full migration of studies export into a `.7z` format. 
This PR may solve ANT-1551 and ANT-1549 all at once.


antarest/core/utils/utils.py
```diff
@@ -180,6 +181,16 @@ def zip_dir(dir_path: Path, zip_path: Path, remove_source_dir: bool = False) ->
         shutil.rmtree(dir_path)
 
 
+def seven_zip_dir(dir_path: Path, seven_zip_path: Path, remove_source_dir: bool = False) -> None:
+    with SevenZipFile(seven_zip_path, "w") as szf:
+        for root, _, files in os.walk(dir_path):
```
laurent-laporte-pro: La façon _moderne_ de rechercher les fichiers récursivement est d'utiliser `rglob("*")` de la bibliothèque `pathlib`.
Ensuite, tu peux utiliser `relative_to` pour calculer le chemin relatif, puis `as_posix()` pour obtenir un chemin POSIX pour le nom d'archive.

laurent-laporte-pro: Mais, en fait, si l'objectif est de compresser un répertoire complet, on peut tout simplement faire :

```python
import py7zr
with py7zr.SevenZipFile(seven_zip_path, mode="w") as archive:
    archive.writeall(dir_path)
```

À tester.

> NOTE : ne pas oublier de lire la [documentation](https://py7zr.readthedocs.io/en/latest/user_guide.html#programming-apis)

antarest/study/storage/abstract_storage_service.py
```diff
@@ -267,24 +267,24 @@ def import_output(
...
         """
         path_study = Path(metadata.path)
         with tempfile.TemporaryDirectory(dir=self.config.storage.tmp_dir) as tmpdir:
             logger.info(f"Exporting study {metadata.id} to temporary path {tmpdir}")
             tmp_study_path = Path(tmpdir) / "tmp_copy"
             self.export_study_flat(metadata, tmp_study_path, outputs)
             stopwatch = StopWatch()
-            zip_dir(tmp_study_path, target)
-            stopwatch.log_elapsed(lambda x: logger.info(f"Study {path_study} exported (zipped mode) in {x}s"))
+            seven_zip_dir(tmp_study_path, target)
```
laurent-laporte-pro: Je préfère éviter l'appel d'une fonction lorsque ça peut tenir sur une ou deux lignes

laurent-laporte-pro (commented):
Lire la doc de 7zip avant de coder 👻

antarest/study/storage/abstract_storage_service.py
```diff
@@ -9,12 +9,13 @@
 
 import numpy as np
 import pandas as pd
+import py7zr
```
flomnes, line 21: Add this module to some requirements-XXX.txt ?

antarest/study/storage/rawstudy/model/filesystem/config/files.py
```diff
@@ -45,6 +47,64 @@ class FileType(Enum):
...
+        with zipfile.ZipFile(root) as zf:
+            try:
+                with zf.open(posix_path) as f:
+                    text = f.read().decode("utf-8")
+                    return text.splitlines(keepends=False)
+            except KeyError:
+                # File not found in the ZIP archive
+                return []
+    elif root.suffix.lower() == ".7z":
+        raise NotImplementedError("7z archive not supported yet")
```
flomnes: Do we want to support 7z files for this operation sometime in the future ?

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -392,7 +395,21 @@ def unarchive(self, study: RawStudy) -> None:
...
+
+        Returns:
+            The full path of the archive file (zip or 7z).
+        """
+        archive_dir: Path = self.config.storage.archive_dir
+        for suffix in [".zip", ".7z"]:
+            path = archive_dir.joinpath(f"{study.id}{suffix}")
+            if path.is_file():
+                return path
+        return archive_dir.joinpath(f"{study.id}.7z")
```
flomnes: This line doesn't make any sense. Either we find the .7z file and return it at l. 411, or we don't and should throw some sort of error. Returning a path for a non-existent file is dangerous if the caller expects a path.

antarest/study/storage/abstract_storage_service.py
laurent-laporte-pro: The Python package `py7zr` is already included in [`requirements.txt`](https://github.com/AntaresSimulatorTeam/AntaREST/blob/b4948ca9ab3e2287586c28e6e2d92b29dd325e6f/requirements.txt#L21).

antarest/study/storage/rawstudy/model/filesystem/config/files.py
laurent-laporte-pro: If an archive is in 7z format, this must of course be handled. It may be an omission or a work in progress...

antarest/study/storage/rawstudy/raw_study_service.py
laurent-laporte-pro: There are two possible uses for this function: either to read an existing archive, or to create a new one. In the second case, the archive doesn't yet exist, and we supply an archive name in 7z format.

laurent-laporte-pro: I think this is indeed a source of confusion, and that it would be preferable to create two separate functions: one to search for a supposedly existing archive, the other to suggest a name for the creation of an archive.

laurent-laporte-pro: Avoir une méthode `find_study_path` pour trouver le chemin de l'archive en effectuant une détection automatique selon l'extension `.zip` ou `.7z`.
Avoir une autre méthode pour créer le chemin `create_archive_path`.

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -61,14 +60,17 @@ def __init__(
...
+    ) -> None:
         """
         Update metadata from study raw metadata
         Args:
             metadata: study
             fallback_on_default: use default values in case of failure
+            study_path: optional study path
         """
-        path = self.get_study_path(metadata)
+        path = study_path if study_path is not None else self.get_study_path(metadata)
```
MartinBelthle: path = study_path or self.get_study_path(metadata)

tests/study/storage/test_abstract_storage_service.py
```diff
@@ -1,124 +1,44 @@
...
-            config=Config(storage=StorageConfig(tmp_dir=tmp_dir)),
-            study_factory=Mock(spec=StudyFactory),
-            patch_service=Mock(spec=PatchService),
-            cache=Mock(spec=ICache),
-        )
+        Returns:
 
-        ## Prepare database objects
+        """
+        # Prepare a dummy study with a `study.antarest` file, and non-empty input and output folder
```
MartinBelthle: study.antares

MartinBelthle (changes requested):
Can you specify which features this PR handles because it's not really clear for me.
From my understanding, the only thing it does is handling the archiving/unarchiving process with 7z (and retro compatibility)

antarest/study/storage/abstract_storage_service.py
```diff
@@ -267,24 +269,25 @@ def import_output(
...
         path_study = Path(metadata.path)
         with tempfile.TemporaryDirectory(dir=self.config.storage.tmp_dir) as tmpdir:
             logger.info(f"Exporting study {metadata.id} to temporary path {tmpdir}")
             tmp_study_path = Path(tmpdir) / "tmp_copy"
             self.export_study_flat(metadata, tmp_study_path, outputs)
             stopwatch = StopWatch()
-            zip_dir(tmp_study_path, target)
-            stopwatch.log_elapsed(lambda x: logger.info(f"Study {path_study} exported (zipped mode) in {x}s"))
+            with py7zr.SevenZipFile(target, "w") as szf:
+                szf.writeall(tmp_study_path, arcname=".")
```
MartinBelthle: The parameter `arcname="."` creates an unwanted folder `.` at the root of the archived study.

MartinBelthle (changes requested):
It's not a major bug but your change in the code didn't fix the issue of the `.` folder inside your `.7z`archive.
Also, you didn't answer my previous comment on what does this PR really achieve.

antarest/study/service.py
```diff
@@ -1143,7 +1143,7 @@ def delete_study(self, uuid: str, children: bool, params: RequestParameters) ->
             self.storage_service.get_storage(study).delete(study)
         else:
             if isinstance(study, RawStudy):
```
laurent-laporte-pro, line 1213: Please add a UT which covers this piece of code.

antarest/study/storage/rawstudy/model/filesystem/config/model.py
```diff
@@ -169,13 +169,13 @@ def __init__(
         self.archive_input_series = archive_input_series or []
         self.enr_modelling = enr_modelling
         self.cache = cache or {}
-        self.zip_path = zip_path
+        self.archive_path = archive_path
 
     def next_file(self, name: str, is_output: bool = False) -> "FileStudyTreeConfig":
         if is_output and name in self.outputs and self.outputs[name].archived:
-            zip_path: t.Optional[Path] = self.path / f"{name}.zip"
+            archive_path: t.Optional[Path] = self.path / f"{name}.zip"
```
laurent-laporte-pro, line 196: Did you take into account in #2039 that you'll have to search for the archive in the output directory and determine whether it's in .zip or .7z format?

antarest/study/storage/rawstudy/model/filesystem/root/input/areas/list.py
```diff
@@ -32,7 +32,7 @@ def get(
         expanded: bool = False,
         formatted: bool = True,
     ) -> List[str]:
-        if self.config.zip_path:
+        if self.config.archive_path:
```
laurent-laporte-pro, line 50: This piece of code is not covered by unit test.

antarest/study/storage/rawstudy/model/filesystem/root/input/areas/list.py
```diff
@@ -32,7 +32,7 @@ def get(
...
             path, tmp_dir = self._extract_file_to_tmp_dir()
```
laurent-laporte-pro: This code is rotten. I think it would be better to use `extract_lines_from_archive`, which is optimized to extract only the necessary information.

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -35,19 +35,19 @@ def _get_real_file_path(
...
+        if self.config.archive_path:
             path, tmp_dir = self._extract_file_to_tmp_dir()
         else:
             path = self.config.path
         return path, tmp_dir
 
     def file_exists(self) -> bool:
-        if self.config.zip_path:
-            str_zipped_path = str(self.config.zip_path)
+        if self.config.archive_path:
```
laurent-laporte-pro, line 57: Test coverage must be improved in #2039

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -128,21 +130,19 @@ def update_name_and_version_from_raw_meta(self, metadata: RawStudy) -> bool:
...
         Args:
-            study: study
+            study: The study to check.
 
         Returns: true if study presents in disk, false else.
-
         """
-        path = self.get_study_path(study)
-
         if study.archived:
```
laurent-laporte-pro, line 154: This code block is not covered by unit tests.

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -309,19 +309,22 @@ def import_study(self, metadata: RawStudy, stream: t.BinaryIO) -> Study:
...
 
         try:
-            extract_zip(stream, path_study)
-            fix_study_root(path_study)
-            self.update_from_raw_meta(metadata)
+            extract_archive(stream, study_path)
+            fix_study_root(study_path)
+            self.update_from_raw_meta(
+                metadata,
+                study_path=study_path,
```
laurent-laporte-pro: style: avoid newline here

tests/storage/business/test_export.py
```diff
@@ -36,7 +37,13 @@ def test_export_file(tmp_path: Path):
 
 
 @pytest.mark.unit_test
-@pytest.mark.parametrize("outputs", [True, False])
+@pytest.mark.parametrize(
```
laurent-laporte-pro: style: avoid newlines ;-)

tests/storage/business/test_export.py
```diff
@@ -36,7 +37,13 @@ def test_export_file(tmp_path: Path):
...
+    "outputs",
+    [
+        True,
+        False,
+    ],
+)
 def test_export_file(tmp_path: Path, outputs: bool):
```
laurent-laporte-pro: Please change the name of this function, it is already used above!

laurent-laporte-pro (changes requested):
Il convient d'améliorer la couverture des tests at de tester avec ZIP et 7z.

Lorsqu'on exporte une étude, on télécharge un fichier au format 7z, mais avec l'extension ".zip". Il faut corriger cela.

antarest/study/service.py
mabw-rte: a technical debt ... not directly related to the new feature

antarest/study/storage/rawstudy/model/filesystem/root/input/areas/list.py
mabw-rte: a technical debt ... not directly related to the new feature

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
mabw-rte: a technical debt ... not directly related to the bug

antarest/study/storage/rawstudy/raw_study_service.py
mabw-rte: a technical debt ... not directly related to the new feature

antarest/core/utils/utils.py
```diff
@@ -203,20 +205,29 @@ def is_zip(path: Path) -> bool:
...
     tmp_dir = tempfile.TemporaryDirectory()
     try:
-        with zipfile.ZipFile(zip_path) as zip_obj:
-            zip_obj.extract(str_inside_zip_path, tmp_dir.name)
+        if archive_path.suffix == ".zip":
+            with zipfile.ZipFile(archive_path) as zip_obj:
+                zip_obj.extract(str_inside_archive_path, tmp_dir.name)
+        elif archive_path.suffix == ".7z":
+            with py7zr.SevenZipFile(archive_path, mode="r") as zip_obj:
+                str_inside_archive_path = (
```
MartinBelthle: Not really sure to understand why you've done this here and not for .zip ?

antarest/core/utils/utils.py
```diff
@@ -24,7 +24,9 @@
 from pathlib import Path
 
 import py7zr
+import redis
 from fastapi import HTTPException
+from py7zr import SevenZipFile
```
MartinBelthle: useless import

antarest/core/utils/utils.py
```diff
@@ -24,7 +24,9 @@
 from pathlib import Path
 
 import py7zr
+import redis
```
MartinBelthle: same

tests/study/storage/test_abstract_storage_service.py
```diff
@@ -146,18 +66,23 @@ def test_export_study(self, tmp_path: Path) -> None:
...
-            # Actually, there is nothing is the ZIP file,
-            # because the Study files doesn't really exist.
-            assert not zf.namelist()
+        # Check that the 7zip file exist and is valid
+        with SevenZipFile(target_path) as szf:
+            # Check that the content of the 7z file is the same as the study folder
+            assert set(content_list) == (
+                set(szf.getnames())
+                - {
+                    ".",
```
MartinBelthle, line 82: Don't understand why when we create a 7z archive we have this . folder. If we can remove it I'll be happy but otherwise it's not really a problem as unarchiving the archive make it disappear

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -402,11 +402,38 @@ def unarchive(self, study: RawStudy) -> None:
...
+            The full path of the archive file (zip or 7z).
+        """
+        archive_dir: Path = self.config.storage.archive_dir
+        for suffix in [".zip", ".7z"]:
+            path = archive_dir.joinpath(f"{study.id}{suffix}")
+            if path.is_file():
+                return path
+        raise FileNotFoundError(f"Study {study.id} archiving process is corrupted (no archive file found).")
+
+    def create_archive_path(self, study: RawStudy) -> Path:
```
MartinBelthle: I don't see the point of this method. You could use find_archive_path instead I believe and it would be clearer as you don't really create a path inside this method. If so, you can remove this method

MartinBelthle (changes requested):
Perhaps I missed it but I think we should have a non regression test on old zipped archived that can still be unarchived.

Also, as we agreed, please rename this PR to explain it only adresses the archiving/unarchiving part

antarest/core/utils/utils.py
mabw-rte: You have both `.zip` and `.7z` handled here

antarest/study/storage/rawstudy/raw_study_service.py
mabw-rte: This is how Laurent wanted it to be called ... but I would agree with you with slight closer name `infer_archive_path`

tests/study/storage/test_abstract_storage_service.py
mabw-rte: I couldn't just have a way around it ?!
Do you have some ideas ?

antarest/core/utils/utils.py
MartinBelthle: Yes but I meant the part where you do this:
`str_inside_archive_path = str_inside_archive_path[1:] if str_inside_archive_path.startswith("/") else str_inside_archive_path)`
You do this only inside the 7z condition and I wanted to know why

tests/study/storage/test_abstract_storage_service.py
MartinBelthle: No I don't but as I said it's not really a problem as unarchiving the study makes the `.` disappear

antarest/study/storage/rawstudy/model/filesystem/inode.py
```diff
@@ -148,17 +148,17 @@ def _extract_file_to_tmp_dir(
             The actual path of the extracted file
             the tmp_dir object which MUST be cleared after use of the file
         """
-        if self.config.zip_path is None:
+        if self.config.archive_path is None:
             raise ShouldNotHappenException()
-        inside_zip_path = str(self.config.path)[len(str(self.config.zip_path)[:-4]) + 1 :]
-        if self.config.zip_path:
-            return extract_file_to_tmp_dir(self.config.zip_path, Path(inside_zip_path))
+        inside_archive_path = str(self.config.path)[len(str(self.config.archive_path)[:-4]) + 1 :]
```
MartinBelthle: Even if you didn't write it it's ugly and false (we should use [:-3] if we are in 7z and not -4 which corresponds to zip).
We should change it to:
`inside_archive_path = self.config.path.relative_to(self.config.archive_path.parent / self.config.study_id)` which is reliable and doesn't need to recast it to a path afterwards

Also with this you will never have the `/` at the beginning of the path so you can remove the weird lines I pointed out in my last comment (line 216 inside utils.py)

antarest/study/storage/rawstudy/model/filesystem/ini_file_node.py
```diff
@@ -105,9 +105,9 @@ def _get(
         url = url or []
         kwargs = self._get_filtering_kwargs(url)
 
-        if self.config.zip_path:
-            with zipfile.ZipFile(self.config.zip_path, mode="r") as zipped_folder:
-                inside_zip_path = self.config.path.relative_to(self.config.zip_path.with_suffix("")).as_posix()
+        if self.config.archive_path:
+            with zipfile.ZipFile(self.config.archive_path, mode="r") as zipped_folder:
```
MartinBelthle: here you should also handle 7z as i tested on my local env and it failed.
You should also write a test to reproduce:
1- Create a study with one area
2- Archive it
3- Use the endpoint GET /raw towards any Ini file.
4- It should succeed and currently it fails with an HTTP 500

MartinBelthle (changes requested):

antarest/study/storage/abstract_storage_service.py
```diff
@@ -278,24 +280,25 @@ def import_output(
...
         """
         path_study = Path(metadata.path)
         with tempfile.TemporaryDirectory(dir=self.config.storage.tmp_dir) as tmpdir:
             logger.info(f"Exporting study {metadata.id} to temporary path {tmpdir}")
             tmp_study_path = Path(tmpdir) / "tmp_copy"
             self.export_study_flat(metadata, tmp_study_path, outputs)
             stopwatch = StopWatch()
-            zip_dir(tmp_study_path, target)
-            stopwatch.log_elapsed(lambda x: logger.info(f"Study {path_study} exported (zipped mode) in {x}s"))
+            with py7zr.SevenZipFile(target, "w") as szf:
```
sylvlecl: Like for the extract, it would be nice to have one method `archive_dir` with an option between the 2 formats ZIP or 7Z

tests/integration/studies_blueprint/assets/test_synthesis/variant_study.synthesis.json
```diff
@@ -983,5 +983,5 @@
   "store_new_set": true,
   "archive_input_series": [],
   "enr_modelling": "aggregated",
-  "zip_path": null
+  "archive_path": null
```
sylvlecl, line 986: Removing zip_path is a breaking change of the API.
Are we sure it does not beak clients ? (front, scripts)

antarest/core/utils/utils.py
```diff
@@ -203,20 +203,29 @@ def is_zip(path: Path) -> bool:
     return path.name.endswith(".zip")
 
 
-def extract_file_to_tmp_dir(zip_path: Path, inside_zip_path: Path) -> t.Tuple[Path, t.Any]:
-    str_inside_zip_path = str(inside_zip_path).replace("\\", "/")
+def extract_file_to_tmp_dir(archive_path: Path, inside_archive_path: Path) -> t.Tuple[Path, t.Any]:
```
sylvlecl: Let's move those methods to a more dedicated module than `utils`: `antarest.core.utils.archives` for example.

antarest/study/storage/rawstudy/model/filesystem/config/files.py
```diff
@@ -65,6 +67,78 @@ class FileType(Enum):
...
+     Args:
+          root: 7zip or ZIP file containing the study.
+          posix_path: Relative path to the file to extract.
+          reader: IniReader object to use for processing the file.
+
+    Returns:
+        The content of the file, processed according to its type:
+        - SIMPLE_INI or MULTI_INI: dictionary of keys/values
+    """
+    if root.suffix.lower() == ".zip":
```
sylvlecl: We have far too much code duplication with the previous method:
why don't we call it here and just apply the reader to the result ?

antarest/study/storage/rawstudy/model/filesystem/config/files.py
```diff
@@ -65,6 +67,78 @@ class FileType(Enum):
     MULTI_INI = "multi_ini"
 
 
+def extract_lines_from_archive(root: Path, posix_path: str) -> t.List[str]:
```
sylvlecl: This method is very generic, we can move it to `antarest.core.utils.archives`

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -402,11 +402,38 @@ def unarchive(self, study: RawStudy) -> None:
...
+            The full path of the archive file (zip or 7z).
+        """
+        archive_dir: Path = self.config.storage.archive_dir
+        for suffix in [".zip", ".7z"]:
+            path = archive_dir.joinpath(f"{study.id}{suffix}")
+            if path.is_file():
+                return path
+        raise FileNotFoundError(f"Study {study.id} archiving process is corrupted (no archive file found).")
+
+    def infer_archive_path(self, study: RawStudy) -> Path:
```
sylvlecl: should be private. Also "infer" sounds like we try to guess something complex, why not just "get" ?

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -402,11 +402,38 @@ def unarchive(self, study: RawStudy) -> None:
...
+        Create archive path of a study.
+
+        Args:
+            study: The study to get the archive path for.
+
+        Returns:
+            The full path of the archive file (7z).
+        """
+        archive_dir: Path = self.config.storage.archive_dir
+        return archive_dir.joinpath(f"{study.id}.7z")
```
sylvlecl: This hard coded ".7z" is too far away from the implementation where we actually create the a 7z archive.
This will be difficult to maintain. If we want to change again the archiving format, we will need to find again all those places.

antarest/study/storage/rawstudy/model/filesystem/config/files.py
```diff
@@ -78,10 +152,10 @@ def build(study_path: Path, study_id: str, output_path: t.Optional[Path] = None)
     Returns:
         An instance of `FileStudyTreeConfig` filled with the study data.
     """
-    is_zip_file = study_path.suffix.lower() == ".zip"
+    is_archive = study_path.suffix.lower() in {".zip", ".7z"}
```
sylvlecl: We should not hard code in this file the supported archive formats. This will be difficult to maintain.

We should have a constant and a method `is_archive` in a common module `antarest.core.utils.archives`

antarest/study/storage/rawstudy/model/filesystem/config/files.py
```diff
@@ -165,14 +197,14 @@ def _extract_data_from_file(
         - SIMPLE_INI or MULTI_INI: dictionary of keys/values
     """
 
-    is_zip_file: bool = root.suffix.lower() == ".zip"
+    is_archive: bool = root.suffix.lower() in {".zip", ".7z"}
```
sylvlecl: See comment about hard coding this in many places

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -402,11 +402,38 @@ def unarchive(self, study: RawStudy) -> None:
...
+        Fetch for archive path of a study if it exists else raise an incorrectly archived study.
+
+        Args:
+            study: The study to get the archive path for.
+
+        Returns:
+            The full path of the archive file (zip or 7z).
+        """
+        archive_dir: Path = self.config.storage.archive_dir
+        for suffix in [".zip", ".7z"]:
```
sylvlecl: Should move that logic to common module too

antarest/study/storage/utils.py
```diff
@@ -79,7 +79,7 @@ def fix_study_root(study_path: Path) -> None:
         study_path: the study initial root path
     """
     # TODO: what if it is a zipped output ?
-    if study_path.suffix == ".zip":
+    if study_path.suffix in {".zip", ".7z"}:
```
sylvlecl: Should call common module too

tests/core/utils/test_extract_zip.py
```diff
@@ -17,7 +17,7 @@
 import py7zr
 import pytest
 
-from antarest.core.utils.utils import BadArchiveContent, extract_zip
+from antarest.core.utils.utils import BadArchiveContent, extract_archive
 
 
 class TestExtractZip:
```
sylvlecl: We must add unit tests for 7zip too

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -402,11 +402,38 @@ def unarchive(self, study: RawStudy) -> None:
         Raises:
             BadArchiveContent: If the archive is corrupted or in an unknown format.
         """
-        with open(self.get_archive_path(study), mode="rb") as fh:
+        with open(self.find_archive_path(study), mode="rb") as fh:
```
sylvlecl, line 405: The unarchive method does not seem to be tested anywhere:
we need to add a test that checks that we can still read both zip and 7z archives

sylvlecl (changes requested):
I think we need to rework the implementation to make the code more maintainable/evolutive.
This handling of the archive formats is too much scattered accross services, we must centralize it more in a common module that we'll be able to change more easily.

In particular, there are lots of if/else, that would call for an interface with  2 implementations, one for zip and one for 7z.

Also:
- the reading of both zip and 7z archives seems not covered by tests
- we introduce a breaking change of the API, we should avoid this, at least checking this has no impact on known clients

tests/integration/studies_blueprint/assets/test_synthesis/variant_study.synthesis.json
mabw-rte: @sylvlecl  

That's right, Is should add `!` to the commit and @skamril  and @hdinia may include the corresponding changes.

antarest/study/storage/rawstudy/model/filesystem/config/files.py
mabw-rte: The problem is that we have a lot of code to be refactored regarding reducing the number of functions ... I guess for this part, it is better to leave for future iterations.
Otherwise, we will get stuck on refactoring all the code.

antarest/study/storage/rawstudy/raw_study_service.py
mabw-rte: As I Explained to @MartinBelthle, the name of this function was proposed by Laurent.
It was `create_archive_path`, then Martin came with another name to finally stick to this.

However, I am going to figure out a way to just not use it.

tests/core/utils/test_extract_zip.py
mabw-rte: It is already tested for `.7z`, I need to rename the class

antarest/study/storage/rawstudy/model/filesystem/inode.py
```diff
@@ -148,17 +148,17 @@ def _extract_file_to_tmp_dir(
...
         """
-        if self.config.zip_path is None:
+        if self.config.archive_path is None:
             raise ShouldNotHappenException()
-        inside_zip_path = str(self.config.path)[len(str(self.config.zip_path)[:-4]) + 1 :]
-        if self.config.zip_path:
-            return extract_file_to_tmp_dir(self.config.zip_path, Path(inside_zip_path))
+        inside_archive_path = self.config.path.relative_to(self.config.archive_path.parent / self.config.study_id)
+        if self.config.archive_path:
+            return extract_file_to_tmp_dir(self.config.archive_path, Path(inside_archive_path))
```
MartinBelthle: You can remove the cast to Path as it's already one

MartinBelthle (dismissed):

antarest/study/storage/rawstudy/model/filesystem/inode.py
MartinBelthle: Otherwise seems good to me, I'll let Sylvain handle the review from now on

antarest/study/storage/rawstudy/raw_study_service.py
mabw-rte: The `import_study` is tested and you cannot really mock it ... and most of what `unarchive` does is handled by `import_study`.

Plus there are `integration` tests for `unarchive`.

antarest/study/storage/rawstudy/model/filesystem/config/files.py
mabw-rte: ok as discussed I added a function `read_file_from_archive`

antarest/study/storage/rawstudy/raw_study_service.py
mabw-rte: As discussed I added a test for `import_study` with an archived study in `.7z` format

sylvlecl (approved):

PR #2017, opened 2024-04-25
ci(hydro): avoid SonarCloud warning about unfilled todo

laurent-laporte-pro:


MartinBelthle (approved):
I don't really understand the TODO but these change is ok for me

PR #2022, opened 2024-04-30
fix(tablemode): hide `adequacy_patch_mode` column from table-mode before v8.3

MartinBelthle:
## Description

This PR addresses an issue with the editing of Area properties in table mode. Specifically, it corrects the behavior of the `AreaOutput` model, which is used to group all properties of an area. 

Previously, the `adequacy-patch-mode` column was being populated even when it contained an empty value. This PR corrects this behavior to ensure that the `adequacy-patch-mode` column is not populated when its value is empty. That way, this column is hidden in the Table Mode view, in the UI.

In addition, this PR improves the unit tests to account for study versions 8.1, 8.3, 8.6, 8.7, and 8.8. Each of these versions adds one or more fields, with default values. The improved unit tests ensure that new fields are properly hidden for earlier versions.

### DIFF
--- a/antarest/study/business/table_mode_management.py
+++ b/antarest/study/business/table_mode_management.py
@@ -83,11 +83,7 @@ class TableModeManager:
     def _get_table_data_unsafe(self, study: RawStudy, table_type: TableModeType) -> TableDataDTO:
         if table_type == TableModeType.AREA:
             areas_map = self._area_manager.get_all_area_props(study)
-            study_version = int(study.version)
-            data = {
-                area_id: area.dict(by_alias=True, exclude={"adequacy_patch_mode"} if study_version < 830 else {})
-                for area_id, area in areas_map.items()
-            }
+            data = {area_id: area.dict(by_alias=True) for area_id, area in areas_map.items()}
         elif table_type == TableModeType.LINK:
             links_map = self._link_manager.get_all_links_props(study)
             data = {
@@ -181,11 +177,8 @@ class TableModeManager:
             # Use AreaOutput to update properties of areas
             area_props_by_ids = {key: AreaOutput(**values) for key, values in data.items()}
             areas_map = self._area_manager.update_areas_props(study, area_props_by_ids)
-            study_version = int(study.version)
-            data = {
-                area_id: area.dict(by_alias=True, exclude={"adequacy_patch_mode"} if study_version < 830 else {})
-                for area_id, area in areas_map.items()
-            }
+            data = {area_id: area.dict(by_alias=True) for area_id, area in areas_map.items()}
+            return data
         elif table_type == TableModeType.LINK:
             links_map = {tuple(key.split(" / ")): LinkOutput(**values) for key, values in data.items()}
             updated_map = self._link_manager.update_links_props(study, links_map)  # type: ignore
@@ -193,6 +186,7 @@ class TableModeManager:
                 f"{area1_id} / {area2_id}": link.dict(by_alias=True)
                 for (area1_id, area2_id), link in updated_map.items()
             }
+            return data
         elif table_type == TableModeType.THERMAL:
             thermals_by_areas: t.MutableMapping[str, t.MutableMapping[str, ThermalClusterInput]]
             thermals_by_areas = collections.defaultdict(dict)
@@ -205,6 +199,7 @@ class TableModeManager:
                 for area_id, thermals_by_ids in thermals_map.items()
                 for cluster_id, cluster in thermals_by_ids.items()
             }
+            return data
         elif table_type == TableModeType.RENEWABLE:
             renewables_by_areas: t.MutableMapping[str, t.MutableMapping[str, RenewableClusterInput]]
             renewables_by_areas = collections.defaultdict(dict)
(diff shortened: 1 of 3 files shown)


antarest/study/business/table_mode_management.py
```diff
@@ -83,7 +83,11 @@ def __init__(
     def _get_table_data_unsafe(self, study: RawStudy, table_type: TableModeType) -> TableDataDTO:
         if table_type == TableModeType.AREA:
             areas_map = self._area_manager.get_all_area_props(study)
-            data = {area_id: area.dict(by_alias=True) for area_id, area in areas_map.items()}
+            study_version = int(study.version)
```
laurent-laporte-pro: It's better not to use special code that depends on the version of a study in the Table Mode manager. This is because the `TableModeManager` class is like a facade; it delegates tasks to other managers. So, it's better to change the `get_all_area_props` method in the `AreaManager` class. This way, it won't pay attention to the `adequacy_patch_mode` parameter if the study version < 830.

antarest/study/business/table_mode_management.py
```diff
@@ -177,16 +181,18 @@ def update_table_data(
             # Use AreaOutput to update properties of areas
             area_props_by_ids = {key: AreaOutput(**values) for key, values in data.items()}
             areas_map = self._area_manager.update_areas_props(study, area_props_by_ids)
-            data = {area_id: area.dict(by_alias=True) for area_id, area in areas_map.items()}
-            return data
+            study_version = int(study.version)
```
laurent-laporte-pro: It's better not to use special code that depends on the version of a study in the Table Mode manager.

tests/integration/study_data_blueprint/test_table_mode.py
```diff
@@ -15,11 +15,9 @@ class TestTableMode:
     which contains the following areas: ["de", "es", "fr", "it"].
     """
 
+    @pytest.mark.parametrize("study_version", [0, 860])
```
laurent-laporte-pro: Rigorously, the versions to be considered for the test are 0, 810, 830 and 860…

tests/integration/study_data_blueprint/test_table_mode.py
```diff
@@ -29,16 +27,17 @@ def test_lifecycle__nominal(
...
-            f"/v1/studies/{study_id}/upgrade",
-            headers={"Authorization": f"Bearer {user_access_token}"},
-            params={"target_version": 860},
-        )
-        assert res.status_code == 200, res.json()
-
-        task_id = res.json()
-        task = wait_task_completion(client, user_access_token, task_id)
-        assert task.status == TaskStatus.COMPLETED, task
+        if study_version != 0:
```
laurent-laporte-pro: or:

```python
if study_version:
    ...
```

tests/integration/study_data_blueprint/test_table_mode.py
```diff
@@ -79,6 +78,9 @@ def test_lifecycle__nominal(
             "adequacyPatchMode",
         }
 
+        table_mode_es = {"spreadSpilledEnergyCost": None}  # not changed
+        if study_version > 830:
```
laurent-laporte-pro: No, the right condition is:

```python
 if study_version >= 830:
    ...
```

tests/integration/study_data_blueprint/test_table_mode.py
```diff
@@ -146,6 +141,11 @@ def test_lifecycle__nominal(
                 "spreadUnsuppliedEnergyCost": 0,
             },
         }
+        if study_version > 830:
```
laurent-laporte-pro: No, the right condition is:

```python
 if study_version >= 830:
    ...
```

tests/integration/study_data_blueprint/test_table_mode.py
```diff
@@ -428,339 +406,350 @@ def test_lifecycle__nominal(
...
+            "it / 02_wind_on": {"group": "Other 1", "nominalCapacity": 1000000, "unitCount": 1},
+            "it / 03_wind_off": {"group": "Other 1", "nominalCapacity": 1000000, "unitCount": 1},
+            "it / 04_res": {"group": "Other 1", "nominalCapacity": 1000000, "unitCount": 1},
+            "it / 05_nuclear": {"group": "Other 1", "nominalCapacity": 1000000, "unitCount": 1},
+            "it / 06_coal": {"group": "Other 1", "nominalCapacity": 1000000, "unitCount": 1},
+            "it / 07_gas": {"group": "Other 1", "nominalCapacity": 1000000, "unitCount": 1},
+            "it / 08_non-res": {"group": "Other 1", "nominalCapacity": 1000000, "unitCount": 1},
+            "it / 09_hydro_pump": {"group": "Other 1", "nominalCapacity": 1000000, "unitCount": 1},
         }
+        if study_version >= 860:
```
laurent-laporte-pro, line 456: Could be simplified with:

```python
if study_version >= 860:
    for key in expected:
        expected[key]["so2"] = 8.25 if key == "de / 01_solar" else 0
```

laurent-laporte-pro (changes requested):

laurent-laporte-pro (approved):

PR #2029, opened 2024-05-14
feat(ui-utils): add overload signatures for `validateNumber` function in validationUtils

skamril:
Modification to allow `validateNumber` function to support currying and composition:

```
rules={{
  validate: (v) => validateNumber(v, { min: 0 }),
}}
```

=>

```
rules={{
  validate: validateNumber({ min: 0 }),
}}
```

---

```
rules={{
  validate: (v: string) => validateNumber(Number(v))
}}
```

=> 

```
rules={{
  validate: R.compose(validateNumber, Number)
}}
```


webapp/src/utils/validationUtils.ts
```diff
@@ -4,16 +4,19 @@ import { t } from "i18next";
 // Types
 ////////////////////////////////////////////////////////////////
 
-interface ValidationOptions {
+interface NumberValidationOptions {
+  min?: number;
+  max?: number;
+}
+
+interface ValidationOptions extends NumberValidationOptions {
```
hdinia: Consolidating shared properties like min and max into a single interface eliminates redundancy but keep in mind the context, here their serve different purposes. I think we should keep them separate to simplify future additions and flexibility

webapp/src/utils/validationUtils.ts
```diff
@@ -150,20 +153,44 @@ export function validatePassword(password: string): string | true {
...
+  valueOrOpts?: number | NumberValidationOptions,
+  options?: NumberValidationOptions,
+): (string | true) | ((value: number) => string | true) {
+  if (typeof valueOrOpts !== "number") {
+    return (v: number) => validateNumber(v, valueOrOpts);
+  }
+
+  const value = valueOrOpts;
+
+  if (isFinite(value)) {
```
hdinia: you're returning an error message when the number is valid (finite). This should be changed to check for the opposite condition.

hdinia (changes requested):
Nice work on integrating currying into the `validateNumber` it makes the usage more flexible and clean. However, I'm curious about the use of function composition here. Could you explain a bit more about why we're mixing it with the validation logic, is that a need you have to use it on another PR? 

thanks for clarifying

hdinia (dismissed):

webapp/src/utils/validationUtils.ts
skamril: I forgot to commit a fix for the length message sorry.
I change this.

webapp/src/utils/validationUtils.ts
skamril: Fixed

skamril:
> Nice work on integrating currying into the `validateNumber` it makes the usage more flexible and clean. However, I'm curious about the use of function composition here. Could you explain a bit more about why we're mixing it with the validation logic, is that a need you have to use it on another PR?

Composition here it's just to show the double avantage to have this signature in general.
In fact the purpose of this PR is to add an example of signature overload in the project.

hdinia (approved):

PR #2031, opened 2024-05-16
feat(raw): refactor aggregation endpoint

MartinBelthle:
1- Speeds up the process (refactoring + new response model)
2- Add possibility to select links ids
3- Add max size limit
4- Change horizon value to fit with the spec
5- Tests added


antarest/study/common/default_values.py
```diff
@@ -33,10 +33,9 @@ class LinkProperties:
 class AreasQueryFile(str, Enum):
```
laurent-laporte-pro: Could you move `AreasQueryFile` enumeration in the `ouput_aggregator_management` module (where it is mostly used)?

antarest/study/common/default_values.py
```diff
@@ -33,10 +33,9 @@ class LinkProperties:
...
     VALUES = "values"
     DETAILS = "details"
-    DETAILS_ST_STORAGE = "details-st-storage"
+    DETAILS_ST_STORAGE = "details-STstorage"
     DETAILS_RES = "details-res"
 
 
 class LinksQueryFile(str, Enum):
```
laurent-laporte-pro: Could you move `LinksQueryFile` enumeration in the `ouput_aggregator_management` module (where it is mostly used)?

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -323,45 +320,46 @@ def aggregate_areas_raw_data(
...
+        - `mc_years`: which Monte Carlo years to be selected if empty all are selected (comma separated)
+        - `links_ids`: which links to be selected (ex: be - fr). If empty all are selected (comma separated)
+        - `columns_names`: which columns to be selected if empty all are selected (comma separated)
 
         Returns:
-            JSON (DF like matrix)  with the aggregated links raw data
-
+            FileResponse that corresponds to a dataframe with the aggregated links raw data
         """
         logger.info(
```
laurent-laporte-pro, line 278: You can avoid code duplication by calling `aggregate_areas_raw_data` directly.

antarest/study/common/default_values.py
laurent-laporte-pro: Actually, I'd like to see the `default_values.py` module disappear, as it only contains classes specific to given modules, e.g. `LinkProperties` is only used in `create_link.py`…

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+"""Index in path parts starting from the Monte Carlo year to determine the Monte Carlo year."""
+AREA_OR_LINK_INDEX = 2
+"""Index in path parts starting from the Monte Carlo year to determine the area/link name."""
+
+logger = logging.getLogger(__name__)
+
+
+def checks_expected_size(nb_files: int, df_bytes_size: int, nb_files_checked: int) -> None:
+    maximum_size = 100  # size in Mo that corresponds to a 15 seconds task.
+    expected_df_size = int(nb_files * df_bytes_size / (nb_files_checked * 10**6))
```
laurent-laporte-pro: To avoid converting to `int`, you can replace the `/` operator by `//`:

```python
expected_df_size = nb_files * df_bytes_size // (nb_files_checked * 10**6)
```

See [What is the difference between '/' and '//' when used for division?](https://stackoverflow.com/a/183870/1513933)

NOTE: this size is not "expected" but "estimated" 😜

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+    if expected_df_size > maximum_size:
+        raise FileTooLargeError(
```
laurent-laporte-pro: A good practice for exceptions is to include the error message in the exception itself, rather than at the place where it is raised (the `raise`), as there are several advantages:
- The `raise` is more concise: only the essential parameters are given.
- The error message is generally more generic, based on a formatted string.
- It's easier to update the error message, as it's in a single place.
- The error message implicitly documents the exception class.
- You can add a `__str__` method to the class to display the error message (this is what I recommend for `HTTPException` to simplify unit testing of error cases.

Here, you could write:

```python
    if expected_df_size > maximum_size:
        raise FileTooLargeError(estimated_size, maximum_size)
```

Note about the phrasing: there is no space between a value and its unit in English, so you should write: `{estimated_size}Mo`.

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+"""Column name for the timestamp."""
+MC_YEAR_INDEX = 0
+"""Index in path parts starting from the Monte Carlo year to determine the Monte Carlo year."""
+AREA_OR_LINK_INDEX = 2
+"""Index in path parts starting from the Monte Carlo year to determine the area/link name."""
+
+logger = logging.getLogger(__name__)
+
+
+def checks_expected_size(nb_files: int, df_bytes_size: int, nb_files_checked: int) -> None:
```
laurent-laporte-pro: This function should be _protected_: its name should start with "_"…

antarest/study/service.py
```diff
@@ -1,6 +1,7 @@
 import base64
 import collections
 import contextlib
+import http
```
laurent-laporte-pro, line 4: The import below should be removed:
```python
from http import HTTPStatus
```

Consider using `http.HTTPStatus` instead.

antarest/study/service.py
```diff
@@ -31,7 +32,7 @@
     UnsupportedOperationOnArchivedStudy,
     UnsupportedStudyVersion,
 )
-from antarest.core.filetransfer.model import FileDownloadTaskDTO
+from antarest.core.filetransfer.model import FileDownloadNotFound, FileDownloadTaskDTO
```
laurent-laporte-pro: The `FileDownloadNotFound` should be moved in the `exceptions` module, no?

antarest/study/service.py
```diff
@@ -63,6 +64,7 @@
 from antarest.study.business.link_management import LinkInfoDTO, LinkManager
 from antarest.study.business.matrix_management import MatrixManager, MatrixManagerError
 from antarest.study.business.optimization_management import OptimizationManager
+from antarest.study.business.output_aggregator_management import AggregatorManager
```
laurent-laporte-pro: The name of the `output_aggregator_management` module is too long for me. Why not `aggregator_management` (like the main class inside it)?

antarest/study/service.py
```diff
@@ -143,7 +144,7 @@
 def get_disk_usage(path: t.Union[str, Path]) -> int:
     """Calculate the total disk usage (in bytes) of a study in a compressed file or directory."""
     path = Path(path)
-    if path.suffix.lower() in {".zip", "7z"}:
+    if path.suffix.lower() in {".zip", ".7z"}:
```
laurent-laporte-pro, line 144: Oops!

antarest/study/service.py
```diff
@@ -327,71 +328,62 @@ def get(
...
-        Returns: data study formatted in json
+        Returns: the aggregated data in a JSON format
 
         """
         study = self.get_study(uuid)
         assert_permission(params.user, study, StudyPermissionType.READ)
-        output = self.storage_service.get_storage(study).aggregate_areas_data(
-            study, output_id, query_file, frequency, mc_years, areas_ids, columns_names
+        study_path = self.storage_service.raw_study_service.get_study_path(study)
+        aggregator_manager = AggregatorManager(
```
laurent-laporte-pro, line 358: I think it would be easier to read if we inserted a newline between each parameter.

antarest/study/service.py
```diff
@@ -327,71 +328,62 @@ def get(
...
-        query_file: AreasQueryFile,
+        query_file: t.Union[AreasQueryFile, LinksQueryFile],
         frequency: MatrixFrequency,
         mc_years: t.Sequence[int],
-        areas_ids: t.Sequence[str],
         columns_names: t.Sequence[str],
+        ids_to_consider: t.Sequence[str],
         params: RequestParameters,
-    ) -> JSON:
+    ) -> FileResponse:
```
laurent-laporte-pro: Could we consider having an `export_format` parameter (like for the export matrix endpoint)?

```python
export_format: TableExportFormat = Query(  # type: ignore
    TableExportFormat.CSV, alias="format", description="Export format", title="Export Format"
),
```

antarest/study/service.py
```diff
@@ -327,71 +328,62 @@ def get(
...
+            f"Exporting aggregated output data for study '{uuid}' as csv file",
+            params.user,
+            use_notification=False,
+            expiration_time_in_minutes=10,
         )
+        export_path = Path(export_file_download.path)
+        export_id = export_file_download.id
 
-        return output
+        try:
```
laurent-laporte-pro: Consider using the same approach as in [get_matrix](https://github.com/AntaresSimulatorTeam/AntaREST/blob/dev/antarest/study/web/raw_studies_blueprint.py#L464).
I think we can do a quick refactoring to move the logic of this file conversion into an isolated module, e.g. a new `antarest.study.df_download` module (called "DataFrame Download").

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+
+        # Retrieves the horizon from the study output
+        horizon_path = self.study_path / HORIZON_TEMPLATE.format(sim_id=self.output_id)
+        launching_config = IniReader().read(horizon_path)
+        horizon = launching_config.get("general", {}).get("horizon", 2018)
+
+        # filters files to consider
+        all_output_files = sorted(self.gather_all_files_to_consider())
+
+        logger.info(
```
laurent-laporte-pro: Split this long line, replace `self.frequency` by `self.frequency.value` to ensure you obtain the string value, and enclose the study name by quotes:

```python
logger.info(
    f"Parsing {len(all_output_files)} {self.frequency.value} files"
    f" to build the aggregated output for study '{self.study_path.name}'"
)
```

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+            df = df.reindex(columns=new_column_order)
+
+            final_df = pd.concat([final_df, df], ignore_index=True)
+
+        # replace np.nan by None
+        final_df = final_df.replace({np.nan: None})
+
+        return final_df
+
+    def aggregate_output_data(self) -> pd.DataFrame:
```
laurent-laporte-pro: All methods, unless `aggregate_output_data`, should be _protected_: their names should start by "_".

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+            new_cols.append(name_to_consider.upper().strip())
+
+        df.index = date
+        df.columns = pd.Index(new_cols)
+        return df
+
+    def filter_ids(self, folder_path: Path) -> t.List[str]:
+        if self.output_type == "areas":
+            # Areas names filtering
+            areas_names = sorted([d.name for d in folder_path.iterdir()])
```
laurent-laporte-pro: Please, could you rename into `area_ids`?

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+            if self.ids_to_consider:
+                areas_names = [area_name for area_name in areas_names if area_name in self.ids_to_consider]
```
laurent-laporte-pro: `area_name` ⇒ `area_id`

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+            return areas_names
+
+        # Links names filtering
+        links_names = sorted(d.name for d in folder_path.iterdir())
```
laurent-laporte-pro: `link_names` ⇒ `link_ids`

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+        return (
```
laurent-laporte-pro: I think, it is easier to understand like that:
```python
if self.ids_to_consider:
    return [link for link in links_names if link in self.ids_to_consider]
return links_names
```

laurent-laporte-pro: > Note that in the output directory, the format used for the links is "{area1_id} - {area2_id}", i.e. with a hyphen as separator, whereas the "%" character is traditionally used. Is the conversion done correctly in `self.ids_to_consider` when we get to this function?

antarest/study/service.py
```diff
@@ -327,71 +328,62 @@ def get(
...
+        Aggregates output data based on several filtering conditions
         Args:
             uuid: study uuid
             output_id: simulation output ID
             query_file: which types of data to retrieve ("values", "details", "details-st-storage", "details-res")
             frequency: yearly, monthly, weekly, daily or hourly.
             mc_years: list of monte-carlo years, if empty, all years are selected
-            areas_ids: list of areas names, if empty, all areas are selected
             columns_names: columns to be selected, if empty, all columns are selected
+            ids_to_consider: list of areas or links ids to consider, if empty, all areas are selected
```
laurent-laporte-pro, line 348: We need to give examples of IDs in the docstring because, for links, we traditionally use the format "{area1_id}%{area2_id}", i.e. with the percent character as a separator.

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+        # replace np.nan by None
+        final_df = final_df.replace({np.nan: None})
+
+        return final_df
+
+    def aggregate_output_data(self) -> pd.DataFrame:
+        """
+        Fetches raw data inside study's output.
+
+        Returns: the aggregated data in a JSON format
```
laurent-laporte-pro: Not in JSON format, but as a DataFrame 😜

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -283,38 +283,35 @@ def aggregate_areas_raw_data(
...
+        - `columns_names`: which columns to be selected. If empty all are selected (comma separated)
 
         Returns:
-            JSON (DF like matrix) object with the aggregated raw output data  for areas
-
+            FileResponse that corresponds to a dataframe with the aggregated areas raw data
         """
         logger.info(
-            f"Aggregate areas raw data at {query_file} (output name = {output_id}) from study {uuid}",
+            f"Aggregating areas output data for study {uuid}, output {output_id}, from files {query_file}-{frequency}.txt",
```
laurent-laporte-pro: Line too long, you can split:
```python
logger.info(
    f"Aggregating areas output data for study {uuid}, output {output_id},"
    f" from files {query_file}-{frequency}.txt",
)
```

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+    def gather_all_files_to_consider(self) -> t.Sequence[Path]:
+        # Monte Carlo years filtering
+        all_mc_years = [d.name for d in self.mc_ind_path.iterdir()]
+        if self.mc_years:
+            all_mc_years = [year for year in all_mc_years if int(year) in self.mc_years]
+        if not all_mc_years:
+            return []
+
+        # Links / Areas ids filtering
+        areas_or_links_names = self.filter_ids(self.mc_ind_path / all_mc_years[0] / self.output_type)
```
laurent-laporte-pro: `areas_or_links_names` ⇒ `areas_or_links_ids`

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def gather_all_files_to_consider(self) -> t.Sequence[Path]:
+        # Monte Carlo years filtering
+        all_mc_years = [d.name for d in self.mc_ind_path.iterdir()]
+        if self.mc_years:
+            all_mc_years = [year for year in all_mc_years if int(year) in self.mc_years]
+        if not all_mc_years:
+            return []
+
+        # Links / Areas ids filtering
```
laurent-laporte-pro: I think we could add a comment explaining why we choose `all_mc_years[0]`.

```python
# The list of areas and links is the same whatever the MC year under consideration:
# we therefore choose the first year by default.
first_mc_year = all_mc_years[0]
...
```

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+    def build_dataframe(self, files: t.Sequence[Path], horizon: int) -> pd.DataFrame:
+        final_df = pd.DataFrame()
+        nb_files = len(files)
+        for k, file_path in enumerate(files):
+            df = self.parse_output_file(file_path)
+
+            # columns filtering
+            if self.columns_names:
+                # noinspection PyTypeChecker
+                df = df[[col for col in df.columns if col in self.columns_names]]
```
laurent-laporte-pro: I could be wrong, but I think the best way to extract columns **efficiently** is to use the `.loc` method, as follows:

```python
available_cols = [c for c in df.columns.tolist() if c in self.columns_names]
df = df.loc[:, available_cols]
```

Consider reading:
- [pandas.DataFrame.loc](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.loc.html)
- [Why use loc in Pandas?](https://stackoverflow.com/q/38886080/1513933)

> With using `.loc`, the execution time is more than 3× times faster!

laurent-laporte-pro: We should also consider case-insensitive column names…

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+
+            # if no columns, no need to continue
+            list_of_df_columns = df.columns.tolist()
+            if not list_of_df_columns:
```
laurent-laporte-pro: Could be replaced by:

```python
if not df.size:
    ...
```

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+
+            # checks if the expected dataframe size does not exceed the limit
+            # This check is performed on 10 aggregated files to have a more accurate view of the final df.
+            # The following formula is the more accurate one compared to the final csv file.
+            expected_binary_size = final_df.memory_usage().sum()
+            if k == 10:
+                checks_expected_size(nb_files, expected_binary_size, k)
+
+            # add column for links/areas
+            df_length = len(df)
```
laurent-laporte-pro: Could you replace "length" with "height", which is more precise.

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+            relative_path_parts = file_path.relative_to(self.mc_ind_path).parts
+            column_name = AREA_COL if self.output_type == "areas" else LINK_COL
+            new_column_order = [column_name, MCYEAR_COL, TIME_ID_COL, TIME_COL] + list_of_df_columns
+            df[column_name] = [relative_path_parts[AREA_OR_LINK_INDEX]] * df_length
```
laurent-laporte-pro: To efficiently insert a single value to a column, you can do:

```python
df[column_name] = relative_path_parts[AREA_OR_LINK_INDEX]
```

antarest/study/business/output_aggregator_management.py
```diff
@@ -0,0 +1,209 @@
...
+
+            # add column to record the Monte Carlo year
+            df[MCYEAR_COL] = [int(relative_path_parts[MC_YEAR_INDEX])] * df_length
+
+            # add a column for the time id
+            df[TIME_ID_COL] = list(range(1, df_length + 1))
```
laurent-laporte-pro: Avoid using a `list` (which pollute the memory with a possibly big temporary array), use a `pd.Series`:

```python
df[TIME_ID_COL] = pd.Series(range(1, df_length + 1))
```

tests/integration/raw_studies_blueprint/test_aggregate_raw_data.py
```diff
@@ -9,188 +9,243 @@
...
+        },
+        "test-03.result.tsv",
+    ),
+    (
+        {
+            "output_id": "20201014-1425eco-goodbye",
+            "query_file": LinksQueryFile.VALUES,
+            "frequency": MatrixFrequency.HOURLY,
+            "mc_years": "1",
+            "links_ids": "de - fr",
```
laurent-laporte-pro, line 108: We need to choose if we use the hyphen or the percent sign as a delimiter.

tests/integration/raw_studies_blueprint/test_aggregate_raw_data.py
```diff
@@ -9,188 +9,243 @@
...
         client: TestClient,
         user_access_token: str,
         study_id: str,
-        params: t.Dict[str, t.Any],
-        expected_result_filename: str,
     ):
         """
         Test the aggregation of areas data
         """
         headers = {"Authorization": f"Bearer {user_access_token}"}
```
laurent-laporte-pro: You can simplify by using `client.headers` directly:

```python
client.headers = {"Authorization": f"Bearer {user_access_token}"}

...

res = client.get(
    f"/v1/studies/{study_id}/areas/aggregate",
    params=params,
)
```

tests/integration/raw_studies_blueprint/test_aggregate_raw_data.py
```diff
@@ -9,188 +9,243 @@
...
-        resource_file.parent.mkdir(exist_ok=True, parents=True)
-        if resource_file.exists():
-            # compare with previous result (non-regression test)
+        for params, expected_result_filename in AREAS_REQUESTS:
+            res = client.get(
+                f"/v1/studies/{study_id}/areas/aggregate",
+                params=params,
+                headers=headers,
+            )
+            assert res.status_code == 200
```
laurent-laporte-pro: Use the following construction to get a JSON error message in case of failure:

```python
```

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -283,38 +283,35 @@ def aggregate_areas_raw_data(
         areas_ids: str = "",
         columns_names: str = "",
         current_user: JWTUser = Depends(auth.get_current_user),
-    ) -> t.Dict[str, t.Any]:
+    ) -> FileResponse:
```
laurent-laporte-pro, line 192: The frontend team prefers having cameCase names:

```python
output_id: str = Query(..., alias="outputId"),
query_file: AreasQueryFile = Query(..., alias="queryFile"),
frequency: MatrixFrequency = Query(...),
mc_years: str = Query("", alias="mcYears"),
areas_ids: str = Query("", alias="areasIds"),
columns_names: str = Query("", alias="columnsNames"),
max_aggregated_files: int = Query(10, alias="maxAggregatedFiles"),
```

> NOTE: the Ellipsis may be omitted, I think.

laurent-laporte-pro: Doing that, you have to change all the HTTP request parameters in the docstring and **unit test** 👀.

laurent-laporte-pro (changes requested):
J'aime la façon dont tu as mis en œuvre l'agrégation des résultats, elle semble plus claire et probablement plus efficace que celle de Mohamed. Cela dit, il a fait un bon travail pour résoudre le problème.

J'ai formulé de nombreux commentaires, dont certains sont à prendre ou à laisser : c'est une question de goût. Néanmoins, je pense que nous pouvons faire un effort supplémentaire pour améliorer un peu.

- Le nom du module `output_machin_chose` me semble trop long,
- Les exceptions peuvent être implémentées plus subtilement et regroupées dans `exceptions`. Les codes d'erreur me semble bon.
- Je n'aime pas le module `default_values.py`, je pense qu'il devrait disparaître à terme : mettre les classes au bon endroit.
- On peut éviter la duplication de code dans les endpoints. Mettre les paramètres Query en camelCase (pour le frontend),
- L'algorithme d'agrégation me semble bon aussi. J'ai fais des suggestions concernant l'utilisation des DataFrame : attention à ne pas trop construire de list temporaire en mémoire et utiliser des pd.Slices / range, etc.
- Mettre les longs messages de logs / d'erreur sur plusieurs lignes.
- Dans les tests, il y a quelques amélioration à faire (notamment les paramètres en camelCase).
- Pour avoir une meilleure couverture de tests, on pourrait ajouter un paramètre `max_aggregated_files` (default = 10) pour tester la fonction `checks_expected_size`.

antarest/study/common/default_values.py
MartinBelthle: Done

antarest/study/common/default_values.py
MartinBelthle: Done

antarest/study/business/output_aggregator_management.py
MartinBelthle: No conversion is done, the endpoint expects ids such as `at - be`. This little example is written in the docstring of the endpoint.

antarest/study/service.py
MartinBelthle: Such examples should exist in the endpoint docstring i believe (there's already one)

antarest/study/business/output_aggregator_management.py
MartinBelthle: Why not but as I use `list_if_df_columns` after in my code, I'll let this the way it is.

tests/integration/raw_studies_blueprint/test_aggregate_raw_data.py
MartinBelthle: I think we can keep this as it is but I don't mind changing it.

antarest/study/web/raw_studies_blueprint.py
MartinBelthle: I don't really understand, can you give a more specific example ?

antarest/study/business/output_aggregator_management.py
MartinBelthle: This doesn't work, it creates NaN values and I would need to give specific index to have the right values so I'd rather keep the list

antarest/study/business/output_aggregator_management.py
MartinBelthle: I agree with the first comment, this was old code that I kept.
For the 2nd comment, I believe that each column name is capitalized so IMO we can keep the behavior as it is.

antarest/study/service.py
MartinBelthle: I asked Alexander and he said he only wanted CSV format. but why not. I'm implementing it.

antarest/study/web/raw_studies_blueprint.py
MartinBelthle: This endpoint is not designed to be used by the front-end. I think we should let this the way it is and do what you recommend if the demand evolves in the future.

antarest/study/service.py
MartinBelthle: I agree it's a bit odd but i'd rather tackle this when we'll refactor the `FileDownload` module which is kinda weird in itself.

antarest/study/service.py
MartinBelthle: I've created a file `df_donwload.py` inside the `antarest/study/storage` folder that merges the 2 approaches.

MartinBelthle:
I've answered every comment and resolved the one where I agreed with you.
For the other ones, I'll let you check what I've responded.
I've removed the `max_aggregated_files` from the endpoint as i think it was not useful and a bit confusing.

laurent-laporte-pro (approved):

PR #2040, opened 2024-05-28
feat(ui-common): refactor and update ImportDialog

skamril:
ANT-1713


webapp/src/components/common/FileTable.tsx
```diff
@@ -238,8 +215,8 @@ function FileTable(props: PropType) {
       {openImportDialog && (
         <ImportDialog
           open={openImportDialog}
-          onClose={() => setOpenImportDialog(false)}
-          onImport={onImport}
+          onCancel={() => setOpenImportDialog(false)}
+          onImport={async (file) => uploadFile?.(file)}
```
hdinia, line 219: optional: `onImport={(file) => uploadFile?.(file) ?? Promise.resolve()}`

webapp/src/components/common/dialogs/ImportDialog.tsx
```diff
@@ -44,31 +62,35 @@ function ImportDialog(props: Props) {
...
-      } catch {
-        // noop
-      } finally {
-        if (isMounted()) {
-          setIsUploading(false);
-          setUploadProgress(-1);
-        }
-      }
+  async function handleDropAccepted(acceptedFiles: File[]) {
+    setInvalidText("");
```
hdinia, line 66: if used to display error messages maybe `setError`

webapp/src/components/common/dialogs/ImportDialog.tsx
```diff
@@ -1,29 +1,47 @@
...
   dropzoneText?: string;
   accept?: Accept;
-  onClose: VoidFunction;
+  onCancel: VoidFunction;
   onImport: (
     file: File,
-    onUploadProgress: (progress: number) => void,
-  ) => Promise<void>;
+    setUploadProgress: (progress: number) => void,
+  ) => PromiseAny;
```
hdinia, line 19: any reason to force any here ? I think in most cases its a void Promise or undefined

hdinia (approved):

webapp/src/components/common/FileTable.tsx
skamril: It's the same with one keyword

webapp/src/components/common/dialogs/ImportDialog.tsx
skamril: I want to make a distinction between submit error (server side) and invalid value (front side) (like for field validation)

webapp/src/components/common/dialogs/ImportDialog.tsx
skamril: The server call can return a value (like `importFile` in 'study.ts')

PR #2052, opened 2024-06-07
feat: enhancements for areas and binding constraints management

hdinia:
#### 1. **Disable 'Areas' Tab When No Areas are Present (`fix/ui`)**
   - **Issue**: Previously, the 'Areas' tab incorrectly displayed the map component when no area data were available.
   - **Fix**: Disabled the 'Areas' tab if no areas are present. This change forces users to create areas before accessing this tab, thus ensuring UI consistency and preventing incorrect displays.

#### 2. **Add Empty Screen for Binding Constraints (`feat/ui-bc`)**
   - **Enhancement**: Implemented an empty message screen that appears when no binding constraints exist, replacing what was previously a blank or erroneous 404 screen. This provides clearer feedback to users about the absence of data.

#### 3. **Ensure Removal of the Last Term in Binding Constraints (`fix/api-bc`)**
   - **Issue**: A bug prevented the deletion of the last term in a binding constraint, without raising any execption.
   - **Fix**: Updated API logic to allow the complete removal of terms, ensuring data integrity and user control over constraint terms.

#### 4. **Minor Style Adjustments on Terms Component (`feat/ui-bc`)**
   - **Enhancement**: Adjusted padding and spacing in the terms component to improve visual alignment and user interface consistency.

#### 5. **Prevent Submitting Empty 'SelectFE' Fields (`fix/ui-bc`)**
   - **Issue**: The dialog for adding constraint terms allowed submissions with empty selects, leading to server errors.
   - **Fix**: Introduced validation to require users to select options from both dropdowns before submission, enhancing form handling and preventing unnecessary server errors.

#### 6. **Prevent '404' Error After Deletion of the Current or Last Constraint (`fix/ui-bc`)**
   - **Issue**: Deleting the currently selected or the last binding constraint led to a '404' error screen.
   - **Fix**: Implemented a post-deletion refresh of the constraints list and adjusted the UI to redirect to the first available constraint if any remain, or handle the absence of constraints more gracefully.

PR #2060, opened 2024-06-10
fix(bc): only remove terms when asked

MartinBelthle:
Rollback to old behavior (prior to https://github.com/AntaresSimulatorTeam/AntaREST/pull/2052)  + handle last term deletion

### DIFF
--- a/antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
+++ b/antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
@@ -79,12 +79,12 @@ class UpdateBindingConstraint(AbstractBindingConstraintCommand):
         updated_cfg = binding_constraints[index]
         updated_cfg.update(obj)
 
-        updated_terms = set(self.coeffs) if self.coeffs else set()
-
-        # Remove the terms not in the current update but existing in the config
-        terms_to_remove = {key for key in updated_cfg if ("%" in key or "." in key) and key not in updated_terms}
-        for term_id in terms_to_remove:
-            updated_cfg.pop(term_id, None)
+        updated_properties = self.dict(exclude={"command_context", "command_name", "version", "id"}, exclude_none=True)
+        # This 2nd check is here to remove the last term.
+        if self.coeffs or updated_properties == {"coeffs": {}}:
+            # Remove terms which IDs contain a "%" or a "." in their name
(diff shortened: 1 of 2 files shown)


antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -79,12 +79,12 @@ def _apply(self, study_data: FileStudy) -> CommandOutput:
         updated_cfg = binding_constraints[index]
         updated_cfg.update(obj)
 
-        updated_terms = set(self.coeffs) if self.coeffs else set()
-
-        # Remove the terms not in the current update but existing in the config
-        terms_to_remove = {key for key in updated_cfg if ("%" in key or "." in key) and key not in updated_terms}
-        for term_id in terms_to_remove:
-            updated_cfg.pop(term_id, None)
+        updated_properties = self.dict(exclude={"command_context", "command_name", "version", "id"}, exclude_none=True)
```
laurent-laporte-pro: You should use:

```python
excluded_fields = frozenset(ICommand.__fields__)
updated_properties = self.dict(exclude=excluded_fields, exclude_none=True)
```

tests/integration/study_data_blueprint/test_binding_constraints.py
```diff
@@ -426,6 +426,23 @@ def test_lifecycle__nominal(self, client: TestClient, user_access_token: str, st
         ]
         assert constraint_terms == expected
 
+        # Update random field, shouldn't remove the term.
+        res = client.put(
+            f"v1/studies/{study_id}/bindingconstraints/{bc_id}",
+            json={"enabled": False},
+            headers=user_headers,
+        )
+        assert res.status_code == 200
```
laurent-laporte-pro: It's better to add a message in case of assertion failure:

```python
```

laurent-laporte-pro (dismissed):
Good job! Some little improvements could be done.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
MartinBelthle: I needed to add the `id` which is not in the `ICommand` fields so the code goes like this : 

```python
excluded_fields = set(ICommand.__fields__) | {"id"}
```

laurent-laporte-pro (approved):

PR #2065, opened 2024-06-13
feat(outputs): remove useless folder `updated_links`

MartinBelthle:
**Current situation:**
The folder `updated_links` is created at the root of each output folder. But it is useless (seen with Thomas Bittar).
Furthermore, due to a little code error, we always create it (`xpansion_mode` is now `""` and not `None`)
**So it's useless AND time consuming !!**

**After this PR:**
AntaresWeb never builds this folder.


laurent-laporte-pro:
"enhancement" n'existe pas dans la [liste des messages de commit-lint](https://github.com/AntaresSimulatorTeam/AntaREST/discussions/1569) 🙂

laurent-laporte-pro (approved):

PR #2069, opened 2024-06-18
feat(ui-theme): increase scrollbar size

skamril:
ANT-1559

### DIFF
--- a/webapp/src/theme.ts
+++ b/webapp/src/theme.ts
@@ -13,14 +13,20 @@ export const PAPER_BACKGROUND_NO_TRANSPARENCY = "#212c38";
 
 export const scrollbarStyle = {
   "&::-webkit-scrollbar": {
-    width: "7px",
-    height: "7px",
+    width: "10px",
+    height: "10px",
   },
   "&::-webkit-scrollbar-track": {
     boxShadow: "inset 0 0 6px rgba(0, 0, 0, 0.3)",
   },
   "&::-webkit-scrollbar-thumb": {
     backgroundColor: SECONDARY_MAIN_COLOR,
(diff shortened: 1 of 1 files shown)


webapp/src/theme.ts
```diff
@@ -13,14 +13,20 @@ export const PAPER_BACKGROUND_NO_TRANSPARENCY = "#212c38";
...
   "&::-webkit-scrollbar-thumb": {
     backgroundColor: SECONDARY_MAIN_COLOR,
+    borderRadius: "10px",
+    border: "2px solid transparent",
+    backgroundClip: "padding-box",
+  },
+  "&::-webkit-scrollbar-thumb:hover": {
+    border: 0,
   },
 };
```
hdinia: The styles using ::-webkit-scrollbar only apply to webkit-based browsers

you can add this to support non webkit-based browsers like Firefox and others:


```javascript
scrollbarWidth: "medium", // Applies to Firefox
scrollbarColor: `${SECONDARY_MAIN_COLOR} #222333`, // Applies to Firefox, thumb and track color
```

here's the list of common browsers and their rendering engine:
Firefox - Gecko
Internet Explorer - Trident
Microsoft Edge (Legacy) - EdgeHTML
Microsoft Edge (Chromium-based) - Blink
Opera (newer versions) - Blink
Google Chrome - Blink
Safari - WebKit

> Note that Blink is a fork of Webkit so it should work properly on it

hdinia (changes requested):

webapp/src/theme.ts
skamril: That break style on WebKit browser

PR #2077, opened 2024-06-25
feat(bc): avoid unnecessary creation of RHS matrices for binding constraints

mabw-rte:
Context: 
We create many matrices, meanwhile some them are useless

Issue:
Over consumption of disk space

Solution:
New paradigm to avoid creating useless matrices

### DIFF
--- a/antarest/core/exceptions.py
+++ b/antarest/core/exceptions.py
@@ -2,6 +2,7 @@ import re
 import typing as t
 from http import HTTPStatus
 
+from fastapi import HTTPException
 from fastapi.exceptions import HTTPException
 
 
@@ -603,3 +604,8 @@ class CannotScanInternalWorkspace(HTTPException):
             HTTPStatus.BAD_REQUEST,
             "You cannot scan the default internal workspace",
         )
+
+
+class ChildNotFoundError(HTTPException):
+    def __init__(self, message: str) -> None:
+        super().__init__(HTTPStatus.NOT_FOUND, message)
--- a/antarest/core/exceptions.py
+++ b/antarest/core/exceptions.py
@@ -2,7 +2,6 @@ import re
 import typing as t
 from http import HTTPStatus
 
-from fastapi import HTTPException
 from fastapi.exceptions import HTTPException
 
 
--- a/antarest/study/business/binding_constraint_management.py
+++ b/antarest/study/business/binding_constraint_management.py
@@ -46,11 +46,11 @@ from antarest.study.storage.variantstudy.business.matrix_constants.binding_const
 from antarest.study.storage.variantstudy.model.command.create_binding_constraint import (
     DEFAULT_GROUP,
     EXPECTED_MATRIX_SHAPES,
-    TERM_MATRICES,
     BindingConstraintMatrices,
     BindingConstraintPropertiesBase,
     CreateBindingConstraint,
     OptionalProperties,
+    TermMatrices,
 )
 from antarest.study.storage.variantstudy.model.command.remove_binding_constraint import RemoveBindingConstraint
 from antarest.study.storage.variantstudy.model.command.update_binding_constraint import UpdateBindingConstraint
@@ -60,10 +60,10 @@ logger = logging.getLogger(__name__)
 
 
 OPERATOR_CONFLICT_MAP = {
-    BindingConstraintOperator.EQUAL: ["less_term_matrix", "greater_term_matrix"],
-    BindingConstraintOperator.GREATER: ["less_term_matrix", "equal_term_matrix"],
-    BindingConstraintOperator.LESS: ["equal_term_matrix", "greater_term_matrix"],
-    BindingConstraintOperator.BOTH: ["equal_term_matrix"],
+    BindingConstraintOperator.EQUAL: [TermMatrices.LESS.value, TermMatrices.GREATER.value],
+    BindingConstraintOperator.GREATER: [TermMatrices.LESS.value, TermMatrices.EQUAL.value],
+    BindingConstraintOperator.LESS: [TermMatrices.EQUAL.value, TermMatrices.GREATER.value],
+    BindingConstraintOperator.BOTH: [TermMatrices.EQUAL.value],
 }
 
 
@@ -254,7 +254,7 @@ class ConstraintCreation(ConstraintInput):
 
     @root_validator(pre=True)
     def check_matrices_dimensions(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
-        for _key in ["time_step"] + TERM_MATRICES:
+        for _key in ["time_step"] + [m.value for m in TermMatrices]:
             _camel = to_camel_case(_key)
             values[_key] = values.pop(_camel, values.get(_key))
 
@@ -272,7 +272,7 @@ class ConstraintCreation(ConstraintInput):
 
         # Collect the matrix shapes
         matrix_shapes = {}
-        for _field_name in ["values"] + TERM_MATRICES:
+        for _field_name in ["values"] + [m.value for m in TermMatrices]:
             if _matrix := values.get(_field_name):
                 _array = np.array(_matrix)
                 # We only store the shape if the array is not empty
@@ -353,7 +353,7 @@ def _get_references_by_widths(
     references_by_width: t.Dict[int, t.List[t.Tuple[str, str]]] = {}
     _total = len(bcs) * len(matrix_id_fmts)
     for _index, (bc, fmt) in enumerate(itertools.product(bcs, matrix_id_fmts), 1):
-        if int(file_study.config.version) >= 870 and fmt not in operator_matrix_file_map.get(bc.operator, []):
+        if int(file_study.config.version) >= 870 and fmt not in operator_matrix_file_map[bc.operator]:
             continue
         bc_id = bc.id
         matrix_id = fmt.format(bc_id=bc.id)
@@ -758,7 +758,7 @@ class BindingConstraintManager:
 
         # Validates the matrices. Needed when the study is a variant because we only append the command to the list
         if isinstance(study, VariantStudy):
-            updated_matrices = [term for term in TERM_MATRICES if getattr(data, term)]
+            updated_matrices = [term for term in [m.value for m in TermMatrices] if getattr(data, term)]
             time_step = data.time_step or existing_constraint.time_step
             command.validates_and_fills_matrices(
                 time_step=time_step, specific_matrices=updated_matrices, version=study_version, create=False
@@ -930,7 +930,7 @@ def _replace_matrices_according_to_frequency_and_version(
             BindingConstraintFrequency.DAILY.value: default_bc_weekly_daily_87,
             BindingConstraintFrequency.WEEKLY.value: default_bc_weekly_daily_87,
         }[data.time_step].tolist()
-        for term in TERM_MATRICES:
+        for term in [m.value for m in TermMatrices]:
             if term not in args:
                 args[term] = matrix
     return args
@@ -941,6 +941,8 @@ def check_attributes_coherence(
     study_version: int,
     existing_operator: t.Optional[BindingConstraintOperator] = None,
 ) -> None:
+    update_operator = data.operator or existing_operator
+
     if study_version < 870:
         if data.group:
             raise InvalidFieldForVersionError(
@@ -950,19 +952,21 @@ def check_attributes_coherence(
             raise InvalidFieldForVersionError("You cannot fill a 'matrix_term' as these values refer to v8.7+ studies")
     elif data.values:
         raise InvalidFieldForVersionError("You cannot fill 'values' as it refers to the matrix before v8.7")
-    elif data.operator:
+    elif update_operator:
         conflicting_matrices = [
-            getattr(data, matrix) for matrix in OPERATOR_CONFLICT_MAP[data.operator] if getattr(data, matrix)
+            getattr(data, matrix) for matrix in OPERATOR_CONFLICT_MAP[update_operator] if getattr(data, matrix)
         ]
         if conflicting_matrices:
             raise InvalidFieldForVersionError(
-                f"You cannot fill matrices '{conflicting_matrices}' while using the operator '{data.operator}'"
+                f"You cannot fill matrices '{OPERATOR_CONFLICT_MAP[update_operator]}' while using the operator '{update_operator}'"
             )
-    elif existing_operator:
-        conflicting_matrices = [
-            getattr(data, matrix) for matrix in OPERATOR_CONFLICT_MAP[existing_operator] if getattr(data, matrix)
-        ]
-        if conflicting_matrices:
-            raise InvalidFieldForVersionError(
-                f"You cannot fill matrices '{conflicting_matrices}' while using the operator '{existing_operator}'"
-            )
+    # TODO: the default operator should be fixed somewhere so this condition can be consistent
+    elif [
+        getattr(data, matrix)
+        for matrix in OPERATOR_CONFLICT_MAP[BindingConstraintOperator.EQUAL]
+        if getattr(data, matrix)
+    ]:
+        raise InvalidFieldForVersionError(
+            f"You cannot fill one of the matrices '{OPERATOR_CONFLICT_MAP[BindingConstraintOperator.EQUAL]}' "
+            "while using the operator '{BindingConstraintOperator.EQUAL}'"
+        )
--- a/antarest/study/business/binding_constraint_management.py
+++ b/antarest/study/business/binding_constraint_management.py
@@ -59,6 +59,14 @@ from antarest.study.storage.variantstudy.model.dbmodel import VariantStudy
 logger = logging.getLogger(__name__)
 
 
+OPERATOR_CONFLICT_MAP = {
+    BindingConstraintOperator.EQUAL: ["less_term_matrix", "greater_term_matrix"],
+    BindingConstraintOperator.GREATER: ["less_term_matrix", "equal_term_matrix"],
+    BindingConstraintOperator.LESS: ["equal_term_matrix", "greater_term_matrix"],
+    BindingConstraintOperator.BOTH: ["equal_term_matrix"],
+}
+
+
 class LinkTerm(BaseModel):
     """
     DTO for a constraint term on a link between two areas.
@@ -335,9 +343,18 @@ def _get_references_by_widths(
     else:
         matrix_id_fmts = {"{bc_id}_eq", "{bc_id}_lt", "{bc_id}_gt"}
 
+    operator_matrix_file_map = {
+        BindingConstraintOperator.EQUAL: ["{bc_id}_eq"],
+        BindingConstraintOperator.GREATER: ["{bc_id}_gt"],
+        BindingConstraintOperator.LESS: ["{bc_id}_lt"],
+        BindingConstraintOperator.BOTH: ["{bc_id}_lt", "{bc_id}_gt"],
+    }
+
     references_by_width: t.Dict[int, t.List[t.Tuple[str, str]]] = {}
     _total = len(bcs) * len(matrix_id_fmts)
     for _index, (bc, fmt) in enumerate(itertools.product(bcs, matrix_id_fmts), 1):
+        if int(file_study.config.version) >= 870 and fmt not in operator_matrix_file_map.get(bc.operator, []):
+            continue
         bc_id = bc.id
         matrix_id = fmt.format(bc_id=bc.id)
         logger.info(f"⏲ Validating BC '{bc_id}': {matrix_id=} [{_index}/{_total}]")
@@ -718,8 +735,9 @@ class BindingConstraintManager:
     ) -> ConstraintOutput:
         file_study = self.storage_service.get_storage(study).get_raw(study)
         existing_constraint = self.get_binding_constraint(study, binding_constraint_id)
+
         study_version = int(study.version)
-        check_attributes_coherence(data, study_version)
+        check_attributes_coherence(data, study_version, existing_constraint.operator)
 
         upd_constraint = {
             "id": binding_constraint_id,
@@ -918,7 +936,11 @@ def _replace_matrices_according_to_frequency_and_version(
     return args
 
 
-def check_attributes_coherence(data: t.Union[ConstraintCreation, ConstraintInput], study_version: int) -> None:
+def check_attributes_coherence(
+    data: t.Union[ConstraintCreation, ConstraintInput],
+    study_version: int,
+    existing_operator: t.Optional[BindingConstraintOperator] = None,
+) -> None:
     if study_version < 870:
         if data.group:
             raise InvalidFieldForVersionError(
@@ -928,3 +950,19 @@ def check_attributes_coherence(data: t.Union[ConstraintCreation, ConstraintInput
             raise InvalidFieldForVersionError("You cannot fill a 'matrix_term' as these values refer to v8.7+ studies")
     elif data.values:
         raise InvalidFieldForVersionError("You cannot fill 'values' as it refers to the matrix before v8.7")
+    elif data.operator:
+        conflicting_matrices = [
+            getattr(data, matrix) for matrix in OPERATOR_CONFLICT_MAP[data.operator] if getattr(data, matrix)
+        ]
+        if conflicting_matrices:
+            raise InvalidFieldForVersionError(
+                f"You cannot fill matrices '{conflicting_matrices}' while using the operator '{data.operator}'"
+            )
+    elif existing_operator:
+        conflicting_matrices = [
+            getattr(data, matrix) for matrix in OPERATOR_CONFLICT_MAP[existing_operator] if getattr(data, matrix)
+        ]
+        if conflicting_matrices:
+            raise InvalidFieldForVersionError(
+                f"You cannot fill matrices '{conflicting_matrices}' while using the operator '{existing_operator}'"
+            )
--- a/antarest/study/business/binding_constraint_management.py
+++ b/antarest/study/business/binding_constraint_management.py
@@ -958,7 +958,8 @@ def check_attributes_coherence(
         ]
         if conflicting_matrices:
             raise InvalidFieldForVersionError(
-                f"You cannot fill matrices '{OPERATOR_CONFLICT_MAP[update_operator]}' while using the operator '{update_operator}'"
+                f"You cannot fill matrices '{OPERATOR_CONFLICT_MAP[update_operator]}' while using the operator "
+                f"'{update_operator}'"
             )
     # TODO: the default operator should be fixed somewhere so this condition can be consistent
     elif [
--- a/antarest/study/business/binding_constraint_management.py
+++ b/antarest/study/business/binding_constraint_management.py
@@ -956,5 +956,5 @@ def check_attributes_coherence(
     ]
     if conflicting_matrices:
         raise InvalidFieldForVersionError(
-            f"You cannot fill matrices '{OPERATOR_CONFLICT_MAP[operator]}' while using the operator " f"'{operator}'"
+            f"You cannot fill matrices '{OPERATOR_CONFLICT_MAP[operator]}' while using the operator '{operator}'"
         )
--- a/antarest/study/business/binding_constraint_management.py
+++ b/antarest/study/business/binding_constraint_management.py
@@ -1,5 +1,4 @@
 import collections
-import itertools
 import json
 import logging
 import typing as t
@@ -338,11 +337,6 @@ def _get_references_by_widths(
         The height of the matrices may vary depending on the time step,
         but the width should be consistent within a group of binding constraints.
     """
-    if int(file_study.config.version) < 870:
-        matrix_id_fmts = {"{bc_id}"}
-    else:
-        matrix_id_fmts = {"{bc_id}_eq", "{bc_id}_lt", "{bc_id}_gt"}
-
     operator_matrix_file_map = {
         BindingConstraintOperator.EQUAL: ["{bc_id}_eq"],
         BindingConstraintOperator.GREATER: ["{bc_id}_gt"],
@@ -351,28 +345,27 @@ def _get_references_by_widths(
     }
 
     references_by_width: t.Dict[int, t.List[t.Tuple[str, str]]] = {}
-    _total = len(bcs) * len(matrix_id_fmts)
-    for _index, (bc, fmt) in enumerate(itertools.product(bcs, matrix_id_fmts), 1):
-        if int(file_study.config.version) >= 870 and fmt not in operator_matrix_file_map[bc.operator]:
-            continue
-        bc_id = bc.id
-        matrix_id = fmt.format(bc_id=bc.id)
-        logger.info(f"⏲ Validating BC '{bc_id}': {matrix_id=} [{_index}/{_total}]")
-        obj = file_study.tree.get(url=["input", "bindingconstraints", matrix_id])
-        matrix = np.array(obj["data"], dtype=float)
-        # We ignore empty matrices as there are default matrices for the simulator.
-        if not matrix.size:
-            continue
+    _total = len(bcs)
+    for _index, bc in enumerate(bcs):
+        matrices_name = operator_matrix_file_map[bc.operator] if file_study.config.version >= 870 else ["{bc_id}"]
+        for matrix_name in matrices_name:
+            matrix_id = matrix_name.format(bc_id=bc.id)
+            logger.info(f"⏲ Validating BC '{bc.id}': {matrix_id=} [{_index+1}/{_total}]")
+            obj = file_study.tree.get(url=["input", "bindingconstraints", matrix_id])
+            matrix = np.array(obj["data"], dtype=float)
+            # We ignore empty matrices as there are default matrices for the simulator.
+            if not matrix.size:
+                continue
 
-        matrix_height = matrix.shape[0]
-        expected_height = EXPECTED_MATRIX_SHAPES[bc.time_step][0]
-        if matrix_height != expected_height:
-            raise WrongMatrixHeightError(
-                f"The binding constraint '{bc.name}' should have {expected_height} rows, currently: {matrix_height}"
-            )
-        matrix_width = matrix.shape[1]
-        if matrix_width > 1:
-            references_by_width.setdefault(matrix_width, []).append((bc_id, matrix_id))
+            matrix_height = matrix.shape[0]
+            expected_height = EXPECTED_MATRIX_SHAPES[bc.time_step][0]
+            if matrix_height != expected_height:
+                raise WrongMatrixHeightError(
+                    f"The binding constraint '{bc.name}' should have {expected_height} rows, currently: {matrix_height}"
+                )
+            matrix_width = matrix.shape[1]
+            if matrix_width > 1:
+                references_by_width.setdefault(matrix_width, []).append((bc.id, matrix_id))
 
     return references_by_width
 
--- a/antarest/study/business/binding_constraint_management.py
+++ b/antarest/study/business/binding_constraint_management.py
@@ -701,7 +701,8 @@ class BindingConstraintManager:
         if bc_id in {bc.id for bc in self.get_binding_constraints(study)}:
             raise DuplicateConstraintName(f"A binding constraint with the same name already exists: {bc_id}.")
 
-        check_attributes_coherence(data, version)
+        # TODO: the default operator should be fixed somewhere so this condition can be consistent
+        check_attributes_coherence(data, version, data.operator or BindingConstraintOperator.EQUAL)
 
         new_constraint = {"name": data.name, **json.loads(data.json(exclude={"terms", "name"}, exclude_none=True))}
         args = {
@@ -737,7 +738,7 @@ class BindingConstraintManager:
         existing_constraint = self.get_binding_constraint(study, binding_constraint_id)
 
         study_version = int(study.version)
-        check_attributes_coherence(data, study_version, existing_constraint.operator)
+        check_attributes_coherence(data, study_version, data.operator or existing_constraint.operator)
 
         upd_constraint = {
             "id": binding_constraint_id,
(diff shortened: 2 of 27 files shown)


antarest/core/exceptions.py
```diff
@@ -2,6 +2,7 @@
 import typing as t
 from http import HTTPStatus
 
+from fastapi import HTTPException
```
MartinBelthle: Remove unused import

antarest/study/business/binding_constraint_management.py
```diff
@@ -59,6 +59,14 @@
 logger = logging.getLogger(__name__)
 
 
+OPERATOR_CONFLICT_MAP = {
```
MartinBelthle, line 61: We should probably change the `TERM_MATRICES` inside `create_binding_constraint.py` to make it an Enum. Or at least reuse these values here instead of duplicating the code.

antarest/study/business/binding_constraint_management.py
```diff
@@ -335,9 +343,18 @@ def _get_references_by_widths(
...
+        BindingConstraintOperator.EQUAL: ["{bc_id}_eq"],
+        BindingConstraintOperator.GREATER: ["{bc_id}_gt"],
+        BindingConstraintOperator.LESS: ["{bc_id}_lt"],
+        BindingConstraintOperator.BOTH: ["{bc_id}_lt", "{bc_id}_gt"],
+    }
+
     references_by_width: t.Dict[int, t.List[t.Tuple[str, str]]] = {}
     _total = len(bcs) * len(matrix_id_fmts)
     for _index, (bc, fmt) in enumerate(itertools.product(bcs, matrix_id_fmts), 1):
+        if int(file_study.config.version) >= 870 and fmt not in operator_matrix_file_map.get(bc.operator, []):
```
MartinBelthle: Do not get `[]` by default as it's not possible to have a key that's not in the dictionnary.

Also don't cast to int as it's already an int (same remark line 341).
And refactor your code to avoid checking the version for each BC as it's the same for all of them.

antarest/study/business/binding_constraint_management.py
```diff
@@ -928,3 +950,19 @@ def check_attributes_coherence(data: t.Union[ConstraintCreation, ConstraintInput
             raise InvalidFieldForVersionError("You cannot fill a 'matrix_term' as these values refer to v8.7+ studies")
     elif data.values:
         raise InvalidFieldForVersionError("You cannot fill 'values' as it refers to the matrix before v8.7")
+    elif data.operator:
```
MartinBelthle: I think this function should only have one operator as a parameter. For creation, `data.operator` and for update `data.operator or existing_operator`

antarest/study/business/binding_constraint_management.py
```diff
@@ -928,3 +950,19 @@ def check_attributes_coherence(data: t.Union[ConstraintCreation, ConstraintInput
...
+        conflicting_matrices = [
+            getattr(data, matrix) for matrix in OPERATOR_CONFLICT_MAP[data.operator] if getattr(data, matrix)
+        ]
+        if conflicting_matrices:
+            raise InvalidFieldForVersionError(
+                f"You cannot fill matrices '{conflicting_matrices}' while using the operator '{data.operator}'"
+            )
+    elif existing_operator:
```
MartinBelthle: For creation, operator being optional you have to take the default value if not given. Otherwise you won't crash when you're supposed to

antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
```diff
@@ -356,11 +356,18 @@ def apply_binding_constraint(
...
-            ["lt", "eq", "gt"],
-        ):
+        operator_matrices_map = {
+            BindingConstraintOperator.EQUAL: [(self.equal_term_matrix, "eq")],
+            BindingConstraintOperator.GREATER: [(self.greater_term_matrix, "gt")],
+            BindingConstraintOperator.LESS: [(self.less_term_matrix, "lt")],
+            BindingConstraintOperator.BOTH: [(self.less_term_matrix, "lt"), (self.greater_term_matrix, "gt")],
+        }
+
+        current_operator = self.operator or BindingConstraintOperator(
```
MartinBelthle: You can simplify this.
`current_operator = self.operator or BindingConstraintOperator(binding_constraints[new_key]["operator"])`

antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
```diff
@@ -356,11 +356,18 @@ def apply_binding_constraint(
...
+            [bc for bc in binding_constraints.values() if bc.get("id") == bd_id][0].get("operator")
+        )
+
+        for matrix_term, matrix_alias in operator_matrices_map.get(current_operator, []):
```
MartinBelthle: Don't return `[]` as it makes no sense.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -16,7 +25,189 @@
...
+
+    Raises:
+        NotImplementedError: if the case is not handled
+    """
+
+    if existing_operator == new_operator:
+        return
+
+    parent_folder_node = file_study.tree.get_node(["input", "bindingconstraints"])
+    assert isinstance(parent_folder_node, FolderNode), f"Node type not handled yet: {type(parent_folder_node)}"
```
MartinBelthle: Remove this check it doesn't make sense

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -16,7 +25,189 @@
...
+    matrix_paths = _infer_input_and_target_paths(
+        parent_folder_node, binding_constraint_id, existing_operator, new_operator
+    )
+
+    # TODO: due to legacy matrices generation, we need to check if the new matrix file already exists
+    #  and if it does, we need to first remove it before renaming the existing matrix file
+
+    if existing_operator != BindingConstraintOperator.BOTH and new_operator != BindingConstraintOperator.BOTH:
+        (matrix_path,) = matrix_paths.matrix_input_paths
+        (new_matrix_path,) = matrix_paths.matrix_target_paths
```
MartinBelthle: It's a weird syntax, I'd prefer you loop through lists it would be clearer

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -16,7 +25,189 @@
...
+        file_study: the file study
+        binding_constraint_id: the binding constraint ID
+        existing_operator: the existing operator
+        new_operator: the new operator
+
+    Raises:
+        NotImplementedError: if the case is not handled
+    """
+
+    if existing_operator == new_operator:
```
MartinBelthle: Due to legacy matrices generation, we should handle such a case by unlinking every unconcerned matrix.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -16,7 +25,189 @@
...
+            matrix_path_eq.rename(matrix_path_lt)
+            # copy the matrix lt to gt
+            shutil.copy(matrix_path_lt, matrix_path_gt)
+        elif existing_operator == BindingConstraintOperator.LESS:
+            matrix_path_gt.unlink(missing_ok=True)
+            shutil.copy(matrix_path_lt, matrix_path_gt)
+        elif existing_operator == BindingConstraintOperator.GREATER:
+            matrix_path_lt.unlink(missing_ok=True)
+            shutil.copy(matrix_path_gt, matrix_path_lt)
+    elif existing_operator == BindingConstraintOperator.BOTH:
```
MartinBelthle: this is a else

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -16,7 +25,189 @@
...
+            (matrix_path_eq,) = matrix_paths.matrix_input_paths
+            matrix_path_lt.unlink(missing_ok=True)
+            matrix_path_gt.unlink(missing_ok=True)
+            matrix_path_eq.rename(matrix_path_lt)
+            # copy the matrix lt to gt
+            shutil.copy(matrix_path_lt, matrix_path_gt)
+        elif existing_operator == BindingConstraintOperator.LESS:
+            matrix_path_gt.unlink(missing_ok=True)
+            shutil.copy(matrix_path_lt, matrix_path_gt)
+        elif existing_operator == BindingConstraintOperator.GREATER:
```
MartinBelthle: Leaving on an `elif` is a really weird design but I'll think you'll fix it while handling operator equality.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -16,7 +25,189 @@
...
+            matrix_path_lt.unlink(missing_ok=True)
+            shutil.copy(matrix_path_gt, matrix_path_lt)
+    elif existing_operator == BindingConstraintOperator.BOTH:
+        matrix_path_lt, matrix_path_gt = matrix_paths.matrix_input_paths
+        if new_operator == BindingConstraintOperator.EQUAL:
+            # TODO: we may retrieve the mean of the two matrices, but here we just copy the lt matrix
```
MartinBelthle: What does this comment mean ?

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -16,7 +25,189 @@
...
+        parent_node: the parent folder node
+        binding_constraint_id: the binding constraint ID
+        existing_operator: the existing operator
+        new_operator: the new operator
+
+    Returns:
+        the matrix input and target paths
+    """
+    # new and existing operators should be different
+    assert existing_operator != new_operator, "Existing and new operators should be different"
```
MartinBelthle: I think you'll change this when handling operator equality

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -16,7 +25,189 @@
...
+
+    matrix_node_lt = parent_node.get_node([f"{binding_constraint_id}_lt"])
+    assert isinstance(
```
MartinBelthle: Remove all these checks they're useless

tests/integration/study_data_blueprint/test_binding_constraints.py
```diff
@@ -542,47 +513,28 @@ def test_lifecycle__nominal(self, client: TestClient, user_access_token: str, st
         assert res.json()["exception"] == "InvalidFieldForVersionError"
         assert res.json()["description"] == "You cannot fill a 'matrix_term' as these values refer to v8.7+ studies"
 
-    @pytest.mark.parametrize("study_type", ["raw", "variant"])
+    @pytest.mark.parametrize(
```
MartinBelthle: You should put that in one line back

tests/integration/study_data_blueprint/test_binding_constraints.py
```diff
@@ -746,13 +718,29 @@ def test_for_version_870(self, client: TestClient, user_access_token: str, study
...
         assert res.json()["group"] == grp_name
 
-        # Update matrix_term
+        # check that updating of a binding constraint that has an operator "equal"
+        # with a greater matrix will raise an error 422
+        res = client.put(
+            f"/v1/studies/{study_id}/bindingconstraints/{bc_id_w_matrix}",
+            json={"greater_term_matrix": matrix_lt3.tolist()},
+        )
+        assert res.status_code == 422, res.json()
```
MartinBelthle, line 724: Could you also check the exception msg or class

tests/integration/study_data_blueprint/test_binding_constraints.py
```diff
@@ -784,17 +772,44 @@ def test_for_version_870(self, client: TestClient, user_access_token: str, study
...
-            res = client.get(
-                f"/v1/studies/{study_id}/raw",
-                params={
-                    "path": f"input/bindingconstraints/{bc_id_w_matrix}_{term_alias}",
-                    "depth": 1,
-                    "formatted": True,
-                },  # type: ignore
-            )
+        for operator in ["less", "equal", "greater", "both"]:
+            if operator != "both":
```
MartinBelthle, line 776: You should use the `BindingConstraintOperator` class and use .value to get the string. The same goes for the rest of the tests

tests/integration/study_data_blueprint/test_binding_constraints.py
```diff
@@ -1031,3 +1058,153 @@ def test_for_version_870(self, client: TestClient, user_access_token: str, study
...
+                bc_id = bc_name.lower()
+
+                required_matrices = REQUIRED_MATRICES[bc_operator]
+                for matrix in required_matrices:
+                    df = preparer.download_matrix(variant_id, f"input/bindingconstraints/{bc_id}_{matrix}")
+                    assert df.shape == (MATRIX_SIZES[bc_time_step], 1)
+                    assert np.allclose(df.values, matrices_by_time_steps[bc_time_step], atol=1e-6)
+
+                superfluous_matrices = {"lt", "gt", "eq"} - required_matrices
+                for matrix in superfluous_matrices:
```
MartinBelthle, line 1204: don't really understand this check. `superfluous_matrices` shouldn't be empty ?

MartinBelthle (changes requested):
Keep going you've already coded a big part of the feature

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -107,6 +108,25 @@ def delete(self, url: Optional[List[str]] = None) -> None:
...
+        if self.get_link_path().exists():
+            return self.get_link_path()
+        elif self.config.path.exists():
+            return self.config.path
+        else:
+            raise ChildNotFoundError(
+                f"Neither link file {self.get_link_path} nor matrix file {self.config.path} exists"
+            )
+
+    def infer_target_path(self, is_link: bool) -> Path:
```
sylvlecl: I don't hink the notion of "target" makes sense here, it only makes sense in the context of a move or copy operation.

Instead of adding such introspection methods here, and have a complicated logic related to links in `update_binding_constraint`, we could offer `copy` and `move` methods, which will take care internally of the handling of links.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -16,7 +25,189 @@
...
+    matrix_node_eq = parent_node.get_node([f"{binding_constraint_id}_eq"])
+    assert isinstance(
+        matrix_node_eq, LazyNode
+    ), f"Node type not handled yet: LazyNode expected, got {type(matrix_node_eq)}"
+    matrix_node_gt = parent_node.get_node([f"{binding_constraint_id}_gt"])
+    assert isinstance(
+        matrix_node_gt, LazyNode
+    ), f"Node type not handled yet: LazyNode expected, got {type(matrix_node_gt)}"
+
+    is_link_eq = matrix_node_eq.infer_is_link_path()
```
sylvlecl: In this file we should not have to manipulate paths directly, we should instead manipulate nodes of the tree.

For example to copy lt file to gt file, we would just call:
```python
matrix_node_lt.copy_to(matrix_node_gt) 
# or maybe just a path in the argument ?
```

This way all the link-related complexities will be handled by the nodes themselves which have the right information internally, and hidden from here.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -16,7 +25,189 @@
...
+        parent_folder_node, binding_constraint_id, existing_operator, new_operator
+    )
+
+    # TODO: due to legacy matrices generation, we need to check if the new matrix file already exists
+    #  and if it does, we need to first remove it before renaming the existing matrix file
+
+    if existing_operator != BindingConstraintOperator.BOTH and new_operator != BindingConstraintOperator.BOTH:
+        (matrix_path,) = matrix_paths.matrix_input_paths
+        (new_matrix_path,) = matrix_paths.matrix_target_paths
+        new_matrix_path.unlink(missing_ok=True)
```
sylvlecl: See comment above: here we should manipulate only nodes through their methods, not filesystem paths directly.

tests/integration/raw_studies_blueprint/test_download_matrices.py
```diff
@@ -101,6 +101,9 @@ def update_general_data(self, internal_study_id: str, **data: t.Any):
         res.raise_for_status()
 
 
+from tests.integration.prepare_proxy import PreparerProxy
```
sylvlecl, line 104: it's very confusing to import in the middle of the file a class that has the same name as the class defined in this file: is it on purpose ?
There must be a better way !

sylvlecl (changes requested):
I think we need to change slightly the implementation:
the level of abstraction in the command is not appropriate, it should manipulate nodes and not paths. Manipulation of paths should be hidden inside the nodes as much as possible.

We need to add copy and move operations to the nodes in order to implement this.

antarest/study/business/binding_constraint_management.py
mabw-rte: Yeas you are right. If a developer add more options to the `Enum` the algorithm should spit out an Error.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: you need to assert the type for the function that comes just after `_infer_input_and_target_paths` 
otherwise `mypy` would not not be ok with it.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: as both `matrix_input_paths` and `matrix_target_paths` can be both of `length` 2, I think it important that the algorithm spit out an error when they are supposed to have `length` of 1.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: I prefer to do the `unlinking` only when needed. It is the case when you have an operator update.
It is not because you want it but because you need (you cannot have two files with the same name in the same folder).

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: In previous discussions, we agreed to give total freedom to the developer to add new options to an `Enum`. 
This pattern enhances that in such case, implementation should be done.

mabw-rte: Anyway, I agree that the crash will happen in `_infer_input_and_target_paths`. We should see how to figure out a solution.

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
mabw-rte: Yeah, the design pattern at this point was poorly handled.

tests/integration/raw_studies_blueprint/test_download_matrices.py
mabw-rte: I agree, but I think that Laurent, wanted to avoid repetition in code. I will see.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: Actualy when switch from `BOTH` to `EQUAL`, we keep the `LESS` matrix rename to `EQUAL` and remove `GREATER` matrix. Some one might argue shouldn't we compute the mean of both matrices and output it as the `EQUAL` matrix? just to give an example.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: It might be stupid but as you have many ifs in functions, this increase the complexity. So I had to handle the equality out of the function. The downside, you should enforce such case, just because you never know what a developer could do.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: `mypy` issues

tests/integration/study_data_blueprint/test_binding_constraints.py
mabw-rte: `superfluous_matrices` are the matrices that should no longer exit following the new PR deployment. 
So if you don't have an Error 404 the `else` will execute raising an error.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: I found an alternative

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: I removed `_infer_input_and_target_paths`

antarest/study/business/binding_constraint_management.py
MartinBelthle: You didn't resolve this comment

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -121,14 +142,33 @@ def save(self, data: Union[str, bytes, S], url: Optional[List[str]] = None) -> N
...
+                shutil.copy2(self.infer_path(), target)
+            else:
+                self.infer_path().rename(target)
+        else:
+            target_path = target.infer_target_path(self.infer_is_link_path())
+            if not force and target_path.exists():
+                raise FileExistsError(f"Target path {target_path} already exists")
+            target_path.unlink(missing_ok=True)
+            if copy:
+                shutil.copy2(self.infer_path(), target_path)
```
MartinBelthle: Don't think we need any file's metadata so we should simply use `copy`

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
MartinBelthle: Okay I see but this is not a TODO, just a comment. If you want to keep track of this you can keep the comment without the TODO

tests/integration/study_data_blueprint/test_binding_constraints.py
MartinBelthle: Okay I see

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -83,35 +85,54 @@ def _get(
...
+                f"Neither link file {self.get_link_path} nor matrix file {self.config.path} exists"
+            )
+
+    def infer_target_path(self, is_link: bool) -> Path:
+        if is_link:
+            return self.get_link_path()
+        else:
+            return self.config.path
+
+    def infer_is_link_path(self) -> bool:
```
MartinBelthle: remove this method, it's only used once and it's just one line. Also I think you should remove the `infer_targer_path` method and adapt your code inside `move_file_to` line 160

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -121,14 +142,33 @@ def save(self, data: Union[str, bytes, S], url: Optional[List[str]] = None) -> N
                 self.config.path.unlink()
             return None
 
-        self.dump(cast(S, data), url)
+        self.dump(t.cast(S, data), url)
         if self.get_link_path().exists():
             self.get_link_path().unlink()
         return None
 
+    def move_file_to(
```
MartinBelthle: I don't really like the fact that this method does the move and can also copy files. I think you should separate things between 2 methods, one for copy and another for moving.

tests/integration/study_data_blueprint/test_binding_constraints.py
```diff
@@ -1,12 +1,23 @@
...
+from requests.exceptions import HTTPError
 from starlette.testclient import TestClient
 
 from antarest.study.business.binding_constraint_management import ClusterTerm, ConstraintTerm, LinkTerm
+from tests.integration.prepare_proxy import PreparerProxy
+
+MATRIX_SIZES = {"hourly": 8784, "daily": 366, "weekly": 366}
+
+
+REQUIRED_MATRICES = {
```
MartinBelthle, line 15: Same remark use BIndingConstraintOperator class

antarest/study/business/binding_constraint_management.py
```diff
@@ -912,13 +930,19 @@ def _replace_matrices_according_to_frequency_and_version(
...
     return args
 
 
-def check_attributes_coherence(data: t.Union[ConstraintCreation, ConstraintInput], study_version: int) -> None:
+def check_attributes_coherence(
+    data: t.Union[ConstraintCreation, ConstraintInput],
+    study_version: int,
+    existing_operator: t.Optional[BindingConstraintOperator] = None,
+) -> None:
+    update_operator = data.operator or existing_operator
```
MartinBelthle: The way you've handled it is disturbing IMO. You should not need to  add this line. When calling this method for creation, give data.operator or the default operator we use (tI think it's less). When calling for an update, use data.operator or existing_constraint.operator. So you'll always have an operator here, it won't be optional and it will simplify your code

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -1,22 +1,108 @@
...
-MatrixType = List[List[MatrixData]]
+MatrixType = t.List[t.List[MatrixData]]
+
+ALIAS_OPERATOR_MAP = {
+    BindingConstraintOperator.EQUAL: "eq",
+    BindingConstraintOperator.LESS: "lt",
+    BindingConstraintOperator.GREATER: "gt",
+}
+
+HANDLED_OPERATORS = [
```
MartinBelthle: This doesn't make sense, you should remove this

antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
```diff
@@ -35,6 +35,12 @@
 }
 
 
+class TermMatrices(str, Enum):
```
MartinBelthle: This is good but I have a lot of warnings in my IDE when calling `TermMatrices.value` because it understands TermMatrices as a string :/ Perhaps try to make this class inherit from Enum only to see if the code still works as it will shush the warnings

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -1,22 +1,108 @@
...
+
+    # TODO: due to legacy matrices generation, we need to check if the new matrix file already exists
+    #  and if it does, we need to first remove it before renaming the existing matrix file
+
+    if (existing_operator not in HANDLED_OPERATORS) or (new_operator not in HANDLED_OPERATORS):
+        raise NotImplementedError(
+            f"Case not handled yet: existing_operator={existing_operator}, new_operator={new_operator}"
+        )
+    elif existing_operator == new_operator:
+        return  # nothing to do
```
MartinBelthle, line 74: Shouldn't we remove matrices that are not concerned by the operator in such a case due to legacy matrices generation ?

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -1,22 +1,108 @@
...
+
+    parent_folder_node = file_study.tree.get_node(["input", "bindingconstraints"])
+    matrix_lt = parent_folder_node.get_node([f"{binding_constraint_id}_lt"])
+    assert isinstance(matrix_lt, LazyNode), f"Node type not handled yet: LazyNode expected, got {type(matrix_lt)}"
+    matrix_eq = parent_folder_node.get_node([f"{binding_constraint_id}_eq"])
+    assert isinstance(matrix_eq, LazyNode), f"Node type not handled yet: LazyNode expected, got {type(matrix_eq)}"
+    matrix_gt = parent_folder_node.get_node([f"{binding_constraint_id}_gt"])
+    assert isinstance(matrix_gt, LazyNode), f"Node type not handled yet: LazyNode expected, got {type(matrix_gt)}"
+
+    # TODO: due to legacy matrices generation, we need to check if the new matrix file already exists
```
MartinBelthle: With this PR this won't be a TODO just a comment to explain the code so you can remove the TODO at the beginning of the line

MartinBelthle (changes requested):

antarest/study/business/binding_constraint_management.py
mabw-rte: I didn't understand what you mean.
Can you be more specific? 

This function is called both on creation and update.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: In previous discussion, I explained that I prefer to delete only when you need to not because you want to.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: Yeah I moved it in side of the corresponding function

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
mabw-rte: ok

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
mabw-rte: `_update_matrices_names`

MartinBelthle: @sylvlecl we have different opinion on the subject with Mohamed so we would like to have yours :)

IMO we should remove this list and the associated check as it's not possible to have other operators since it's an enum.

On the other hand, Mohamed says it will be useful if someone wants to add another value to the Enum as it would crash (with my solution it won't crash here, it will enter the else and do something it shouldn't) and force the developer to adapt this code.

antarest/study/business/binding_constraint_management.py
```diff
@@ -928,3 +951,10 @@ def check_attributes_coherence(data: t.Union[ConstraintCreation, ConstraintInput
             raise InvalidFieldForVersionError("You cannot fill a 'matrix_term' as these values refer to v8.7+ studies")
     elif data.values:
         raise InvalidFieldForVersionError("You cannot fill 'values' as it refers to the matrix before v8.7")
+    conflicting_matrices = [
+        getattr(data, matrix) for matrix in OPERATOR_CONFLICT_MAP[operator] if getattr(data, matrix)
+    ]
+    if conflicting_matrices:
+        raise InvalidFieldForVersionError(
+            f"You cannot fill matrices '{OPERATOR_CONFLICT_MAP[operator]}' while using the operator " f"'{operator}'"
```
MartinBelthle: You can merge the 2 f-strings it's easier to read and shush my IDE warning

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -121,14 +139,24 @@ def save(self, data: Union[str, bytes, S], url: Optional[List[str]] = None) -> N
                 self.config.path.unlink()
             return None
 
-        self.dump(cast(S, data), url)
+        self.dump(t.cast(S, data), url)
         if self.get_link_path().exists():
             self.get_link_path().unlink()
         return None
 
+    def rename_filename(self, target: t.Union[Path, "LazyNode[t.Any, t.Any, t.Any]"]) -> None:
```
MartinBelthle: I don't really like the repetition of `name`, I would prefer `rename_file`

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -148,7 +176,7 @@ def load(
         Args:
             url: data path to retrieve
             depth: after url is reached, node expand tree until matches depth asked
-            expanded: context parameter to determine if current node become from a expansion
+            expanded: context parameter to determine if current node become from an expansion
```
MartinBelthle: Should be `comes` instead of `become` no ?

antarest/study/business/binding_constraint_management.py
MartinBelthle: You still didn't resolve this comment

MartinBelthle (changes requested):
Last round of comment for me

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
mabw-rte: You should ask the one who wrote it ... I guess you right, It should be even `comes` with an `s` 

I'mma correct this

antarest/study/business/binding_constraint_management.py
mabw-rte: Why ? 

I think we should discuss this for tomorrow ...

for versions < 870, you should skip this condition as you having  only one file `{bc_id}` template.
for versions >= 870 you should avoid unnecessary matrices 

That's why you need to check for version.


The casting was made in line `341`, I guess I thought it should be the same. 

But on removing the default `[]` I agree it should be removed, this way add more options the `Enum` will force the code to crash.

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
mabw-rte: yes

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -121,14 +139,24 @@ def save(self, data: Union[str, bytes, S], url: Optional[List[str]] = None) -> N
                 self.config.path.unlink()
             return None
 
-        self.dump(cast(S, data), url)
+        self.dump(t.cast(S, data), url)
         if self.get_link_path().exists():
             self.get_link_path().unlink()
         return None
 
+    def rename_file(self, target: t.Union[Path, "LazyNode[t.Any, t.Any, t.Any]"]) -> None:
```
sylvlecl: We should add unit tests in `test_lazy_node.py` to check the behaviour of the new methods in both cases (link or actual file).

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -121,14 +139,24 @@ def save(self, data: Union[str, bytes, S], url: Optional[List[str]] = None) -> N
...
+        target_path = target if isinstance(target, Path) else target._infer_target_path(self.get_link_path().exists())
+        target_path.unlink(missing_ok=True)
+        self._infer_path().rename(target_path)
```
sylvlecl, line 150: I don't think this will work correctly in all cases:
if self is a link, but target is an actual file, we will copy the link to the wrong place

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -121,14 +139,24 @@ def save(self, data: Union[str, bytes, S], url: Optional[List[str]] = None) -> N
...
+
+    def copy_file(self, target: t.Union[Path, "LazyNode[t.Any, t.Any, t.Any]"]) -> None:
+        target_path = target if isinstance(target, Path) else target._infer_target_path(self.get_link_path().exists())
+        target_path.unlink(missing_ok=True)
+        shutil.copy(self._infer_path(), target_path)
```
sylvlecl, line 155: Same comment as above, I think if one is a link and the other is a file, this will not work as expected.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -1,22 +1,108 @@
...
+MatrixType = t.List[t.List[MatrixData]]
+
+ALIAS_OPERATOR_MAP = {
+    BindingConstraintOperator.EQUAL: "eq",
+    BindingConstraintOperator.LESS: "lt",
+    BindingConstraintOperator.GREATER: "gt",
+}
+
+
+def _update_matrices_names(
```
sylvlecl, line 32: We should add unit tests in `test_manage_binding_constraints.py`

sylvlecl (commented):
Implementation is better but I think we should add unit tests, not only integration tests!
In particular there might be an issue for the new copy and rename methdos

tests/integration/study_data_blueprint/test_binding_constraints.py
```diff
@@ -784,17 +772,44 @@ def test_for_version_870(self, client: TestClient, user_access_token: str, study
...
+            else:
+                res = client.put(
+                    f"/v1/studies/{study_id}/bindingconstraints/{bc_id_w_matrix}",
+                    json={
+                        "operator": operator,
+                        "greater_term_matrix": expected_matrix.tolist(),
+                        "less_term_matrix": expected_matrix.tolist(),
+                    },
+                )
             assert res.status_code == 200, res.json()
```
sylvlecl, line 790: All combinations of ("old operator", "new operator") are not tested, it seems we need to add some.
Also, it would be better to check that the content is correctly copied with something else than zeros, which could be some kind of default values.

sylvlecl: note: those combinations could be tested in unit tests, not in integration tests, instead

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
mabw-rte: If you decide top pass a target as a `Path` and not as `LazyNode` ,  the copy or the renaming will be done using that path. 
So I'm going to remove it.

antarest/study/business/binding_constraint_management.py
MartinBelthle: You don't need to cast as an int as it's already one.
Also, as I said in my comment you should check study version before looping on constraints as you'll do a lot of useless checks

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
MartinBelthle: @mabw-rte I've seen with Sylvain and he agrees with you so I resolve this comment.

antarest/study/business/binding_constraint_management.py
mabw-rte: Why a lot useless checks, technically this should be even faster than what existed ...

sylvlecl (dismissed):

PR #2079, opened 2024-06-28
test: rename 'study_id' fixture to 'internal_study_id' to avoid confusion with `study_id` variable

laurent-laporte-pro:
## Description

In the context of API endpoint integration tests, we have a pytest fixture that allows us to locate the "STA-mini" study among the internalized studies (currently, there is only one). As a reminder, this study is automatically detected by the watcher when launching the FastAPI test application. Therefore, it is always available. The fixture searches for the internalized study in the database and returns its ID.

The aim of this PR is to rename the "study_id" fixture to "internal_study_id" in order to avoid any confusion between an arbitrary study identifier and the "STA-mini" test study. This confusion occurs in tests that do not use the internalized study, as it is common to name a study identifier `study_id`. There is also confusion caused by a PyCharm plugin when detecting fixtures not present in the parameters of test functions. Currently, this plugin generates false positives because it considers the `study_id` variable as a fixture, whereas in these tests that do not use the internalized study, it is a regular variable (for example, the ID of a newly created study for the test).

## How to check this PR?

To verify the validity of this PR, the reviewer should:

1. **Review the Code**: Ensure that the code has been correctly modified to reflect the fixture name change from 'study_id' to 'internal_study_id'. This includes all places where the fixture is used in the tests.

2. **Run the Tests**: Execute all unit and integration tests to ensure they still pass with the new fixture name. This includes tests that use the internalized study and those that do not.

3. **Check for Consistency**: Ensure that the new fixture name 'internal_study_id' is consistent with its usage and function in the code.

4. **Review False Positives**: Check if the PyCharm plugin no longer generates false positives by considering `study_id` as a fixture in tests that do not use the internalized study.

5. **Read the Documentation**: Ensure that any documentation or comments related to this fixture have been updated to reflect the name change.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -120,7 +120,7 @@ def user_access_token_fixture(
     return t.cast(str, credentials["access_token"])
 
 
-@pytest.fixture(name="internal_study")
+@pytest.fixture(name="internal_study_id")
 def internal_study_fixture(
     client: TestClient,
     user_access_token: str,
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -120,12 +120,12 @@ def user_access_token_fixture(
     return t.cast(str, credentials["access_token"])
 
 
-@pytest.fixture(name="study_id")
-def study_id_fixture(
+@pytest.fixture(name="internal_study")
+def internal_study_fixture(
     client: TestClient,
     user_access_token: str,
 ) -> str:
-    """Get the ID of the study stored in database"""
+    """Get the ID of the internal study which is scanned by the watcher"""
     res = client.get(
         "/v1/studies",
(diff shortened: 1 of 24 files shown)


mabw-rte (dismissed):
Sounds good for me ... Just should discuss the naming `internal_study`, I think that you need to put `id` (even if it gonna make it long) otherwise we may get confused that the variable is an instance of the object `Study`

laurent-laporte-pro:
> Sounds good for me ... Just should discuss the naming `internal_study`, I think that you need to put `id` (even if it gonna make it long) otherwise we may get confused that the variable is an instance of the object `Study`

Agree: I'll change to `internal_study_id` which is more explicit.

mabw-rte (approved):

PR #2081, opened 2024-07-04
docs(user-guide): improve the user guide and add "How to Create a New Study?" topic

laurent-laporte-pro:
This PR adds several enhancements to the documentation, focusing on the creation, upgrade, and import processes of studies in Antares Web. Below are the key changes included in this PR:

## Added
- A new topic "How to Create a New Study?" in `docs/how-to/studies-create.md`, providing a detailed guide on creating a new study in Antares Web.
- Reference to the "How to Upgrade a Study?" section in `docs/how-to/studies-upgrade.md`, ensuring users are aware that upgrades are only applicable for raw studies without variants.
- A section on importing a study using the API endpoint in `docs/how-to/studies-import.md`, demonstrating the process with a Python script.

## Updated
- Enhanced the tags in the header of `docs/how-to/studies-create.md` to include more relevant topics such as `tutorial`, `guide`, `create`, `version`, `new`, `study management`, `workspace`, `energy simulation`, `Antares`, and `study management`.
- General improvements to the User Guide documentation to make it more comprehensive and user-friendly.

These updates aim to improve the accessibility and usability of the documentation, making it easier for users to navigate and perform essential tasks within Antares Web.


hdinia (changes requested):
Adding images to the repository for docs could become problematic with time, instead consider using a dedicated storage like AWS S3, Google Cloud Storage, or a CDN, and link them into the docs. Why ?

- Performance degradation:
Pulling and fetching updates becomes slower as the repository grows.
Checkout times increase when branches contain different versions of large files.

- Version control inefficiency with binary files:
Git stores the entire file for each change, causing rapid repository growth.

- Repository bloat:
 Git retains all versions of files, leading to irreversible growth even if images are later removed.

- CI/CD Pipeline efficiency:
Larger repositories delay CI/CD processes due to longer fetch times, affecting build times and the delivery of automated tests and deployments.

you can also check out git LFS that is a popular solution to this problem:
[Git Large File Storage](https://git-lfs.com/)

hdinia:
Also normally docs websites are separated from the source code of the app due to those issues on managing many images and assets like MD files, videos, gifs etc.

I think we should split our docs into a separate repo and use [Docusaurus](https://docusaurus.io/fr/) (free and open source) to manage it as a static site. This way, we can keep things tidy and it'll be easier to update. 
Plus, the performance boost is a nice bonus. 
@laurent-laporte-pro  @makdeuneuv

MartinBelthle:
Even though you're right @hdinia, the solutions you've proposed are build to deal with GB of files (cf. the first lines of the link you shared). I doubt that our documentation will ever be that huge. It's an issue to keep in mind but I don't think we need to tackle it now.

makdeuneuv:
@hdinia , @MartinBelthle I think we can accept the modifications as they are, knowing that it is very likely that we will migrate our documentation to xwiki (to be confirmed of course)

MartinBelthle (approved):

PR #2083, opened 2024-07-05
feat(raw): add arrow support in backend

MartinBelthle:
There are few things to note when using the `arrow` format:
1- For output files the index is now inside the dataframe as the column "Index".
2- The columns names are now in string (forced by pyarrow)


antarest/study/web/raw_studies_blueprint.py
```diff
@@ -59,6 +60,13 @@
     ".json": ("application/json", "utf-8"),
 }
 
+
+class MATRIX_FORMAT(EnumIgnoreCase):
+    JSON = "json"
```
laurent-laporte-pro, line 65: Selon les cas d'usage dans l'application, vous pouvez avoir "json-split" (cas général) ou "json-index" (cas des Table Form du scenario builder, par exemple).

Exemple :
```python
>>> import pandas as pd
>>> df = pd.DataFrame(data=[[1, 2, 3.14], [4, 5, 6.18]], index=["00:00", "01:00"], columns=["TS-1", "TS-2", "TS-3"])
>>> df.to_json(orient="split")
'{"columns":["TS-1","TS-2","TS-3"],"index":["00:00","01:00"],"data":[[1,2,3.14],[4,5,6.18]]}'
>>> df.to_json(orient="index")
'{"00:00":{"TS-1":1,"TS-2":2,"TS-3":3.14},"01:00":{"TS-1":4,"TS-2":5,"TS-3":6.18}}'
```

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -59,6 +60,13 @@
...
+    BYTES = "bytes"
```
laurent-laporte-pro: ou "TSV"

laurent-laporte-pro (approved):

antarest/study/storage/abstract_storage_service.py
```diff
@@ -116,21 +116,14 @@ def get_study_information(
...
-    ) -> JSON:
+    def get(self, metadata: T, url: str = "", depth: int = 3, format: str = "json", use_cache: bool = True) -> JSON:
         """
         Entry point to fetch data inside study.
         Args:
             metadata: study
             url: path data inside study to reach
             depth: tree depth to reach after reach data path
-            formatted: indicate if raw files must be parsed and formatted
+            format: indicate if raw files must be parsed and formatted
```
sylvlecl: we should update this sentence

antarest/study/storage/rawstudy/model/filesystem/matrix/matrix.py
```diff
@@ -89,15 +90,11 @@ def denormalize(self) -> None:
...
-        depth: int = -1,
-        expanded: bool = False,
-        formatted: bool = True,
-    ) -> Union[bytes, JSON]:
+    def load(  # type: ignore
+        self, url: Optional[List[str]] = None, depth: int = -1, expanded: bool = False, format: str = "json"
+    ) -> Union[bytes, JSON, pd.DataFrame]:
         file_path, tmp_dir = self._get_real_file_path()
-        if not formatted:
+        if format == "bytes":
```
sylvlecl: I think it's more clear to make format optional and return the raw byte when it's not defined.
"bytes" is not a format, bytes could contain anything (some TSV, some JSON, etc).

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
```diff
@@ -73,14 +75,21 @@ def parse(
...
-            data = cast(JSON, matrix.to_dict(orient="split"))
-            stopwatch.log_elapsed(lambda x: logger.info(f"Matrix to dict in {x}s"))
+            if format == "json":
+                matrix_json = cast(JSON, matrix.to_dict(orient="split"))
+                stopwatch.log_elapsed(lambda x: logger.info(f"Matrix to dict in {x}s"))
+                return matrix_json
+
+            with io.BytesIO() as buffer:
+                matrix.columns = matrix.columns.map(str)
+                matrix.to_feather(buffer, compression="uncompressed")
```
sylvlecl, line 89: Don't we need the index here too?

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -117,12 +127,23 @@ def get_study(
...
+        if format:
+            real_format = format.value
+        else:
+            real_format = "json" if formatted else "bytes"
+
+        output = study_service.get(uuid, path, depth=depth, format=real_format, params=parameters)
 
         if isinstance(output, bytes):
+            if real_format == MatrixFormat.ARROW:
+                return Response(content=output, media_type="application/octet-stream")
```
sylvlecl: arrow project recommends to use the following mime type "application/vnd.apache.arrow.file"

see https://arrow.apache.org/faq/

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -117,12 +127,23 @@ def get_study(
...
         )
         parameters = RequestParameters(user=current_user)
-        output = study_service.get(uuid, path, depth=depth, formatted=formatted, params=parameters)
+
+        if format:
+            real_format = format.value
+        else:
+            real_format = "json" if formatted else "bytes"
+
+        output = study_service.get(uuid, path, depth=depth, format=real_format, params=parameters)
```
sylvlecl: From a global point of view, I think it's weird to push the formatting down all the call stack to serialize the matrix there.

Usually, application manipulate "objects" in memory, and then we serialize it according to the requested format only when we need it.

So here for example, I think it would make more sense to retrieve a `DataFrame` object, and serialize it here.

Do you think it's possible ? Maybe in another PR because it has a bigger impact on the codebase ?
In particular, we still want matrices to be formatted as JSON when getting a whole tree from the file tree.

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
```diff
@@ -174,33 +155,63 @@ def test_get_study(
 
         # If we ask for a matrix, we should have a JSON content if formatted is True
         rel_path = "/input/links/de/fr"
```
sylvlecl: We should probably test it with an output matrix too, since they are handled by different code paths

antarest/study/web/raw_studies_blueprint.py
sylvlecl: Agreed, "bytes" is not a format, it could be anything inside.
However, output matrices will not return TSV without further handling, so maybe we should just drop this value.

If format is not defined, then we return the raw unformatted bytes.
An alternative is to keep both `formatted` and `format` arguments in the API to make it clearer. If `formatted` is False we return the raw bytes, else we look at format value to choose what to return (json or arrow)

sylvlecl (changes requested):

antarest/study/web/raw_studies_blueprint.py
MartinBelthle: I prefer the alternative. I'll implement that

antarest/study/web/raw_studies_blueprint.py
MartinBelthle: I think it's possible and I agree it's a better way to do it. I tried inside this PR but it's too big of a work so we should tackle it inside another PR.

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
MartinBelthle: For now, we don't return index inside input matrices so we shouldn't do it here.
Once we will, we'll have to change this code indeed.

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
MartinBelthle: I've duplicated the input tests to do the output matrix ones. The code of the test is a bit more complicated but it doesn't bother me.

antarest/study/web/raw_studies_blueprint.py
laurent-laporte-pro: Remember that, some binary files in the study directory are not DataFrame at all : `*.ico`, XML files, user-defined files stored in the `user` directory…

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
```diff
@@ -73,14 +75,21 @@ def parse(
                     raise ChildNotFoundError(f"File '{relpath}' not found in the study '{study_id}'") from e
 
             stopwatch.log_elapsed(lambda x: logger.info(f"Matrix parsed in {x}s"))
-            matrix.dropna(how="any", axis=1, inplace=True)
+            matrix = matrix.dropna(how="any", axis=1)
```
TheoPascoli, line 78: Why you decided to do this. I think the original way to remove the na is more efficient

antarest/study/business/binding_constraint_management.py
```diff
@@ -351,7 +351,7 @@ def _get_references_by_widths(
         for matrix_name in matrices_name:
             matrix_id = matrix_name.format(bc_id=bc.id)
             logger.info(f"⏲ Validating BC '{bc.id}': {matrix_id=} [{_index+1}/{_total}]")
-            obj = file_study.tree.get(url=["input", "bindingconstraints", matrix_id])
+            obj = file_study.tree.get(url=["input", "bindingconstraints", matrix_id], format="json")
```
TheoPascoli, line 354: Do you think it is worth it to set the format as an enum everywhere ?

MartinBelthle:
This PR was too old so I created a new one: https://github.com/AntaresSimulatorTeam/AntaREST/pull/2624

PR #2085, opened 2024-07-09
build: Update tables and PyYaml versions for python > 3.8

sylvlecl:
**Description**
When working in python > 3.8 environment, a few dependencies are not available.

This PR allows to work with python >= 3.9 by having a conditional versioning for:
- tables
- pyyaml

For python 3.8, versions are unchanged.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,10 +22,8 @@ pydantic~=1.9.0
 PyQt5~=5.15.6
 python-json-logger~=2.0.7
 python-multipart~=0.0.5
-# See https://github.com/yaml/pyyaml/issues/724#issuecomment-1638801106
-PyYAML>=5.3.1; python_version <= '3.5'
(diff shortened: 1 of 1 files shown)


TheoPascoli (approved):

MartinBelthle (commented):
I didn't know we could do such a thing inside requirements, I thought we add to create specific files for each version, nice :smile: 

Shouldn't we also update the github workflow `main.yaml` to run it with different python versions ?

sylvlecl:
> I didn't know we could do such a thing inside requirements, I thought we add to create specific files for each version, nice 😄
> 
> Shouldn't we also update the github workflow `main.yaml` to run it with different python versions ?

Indeed it would be intesting to run our tests with other python versions.
However the full matrix of [OS*python versions] will take too long to execute for each push.

What we could do is:
- keeping only one python version for PR workflows
- adding other python versions for merges on develop, or for a nightly workflow

MartinBelthle (approved):

PR #2088, opened 2024-07-10
feat(output): enhance output synthesis view

MartinBelthle:
Will fix [ANT-1863]


antarest/study/storage/rawstudy/model/filesystem/root/output/simulation/mode/mcall/grid.py
```diff
@@ -1,12 +1,88 @@
...
+
+    def load(
+        self,
+        url: t.Optional[t.List[str]] = None,
+        depth: int = -1,
+        expanded: bool = False,
+        formatted: bool = True,
+    ) -> JSON:
+        file_path = self.config.path
+        with open(file_path, "r") as f:
```
laurent-laporte-pro, line 80: The solution you propose is very good, this is exactly what I wanted to propose you.
For more maintainability, it is better to write a small function next to the main function, which will allow to test this function separately, and will make it more readable and evolutive.
Finally, it is always good to add a docstring to give some business context for developers who do not know the `digest.txt` file.

Here is an improved version of your code:

```python
def scan_digest_file(digest_file: typing.TextIO) -> pd.DataFrame:
    """
    Parse a digest file as a whole and return a single DataFrame.

    The `digest.txt` file is a TSV file containing synthetic results of the simulation.
    This file contains several data tables, each being separated by empty lines
    and preceded by a header describing the nature and dimensions of the table.

    Note that rows in the file may have different number of columns.
    """
    # Read the file and find the maximum number of columns in any row
    data = [line.rstrip("\n").split("\t") for line in digest_file]
    max_cols = max(len(row) for row in data)
    # Adjust the number of columns in each row
    data = [row + [""] * (max_cols - len(row)) for row in data]
    # Create a DataFrame from the data (do not convert values to float)
    df = pd.DataFrame(data, dtype=object)
    return df

...

    with file_path.open(mode="r") as digest_file:
        df = scan_digest_file(digest_file)
```

laurent-laporte-pro (approved):

antarest/study/storage/rawstudy/model/filesystem/root/output/simulation/mode/mcall/grid.py
MartinBelthle: I agree so I modified the code as you suggested. I've added an integration test but no UT.

MartinBelthle:
I've encountered a weird case in the front-end, we're going to investigate with Hatim so the PR is not ready for merge yet.

antarest/study/storage/rawstudy/model/filesystem/matrix/output_series_matrix.py
```diff
@@ -160,11 +138,7 @@ def load(
             ) from e
 
     def dump(self, data: Union[bytes, JSON], url: Optional[List[str]] = None) -> None:
-        if isinstance(data, bytes):
```
sylvlecl, line 163: Note that this change is not related to digest, would have been slightly clearer to do it in another feature to have a clear history of changes (but let's keep it, not a big deal).

hdinia (approved):

sylvlecl (approved):

antarest/study/storage/rawstudy/model/filesystem/matrix/output_series_matrix.py
MartinBelthle: Indeed.

PR #2089, opened 2024-07-10
fix(ui-commons): prompt form Form displayed on dialog validation

skamril:
ANT-1858

### DIFF
--- a/webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
+++ b/webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
@@ -58,7 +58,7 @@ function CreateVariantDialog(props: Props) {
 
   return (
     <FormDialog
-      maxWidth="sm"
+      maxWidth="sm" // Study name source can be long
       title={t("studies.createNewStudy")}
       titleIcon={AddCircleIcon}
       open={open}
@@ -68,7 +68,7 @@ function CreateVariantDialog(props: Props) {
       config={{ defaultValues }}
     >
       {({ control }) => (
-        <Fieldset fieldWidth={550}>
+        <Fieldset fullFieldWidth>
           <StringFE
             label={t("variants.newVariant")}
             name="name"
@@ -80,14 +80,14 @@ function CreateVariantDialog(props: Props) {
           />
           <SelectFE
             label={t("study.versionSource")}
+            name="sourceId"
             variant="outlined"
             options={sourceList.map((ver) => ({
               label: ver.name,
               value: ver.id,
             }))}
-            name="sourceId"
             control={control}
-            required
+            rules={{ required: true }}
           />
         </Fieldset>
       )}
--- a/webapp/src/components/common/Form/index.tsx
+++ b/webapp/src/components/common/Form/index.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { FormEvent, useEffect, useMemo, useRef } from "react";
+import { FormEvent, useEffect, useMemo, useRef, useState } from "react";
 import {
   DeepPartial,
   FieldPath,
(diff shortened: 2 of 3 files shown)


webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
```diff
@@ -58,7 +58,7 @@ function CreateVariantDialog(props: Props) {
 
   return (
     <FormDialog
-      maxWidth="sm"
+      maxWidth="xs"
```
hdinia: please revert that change, studies names can be quite long so `xs` width is too small in most cases. And their is no width constraints when opening the dialog

webapp/src/components/common/Form/index.tsx
```diff
@@ -116,6 +116,7 @@ function Form<TFieldValues extends FieldValues, TContext>(
   const { t } = useTranslation();
   const autoSubmitConfig = toAutoSubmitConfig(autoSubmit);
 
+  const [isInProgress, setIsInProgress] = useState(false);
```
hdinia, line 119: consider this instead:
 ```javascript
  const [isSubmitting, setIsSubmitting] = useState(false);
```

File: webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx (hunk 1 above)
Comment (skamril) [reply webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx]: ok

File: webapp/src/components/common/Form/index.tsx (hunk 2 above)
Comment (skamril) [reply webapp/src/components/common/Form/index.tsx:119]: `isSubmitting` already exist...

File: webapp/src/components/common/Form/index.tsx [hunk 3]
```diff
@@ -224,7 +224,7 @@ function Form<TFieldValues extends FieldValues, TContext>(
   // Prevent browser close if a submit is pending
   useEffect(() => {
     const listener = (event: BeforeUnloadEvent) => {
-      if (preventClose.current) {
+      if (isInProgress) {
```
hdinia, line 227: `isSubmitting`

webapp/src/components/common/Form/index.tsx
```diff
@@ -238,14 +238,14 @@ function Form<TFieldValues extends FieldValues, TContext>(
...
+  }, [t, isInProgress, isDirty]);
 
   useUpdateEffect(() => onStateChange?.(formState), [formState]);
 
   useEffect(() => setRef(apiRef, formApiPlus));
 
-  usePrompt(t("form.submit.inProgress"), preventClose.current);
-  usePrompt(t("form.changeNotSaved"), isDirty);
+  usePrompt(t("form.submit.inProgress"), isInProgress);
+  usePrompt(t("form.changeNotSaved"), isDirty && !isInProgress);
```
hdinia, line 248: optional: maybe you can wrap those in a useEffect, use strict conditions, and add a cleanup for the prompt
(this can potentially lead to overlapping or competing prompts)

```javascript
  useEffect(() => {
    if (isSubmitting) {
      usePrompt(t("form.submit.inProgress"), true);
    } else if (isDirty) {
      usePrompt(t("form.changeNotSaved"), true);
    } else {
      usePrompt(null, false); // Clear the prompt when no conditions are met
    }

    // Cleanup function to clear the prompt when the component unmounts
    return () => usePrompt(null, false);
  }, [isSubmitting, isDirty, t]);
```

webapp/src/components/common/Form/index.tsx
hdinia: In the same component? Ok mybad didnt see that

hdinia: maybe you can just add the 'submit' to indicate it refers to a submitting state like `isSubmitInProgress` (it's a detail tho feel free to skip)

hdinia (approved):

PR #2092, opened 2024-07-15
feat(aggregation-api)!: add new endpoint for `economy/mc-all` aggregation

mabw-rte:
Context: 
for the aggregation `apis`, we want to add additional endpoints that perform aggregation on the folder `economy/mc-all`


MartinBelthle (changes requested):
I don't see the part concerning the regex for column names (see JIRA issue).
You should add and test it.
Also, for `mc-all`, there are new type of files that can be requested (see JIRA issue) so you need to implement that.

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -201,7 +206,7 @@ def aggregate_areas_raw_data(
         - `frequency`: "hourly", "daily", "weekly", "monthly", "annual"
         - `mc_years`: which Monte Carlo years to be selected. If empty, all are selected (comma separated)
         - `areas_ids`: which areas to be selected. If empty, all are selected (comma separated)
-        - `columns_names`: which columns to be selected. If empty, all are selected (comma separated)
+        - `columns_names`: regexes (if details) or columns to be selected, if empty, all columns are selected
```
MartinBelthle: I don't find this explanation clear. Perhaps say `regexes (comma separated) if the query_file is of 'details' type ...`

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -311,6 +316,137 @@ def aggregate_links_raw_data(
...
+        export_format: TableExportFormat = DEFAULT_EXPORT_FORMAT,  # type: ignore
+        current_user: JWTUser = Depends(auth.get_current_user),
+    ) -> FileResponse:
+        """
+        Create an aggregation of links in mc-all
+
+        Parameters:
+        - `uuid`: study ID
+        - `output_id`: the output ID aka the simulation ID
+        - `query_file`: "values" (currently the only available option)
```
MartinBelthle: Not true as you can also have id.
Also for the mc-ind endpoint, this sentence implies it will evolve and it's not the case so you can just put: only option.

antarest/study/service.py
```diff
@@ -357,18 +363,44 @@ def aggregate_output_data(
...
+            ids_to_consider: list of areas or links ids to consider, if empty, all areas are selected
+            params: request parameters
+
+        Returns: the aggregated data as a DataFrame
+
+        """
+        study = self.get_study(uuid)
+        assert_permission(params.user, study, StudyPermissionType.READ)
+        study_path = self.storage_service.raw_study_service.get_study_path(study)
+        # fmt: off
```
MartinBelthle: Don't really see why you used `fmt: off` here it's not even in a singe line you can remove it. Same for line 366.

antarest/study/business/aggregator_management.py
```diff
@@ -54,28 +78,80 @@ def _checks_estimated_size(nb_files: int, df_bytes_size: int, nb_files_checked:
...
+        return list(range(1, len(df) + 1))
+
+
 class AggregatorManager:
     def __init__(
         self,
         study_path: Path,
         output_id: str,
-        query_file: t.Union[AreasQueryFile, LinksQueryFile],
+        query_file: t.Union[MCIndAreasQueryFile, MCAllAreasQueryFile, MCIndLinksQueryFile, MCAllLinksQueryFile],
```
MartinBelthle, line 128: You don't have to type args here, it's hard to read. This goes for the other following args.

antarest/study/business/aggregator_management.py
```diff
@@ -54,28 +78,80 @@ def _checks_estimated_size(nb_files: int, df_bytes_size: int, nb_files_checked:
...
+            sim_id=self.output_id, mc_root=MCRoot.MC_ALL.value
+        )
+        self.mc_root = (
+            MCRoot.MC_IND
+            if (isinstance(query_file, MCIndAreasQueryFile) or isinstance(query_file, MCIndLinksQueryFile))
+            else MCRoot.MC_ALL
+        )
 
-    def _parse_output_file(self, file_path: Path) -> pd.DataFrame:
+    def _parse_output_file(self, file_path: Path, normalize: bool = True) -> pd.DataFrame:
```
MartinBelthle: The keyword `normalize` is already used widely in the code for matrices (it's also a boolean) but in another context so can you please change it as it could be misleading

antarest/study/business/aggregator_management.py
```diff
@@ -196,6 +360,9 @@ def aggregate_output_data(self) -> pd.DataFrame:
         Aggregates the output data of a study and returns it as a DataFrame
         """
 
+        # check that the `mc_years` is Sequence[int]
+        assert self.mc_years is not None, "mc_years should be a `Sequence` of integers"
```
MartinBelthle: I think you should merge the 2 aggregation function as the code is almost similar. Also when you do so, you could remove this ugly check

antarest/study/business/aggregator_management.py
```diff
@@ -144,20 +227,97 @@ def _gather_all_files_to_consider(self) -> t.Sequence[Path]:
...
+            df = pd.DataFrame(new_obj).reindex(columns=columns_order).sort_values(by=[TIME_ID_COL, CLUSTER_ID_COL])
+            df.index = pd.Index(list(range(1, len(df) + 1)))
+
+            return df
+
+        else:
+            return self._parse_output_file(file_path)
+
     def _build_dataframe(self, files: t.Sequence[Path], horizon: int) -> pd.DataFrame:
+        assert self.mc_root in [MCRoot.MC_IND, MCRoot.MC_ALL], f"Unknown Monte Carlo root: {self.mc_root}"
```
MartinBelthle: Not sure assertion is the way to go, you should probably raise an Exception as assertions are removed when optimization are requested from Python

antarest/study/business/aggregator_management.py
```diff
@@ -54,28 +78,80 @@ def _checks_estimated_size(nb_files: int, df_bytes_size: int, nb_files_checked:
...
+    if is_details:
+        org_cols = [col for col in org_cols if col != CLUSTER_ID_COL and col != TIME_ID_COL]
+    if mc_root == MCRoot.MC_IND:
+        new_column_order = (
+            [column_name] + ([CLUSTER_ID_COL] if is_details else []) + [MCYEAR_COL, TIME_ID_COL, TIME_COL] + org_cols
+        )
+    elif mc_root == MCRoot.MC_ALL:
+        new_column_order = [column_name] + ([CLUSTER_ID_COL] if is_details else []) + [TIME_ID_COL, TIME_COL] + org_cols
+    else:
+        raise NotImplementedError(f"Unknown Monte Carlo root: {mc_root}")
```
MartinBelthle: You checked this earlier in the code with your assertion so you can remove it here

antarest/study/business/aggregator_management.py
```diff
@@ -54,28 +78,80 @@ def _checks_estimated_size(nb_files: int, df_bytes_size: int, nb_files_checked:
...
+            col for col in org_cols if col not in {column_name, CLUSTER_ID_COL, MCYEAR_COL, TIME_ID_COL, TIME_COL}
+        ]
+    if is_details:
+        org_cols = [col for col in org_cols if col != CLUSTER_ID_COL and col != TIME_ID_COL]
+    if mc_root == MCRoot.MC_IND:
+        new_column_order = (
+            [column_name] + ([CLUSTER_ID_COL] if is_details else []) + [MCYEAR_COL, TIME_ID_COL, TIME_COL] + org_cols
+        )
+    elif mc_root == MCRoot.MC_ALL:
+        new_column_order = [column_name] + ([CLUSTER_ID_COL] if is_details else []) + [TIME_ID_COL, TIME_COL] + org_cols
```
MartinBelthle, line 92: I think it would be easier to read if you'd construct the list with MCYEAR_COL inside and then if mc_root == MCRoot.MC_IND: new_column_order.remove(MCYEAR_COL)

MartinBelthle (changes requested):
I didn't read everything but overall the code is super complex and it's really hard to understand your logic so could you add some doc inside the code to make it more readable and maintainable.

antarest/study/business/aggregator_management.py
mabw-rte: you need to handle all types of aggregation so you cannot go around it

antarest/study/business/aggregator_management.py
mabw-rte: I will handle in an other way

antarest/study/business/aggregator_management.py
mabw-rte: I will do something different

antarest/study/business/aggregator_management.py
MartinBelthle: Sorry I put my comment on the wrong line I meant line 126 to 134 inside the initialization function you don't have to retype your args

antarest/study/business/aggregator_management.py
```diff
@@ -196,18 +359,33 @@ def aggregate_output_data(self) -> pd.DataFrame:
...
-        # Checks if mc-ind results exist
-        if not self.mc_ind_path.exists():
-            raise OutputNotFound(self.output_id)
+        if self.mc_root == MCRoot.MC_IND:
+            # check that the `mc_years` is Sequence[int]
+            assert self.mc_years is not None, "mc_years should be a `Sequence` of integers"
+
+            # Checks if mc-ind results exist
+            if not self.mc_ind_path.exists():
+                raise OutputNotFound(self.output_id)
```
MartinBelthle: It's not something you changed in this dev but I believe we should be more specific with these exceptions.
3 things can happen:
1- The `output` folder doesn't exist, we should raise an OutputNotFound exception
2- The `mc-ind` folder doesn't exist, we should specify that it's just this folder missing
3- The `mc-all` folder doesn't exist, same as 2-

antarest/study/business/aggregator_management.py
```diff
@@ -196,18 +359,33 @@ def aggregate_output_data(self) -> pd.DataFrame:
...
+        elif self.mc_root == MCRoot.MC_ALL:
+            # Checks if mc-all results exist
+            if not self.mc_all_path.exists():
+                raise OutputNotFound(self.output_id)
+
+            # filters files to consider
+            all_output_files = sorted(self._gather_all_files_to_consider__all())
+
+        else:
+            raise InvalidFieldForVersionError(f"Unknown Monte Carlo root: {self.mc_root}")
```
MartinBelthle: This exception has nothing to do here as you can see in its name.

antarest/study/business/aggregator_management.py
```diff
@@ -144,20 +225,98 @@ def _gather_all_files_to_consider(self) -> t.Sequence[Path]:
         ]
         return all_output_files
 
+    def _gather_all_files_to_consider__all(self) -> t.Sequence[Path]:
+        # Links / Areas ids filtering
+        areas_or_links_ids = self._filter_ids(self.mc_all_path / self.output_type)
+
+        # Frequency and query file filtering
+        folders_to_check = [self.mc_all_path / self.output_type / id for id in areas_or_links_ids]
+        filtered_files: t.Dict[str, t.MutableSequence[str]] = {}
```
MartinBelthle: This little iterdir part of the code is exactly the same inside the loop for mc_ind so you could probably extract it to avoid code duplication

antarest/study/business/aggregator_management.py
```diff
@@ -196,18 +359,33 @@ def aggregate_output_data(self) -> pd.DataFrame:
         Aggregates the output data of a study and returns it as a DataFrame
         """
 
-        # Checks if mc-ind results exist
-        if not self.mc_ind_path.exists():
-            raise OutputNotFound(self.output_id)
+        if self.mc_root == MCRoot.MC_IND:
+            # check that the `mc_years` is Sequence[int]
+            assert self.mc_years is not None, "mc_years should be a `Sequence` of integers"
```
MartinBelthle: You should remove this check as we can have a None here which means we want all mc_years (see line 199 of the file)

antarest/study/business/aggregator_management.py
```diff
@@ -144,20 +225,98 @@ def _gather_all_files_to_consider(self) -> t.Sequence[Path]:
...
+            df = pd.DataFrame(new_obj).reindex(columns=columns_order).sort_values(by=[TIME_ID_COL, CLUSTER_ID_COL])
+            df.index = pd.Index(list(range(1, len(df) + 1)))
+
+            return df
+
+        else:
+            return self._parse_output_file(file_path)
+
     def _build_dataframe(self, files: t.Sequence[Path], horizon: int) -> pd.DataFrame:
+        if self.mc_root not in [MCRoot.MC_IND, MCRoot.MC_ALL]:
```
MartinBelthle, line 324: You can't run into this if as you've already checked that we were in one of the 2 cases, remove this

antarest/study/business/aggregator_management.py
```diff
@@ -144,20 +225,98 @@ def _gather_all_files_to_consider(self) -> t.Sequence[Path]:
...
+                    c for c in df.columns.tolist() if any(regex in c.lower() for regex in lower_case_columns)
+                ]
+            else:
+                filtered_columns = [c for c in df.columns.tolist() if c.lower() in lower_case_columns]
+            df = df.loc[:, filtered_columns]
+        return df
+
+    def _process_df(self, file_path: Path, is_details: bool) -> pd.DataFrame:
+        if is_details:
+            un_normalized_df = self._parse_output_file(file_path, normalize_column_name=False)
```
MartinBelthle, line 283: Please add some comments inside complex methods

antarest/study/business/aggregator_management.py
```diff
@@ -144,20 +225,98 @@ def _gather_all_files_to_consider(self) -> t.Sequence[Path]:
...
+            df_len = len(un_normalized_df)
+            cluster_dummy_product_cols = sorted(
+                set([(x[CLUSTER_ID_COMPONENT], x[DUMMY_COMPONENT]) for x in un_normalized_df.columns])
+            )
+            actual_cols = sorted(set(un_normalized_df.columns.map(lambda x: x[ACTUAL_COLUMN_COMPONENT])))
+            new_obj: t.Dict[str, t.Any] = {k: [] for k in actual_cols}
```
MartinBelthle: You can do actual_cols + [TIME_ID_COL, CLUSTER_ID_COL] to remove 2 lines

antarest/study/business/aggregator_management.py
```diff
@@ -54,28 +78,78 @@ def _checks_estimated_size(nb_files: int, df_bytes_size: int, nb_files_checked:
...
+        org_cols = [col for col in org_cols if col not in {column_name, MCYEAR_COL, TIME_COL}]
+        new_column_order = [column_name] + ([CLUSTER_ID_COL] if is_details else []) + [TIME_ID_COL, TIME_COL] + org_cols
+    else:
+        raise InvalidFieldForVersionError(f"Unknown Monte Carlo root: {mc_root}")
+
+    return new_column_order
+
+
+def _infer_production_column(cols: t.Sequence[str]) -> t.Optional[str]:
+    prod_col = None
```
MartinBelthle: You can replace this function by:
```python
next((c for c in cols if PRODUCTION_COLUMN_REGEX in c.lower().strip()), None)
```

antarest/study/business/aggregator_management.py
MartinBelthle: If you read my comment, you'll see that this was for all args

antarest/study/business/aggregator_management.py
```diff
@@ -144,20 +223,98 @@ def _gather_all_files_to_consider(self) -> t.Sequence[Path]:
...
+
+    def columns_filtering(self, df: pd.DataFrame, is_details: bool) -> pd.DataFrame:
+        # columns filtering
+        lower_case_columns = [c.lower() for c in self.columns_names]
+        if lower_case_columns:
+            if is_details:
+                filtered_columns = [CLUSTER_ID_COL, TIME_ID_COL] + [
+                    c for c in df.columns.tolist() if any(regex in c.lower() for regex in lower_case_columns)
+                ]
+            elif self.mc_root == MCRoot.MC_ALL:
```
MartinBelthle, line 255: I don't really understand why you separate MC_ALL from MC_IND here.
For me you should do the same thing for both (the MC_IND way) but I'm perhaps missing something.

MartinBelthle (changes requested):

antarest/study/business/aggregator_management.py
mabw-rte: The first if is not for `MC_IND`

If you mean `regex` vs exact `match`, I thought that `regex` makes much sense for the `MC_ALL`

antarest/study/business/aggregator_management.py
mabw-rte: Bien vu

antarest/study/business/aggregator_management.py
MartinBelthle: What I meant was that I think you can remove the `elif self.mc_root == MCRoot.MC_ALL: ` check but I'm perhaps not understanding something

antarest/study/business/aggregator_management.py
mabw-rte: No because, a developer could modify a function without knowing anything about what happens in an other one ... Thus you should keep it or merge functions.

antarest/study/business/aggregator_management.py
mabw-rte: I don't understand you either, I told that I prefer keeping the regex search for `MC_ALL`.
If you remove the `elif` you get an `exact match` for the `MC_ALL`.

MartinBelthle: Okay. But then why do you want an exact match for MC_IND ?

mabw-rte: Basically, in our previous discussions we wanted to have:
- `regex` if details
-  else `exact match` 

In the specific case of `MC_ALL`, I noticed that you may have many columns with the same initial part, e.g. `X EXP`, `X MIN`, `X MAX`; That's why I thought I should keep the `regex` for this.

MartinBelthle: Okay, I'll see with you on Monday for the thing you've noticed for MC_ALL it will be easier to understand.

antarest/study/business/aggregator_management.py
MartinBelthle: In general I understand your point but here this private function is only used once so IMO it doesn't apply. If you follow this logic, you'll have to also do this check inside `_process_df` etc.

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -175,14 +180,14 @@ def get_study(
         return Response(content=json_response, media_type="application/json")
 
     @bp.get(
-        "/studies/{uuid}/areas/aggregate/{output_id}",
```
sylvlecl, line 178: The plain deletion of those endoints are a breaking change of the API, without changing it version, which is dangerous.

We must make sure with users that it's really intended, and we must mention it in a special section in the changelog.

sylvlecl (commented):
We must not merge this before making sure the API breaking change is intended and documented in the changelog.

MartinBelthle (approved):

PR #2098, opened 2024-07-19
fix(sonar): resolve sonar security issues

MartinBelthle:
RTE has changed its policy concerning Sonar. From September we'll have to have an A score in security to be able to deploy the code in production.

This PR solves every security issue (so has an A score) but introduces a new security hotspot (already existed but Sonar didn't care I don't know why) which is dumb IMO cause the path is generated via fastAPI and therefore cannot be a zip bomb.


antarest/core/tasks/service.py
```diff
@@ -302,32 +303,36 @@ def list_db_tasks(self, task_filter: TaskListFilter, request_params: RequestPara
         return self.repo.list(task_filter, user)
 
     def await_task(self, task_id: str, timeout_sec: int = DEFAULT_AWAIT_MAX_TIMEOUT) -> None:
-        if task_id in self.tasks:
+        sanitized_task_id = base64.b64decode(base64.b64encode(task_id.encode("utf-8"))).decode("utf-8")
```
sylvlecl: This is not the right logic: the encode/decodes here only lead to having a sanitized ID identical to the input ID.

The rule for logs only targets end of lines, I think we should just have a sanity check which checks that the ID does not contain such characters.

The solution using base64 encoding as illustrated in sonar example should not be used when the input is correct, because it changes it (by definition, its aim is to obtain something where issues have been removed).
For example, base64 encoding of ID '7ac92d48-7be8-4961-a5ba7430d4b16fc' gives 'N2FjOTJkNDgtN2JlOC00OTYxLWE1YmE3NDMwZDRiMTZmY2Y='

antarest/matrixstore/web.py
```diff
@@ -138,8 +138,9 @@ def download_matrix(
         )
         params = RequestParameters(user=current_user)
         service.download_matrix(matrix_id, tmp_export_file, params)
+        sanitized_path = tmp_export_file.resolve()
         return FileResponse(
-            tmp_export_file,
+            sanitized_path,
             headers={"Content-Disposition": f'attachment; filename="matrix-{matrix_id}.txt'},
```
sylvlecl, line 144: Wasn't the issue the matrix_id part ?

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -569,8 +570,9 @@ def generate_task(
         denormalize: bool = False,
         from_scratch: bool = False,
     ) -> str:
+        sanitized_study_id = base64.b64decode(base64.b64encode(metadata.id.encode("utf-8"))).decode("utf-8")
```
sylvlecl: Same comment as for other log, this is not the right solution

antarest/study/storage/rawstudy/ini_reader.py
```diff
@@ -166,12 +166,13 @@ def __repr__(self) -> str:  # pragma: no cover
 
     def read(self, path: t.Any, **kwargs: t.Any) -> JSON:
         if isinstance(path, (Path, str)):
+            sanitized_path = Path(path).resolve()
```
sylvlecl: I don't think resolves are enough to check our input paths, although it could satisfy sonar.
I think instead we should check that they don't lead to unsecure paths for us like paths that go in undesired folders.

We need to think a little about this!

sylvlecl (changes requested):
The solution for logs is not right!

For paths, we need to think a little more about what we should check

sylvlecl:
Some of the paths identified as problematic by sonar are **false positives**:
they are actually temporary paths generated by a safe function see for example the download_matrix endpoint.

For those, the best solution would be to add an inline comment to ignore this specific rule ... but it is not supported by sonar (in python). We are left with identifying it as a false positive in sonarcloud.

sylvlecl:
I have identified the temporary paths as false positives in sonar.

I think the 2 other cases in ini_reader are also false positives: sonar seems to get confused in its analysis between methods of the study service and methods of "****_management" classes ...

sylvlecl:
I have also tagged the 2 path issues in ini_reader as false positive, indeed sonar analysis is wrong: it thinks the path is built from the user input, whereas we get it from the database, which is safe.

See false positives [here](https://sonarcloud.io/project/issues?issueStatuses=FALSE_POSITIVE&id=AntaresSimulatorTeam_api-iso-antares)

So only the log issues remain!

antarest/core/tasks/web.py
```diff
@@ -61,6 +63,13 @@ def get_task(
         Returns:
             TaskDTO: Information about the specified task.
         """
+        expected_pattern = "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
```
sylvlecl: - I think we should change the existing method "sanitize_uuid" to do this instead, and reuse it where sonar is complaining.
- HTTP status should be bad request 400
- We should compile the regex only once in a "global" variable, for performance reason.

sylvlecl (approved):

PR #2099, opened 2024-07-22
config.prod.yaml: restore content

insatomcat:
This file has probably been emptied by mistake.
Since it's the default used by the server deployment, it is pretty useful.

### DIFF
--- a/resources/deploy/config.prod.yaml
+++ b/resources/deploy/config.prod.yaml
@@ -1,0 +1,87 @@
+security:
+  disabled: false
+  jwt:
+    key: secretkeytochange
+  login:
+    admin:
+      pwd: admin
+  external_auth:
+    url: ""
+    default_group_role: 10
+
+db:
+  url: "postgresql://postgres:somepass@postgresql:5432/postgres"
+  admin_url: "postgresql://postgres:somepass@postgresql:5432/postgres"
+  pool_recycle: 3600
+
+storage:
+  tmp_dir: /antarest_tmp_dir
+  archive_dir: /studies/archives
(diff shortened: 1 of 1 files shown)


MartinBelthle:
Even though the name is misleading, we now use the `config.yaml` file (see the `package_antares_web.sh` script) in our local env and when building the Desktop version. 
So IMO you can remove this file and replace `config.prod.yaml` by `config.yaml` in the `docker.compose.yaml` file.

insatomcat:
well config.yaml points to an sqllite db, whereas docker/prod deployment aims at a postgresql db if I understand correctly.

https://github.com/AntaresSimulatorTeam/AntaREST/blob/6ef9d3675994235ad9997b2101c6134a2bc7ff06/resources/deploy/config.yaml#L8

This file makes antarest crash on docker:
`sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) unable to open database file
`

sylvlecl:
The docker-compose file we provide indeed starts a postgresql service, so it's expected to have the corresponding config:
indeed it was probably a mistake to erase the config.prod.yaml content, otherwise we could have simply removed it.

Our [deployment doc](https://antares-web.readthedocs.io/en/latest/install/2-DEPLOY/) also mentions postgresql for this example.

Thanks for the fix!

sylvlecl (approved):

sylvlecl:
(However totally agree @MartinBelthle that all those files have a misleading organisation, actually this file is not at all used in production anywhere, but only serves as an example to have a quick working environment with all components)

PR #2101, opened 2024-07-23
feat(ui-study): add button to display display 'digest' file on successful tasks in task list

skamril:
ANT-1905

### DIFF
--- a/webapp/src/components/App/Singlestudy/HomeView/InformationView/LauncherHistory/DigestDialog.tsx
+++ b/webapp/src/components/App/Singlestudy/HomeView/InformationView/LauncherHistory/DigestDialog.tsx
@@ -1,0 +1,54 @@
+import { Skeleton } from "@mui/material";
+import OkDialog, {
+  OkDialogProps,
+} from "../../../../../common/dialogs/OkDialog";
+import EditableMatrix from "../../../../../common/EditableMatrix";
+import UsePromiseCond from "../../../../../common/utils/UsePromiseCond";
+import type { LaunchJob } from "../../../../../../common/types";
+import { getStudyData } from "../../../../../../services/api/study";
+import usePromise from "../../../../../../hooks/usePromise";
+import { useTranslation } from "react-i18next";
+
+interface Props extends Pick<OkDialogProps, "open" | "onOk" | "onClose"> {
+  studyId: LaunchJob["studyId"];
+  outputId: LaunchJob["outputId"];
+}
+
+function DigestDialog({ studyId, outputId, ...dialogProps }: Props) {
+  const { t } = useTranslation();
+
+  const synthesisRes = usePromise(
+    () =>
+      getStudyData(studyId, `output/${outputId}/economy/mc-all/grid/digest`),
+    {
+      deps: [studyId, outputId],
+    },
+  );
+
+  return (
+    <OkDialog
+      {...dialogProps}
+      title="Digest"
+      okButtonText={t("global.close")}
+      fullScreen
+      sx={{ m: 5 }}
+    >
+      <UsePromiseCond
+        response={synthesisRes}
+        ifPending={() => <Skeleton sx={{ height: 1, transform: "none" }} />}
+        ifResolved={(matrix) =>
+          matrix && (
+            <EditableMatrix
+              matrix={matrix}
+              columnsNames={matrix.columns}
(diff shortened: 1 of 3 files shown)


webapp/src/components/App/Singlestudy/HomeView/InformationView/LauncherHistory/DigestDialog.tsx
```diff
@@ -0,0 +1,63 @@
...
+function DigestDialog({ studyId, outputId, ...dialogProps }: Props) {
+  const { t } = useTranslation();
+
+  const synthesisRes = usePromise(
+    async () => {
+      await new Promise((res) => {
+        setTimeout(() => {
+          res(1);
+        }, 5000);
+      });
```
hdinia: is that for dev testing ? consider removing it

webapp/src/components/App/Singlestudy/HomeView/InformationView/LauncherHistory/JobStepper.tsx
```diff
@@ -37,6 +39,15 @@ export const ColorStatus = {
   failed: "error.main",
 };
 
+const iconSx = {
```
hdinia: optional: `iconStyles`

webapp/src/components/App/Singlestudy/HomeView/InformationView/LauncherHistory/JobStepper.tsx
```diff
@@ -68,6 +79,9 @@ export default function VerticalLinearStepper(props: Props) {
   const [openConfirmationDialog, setOpenConfirmationDialog] =
     useState<boolean>(false);
   const [jobIdKill, setJobIdKill] = useState<string>();
+  const [openDigestDialog, setOpenDigestDialog] = useState<LaunchJob | null>(
+    null,
+  );
```
hdinia: Instead of using a single state to manage both the visibility and the data of the dialog, you should use two states: one for controlling the visibility and another for storing the data to be passed to the dialog.

```typescript
 const [openDigestDialog, setOpenDigestDialog] = useState(false);
 const [selectedJob, setSelectedJob] = useState<LaunchJob | null>(null);
```
then adapt the rest:

```typescript
// add the event handler
  const handleOpenDigestDialog = (job: LaunchJob) => {
    setSelectedJob(job);
    setOpenDigestDialog(true);
  };
  
  // use it here
  {job.status === "success" && (
      <Tooltip title="Digest">
        <EqualizerIcon
           onClick={handleOpenDigestDialog}
            sx={iconSx}
             />
        </Tooltip>
  )}
```

adapt this too/
```typescript
{openDigestDialog && selectedJob && (
        <DigestDialog
          open
          studyId={selectedJob.studyId}
          outputId={selectedJob.outputId}
          onOk={() => setOpenDigestDialog(false)}
        />
      )}
```

hdinia (changes requested):

webapp/src/components/App/Singlestudy/HomeView/InformationView/LauncherHistory/DigestDialog.tsx
skamril: Thanks

webapp/src/components/App/Singlestudy/HomeView/InformationView/LauncherHistory/JobStepper.tsx
skamril: Ok

webapp/src/components/App/Singlestudy/HomeView/InformationView/LauncherHistory/JobStepper.tsx
skamril: I use an easy way because the view must be redesign, so I wanted to change few codes.

The good way would to have a common state for all the dialog like this for example:

```typescript
const [dialogState, setDialogState] = useState<{ type?: string; data?: LaunchJob }>({});
```

Two states for this is not necessary.

skamril: I wanted to change few codes because the view must be redesigned, but I will make a change.

hdinia (approved):

PR #2105, opened 2024-07-26
feat(bc): use `update_config` instead of `update_bc` for multiple updates

MartinBelthle:
Will fix [ANT-1936]


antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
```diff
@@ -373,7 +373,8 @@ def apply_binding_constraint(
             old_groups = old_groups or set()
             new_groups = {bd.get("group", DEFAULT_GROUP).lower() for bd in binding_constraints.values()}
             removed_groups = old_groups - new_groups
-            remove_bc_from_scenario_builder(study_data, removed_groups)
+            if removed_groups:
```
sylvlecl: This optimization should go into the method implementation so that it benefits other callers

antarest/study/business/binding_constraint_management.py
```diff
@@ -807,17 +821,105 @@ def update_binding_constraints(
...
+        command = UpdateConfig(
+            target="input/bindingconstraints/bindingconstraints",
+            data=config,
+            command_context=command_context,
+        )
+        commands.append(command)
+        execute_or_add_commands(study, file_study, commands, self.storage_service)
         return updated_constraints
 
+    @staticmethod
```
sylvlecl: detail: in general we should prefer free methods instead of static ones

PR #2109, opened 2024-08-01
feat(bc): show existing matrices only

MartinBelthle:
We should make sure that this doesn't impact the R scripts before merging even though I'm 99% sure it won't.


antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -72,37 +58,48 @@ def _update_matrices_names(
...
+        BindingConstraintOperator.EQUAL: "eq",
+        BindingConstraintOperator.GREATER: "gt",
+        BindingConstraintOperator.LESS: "lt",
+    }
+
+    def get_matrix_path(constraint_folder: Path, matrix_id: str) -> Path:
+        raw_file = constraint_folder / f"{matrix_id}.txt"
+        return raw_file if raw_file.exists() else raw_file.with_suffix(".txt.link")
+
+    def get_target_path(existing_path: Path, constraint_folder: Path, matrix_id: str) -> Path:
```
sylvlecl: It was better to not manipulate directly files in this layer of the code (commands implementation): the implementation details of handling links or actual files should be left in the file tree code. Possibly, we'll be able to re-use copy and rename methods elsewhere in the future.

MartinBelthle: I agree it was better but with this dev some nodes won't exist so `copy` and `rename` methods won't work anymore.

antarest/study/storage/rawstudy/model/filesystem/root/input/bindingconstraints/bindingcontraints.py
```diff
@@ -66,3 +68,8 @@ def build(self) -> TREE:
         )
 
         return children
+
+    def _matrix_exists(self, matrix_id: str) -> bool:
+        matrix_wo_link = self.config.path / f"{matrix_id}.txt"
+        matrix_w_link = self.config.path / f"{matrix_id}.txt.link"
+        return matrix_wo_link.exists() or matrix_w_link.exists()
```
sylvlecl: In the initial design, the `build` method seemed to be working only based on the previously build `config` which should contain all the structural information of the study (see comment on `FileTreeStudyConfig` class), instead of going to the filesystem.

I don't really like it to change this principle without being sure of the consequences.
For example, the `check_errors` feature of `INode` which is supposed to detect inconsistencies between the config and the actual files on the filesystem does not make sense anymore (because the config does not contain the necessary information to check the structure), it will not work correctly.
In particular here, when for example we have an "equal" binding constraint, we MUST have an associated _eq matrix in the structure. If we just base the structure on the files that are actually present, and if the study is invalid (missing file), the structure will be wrong too.

If we want to stick with the initial design we should add the operator in the config, and base the children list on it.

antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
```diff
@@ -371,12 +371,16 @@ def apply_binding_constraint(
...
+        existing_files = study_data.tree.get(["input", "bindingconstraints"], depth=1)
         for matrix_term, matrix_alias in operator_matrices_map[current_operator]:
             if matrix_term:
                 if not isinstance(matrix_term, str):  # pragma: no cover
                     raise TypeError(repr(matrix_term))
                 if version >= 870:
-                    study_data.tree.save(matrix_term, ["input", "bindingconstraints", f"{bd_id}_{matrix_alias}"])
+                    matrix_id = f"{bd_id}_{matrix_alias}"
+                    if matrix_id not in existing_files:
+                        (study_data.config.study_path / "input" / "bindingconstraints" / f"{matrix_id}.txt").touch()
```
sylvlecl: Here too we should not manipulate files directly, we should try to stick with the save mechanism. Probably requires to change the `build` method so that it's not based on existing files, as proposed in the above comment.

antarest/study/storage/variantstudy/model/command/update_binding_constraint.py
```diff
@@ -72,37 +54,46 @@ def _update_matrices_names(
...
-        ), f"Node type not handled yet: LazyNode expected, got {type(new_matrix_node)}"
-        matrix_node.rename_file(new_matrix_node)
+
+    operator_matrices_map = {
+        BindingConstraintOperator.EQUAL: "eq",
+        BindingConstraintOperator.GREATER: "gt",
+        BindingConstraintOperator.LESS: "lt",
+    }
+
+    def _get_matrix_url(node: InputSeriesMatrix, bc_term: str) -> t.List[str]:
```
sylvlecl: Here again we should try to hide more filesystem details (suffixes etc) from the command layer.
To achieve this maybe we can simplify what we have done in the `InputSeriesMatrix` class and only support renaming/copying in the same folder, this way we don't need to cope with path issues, for example:
```python
current_node.copy(target="lt")
```

antarest/study/storage/rawstudy/model/filesystem/root/input/bindingconstraints/bindingcontraints.py
sylvlecl: (Also more importantly this node should not need to know the inner details of lazy nodes, with links etc)

sylvlecl (changes requested):
See previous comments, I think for now we should stick with the current approach, for consistency sake:
- operator in the config
- children based on the operator defined in the config

sylvlecl (approved):

PR #2110, opened 2024-08-01
feat(installer): add installer as a submodule

maugde:
linked to https://gopro-tickets.rte-france.com/browse/ANT-1966


skamril (dismissed):

.github/workflows/deploy.yml
```diff
@@ -48,6 +48,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install pydantic --no-binary pydantic
           pip install -r requirements-dev.txt
+          pip install -r installer/requirements.txt
```
sylvlecl: This seems to cause problems:
here we install the installer dependencies and the web server dependencies in the same place, but they are not compatible.

When I start the server, after having installed it with the installer (both generated by the script), I get the following error message:
```
pydantic/_internal/_config.py:334: UserWarning: Valid config keys have changed in V2:
* 'anystr_strip_whitespace' has been renamed to 'str_strip_whitespace'
* 'min_anystr_length' has been renamed to 'str_min_length'
```

Indeed, the installer depends on pydantic v2, and the web server on pydantic v1.

To solve this, we need to isolate the 2 build environments from one another.
I would propose for this to:
 - remove dependencies install from here
 - in the shell script:
    - create a venv or the web server and install its dependencies here
    - create a second venv for the installer and install its dependencies there

scripts/package_antares_web.sh
```diff
@@ -45,6 +46,18 @@ else
   popd
 fi
 
+echo "INFO: Generating the Installer for the Desktop application..."
+echo which python
```
sylvlecl: we can remove this line

sylvlecl (changes requested):
The script does not work as intended because we mix dependencies for the server with dependencies for the installer, therefore we get an invalid server executable.

Note that to validate the workflow behaviour you can add this branch in the "on: push: branches" list in `deploy.yml`. This way we'll be able to download the generated artifacts from the actions page.

.github/workflows/deploy.yml
maugde: I managed all installer dependencies in another workflow step within its own virtualenv, as you mentioned. I wrote another script that build the installer which is called before the archive step and after the desktop version build.

.github/workflows/deploy.yml
```diff
@@ -53,8 +54,17 @@ jobs:
         if: matrix.os == 'windows-latest'
         run: pip install -r requirements-windows.txt
 
-      - name: 📦 Packaging
-        run: bash ./package_antares_web.sh
+      - name: Initialize installer environment
+        run: |
+          python -m venv .installer
+          source .installer/bin/activate
```
sylvlecl: This does not seem to work on windows (see workflow failure):
 - the simpler solution will probably be to add `shell: bash` to this step.
 - otherwise syntax seems to be `<venv>\Scripts\Activate.ps1` for powershell which is the default shell on windows (see [doc](https://docs.python.org/3/library/venv.html))

.github/workflows/deploy.yml
```diff
@@ -53,8 +54,17 @@ jobs:
...
+          python -m pip install -r requirements.txt
+        working-directory: scripts
+
+      - name: Build installer
+        run: |
+          bash ./package_antares_installer.sh
```
sylvlecl: I think you don't call anymore the main packaging script which builds antares desktop  :) 
It still needs to be called after the installer build, since it creates the final zip that will contain everything

sylvlecl (dismissed):

sylvlecl (approved):

PR #2112, opened 2024-08-02
feat(ts-gen): add timeseries generation

MartinBelthle:
Here's an exemple of generation duration time with this PR performed after the commit `add little tests`.
It's clear to see why we need to use `tasks`.

duree_generation_v1

We'll probably encounter some timeouts during variants safe generation. To avoid that, we'll have to increase this limit which is hard-coded at 60 seconds inside the code


antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -0,0 +1,149 @@
...
+
+class GenerateThermalClusterTimeSeries(ICommand):
+    """
+    Command used to generate thermal cluster timeseries for an entire study
+    """
+
+    command_name = CommandName.GENERATE_THERMAL_CLUSTER_TIMESERIES
+    version = 1
+
+    _INNER_MATRICES: t.List[str] = []
```
sylvlecl: probably better to keep lower case ?

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -0,0 +1,149 @@
...
+        return CommandOutput(status=True, message="Nothing to do"), {}
+
+    def _apply(self, study_data: FileStudy) -> CommandOutput:
+        study_path = study_data.config.study_path
+        with tempfile.TemporaryDirectory(
+            suffix=".thermal_timeseries_gen.tmp", prefix="~", dir=study_path.parent
+        ) as path:
+            tmp_dir = Path(path)
+            try:
+                shutil.copytree(study_path / "input" / "thermal" / "series", tmp_dir, dirs_exist_ok=True)
```
sylvlecl, line 47: I guess the dir should not exist since the parent has just been created --> we can remove last arg.

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def _build_timeseries(self, study_data: FileStudy, tmp_path: Path) -> None:
+        # 1- Get the seed and nb_years to generate
+        # NB: Default seed in IHM Legacy: 5489, default seed in web: 3005489.
+        general_data = study_data.tree.get(["settings", "generaldata"], depth=3)
+        thermal_seed = general_data["seeds - Mersenne Twister"]["seed-tsgen-thermal"]
+        nb_years = general_data["general"]["nbtimeseriesthermal"]
+        # 2 - Build the generator
+        rng = MersenneTwisterRNG(seed=thermal_seed)
+        generator = ThermalDataGenerator(rng=rng, days=365)
```
sylvlecl, line 64: Maybe you already checked this: should we read the number of days from the simulation configuration ?

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -0,0 +1,149 @@
...
+            sorted_thermals = sorted(area.thermals, key=lambda x: x.id)
+            for thermal in sorted_thermals:
+                # 5 - Filters out clusters with no generation
+                if thermal.gen_ts == LocalTSGenerationBehavior.FORCE_NO_GENERATION:
+                    continue
+                # 6- Build the cluster
+                modulation_matrix = study_data.tree.get(
+                    ["input", "thermal", "prepro", area_id, thermal.id.lower(), "modulation"]
+                )["data"]
+                modulation_capacity = np.array([row[2] for row in modulation_matrix])
```
sylvlecl: We should use an enum or constants for the various column numbers

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -0,0 +1,149 @@
...
+        sorted_areas = {k: areas[k] for k in sorted(areas)}
+        for area_id, area in sorted_areas.items():
+            # 4- Loop through thermal clusters in alphabetical order
+            sorted_thermals = sorted(area.thermals, key=lambda x: x.id)
+            for thermal in sorted_thermals:
+                # 5 - Filters out clusters with no generation
+                if thermal.gen_ts == LocalTSGenerationBehavior.FORCE_NO_GENERATION:
+                    continue
+                # 6- Build the cluster
+                modulation_matrix = study_data.tree.get(
```
sylvlecl: To improve a little performances we can use the "return dataframe" parameter of the InputSeriesMatrix.parse method, and then only manipulate pandas and numpy arrays.

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -0,0 +1,149 @@
...
+                    ["input", "thermal", "prepro", area_id, thermal.id.lower(), "modulation"]
+                )["data"]
+                modulation_capacity = np.array([row[2] for row in modulation_matrix])
+                ts_generator_matrix = study_data.tree.get(
```
sylvlecl: Same here for the dataframe

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -0,0 +1,149 @@
...
+                    fo_duration=fo_duration,
+                    fo_rate=fo_rate,
+                    po_duration=po_duration,
+                    po_rate=po_rate,
+                    npo_min=npo_min,
+                    npo_max=npo_max,
+                )
+                # 7- Generate the time-series
+                results = generator.generate_time_series(cluster, nb_years)
+                generated_matrix = results.available_power.T.tolist()
```
sylvlecl: For performances we'd better not call `tolist`, the create method seems to accept a numpy array.

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -0,0 +1,149 @@
...
+                # 8- Generates the UUID for the `get_inner_matrices` method
+                uuid = study_data.tree.context.matrix.create(generated_matrix)
+                self._INNER_MATRICES.append(uuid)
+                # 9- Write the matrix inside the input folder.
```
sylvlecl: It's a shame that we need to:
1. write the matrix twice, in the matrix store AND as a file here
2. duplicate here the code that writes a matrix to a file: it's only 1 line but we need to ensure consistency between here and other places where we write matrice, it's error-prone for future evolutions

For 2, we should probably have a small utility method shared at least beween `matrix.py` and here.
For 1, I have not easy solution, we probaby need to postpone this ... Note that this kind of direct write will be an issue the day we want to store input matrices with a different format.

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -0,0 +1,149 @@
...
+        original_path = study_path / "input" / "thermal" / "series"
+        original_path.rename(backup_dir)
+        tmp_path.rename(original_path)
+        shutil.rmtree(backup_dir)
+
+    @staticmethod
+    def _build_matrix_path(matrix_path: Path) -> Path:
+        real_path = matrix_path / "series.txt"
+        if not real_path.exists():
+            (matrix_path / "series.txt.link").rename(real_path)
```
sylvlecl, line 140: Are we sure this link file exists ? Cannot we have no file at all ?

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -0,0 +1,149 @@
...
+        # Only used inside the cli app that no one uses I believe.
+        raise NotImplementedError()
+
+    def get_inner_matrices(self) -> t.List[str]:
+        # This is used to get used matrices and not remove them inside the garbage collector loop.
+        return self._INNER_MATRICES
+
+    @staticmethod
+    def _replace_safely_original_files(study_path: Path, tmp_path: Path) -> None:
+        backup_dir = Path(
```
sylvlecl: Not clear to me what the backup dir is for, it seems that we never restore it:
should we add a try / catch to try to restore it if we get an exception when we perform the move operation ?

sylvlecl (changes requested):
Suggestion to use more numpy/pandas for performance reason + question on backup mechanism + a few minor comments.

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
MartinBelthle: In fact I need it as otherwise the copytree fails as the tmp_dir already exists. I could change it but that would mean adding the line `tmpdir.rmdir()` just above and I think it doesn't add any value

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
MartinBelthle: I didn't understand it at first but since we always want to generate 8760 rows (Thibault says), we need to have 365 days (*24 inside the generation package) I believe.

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
MartinBelthle: Smart

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
MartinBelthle: Indeed

MartinBelthle: As numpy operator T returns a Fortran-contiguous array we have to copy it to make it C-contiguous. To avoid this however, we can change the code of the generation package to make it build the matrix the right way. 
I'll do a release of the package that does so.

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
MartinBelthle: I spoke with Florian and he says the Simulator needs a file. So I don't think we can have no file at all (also we create automatically a file when creating a thermal cluster inside AntaresWeb)

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
MartinBelthle: As we discussed, I removed the backup_dir that wasn't used.

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
MartinBelthle: I did the release so I don't need the `tolist()` method anymore

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
MartinBelthle: Now that we don't have inner_matrices anymore, the 1. is no longer relevant.
Also for 2. I created a function dump_dataframe inside matrix.py that I reused here

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
MartinBelthle: I used constants

sylvlecl (approved):

PR #2122, opened 2024-08-19
build: move all build and project configuration to pyproject.toml

sylvlecl:
Configuration were arbitrarily dispatched between setup.py, setup.cfg, and the newer pyproject.toml.


MartinBelthle (commented):
The only change i think you should do is inside file `update_version.py` as it still uses `setup.py` line 120.

MartinBelthle (approved):

PR #2128, opened 2024-08-28
fix(launcher-api): remove orphan JobResults visibility permissions

mabw-rte:
Context: 
In the job listing of `AntaREST`, we call two endpoints: 
- `GET`, `v1/launcher/jobs`
- `GET`, `v1/launcher/jobs/{job_id}/progress`
It turns out that when user does not have a permission for some job, this job, though listed (using the first endpoint of `jobs`), spits out an error `403` when it is called for the second endpoint `progress`.

Issue:
For the endpoint `jobs`, the logic beyond the permissions includes a second option where when an orphan job is created recently enough would become visible to everyone. This logic was not included in the `progress` endpoint which leads to the `Exception` above. 

Solution:
After consulting the application users, they think that this `orphan` visibility logic is not that much relevant and should be removed (at least temporarily).

### DIFF
--- a/antarest/launcher/service.py
+++ b/antarest/launcher/service.py
@@ -2,7 +2,7 @@ import functools
 import logging
 import os
 import shutil
-from datetime import datetime, timedelta
+from datetime import datetime
 from http import HTTPStatus
 from pathlib import Path
 from typing import Dict, List, Optional, cast
@@ -59,7 +59,6 @@ class LauncherServiceNotAvailableException(HTTPException):
         )
 
 
-ORPHAN_JOBS_VISIBILITY_THRESHOLD = 10  # days
 LAUNCHER_PARAM_NAME_SUFFIX = "output_suffix"
 EXECUTION_INFO_FILE = "execution_info.ini"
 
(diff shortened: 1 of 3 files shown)


antarest/launcher/service.py
```diff
@@ -305,7 +305,6 @@ def _filter_from_user_permission(self, job_results: List[JobResult], user: Optio
         if not user:
             return []
 
-        orphan_visibility_threshold = datetime.utcnow() - timedelta(days=ORPHAN_JOBS_VISIBILITY_THRESHOLD)
```
MartinBelthle, line 308: You can also remove the constant as we don't use it anymore

MartinBelthle (changes requested):
Seems good to me just a minor change

MartinBelthle:
You can also remove the import of `timedelta` that we don't use anymore I believe

sylvlecl (approved):

MartinBelthle (approved):

mabw-rte (commented):
@sylvlecl, il faudrait priviligier de `rebase` en local ... Tu pourrais voir avec @skamril, on a deja eu la discussion

PR #2130, opened 2024-08-30
chore(scripts): add license headers for projects files

MartinBelthle:
Fix [ANT-1916] for python files only

This PR adds a license header for every python files inside folders `antarest` and `tests`.
It also introduces a new github action that checks that every one of these files contains a header.
It's based on a home-made Python script (ugly and unoptimized but runs quickly enough) that I wrote and put inside the `scripts` folder.


sylvlecl (dismissed):
Nice to have automated this in CI :)

sylvlecl:
We need to add the AUTHORS.txt the header refers to.

PR #2131, opened 2024-09-02
ci: parallel pytest execution

sylvlecl:
Experiment the use of pytest-xdist to speed up test execution in CI and dev environment:
see for example [this job](https://github.com/AntaresSimulatorTeam/AntaREST/actions/runs/10668008235/job/29566771799) where tests are executed in under 10min on ubuntu.

Following issues with parallel execution of tests are solved :
- some checks depended on the consistency at 2 moments in time of the result of `disk_usage`, which was anyway not a good idea (dependent on what happens elsewhere on the machine). It's now replaced with mocking
- some parameterized tests had random sort order, which is not allowed by pytest-xdist (all workers must have exactly the same tests in the same order)
- the `command_factory` test checked something in its teardown, assuming all tests were executed in the same process, which is not the case anymore. The same check is done as a separate unit tests.

### DIFF
--- a/antarest/login/auth.py
+++ b/antarest/login/auth.py
@@ -16,13 +16,13 @@ from datetime import timedelta
 from typing import Any, Callable, Coroutine, Dict, Optional, Tuple, Union
 
 from fastapi import Depends
+from fastapi_jwt_auth import AuthJWT  # type: ignore
 from pydantic import BaseModel
 from ratelimit.types import Scope  # type: ignore
 from starlette.requests import Request
 
 from antarest.core.config import Config
 from antarest.core.jwt import DEFAULT_ADMIN_USER, JWTUser
-from fastapi_jwt_auth import AuthJWT  # type: ignore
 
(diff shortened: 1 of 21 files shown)


antarest/login/auth.py
mabw-rte, line 1: Do you have an idea why this was not corrected by `isort` ?

mabw-rte (commented):
@sylvlecl  

- For the `rebase` it should be launched locally (always). 
- The `commit` messages should be corrected.
- The branch naming does not respect the `git flow` we fixed.

antarest/login/auth.py
sylvlecl: Actually there is no change on import sort order in the final PR. Intermediate commits introduced unwanted changes (related to another PR).

sylvlecl:
Closed in favor of https://github.com/AntaresSimulatorTeam/AntaREST/pull/2133 which fixes naming issues

PR #2134, opened 2024-09-05
feat(ui-common): integrate `GlideDataGrid` into `MatrixGrid`

hdinia:


webapp/vite.config.ts
```diff
@@ -30,7 +30,7 @@ export default defineConfig(({ mode }) => {
       },
     },
     test: {
-      globals: true, // Use the APIs globally
+      globals: true, // Use the APIs globally,
```
skamril, line 33: ?

webapp/src/components/common/MatrixGrid/Matrix.tsx
```diff
@@ -0,0 +1,106 @@
...
+    redo,
+    canUndo,
+    canRedo,
+  } = useMatrix(study.id, url, enableTimeSeriesColumns, enableAggregateColumns);
+
+  ////////////////////////////////////////////////////////////////
+  // JSX
+  ////////////////////////////////////////////////////////////////
+
+  if (isLoading) {
```
skamril, line 52: Use `UsePromiseCond` component

skamril (dismissed):

webapp/vite.config.ts
hdinia: with that you can use Vitest methods globally see: https://vitest.dev/config/#globals

hdinia: you don't have to import like vi, test, describe etc each time

skamril (approved):

webapp/src/components/common/MatrixGrid/Matrix.tsx
hdinia: I prefer not to use it in that case, I handled the errors and loading states

PR #2135, opened 2024-09-06
feat(variant): add new endpoint to clear snapshots

maugde:
Adding the new endpoint "clear-snapshots" will allow admin user to clear old snapshots in order to easily gain disk usage.


antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1050,3 +1051,35 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
                 exc_info=e,
             )
             return False
+
+    def clear_all_snapshots(self, limit: int, params: t.Optional[RequestParameters] = None) -> None:
```
mabw-rte: For `Mypy` you should deal first with the case `params=None` (it is an Optional parameter)

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1050,3 +1051,35 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
...
+            limit (integer): number of hours
+            params: request parameters used to identify the user status
+        Returns: None
+
+        Raises:
+            UserHasNotPermissionError
+            HTTPException
+        """
+        if limit < 0:
+            raise HTTPException(status_code=400, detail=f"Limit cannot be negative (limit={limit})")
```
mabw-rte: The Exception should be a `422` instead of `400`. You can look for a suitable Exception in `antarest/core/exceptions.py`.
If you think none of the `422` is ok for this case, you might create your own.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1050,3 +1051,35 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
...
+                    access_permissions=AccessPermissions(is_admin=True),
+                )
+            )
+            for variant in result:
+                if variant.updated_at < datetime.now() - timedelta(hours=limit) or (
+                    variant.last_access and variant.last_access < datetime.now() - timedelta(hours=limit)
+                ):
+                    self.clear_snapshot(variant)
+        else:
+            raise UserHasNotPermissionError
```
mabw-rte: You should raise an instance of the class. I am not sure but that should be `UserHasNotPermissionError()`.

antarest/study/web/variant_blueprint.py
```diff
@@ -421,4 +422,32 @@ def create_from_variant(
         params = RequestParameters(user=current_user)
         raise NotImplementedError()
 
+    @bp.put(
```
mabw-rte, line 419: You should add some integration tests for this endpoint

antarest/study/storage/variantstudy/variant_study_service.py
mabw-rte: Add some Unit tests for this function.

antarest/study/storage/variantstudy/variant_study_service.py
sylvlecl: Note that it's also possible to use a `Query` annotation on the limit argument, with "ge=0" argument to specify that it should be greater or equal to 0.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1050,3 +1051,35 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
...
+            raise HTTPException(status_code=400, detail=f"Limit cannot be negative (limit={limit})")
+        if params.user.is_site_admin():
+            result = self.repository.get_all(
+                study_filter=StudyFilter(
+                    variant=True,
+                    access_permissions=AccessPermissions(is_admin=True),
+                )
+            )
+            for variant in result:
+                if variant.updated_at < datetime.now() - timedelta(hours=limit) or (
```
sylvlecl: Maybe it's better to have an "and" condition: if it has been modified or read recently, we don't delete it.
For example, if it has been read recently, although it has not been modified, it makes sense to keep it.

antarest/study/web/variant_blueprint.py
```diff
@@ -421,4 +422,32 @@ def create_from_variant(
...
+        "/studies/variants/clear-snapshots",
+        tags=[APITag.study_variant_management],
+        summary="Clear variant snapshots",
+        responses={
+            200: {
+                "description": "Delete snapshots older than a specific number of hours. By default, this number is 24."
+            }
+        },
+    )
+    def clear_variant_snapshots(
```
sylvlecl, line 429: After more thinking:
it could be relevant to execute this cleanup as a `Task` by the task service, because it will take a long time to execute.

Then here we should return the task ID. You can for example look at the `THERMAL_CLUSTER_SERIES_GENERATION` that Martin added recently, to see how it's done! it's quite simple you need to add a new task type and provide the function to be executed to the task service, instead of executing it directly in the variant service.

sylvlecl (commented):
See comment: the best option will be to execute the function as a task because it's a long processing

antarest/study/storage/variantstudy/variant_study_service.py
mabw-rte: You don't need `Query` it will create a duplicate doc for your `argument`.
What I meant is that the type of your `HTTPSExpection` should be a `422` not `400`.

tests/integration/variant_blueprint/test_variant_manager.py
```diff
@@ -45,6 +47,60 @@ def variant_id_fixture(
...
+            m.setattr("antarest.study.storage.variantstudy.variant_study_service.datetime", FakeDatetime)
+
+            for index, different_time in enumerate([older_time, old_time, recent_time]):
+                FakeDatetime.fake_time = different_time
+                res = client.post(f"/v1/studies/{base_study_id}/variants?name=variant{index}", headers=admin_headers)
+                variant_ids.append(res.json())
+
+                # Generate snapshot for each variant
+                client.put(f"/v1/studies/{variant_ids[index]}/generate", headers=admin_headers)
+        time.sleep(0.1)  # wait for the filesystem to be updated
```
sylvlecl: You can use the `wait_task_completion` method to make sure that the generation task has ended (you will need to get the id of the task from the `PUT` call above).

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -239,3 +242,168 @@ def test_generate_task(
...
+        # Check if everything was correctly initialized
+        assert len(list(variant_study_path.iterdir())) == 3
+
+        for variant in variant_study_path.iterdir():
+            assert variant.is_dir()
+            assert list(variant.iterdir())[0].name == "snapshot"
+
+        # Begin tests
+        # A user without rights cannot clear snapshots
+        try:
```
sylvlecl: you may use `with pytest.raises(UserHasNotPermissionError):` instead of try/catch, to check that this exception is raised.

tests/integration/variant_blueprint/test_variant_manager.py
```diff
@@ -330,3 +386,29 @@ def test_outputs(client: TestClient, admin_access_token: str, variant_id: str, t
...
+    # Check intitial data
+    assert older.exists() and old.exists() and recent.exists()
+
+    # Try to call the endpoint with a negative value. Must return a 422 error code
+    res = client.put(f"v1/studies/variants/clear-snapshots?limit=-1", headers=admin_headers)
+
+    assert res.status_code == 422
+    assert res.json().get("exception") == "VariantAgeMustBePositive"
+
+    assert older.exists() and old.exists() and recent.exists()
```
sylvlecl: would be nice to check also a valid call, not only the invalid value.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1050,3 +1054,90 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
                 exc_info=e,
             )
             return False
+
+    def clear_all_snapshots(self, limit: int, params: t.Optional[RequestParameters]) -> str:
```
sylvlecl: for clarity, we could rename `limit` to `retention_time` and change the type to `datetime.timedelta` in order to be independent from the unit used (hours).
The conversion int --> timedelta will find a good home in the web layer, which translates request parameters to service arguments.

antarest/study/web/variant_blueprint.py
```diff
@@ -421,4 +421,32 @@ def create_from_variant(
...
+        limit: int = 24,
```
sylvlecl: For clarity in the API we can name it `retention_hours`

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1050,3 +1054,90 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
...
+        self._variant_list = variant_list
+        self._limit = limit
+
+    def _clear_all_snapshots(self) -> None:
+        with db():
+            for variant in self._variant_list:
+                if variant.updated_at < datetime.now() - timedelta(hours=self._limit):
+                    if variant.last_access and variant.last_access < datetime.now() - timedelta(hours=self._limit):
+                        logger.info(f"Variant {variant.id} detected.")
+                        path = self._variant_study_service.get_study_path(variant)
```
sylvlecl: we'd better use the existing method `clear_snapshot`, which in particular also invalidates the cache of the study

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1050,3 +1054,90 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
...
+        if not result:
+            msg = f"No variant updated or accessed {limit} hours ago were found."
+            logger.info(msg)
+            return msg
+
+        # Make sure the task is not running
+        study_tasks = self.task_service.list_tasks(
+            TaskListFilter(
+                ref_id="snapshot_cleaning",
+                type=[TaskType.THERMAL_CLUSTER_SERIES_GENERATION],
```
sylvlecl: wrong type here

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1050,3 +1054,90 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
...
+                access_permissions=AccessPermissions(is_admin=True),
+            )
+        )
+
+        if not result:
+            msg = f"No variant updated or accessed {limit} hours ago were found."
+            logger.info(msg)
+            return msg
+
+        # Make sure the task is not running
```
sylvlecl: Actually, I don't think we need to check that another clean is running.
It's ok to run multiple cleaning independently, there is no real reason to forbid it.

At worse, some variants will have their snapshot cleared multiple times, which will just do nothing after the first time.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1050,3 +1054,90 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
...
+        Args:
+            limit (integer): number of hours
+            params: request parameters used to identify the user status
+        Returns: None
+
+        Raises:
+            UserHasNotPermissionError
+            VariantAgeMustBePositive
+        """
+        if params is None:
```
sylvlecl: we should also check that `user.is_site_admin()`, to allow only admins to use this feature

sylvlecl (changes requested):
A few changes required:
- check that user is admin
- not necessary to check for already running tasks
- advise on naming in particular for the API

Also the installer submodule is updated: it should not be the case in this PR.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1053,3 +1055,68 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
...
+
+        Args:
+            retention_hours: number of retention hours
+            params: request parameters used to identify the user status
+        Returns: None
+
+        Raises:
+            UserHasNotPermissionError
+        """
+        if params is None or (params.user and not params.user.is_site_admin()):
```
MartinBelthle: You should add `or user.is_admin_token()` to support the fact that an admin calls this endpoint via a script

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1053,3 +1055,68 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
...
+    def _clear_all_snapshots(self) -> None:
+        with db():
+            variant_list = self._variant_study_service.repository.get_all(
+                study_filter=StudyFilter(
+                    variant=True,
+                    access_permissions=AccessPermissions(is_admin=True),
+                )
+            )
+            for variant in variant_list:
+                if variant.updated_at < datetime.now() - self._retention_hours:
```
MartinBelthle: This comment goes for this line and the following:

Not sure and you've probably tested it inside the tests but there could be a problem of timezone UTC, we should check how we save the field info to be sure it's not a problem

Other remark, are we sure we'll always have the field ? Cause line 1109 you don't check if you have it and line 1110 you check it.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1053,3 +1055,68 @@ def initialize_additional_data(self, variant_study: VariantStudy) -> bool:
...
+                    if variant.last_access and variant.last_access < datetime.now() - self._retention_hours:
+                        logger.info(f"Variant {variant.id} detected.")
```
MartinBelthle: Not a huge fan of this log as it's not really precise. IMO we should remove it as the log inside the clear_snapshot method is enough

antarest/study/web/variant_blueprint.py
```diff
@@ -416,4 +416,33 @@ def create_from_variant(
...
+    def clear_variant_snapshots(
+        hours: int = 24,
+        current_user: JWTUser = Depends(auth.get_current_user),
+    ) -> str:
+        """
+        Endpoint that clear `limit` hours old and older variant snapshots.
+
+        Args: limit (int, optional): Number of hours to clear. Defaults to 24.
+
+        Returns: None
```
MartinBelthle: It doesn't return None it returns the created task id as you typed it above.

antarest/study/web/variant_blueprint.py
```diff
@@ -416,4 +416,33 @@ def create_from_variant(
...
+        """
+        retention_hours = datetime.timedelta(hours=hours)
+        logger.info(
+            f"Delete all variant snapshots older than {retention_hours} hours (24 by default)",
```
MartinBelthle: Don't really see why you add the (24 by default) inside the log as it doesn't bring any info IMO

tests/integration/variant_blueprint/test_variant_manager.py
```diff
@@ -45,6 +48,66 @@ def variant_id_fixture(
...
+def generate_snapshot_fixture(
+    client: TestClient,
+    admin_access_token: str,
+    base_study_id: str,
+    monkeypatch: pytest.MonkeyPatch,
+    caplog: t.Any,
+) -> [str]:
+    """Generate some snapshots with different date of update and last access"""
+
+    class FakeDatetime:
```
MartinBelthle, line 61: nice way to do this

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -239,3 +242,181 @@ def test_generate_task(
...
+        variant_study_path = Path(tmp_path).joinpath("internal_studies")
+
+        # Check if everything was correctly initialized
+        assert len(list(variant_study_path.iterdir())) == 3
+
+        for variant in variant_study_path.iterdir():
+            assert variant.is_dir()
+            assert list(variant.iterdir())[0].name == "snapshot"
+
+        # Begin tests
```
MartinBelthle: Perhaps just separate the set up and the tests with a bigger sign to ensure better readability. In other tests, we used something like this:
```python
# =============================
#  SET UP
# ============================= 
```

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -239,3 +242,181 @@ def test_generate_task(
...
+        task_id = variant_study_service.clear_all_snapshots(
+            datetime.timedelta(hours=-1),
+            Mock(
+                spec=RequestParameters,
+                user=DEFAULT_ADMIN_USER,
+            ),
+        )
+        variant_study_service.task_service.await_task(task_id)
+
+        # Check if all snapshots was cleared
```
MartinBelthle: were

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -239,3 +242,181 @@ def test_generate_task(
...
+        nb_snapshot_dir = 0  # after the for iterations, must equal 0
+        for variant_path in variant_study_path.iterdir():
+            if variant_path.joinpath("snapshot").exists():
```
MartinBelthle: Better way to do it is `assert not variant_path.joinpath("snapshot").exists()` instead of introducing a counter and you can even do it inside one line

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -239,3 +242,181 @@ def test_generate_task(
...
+        task_id = variant_study_service.clear_all_snapshots(
+            datetime.timedelta(hours=5),
+            Mock(
+                spec=RequestParameters,
+                user=DEFAULT_ADMIN_USER,
+            ),
+        )
+        variant_study_service.task_service.await_task(task_id)
+
+        # Check if old snapshots was successfully cleared
```
MartinBelthle, line 409: were

MartinBelthle (commented):
Seems good, just minor changes/questions , you've added many tests it's nice :)

antarest/study/storage/variantstudy/variant_study_service.py
maugde: I indeed checked only on the `last_access` field because it seemed that the `updated_at` field was always set after a variant creation. In order to remain consistant, i'll add the same test as `last_access`.

sylvlecl (approved):

PR #2136, opened 2024-09-09
fix(standalone): fix api path when embedding front-end

sylvlecl:
**Context**

When including the front-end application in the python server (standalone or "desktop" mode),
we want to be able to append a /api prefix to backend URLs.
So far, this has not been correctly done:
`root_path` has been used, whereas it does not actually change the URLs, its only purpose
is for the server to know that it's behind a proxy which exposes this prefix to the client,
see [fastapi doc](https://fastapi.tiangolo.com/advanced/behind-a-proxy/) about this.

This has more or less worked so far because of a bug in starlette: the server indeed served
the services at both URLs: with a without the prefix.

With newer fastapi and starlette versions, this does not work anymore: **the desktop application
is broken**. We need to clean up the implementation to actually serve the services with a prefix.

**Implementation**

Another configuration property `api_prefix` is introduced, which is used to prefix all
backend routes (through a "root" `APIRouter`).

All front related logic is moved to a dedicated module `antarest.front`.

### DIFF
--- a/antarest/study/business/all_optional_meta.py
+++ b/antarest/study/business/all_optional_meta.py
@@ -17,8 +17,10 @@ from pydantic import BaseModel, create_model
 
 from antarest.core.utils.string import to_camel_case
 
+ModelClass = t.TypeVar("ModelClass", bound=BaseModel)
 
-def all_optional_model(model: t.Type[BaseModel]) -> t.Type[BaseModel]:
+
+def all_optional_model(model: t.Type[ModelClass]) -> t.Type[ModelClass]:
     """
     This decorator can be used to make all fields of a pydantic model optionals.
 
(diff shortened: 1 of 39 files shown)


antarest/study/business/all_optional_meta.py
```diff
@@ -58,4 +60,5 @@ def camel_case_model(model: t.Type[BaseModel]) -> t.Type[BaseModel]:
         field.alias = new_alias
         field.validation_alias = new_alias
         field.serialization_alias = new_alias
+        field.alias_priority = 2
```
mabw-rte: I prefer defining constants like these as global variables for the module.
This enforces maintainability and gives an explicit documentation for the variable using via the naming.
Example:
`DEFAULT_ALIAS_PRIORITY=2`

mabw-rte (commented):
I like the fact that you eliminated a lot of duplicate code.

antarest/study/business/all_optional_meta.py
sylvlecl: Indeed! Actually I just removed the line, I left it by mistake, this change is not needed.

PR #2144, opened 2024-09-18
chore(variants): increase timeout duration for variant generation

MartinBelthle:


antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -497,7 +497,7 @@ def get(
 
         Returns: study data formatted in json
         """
-        self._safe_generation(metadata, timeout=60)
+        self._safe_generation(metadata, timeout=600)
```
sylvlecl, line 500: To be honest I am not sure why we have a timeout here.

If the generation task hangs forever we anyway have a problem, so it does not solve much (the task thread will keep busy).
If it's just "long", which may happen, the client will have its own timeout.

sylvlecl (approved):

antarest/study/storage/variantstudy/variant_study_service.py
sylvlecl: Let's keep it for now .... it can have the small benefit of not keeping a fastapi thread busy for too long ...

PR #2146, opened 2024-09-18
fix(links): endpoint links create

TheoPascoli:


MartinBelthle:
I see that there are still 68 tests that fail (probably all for the same reason) so I'll let you fix that before reviewing your work :)

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -30,8 +31,8 @@ class LinkProperties:
     LOOP_FLOW: bool = False
     USE_PHASE_SHIFTER: bool = False
     DISPLAY_COMMENTS: bool = True
-    TRANSMISSION_CAPACITIES: str = "enabled"
-    ASSET_TYPE: str = "ac"
+    TRANSMISSION_CAPACITIES: TransmissionCapacity = "enabled"
```
MartinBelthle: I think it should be `TransmissionCapacity.ENABLED` and `AssetType.AC`. It shushes mypy issues

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -17,6 +17,7 @@
 from antarest.core.model import JSON
 from antarest.core.utils.utils import assert_this
 from antarest.matrixstore.model import MatrixData
+from antarest.study.storage.rawstudy.model.filesystem.config.links import AssetType, LinkStyle, TransmissionCapacity
```
MartinBelthle: `LinkStyle` import is unused

tests/storage/business/test_arealink_manager.py
```diff
@@ -37,10 +44,11 @@
 from antarest.study.storage.storage_service import StudyStorageService
 from antarest.study.storage.variantstudy.business.matrix_constants_generator import GeneratorMatrixConstants
 from antarest.study.storage.variantstudy.command_factory import CommandFactory
-from antarest.study.storage.variantstudy.model.command.common import CommandName
+from antarest.study.storage.variantstudy.model.command.common import CommandName, FilteringOptions
 from antarest.study.storage.variantstudy.model.dbmodel import VariantStudy
 from antarest.study.storage.variantstudy.model.model import CommandDTO
 from antarest.study.storage.variantstudy.variant_study_service import VariantStudyService
+from tests.conftest_services import study_storage_service_fixture
```
MartinBelthle: Unused import

tests/storage/business/test_arealink_manager.py
```diff
@@ -24,11 +24,18 @@
 from antarest.matrixstore.repository import MatrixContentRepository
 from antarest.matrixstore.service import SimpleMatrixService
 from antarest.study.business.area_management import AreaCreationDTO, AreaManager, AreaType, UpdateAreaUi
-from antarest.study.business.link_management import LinkInfoDTO, LinkManager
+from antarest.study.business.link_management import (
+    LinkInfoDTO820,
+    LinkInfoDTOBase,
+    LinkInfoDTOType,
```
MartinBelthle: Unused import

antarest/study/storage/rawstudy/model/filesystem/config/links.py
```diff
@@ -66,6 +66,42 @@ class TransmissionCapacity(EnumIgnoreCase):
...
+        DOT_DASH: Represents a line style with alternating dots and dashes.
+    """
+
+    DOT = "dot"
+    PLAIN = "plain"
+    DASH = "dash"
+    DOT_DASH = "dotdash"
+
+
+class FilterOption(EnumIgnoreCase):
```
MartinBelthle, line 87: You should probably use the class `FilteringOptions` inside file /variantstudy/model/command/common.py

antarest/study/business/link_management.py
```diff
@@ -14,30 +14,152 @@
...
+    filter_synthesis: t.Optional[str] = None
+    filter_year_by_year: t.Optional[str] = None
+
+
+LinkInfoDTOType = t.Union[LinkInfoDTO820, LinkInfoDTOBase]
+
+
+class LinkInfoFactory:
+    @staticmethod
+    def create_link_info(version: int, **kwargs) -> LinkInfoDTOType:
```
MartinBelthle: type `**kwargs: t.Any` it will solve mypy issue (same for other warnings in this file)

antarest/study/business/link_management.py
```diff
@@ -14,30 +14,152 @@
...
+        filters_provided = kwargs.get("filter_synthesis") is not None or kwargs.get("filter_year_by_year") is not None
+        if version < 820 and filters_provided:
+            raise InvalidFieldForVersionError(
+                "Filters filter_synthesis and filter_year_by_year cannot be used for study versions lower than 820."
+            )
+
+    @staticmethod
+    def create_parameters(
+        study_version: int, link_creation_info: LinkInfoDTOType
+    ) -> t.Dict[str, t.Union[str, bool, float]]:
```
MartinBelthle: As filter_synthesis and filter_year_by_year are optional = None you should add None inside the typing Union to fix the mypy issue.
Also change it inside the little following doc

antarest/study/business/link_management.py
```diff
@@ -52,38 +174,81 @@ class LinkManager:
...
+            filter_year_by_year=link_creation_info.filter_year_by_year if study_version >= 820 else None,
+        )
+
         storage_service = self.storage_service.get_storage(study)
         file_study = storage_service.get_raw(study)
+
         command = CreateLink(
             area1=link_creation_info.area1,
             area2=link_creation_info.area2,
+            parameters=LinkInfoFactory.create_parameters(int(study.version), link_info_dto),
```
MartinBelthle: You could use the variable `study_version` that you've introduced above.

antarest/study/business/link_management.py
```diff
@@ -52,38 +174,81 @@ class LinkManager:
...
+            use_phase_shifter=link_creation_info.use_phase_shifter,
+            transmission_capacities=link_creation_info.transmission_capacities,
+            asset_type=link_creation_info.asset_type,
+            display_comments=link_creation_info.display_comments,
+            colorr=link_creation_info.colorr,
+            colorb=link_creation_info.colorb,
+            colorg=link_creation_info.colorg,
+            link_width=link_creation_info.link_width,
+            link_style=link_creation_info.link_style,
+            filter_synthesis=link_creation_info.filter_synthesis if study_version >= 820 else None,
```
MartinBelthle: As you didn't cast link_creation_dto as a LinkInfoDTO820 instance, mypy doesn't know you can ask for this field. 
Little workaround to fix the issue: 
```python
filter_synthesis = None
filter_year_by_year = None
if study_version >= 820:
    link_creation_info = t.cast(LinkInfoDTO820, link_creation_info)
    filter_synthesis = link_creation_info.filter_synthesis
    filter_year_by_year = link_creation_info.filter_year_by_year
```

MartinBelthle (changes requested):
I'll take a deeper look tomorrow.
I'm not a big fan of having to give every field inside the `create_link_info` method it's a bit heavy in the code IMO (but I don't know how to do differently yet)

MartinBelthle:
I believe you should use the pydantic method `model_dump` instead of using the factory.
This method converts a Pydantic object (any class inheriting from BaseModel) into a Dict[str, Any]. 
We use this method everywhere in the code you can take a look to see how it works :)

tests/variantstudy/model/command/test_create_link.py
```diff
@@ -52,15 +47,6 @@ def test_validation(self, empty_study: FileStudy, command_context: CommandContex
             }
         ).apply(empty_study)
 
-        with pytest.raises(ValidationError):
```
MartinBelthle, line 55: You should keep this test it's important for people using directly the command

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -0,0 +1,173 @@
...
+            "filter-year-by-year": "hourly, daily, weekly, monthly, annual",
+            "hurdles-cost": False,
+            "link-style": "plain",
+            "link-width": 1.0,
+            "loop-flow": False,
+            "transmission-capacities": "enabled",
+            "use-phase-shifter": False,
+        }
+        assert expected == res.json()
+        client.delete(f"/v1/studies/{study_id}/links/{area1_id}/{area2_id}")
```
MartinBelthle: just do `res =...` and `res.raise_for_status()` to ensure the deletion worked.
The same goes for the other calls inside this test

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # Test create link with parameters
+
+        res = client.post(
+            f"/v1/studies/{study_id}/links",
+            json={
```
MartinBelthle: You could probably store this inside a variable and only add `filter-year-by-year` inside the expected output to avoid duplicated your code.

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+@pytest.mark.unit_test
+class TestLink:
+    @pytest.mark.parametrize("study_type", ["raw", "variant"])
+    def test_link_820(self, client: TestClient, user_access_token: str, study_type: str) -> None:
+        client.headers = {"Authorization": f"Bearer {user_access_token}"}  # type: ignore
+
+        preparer = PreparerProxy(client, user_access_token)
+        study_id = preparer.create_study("foo", version=880)
```
MartinBelthle: I'd prefer you use the version 820 or rename the test

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -0,0 +1,173 @@
...
+
+import pytest
+from starlette.testclient import TestClient
+
+from tests.integration.prepare_proxy import PreparerProxy
+
+
+@pytest.mark.unit_test
+class TestLink:
+    @pytest.mark.parametrize("study_type", ["raw", "variant"])
```
MartinBelthle, line 22: It's a good idea to test for `raw` and `variants` but this is not a real fixture. You'll have to do a check at the beginning of the file to really create a variant like this: 

```python
if study_type == "variant":
    study_id = preparer.create_variant(study_id, name="Variant 1")
```
otherwise you're just doing the same test twice.

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+
+        assert res.status_code == 422, res.json()
+        expected = {
+            "description": "Invalid value(s) in filters: centurial. Allowed values are: hourly, daily, weekly, monthly, annual.",
+            "exception": "LinkValidationError",
+        }
+        assert expected == res.json()
+
+    @pytest.mark.parametrize("study_type", ["raw", "variant"])
```
MartinBelthle: I think we don't need to parametrize this test, it will save some time

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -0,0 +1,173 @@
...
+        res = client.delete(f"/v1/studies/{study_id}/links/{area1_id}/{area3_id}")
+        assert res.status_code == 200, res.json()
+
+        res = client.get(f"/v1/studies/{study_id}/links")
+
+        assert res.status_code == 200, res.json()
+        assert 1 == len(res.json())
+        client.delete(f"/v1/studies/{study_id}/links/{area1_id}/{area2_id}")
+
+        # Test create link with same area
```
MartinBelthle, line 114: It could be good to add another test failing because of a wrongly typed enum parameter such as `transmission-capacities`

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -9,37 +9,80 @@
...
-    COLORR: int = 112
-    COLORG: int = 112
-    COLORB: int = 112
+class AreaInfo(BaseModel):
+    area1: str
+    area2: str
+
+
+class LinkInfoProperties(BaseModel):
+    hurdles_cost: Optional[bool] = Field(False, alias="hurdles-cost")
```
MartinBelthle: I believe you could use the method `to_kebab_case` I introduced here to avoid writing aliases at each line.

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -9,37 +9,80 @@
...
+    display_comments: Optional[bool] = Field(True, alias="display-comments")
+    colorr: Optional[int] = DEFAULT_COLOR
+    colorb: Optional[int] = DEFAULT_COLOR
+    colorg: Optional[int] = DEFAULT_COLOR
+    link_width: Optional[float] = Field(1, alias="link-width")
+    link_style: Optional[LinkStyle] = Field(LinkStyle.PLAIN, alias="link-style")
+
+    @model_validator(mode="before")
+    def validate_colors(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
+        if type(values) is dict:
```
MartinBelthle: We often do `if isinstance(values, dict)` instead of checking the type.
I think it's the same thing it's just to keep code consistency

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -9,37 +9,80 @@
...
+            colors = {
+                "colorr": values.get("colorr"),
+                "colorb": values.get("colorb"),
+                "colorg": values.get("colorg"),
+            }
+            for color_name, color_value in colors.items():
+                if color_value is not None and (color_value < 0 or color_value > 255):
```
MartinBelthle: I don't think `color_value` can be none here but I might be wrong.

Also you could try to remove this validator and just change 
`colorr: Optional[int] = DEFAULT_COLOR`  to 
`colorr: Optional[int] = Field(default=DEFAULT_COLOR, gt=0, lt=255)` it should do the same thing.

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -9,37 +9,80 @@
...
+                    raise LinkValidationError(f"Invalid value for {color_name}. Must be between 0 and 255.")
+
+        return values
+
+
+class LinkInfoProperties820(LinkInfoProperties):
+    filter_synthesis: Optional[str] = Field(None, alias="filter-synthesis")
```
MartinBelthle: You could also use `to_kebab_case` I think

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -9,37 +9,80 @@
...
+    filter_year_by_year: Optional[str] = Field(None, alias="filter-year-by-year")
+
+    @field_validator("filter_synthesis", "filter_year_by_year", mode="before")
+    def validate_individual_filters(cls, value: Optional[str], field: Any) -> Optional[str]:
```
MartinBelthle: `field` should be typed as an optional str I believe

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -9,37 +9,80 @@
...
+        if value is not None:
+            filter_options = ["hourly", "daily", "weekly", "monthly", "annual"]
```
MartinBelthle: You could import `FILTER_OPTIONS` to avoid hard-coding this list

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -71,12 +114,6 @@ def validate_series(
         new_values = values if isinstance(values, dict) else values.data
         return validate_matrix(v, new_values) if v is not None else v
 
-    @model_validator(mode="after")
```
MartinBelthle: Same as a comment i wrote already but you should keep this validation for the script team using directly the commands and you could use mode="before" it's always better

antarest/study/business/link_management.py
```diff
@@ -52,38 +56,73 @@ class LinkManager:
...
-                if with_ui and links_config and link in links_config:
-                    ui_info = LinkUIDTO(
-                        color=f"{links_config[link].get('colorr', '163')},{links_config[link].get('colorg', '163')},{links_config[link].get('colorb', '163')}",
-                        width=links_config[link].get("link-width", 1),
-                        style=links_config[link].get("link-style", "plain"),
+                link_properties = links_config[link]
+
+                link_creation_data: t.Dict[str, t.Any] = {"area1": area_id, "area2": link}
+                link_creation_data.update(link_properties)
+                if not with_ui:
```
MartinBelthle: I've seen with Sylvain we decided that it was better to remove the parameter `with_ui`and always return everything.
So you can adapt this code and the front-end will have to adapt also.

antarest/study/business/link_management.py
```diff
@@ -52,38 +56,73 @@ class LinkManager:
...
+                    link_data = LinkInfoDTO820(**link_creation_data)
+
+                result.append(link_data)
 
         return result
 
-    def create_link(self, study: RawStudy, link_creation_info: LinkInfoDTO) -> LinkInfoDTO:
+    def create_link(self, study: RawStudy, link_creation_info: LinkInfoDTOType) -> LinkInfoDTOType:
+        if link_creation_info.area1 == link_creation_info.area2:
+            raise LinkValidationError("Cannot create link on same node")
```
MartinBelthle: IMO it's better to log `area` instead of `node` and also add the concerned area name

antarest/study/business/link_management.py
```diff
@@ -52,38 +56,73 @@ class LinkManager:
...
+
+        study_version = int(study.version)
```
MartinBelthle: I think you can simplify the code form line 97 to 112 to 
```python
link_data: LinkInfoDTOType
if study_version < 820:
    if isinstance(link_creation_info, LinkInfoDTO820):
        raise LinkValidationError("Cannot specify a filter value for study's version earlier than v8.2")
    link_data = link_creation_info
else:
    link_data = t.cast(LinkInfoDTO820, link_creation_info)
    link_data.filter_synthesis = link_data.filter_synthesis or FilteringOptions.FILTER_SYNTHESIS
    link_data.filter_year_by_year = link_data.filter_year_by_year or FilteringOptions.FILTER_YEAR_BY_YEAR
```

MartinBelthle (changes requested):
Last round of changes :)
Also, rename your PR with a title fitting the commitlint for when we merge it

antarest/study/storage/variantstudy/model/command/create_link.py
TheoPascoli: Indeed it works

antarest/study/storage/variantstudy/model/command/create_link.py
TheoPascoli: it triggers a circular import

MartinBelthle: Arf okay so we can keep the code as it is

antarest/study/business/link_management.py
```diff
@@ -51,38 +56,63 @@ class LinkManager:
...
-                        style=links_config[link].get("link-style", "plain"),
-                    )
-                result.append(LinkInfoDTO(area1=area_id, area2=link, ui=ui_info))
+                link_properties = links_config[link]
+
+                link_creation_data: t.Dict[str, t.Any] = {"area1": area_id, "area2": link}
+                link_creation_data.update(link_properties)
+
+                link_data: LinkInfoDTOType
+                if int(study.version) < 820:
```
MartinBelthle: Due to this PR https://github.com/AntaresSimulatorTeam/AntaREST/pull/2156 we handle versions differently now. You should do this instead (this goes for every version check you did in this PR):
`StudyVersion.parse(study.version) < STUDY_VERSION_8_2`

antarest/study/business/link_management.py
```diff
@@ -12,31 +12,37 @@
...
 from antarest.study.storage.variantstudy.model.command.update_config import UpdateConfig
 
 _ALL_LINKS_PATH = "input/links"
 
 
-class LinkUIDTO(AntaresBaseModel):
-    color: str
-    width: float
-    style: str
+class LinkInfoDTOBase(AreaInfo, LinkInfoProperties):
```
sylvlecl, line 38: Class `AreaInfo` seems not used elsewhere, it should be defined here.
Not a big fan of using inheritance that much, areas could also got just in the definition of those 2 classes. But Ok to leave it this way.

sylvlecl, line 38: There is an issue with inheriting from `LinkInfoProperties`: we now use kebab case in the API DTOs, whereas we use camel case everywhere for this. We need to keep consistent with this.

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -0,0 +1,179 @@
...
+
+        expected = {
+            "area1": "area 1",
+            "area2": "area 2",
+            "asset-type": "ac",
+            "colorb": 112,
+            "colorg": 112,
+            "colorr": 112,
+            "display-comments": True,
+            "filter-synthesis": "hourly, daily, weekly, monthly, annual",
```
sylvlecl: We should use camel case in returned DTOs, for consistency sake with other endpoints

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -9,38 +9,76 @@
...
+    link_style: LinkStyle = LinkStyle.PLAIN
+
+    class Config:
+        alias_generator = to_kebab_case
+        populate_by_name = True
+
+
+class LinkInfoProperties820(LinkInfoProperties):
+    filter_synthesis: t.Optional[str] = None
+    filter_year_by_year: t.Optional[str] = None
```
sylvlecl: I think the best approach here, for both fields, would be to use an annotated type with a [custom serializer](https://docs.pydantic.dev/latest/concepts/serialization/#custom-serializers) and validation method.
This way, the fields can have their "real" type which is a `List[FilterOption]` while still being serialized and parsed as a `str`.

This will allow to re-use this annotated type in other places if needed.

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -213,8 +197,11 @@ def _apply(self, study_data: FileStudy) -> CommandOutput:
         area_from = data["area_from"]
         area_to = data["area_to"]
 
-        self.parameters = self.parameters or {}
-        link_property = CreateLink.generate_link_properties(self.parameters)
+        properties = LinkProperties.model_validate(self.parameters or {})
```
sylvlecl: I think here we should parse according to the version, instead of exluding "by hand" the fields later.
If we parse the right version, it will be serialized as expected for this study version.

antarest/study/business/link_management.py
```diff
@@ -51,38 +57,62 @@ class LinkManager:
...
+                raise LinkValidationError("Cannot specify a filter value for study's version earlier than v8.2")
+
+        link_info = link_creation_info.model_dump(exclude_none=True, by_alias=True)
+
+        link_data: LinkInfoDTOType
+        if StudyVersion.parse(study.version) < STUDY_VERSION_8_2:
+            link_data = LinkInfoDTOBase(**link_info)
+        else:
+            link_data = LinkInfoDTO820(**link_info)
+            if isinstance(link_data, LinkInfoDTO820) and link_data.filter_synthesis is None:
```
sylvlecl: we don't need to check the type here, we know it's an instancie of this

antarest/study/business/link_management.py
```diff
@@ -51,38 +57,62 @@ class LinkManager:
...
+                link_data.filter_synthesis = FilteringOptions.FILTER_SYNTHESIS
+            if isinstance(link_data, LinkInfoDTO820) and link_data.filter_year_by_year is None:
```
sylvlecl: Same, type check is redundant

antarest/study/business/link_management.py
```diff
@@ -51,38 +57,62 @@ class LinkManager:
...
 
         return result
 
-    def create_link(self, study: RawStudy, link_creation_info: LinkInfoDTO) -> LinkInfoDTO:
+    def create_link(self, study: RawStudy, link_creation_info: LinkInfoDTOType) -> LinkInfoDTOType:
+        if link_creation_info.area1 == link_creation_info.area2:
+            raise LinkValidationError(f"Cannot create link on same area: {link_creation_info.area1}")
+
+        if StudyVersion.parse(study.version) < STUDY_VERSION_8_2 and isinstance(link_creation_info, LinkInfoDTO820):
+            if link_creation_info.filter_synthesis is not None or link_creation_info.filter_year_by_year is not None:
```
sylvlecl: I think we should rely on pydantic here instead:
we parse the object depending on the version, pydantic knows what are the valid fields and which ones are not.

antarest/study/business/link_management.py
```diff
@@ -51,38 +57,62 @@ class LinkManager:
...
+
+        link_info = link_creation_info.model_dump(exclude_none=True, by_alias=True)
+
+        link_data: LinkInfoDTOType
+        if StudyVersion.parse(study.version) < STUDY_VERSION_8_2:
+            link_data = LinkInfoDTOBase(**link_info)
+        else:
+            link_data = LinkInfoDTO820(**link_info)
+            if isinstance(link_data, LinkInfoDTO820) and link_data.filter_synthesis is None:
+                link_data.filter_synthesis = FilteringOptions.FILTER_SYNTHESIS
```
sylvlecl: Shouldn't this be the default value in the class instead ?

sylvlecl (changes requested):
Some change requests, mainly:
- we should use camelCase in the API return, for consistency with other endpoints
- I think we can rely more on pydantic mechanisms for validation / default values of DTOs

antarest/study/business/link_management.py
TheoPascoli: The check is just to prevent a mypy error

antarest/study/storage/variantstudy/model/command/create_link.py
TheoPascoli: I added an annotated field that performs validation and serialization but the methods are not reusable

antarest/study/business/link_management.py
TheoPascoli: I overrided the config with camelcase in this class

TheoPascoli: I had troubles with some tests when removing the area value, especially commands tests. I left it that way

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -26,22 +44,76 @@
...
+    colorr: int = Field(default=DEFAULT_COLOR, gt=0, lt=255)
+    colorb: int = Field(default=DEFAULT_COLOR, gt=0, lt=255)
+    colorg: int = Field(default=DEFAULT_COLOR, gt=0, lt=255)
+    link_width: float = 1
+    link_style: LinkStyle = LinkStyle.PLAIN
+
+    model_config = ConfigDict(alias_generator=to_kebab_case, populate_by_name=True)
+
+
+comma_separated_enum_list = Annotated[
```
sylvlecl, line 102: Nice!
I think we should name it following type conventions, in CamelCase, for example `FilterOptionList`

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -26,22 +44,76 @@
...
+    t.List[FilterOption],
+    BeforeValidator(lambda x: validate_filters(x, FilterOption)),
+    PlainSerializer(lambda x: join_with_comma(x)),
```
sylvlecl, line 105: detail: I think the lambda is not needed:
```python
PlainSerializer(join_with_comma),
```

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -26,22 +44,76 @@
...
+    FilterOption.HOURLY,
+    FilterOption.DAILY,
+    FilterOption.WEEKLY,
+    FilterOption.MONTHLY,
+    FilterOption.ANNUAL,
+]
+
+
+def validate_filters(
+    filter_value: Union[List[FilterOption], str], enum_cls: t.Type[FilterOption]
```
sylvlecl, line 58: The function only works for FitlerOption class, so I don't think the second argument is useful ?
It would be if we wanted to use this function for other enum types, but here we have hard coded messages, so I don't think we will do it.

antarest/study/storage/variantstudy/model/command/create_link.py
sylvlecl: Also we can move it to the same module where `FilterOption` is defined, because it's likely that we will want to re-use it where `FilterOption` is used (for example in `LinkProperties` etc, but this can be done in a future work, I don't know if it will have more impacts).

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -26,22 +44,76 @@
...
+    colorg: int = Field(default=DEFAULT_COLOR, gt=0, lt=255)
+    link_width: float = 1
+    link_style: LinkStyle = LinkStyle.PLAIN
+
+    model_config = ConfigDict(alias_generator=to_kebab_case, populate_by_name=True)
+
+
+comma_separated_enum_list = Annotated[
+    t.List[FilterOption],
+    BeforeValidator(lambda x: validate_filters(x, FilterOption)),
```
sylvlecl, line 104: detail: if we remove the second argument, lambda is not needed anymore

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -26,22 +44,76 @@
 from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
 from antarest.study.storage.variantstudy.model.model import CommandDTO
 
+DEFAULT_COLOR = 112
+FILTER_VALUES: t.List[FilterOption] = [
```
sylvlecl, line 48: maybe `DEFAULT_FILTERS` for homogeneity with the above var? indeed it's only used as the default value

antarest/study/storage/variantstudy/model/command/create_link.py
sylvlecl: Note:
I found out there are similar methods in `antarest/study/storage/rawstudy/model/filesystem/config/field_validators.py`, we should try to clean this code duplication at some point (we can create a specific task for this if not in that PR)

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -98,51 +170,6 @@ def _create_link_in_config(self, area_from: str, area_to: str, study_data: FileS
             ],
         )
 
-    @staticmethod
-    def generate_link_properties(parameters: JSON) -> JSON:
```
sylvlecl, line 102: Great to have removed this :)

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -211,12 +229,18 @@ def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = N
...
         if not output.status:
             return output
+
+        properties: LinkInfoProperties
+        if StudyVersion.parse(version) >= STUDY_VERSION_8_2:
+            properties = LinkInfoProperties820.model_validate(self.parameters or {})
+        else:
+            properties = LinkInfoProperties.model_validate(self.parameters or {})
+
+        link_property = properties.model_dump(mode="json", exclude={"area1", "area2"}, by_alias=True, exclude_none=True)
```
sylvlecl, line 239: The excludes are not needed I think, the properties classes don't have information about areas

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -211,12 +229,18 @@ def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = N
         output, data = self._apply_config(study_data.config)
         if not output.status:
             return output
+
+        properties: LinkInfoProperties
+        if StudyVersion.parse(version) >= STUDY_VERSION_8_2:
+            properties = LinkInfoProperties820.model_validate(self.parameters or {})
```
sylvlecl, line 235: Taking a step back about the architecture of our classes:
our logic is wrong here, the parameters were created by dumping a `LinkInfoDTO`, but we read it as a `LinkInfoProperties`. There is a kind of implicit conversion.

Instead, we should read it again as a DTO, and have an explicit conversion function to get a "properties" class that will be targetted at being serialized in the INI file (even if it only does a modeul_dump/model_validate to achieve that conversion).

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -211,12 +229,18 @@ def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = N
...
+
+        link_property = properties.model_dump(mode="json", exclude={"area1", "area2"}, by_alias=True, exclude_none=True)
+
         area_from = data["area_from"]
         area_to = data["area_to"]
 
-        self.parameters = self.parameters or {}
-        link_property = CreateLink.generate_link_properties(self.parameters)
-
         study_data.tree.save(link_property, ["input", "links", area_from, "properties", area_to])
```
sylvlecl, line 244: Again, taking a step back:
here we are serializing basically the content of a `LinkInfoProperties` into an ini file.
This seems like a duplication of the role of the `LinkProperties` class in `antarest/study/storage/rawstudy/model/filesystem/config/links.py`.

It looks like maybe the new classes you have created should go there and replace the old one.
OR, we should have a conversion function from the new classes to this one, and use it for serialization.

**I am not sure of the best choice right now, we need to think about this from a general point of view, so that we use common principles everywhere.** 
I think we should use this work as an opportunity to clarify and agree about this design of classes with all the team.

sylvlecl (changes requested):
Sorry, further comments and thoughts, this work makes me realize we should clarify the architecture of our classes to have a common understanding (*Properties, *DTO, their roles, ...).

antarest/study/web/study_data_blueprint.py
```diff
@@ -151,15 +151,14 @@ def get_areas(
     )
     def get_links(
         uuid: str,
-        with_ui: bool = False,
         current_user: JWTUser = Depends(auth.get_current_user),
     ) -> t.Any:
```
skamril, line 155: `t.List[LinkInfoDTO]` instead?

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -26,22 +44,76 @@
...
+class AreaInfo(AntaresBaseModel):
+    area1: str
+    area2: str
+
+
+class LinkInfoProperties(AntaresBaseModel):
+    hurdles_cost: bool = False
+    loop_flow: bool = False
+    use_phase_shifter: bool = False
+    transmission_capacities: TransmissionCapacity = TransmissionCapacity.ENABLED
```
skamril, line 90: `transmission_capacity` in singular (there is an another enum named TransmissionCapacities)

MartinBelthle: The issue is the field is written `transmission-capacities` inside the ini file :/ 
It's a shame that there's another field named the same way inside the generaldata.ini file ...

PR #2147, opened 2024-09-18
feat(debug): allow files and folder deletion inside debug view

MartinBelthle:
Introduces a new endpoint inside the back-end to fix [ANT-2083]
Samir will also add the front-end part

*NB*: I didn't specify inside the endpoint signature that it was for the User folder because I think we might use it for other folders such as the one for .yaml models that comes with Andromede modeling

### DIFF
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -35,6 +35,7 @@ from antarest.core.exceptions import (
     BadEditInstructionException,
     ChildNotFoundError,
     CommandApplicationError,
+    FileDeletionNotAllowed,
     IncorrectPathError,
     NotAManagedStudyException,
     ReferencedObjectDeletionNotAllowed,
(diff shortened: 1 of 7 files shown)


antarest/study/service.py
```diff
@@ -2640,3 +2641,37 @@ def asserts_no_thermal_in_binding_constraints(
...
+        Raises:
+            FileDeletionNotAllowed: if the path does not comply with the above rules
+        """
+        study = self.get_study(study_id)
+        assert_permission(current_user, study, StudyPermissionType.WRITE)
+
+        url = [item for item in path.split("/") if item]
+        if len(url) < 2 or url[0] != "user":
+            raise FileDeletionNotAllowed(f"the targeted data isn't inside the 'User' folder: {path}")
+        if url[1] == "expansion":
```
sylvlecl: We should probablu used registered files instead of hard coding it here

antarest/study/service.py
```diff
@@ -2640,3 +2641,37 @@ def asserts_no_thermal_in_binding_constraints(
...
+
+        study_tree = self.storage_service.raw_study_service.get_raw(study, True).tree.build()
+        try:
+            study_tree["user"].delete(url[1:])
+        except ChildNotFoundError as e:
+            raise FileDeletionNotAllowed(f"the given path doesn't exist: {e.detail}")
+
+        # update cache
+        cache_id = f"{CacheConstants.RAW_STUDY}/{study.id}"
+        updated_tree = self.storage_service.raw_study_service.get_raw(study, False).tree.get()
```
sylvlecl: Why not update with just `study_tree.get()`, I think it should be up to date ?

Or at least use `use_cache=True`, the config is not modified by deletions in user directory so we don't need to rebuild it from scratch.

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -191,6 +191,21 @@ def get_study(
...
+        summary="Delete files or folders located inside the 'User' folder",
+        response_model=None,
+    )
+    def delete_file(
+        uuid: str,
+        path: str = Param("/", examples=["user/wind_solar/synthesis_windSolar.xlsx"]),  # type: ignore
+        current_user: JWTUser = Depends(auth.get_current_user),
+    ) -> t.Any:
+        uuid = sanitize_uuid(uuid)
+        logger.info(f"Deleting some data for study {uuid}", extra={"user": current_user.id})
```
sylvlecl: maybe log the path instead of "some data"

antarest/study/service.py
```diff
@@ -2640,3 +2641,37 @@ def asserts_no_thermal_in_binding_constraints(
...
+        if len(url) < 2 or url[0] != "user":
+            raise FileDeletionNotAllowed(f"the targeted data isn't inside the 'User' folder: {path}")
+        if url[1] == "expansion":
+            raise FileDeletionNotAllowed(f"you cannot delete a file/folder inside 'expansion' folder : {path}")
+
+        study_tree = self.storage_service.raw_study_service.get_raw(study, True).tree.build()
+        try:
+            study_tree["user"].delete(url[1:])
+        except ChildNotFoundError as e:
+            raise FileDeletionNotAllowed(f"the given path doesn't exist: {e.detail}")
```
sylvlecl: We should use "raise from e" (so that e is correctly identified as the direct cause)

sylvlecl (changes requested):
A few change requests

antarest/study/service.py
MartinBelthle: Indeed

antarest/study/service.py
MartinBelthle: I used `RegisteredFile`.
IMO it doesn't bring a lot as the behavior is still a bit hard coded but it has the merit of using the class and showing how for future devs.

PR #2149, opened 2024-09-19
ci(tests): reduce number of workers for tests

sylvlecl:
The use of as many workers as logical cores
seems to cause significant increases of some
monitored durations, including on Linux, which
causes irrelevant CI failures.

We revert to using as many workers as actual CPUs.

Also, load balancing of tests is improved using worksteal,
which works far better with our tests which have very different
execution times (integration tests are far longer than unit tests).


MartinBelthle:
Nice the `worksteal` arg

MartinBelthle (approved):

PR #2150, opened 2024-09-19
chore(license): add a new ESLint rule to check license header

skamril:


MartinBelthle (dismissed):
I approve as it seems okay even though I don't really understand everything. I'll let Hatim decide

webapp/.eslintrc.cjs
```diff
@@ -12,10 +12,17 @@ module.exports = {
     "plugin:react/jsx-runtime",
     "plugin:react-hooks/recommended",
     "plugin:jsdoc/recommended-typescript",
-    "plugin:prettier/recommended",
+    "plugin:prettier/recommended", // Must be the last one
```
hdinia, line 15: what you mean, in order it should be the last ?

hdinia (changes requested):
could you also commit the `license-header.js` file ?

MartinBelthle:
@hdinia I already merged the `license-header.js` file on dev in the last commit

webapp/.eslintrc.cjs
skamril: cf. https://github.com/prettier/eslint-plugin-prettier?tab=readme-ov-file#configuration-legacy-eslintrc

hdinia:
> @hdinia I already merged the `license-header.js` file on dev in the last commit

ok nice then

hdinia (approved):

PR #2151, opened 2024-09-20
feat(aggregation-api): delete index from the response file 

mabw-rte:
The scripts team want to exclude the index from the response file of an aggregation api. This is because, they get an `unnamed` column that contains the index which will not be processed.

 

NB:

We explained that if the front comes to use these endpoints, there may be problem because we need (at least for now, pending upcoming updates) to feed the index to the front.


MartinBelthle (approved):

MartinBelthle:
Just rename your PR to fit with the commitlint

tests/integration/raw_studies_blueprint/test_aggregate_raw_data.py
```diff
@@ -603,8 +601,7 @@ def test_empty_columns(self, client: TestClient, user_access_token: str, interna
             },
         )
         assert res.status_code == 200, res.json()
-        df = pd.read_csv(io.BytesIO(res.content), index_col=0, sep=",")
-        assert df.empty
+        assert res.content == b"\n"
```
MartinBelthle: Fix this to be OS agnostic

MartinBelthle (changes requested):

MartinBelthle (approved):

PR #2156, opened 2024-09-24
feat(version): use class StudyVersion to handle versions

MartinBelthle:
This PR uses the class `StudyVersion` defined inside the `antares-study-version` package to handle study versions widely inside the code.

It allows users to import studies in v9.0.
It writes the version as 9.0 in the database (compared to 880 before)


MartinBelthle:
*NB:* Use StudyVersion eveywhere instead of comparing with `int` or `str`

antarest/launcher/service.py
```diff
@@ -228,7 +228,7 @@ def run_study(
         job_uuid = self._generate_new_id()
         logger.info(f"New study launch (study={study_uuid}, job_id={job_uuid})")
         study_info = self.study_service.get_study_information(uuid=study_uuid, params=params)
-        solver_version = study_version or study_info.version
+        solver_version = study_version or study_info.version.__format__(format_spec="ddd")
```
sylvlecl: I think `__format__` should not be used directly, but rather through the usual formatting syntaxes like `f"{study_info.version:ddd}"`

antarest/study/model.py
```diff
@@ -43,26 +44,44 @@
...
-    "810": "empty_study_810.zip",
-    "820": "empty_study_820.zip",
-    "830": "empty_study_830.zip",
-    "840": "empty_study_840.zip",
-    "850": "empty_study_850.zip",
-    "860": "empty_study_860.zip",
-    "870": "empty_study_870.zip",
-    "880": "empty_study_880.zip",
+NEW_DEFAULT_STUDY_VERSION: StudyVersion = StudyVersion.parse(880)
+STUDY_VERSION_600 = StudyVersion.parse("600")
```
sylvlecl: I would propose to:
- respect the dots also in variables names for example with underscors
- remove the patch version which is not useful
- use a true version string in the right hand side

Otherwise we will not be able to keep the same syntax for longer versions (8.10 ...).
This will make clear that study versions are only major.minor versions

```python
STUDY_VERSION_6_1 = StudyVersion.parse("6.1")
```

antarest/study/storage/matrix_profile.py
```diff
@@ -191,11 +193,11 @@ def adjust_matrix_columns_index(
         study_version: The version of the study.
     """
     # Get the matrix profiles for a given study version
-    if study_version < 820:
+    if study_version < STUDY_VERSION_820:
         matrix_profiles = _SPECIFIC_MATRICES
-    elif study_version < 860:
+    elif study_version < STUDY_VERSION_860:
         matrix_profiles = _SPECIFIC_MATRICES_820
```
sylvlecl: We could rename those also to _SPECIFIC_MATRICES_8_2

antarest/launcher/service.py
```diff
@@ -252,7 +252,7 @@ def run_study(
         self.launchers[launcher].run_study(
             study_uuid,
             job_uuid,
-            str(solver_version),
+            solver_version,
```
sylvlecl, line 256: I think here we should change the signature to accept a `SolverVersion` object, to clarify what kind of version it is

antarest/study/model.py
MartinBelthle: Indeed it's better

antarest/launcher/service.py
MartinBelthle: I used `SolverVersion` and ultimately made a if/else to parse it as a string inside the `study.antares` file

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -644,11 +645,15 @@ def kill_job(self, job_id: str, dispatch: bool = True) -> None:
...
+def _override_solver_version(study_path: Path, version: SolverVersion) -> None:
     study_info_path = study_path / "study.antares"
     study_info = IniReader().read(study_info_path)
     if "antares" in study_info:
-        study_info["antares"]["solver_version"] = version
+        if version.major < 9:  # should be written as XYZ
+            version_to_write = f"{version:ddd}"
+        else:  # should be written as X.Y
+            version_to_write = f"{version:2d}"
+        study_info["antares"]["solver_version"] = version_to_write
```
sylvlecl, line 656: We need an evolution in antares-launcher to handle this correctly

sylvlecl (dismissed):

PR #2157, opened 2024-09-25
feature(installer): update installer version and improve desktop version launcher

sylvlecl:
- updates installer submodule to benefit from last fixes and v2.18 handling
 - make application exit when the server process ends, instead of hanging
 - detect when server is already running
 - separate one popup for "starting" and another for "started"
 - fix alembic handling in pyinstaller:
   alembic must not be defined as a script, otherwise it's executed when the application
   quits. Instead, it must just be analyzed separately to get its dependencies


.github/workflows/deploy.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - "master"
       - "hotfix/**"
+      - "feature/update-installer"
```
MartinBelthle: Should be removed or we could add more generic branch name such as `installer/**` or `deploy/**`

AntaresWebLinux.spec
```diff
@@ -1,25 +1,27 @@
...
-from PyInstaller.utils.hooks import collect_dynamic_libs
 
 block_cipher = None
 
-# We need to analyze all alembic files to be sure the migration phase works fine
-migrations_dir = Path('alembic/versions')
-migration_files = [str(f) for f in migrations_dir.iterdir() if f.is_file() and f.suffix == '.py']
+# We need to analyze all alembic files to be sure the migration phase works fine:
+# alembic loads version files by their path, so we need to add them as "data" to the package,
+# but all the dependencies they use need to be included also, wo we need to perform a
```
MartinBelthle, line 8: typo: so we need

AntaresWebLinux.spec
```diff
@@ -1,25 +1,27 @@
...
 
-binaries = [('./alembic.ini', './alembic.ini')]  + collect_dynamic_libs('tables')
-
-antares_web_server_a = Analysis(['antarest/gui.py', 'alembic/env.py'] + migration_files,
+antares_web_server_a = Analysis(['antarest/gui.py'],
              pathex=[],
-             binaries=binaries,
-             datas=[('./resources', './resources'), ('./alembic', './alembic')],
+             binaries=[],
+             datas=[('./resources', './resources'), ('./alembic', './alembic'), ('./alembic.ini', './')],
```
MartinBelthle, line 17: Why did you add `./` ?

AntaresWebWin.spec
```diff
@@ -1,25 +1,27 @@
...
-from PyInstaller.utils.hooks import collect_dynamic_libs
 
 block_cipher = None
 
-# We need to analyze all alembic files to be sure the migration phase works fine
-migrations_dir = Path('alembic/versions')
-migration_files = [str(f) for f in migrations_dir.iterdir() if f.is_file() and f.suffix == '.py']
+# We need to analyze all alembic files to be sure the migration phase works fine:
+# alembic loads version files by their path, so we need to add them as "data" to the package,
+# but all the dependencies they use need to be included also, wo we need to perform a
```
MartinBelthle, line 8: same typo

antarest/gui.py
```diff
@@ -39,67 +41,139 @@ def run_server(config_file: Path) -> None:
...
+    open_app_action = menu.addAction("Open application")
+    assert open_app_action is not None
+    open_app_action.triggered.connect(open_app)
+    # To quit the app
+    quit_action = menu.addAction("Quit")
+    assert quit_action is not None
+    quit_action.triggered.connect(app.quit)
+
+    # Adding options to the System Tray
+    def handle_action(reason: int) -> None:
```
MartinBelthle: This function/design is really weird but I'm far from understanding this file

MartinBelthle (dismissed):
The `gui.py` file is really hard to understand IMO.
I wouldn't mind more doc cause I'm not familiar with this code :smiley:

AntaresWebLinux.spec
sylvlecl: Before we had this: `binaries = [('./alembic.ini', './alembic.ini')]` which created a folder called "alembic.ini" in which the file was placed.

Now the file is correctly placed at the root

PR #2164, opened 2024-09-30
build(python): bump python version to use v3.11

MartinBelthle:
List of changes that were made:

- Run the CI with Python 3.11
- Update project requirements to work with Python 3.11 (pandas 2.2.3, numpy 1.26.4, psycopg2 2.9.9, pyinstaller 6.10.0)
- Use class `enum.StrEnum` instead of making a class inherit from `str` and `enum.Enum`
- Use `model_dump` with `mode='json'` everywhere. Probably not needed but more secure.
- Fix miscellaneous issues

:warning: Due to `enum.StrEnum` (and probably the requirements) the code is no longer compatible with Python versions prior to v3.11

:warning: This PR shouldn't be merged before it's tested with a branch on antares-ci that deploys Python3.11 on the environments.

:information_source: Every Desktop version has been tested and works :)


.github/workflows/deploy.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - "master"
       - "hotfix/**"
+      - "bump-python/**"
```
MartinBelthle, line 7: Remove when the PR is approved

.github/workflows/worker.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - "master"
       - "worker/**"
+      - "bump-python/**"
```
MartinBelthle, line 7: Same remark

PR #2165, opened 2024-10-01
perf(scripts): improve load balancing

mabw-rte:
Context:
@sylvlecl reported that when `gunicorn`, is wrapped over an ASGI, it does not perform the load balancing as it should.

Solution:
Mount the load balancing to `nginx`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,6 @@ services:
       - ./resources/deploy/logs:/logs
       - ./resources/deploy/gunicorn.py:/conf/gunicorn.py
       - ./antares-8.2.2-Ubuntu-20.04/bin:/antares_simulator
-    command: ["./scripts/start.sh", "--no-gunicorn"]
   antares-antarest-watcher:
     image: antarest:latest
     container_name: antarest-watcher
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,7 @@ services:
       - ./resources/deploy/logs:/logs
       - ./resources/deploy/gunicorn.py:/conf/gunicorn.py
       - ./antares-8.2.2-Ubuntu-20.04/bin:/antares_simulator
+    command: ./scripts/start.sh --no-gunicorn
   antares-antarest-watcher:
     image: antarest:latest
     container_name: antarest-watcher
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       - ./resources/deploy/logs:/logs
       - ./resources/deploy/gunicorn.py:/conf/gunicorn.py
       - ./antares-8.2.2-Ubuntu-20.04/bin:/antares_simulator
-    command: ./scripts/start.sh --no-gunicorn
+    command: ["./scripts/start.sh", "--no-gunicorn"]
   antares-antarest-watcher:
     image: antarest:latest
     container_name: antarest-watcher
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -9,7 +9,7 @@ min() {
     echo $(( $1 < $2 ? $1 : $2 ))
 }
 
-workers=$(min 30 ${NB_WORKERS_ANTARES:-17}) # default to 17 and max is 30
+workers=$(min 30 ${ANTARES_NB_WORKERS:-$((2*$(nproc) + 1))}) # default (2*nproc + 1) and max is 30
 
 # Check for --no-gunicorn or --multiple-ports argument
 use_gunicorn=true
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -11,18 +11,31 @@ min() {
 
 workers=$(min 30 ${NB_WORKERS_ANTARES:-17}) # default to 17 and max is 30
 
+# Check for --no-gunicorn or --multiple-ports argument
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -154,4 +154,8 @@ resources/commit_id
 watcher
 *.lock
 
-**/bucket
\ No newline at end of file
+**/bucket
+
+# local scripts
```
MartinBelthle: Remove this

scripts/start.sh
```diff
@@ -5,9 +5,24 @@ set -e
 CUR_DIR=$(cd "$(dirname "$0")" && pwd)
 BASE_DIR=$(dirname "$CUR_DIR")
 
+min() {
+    echo $(( $1 < $2 ? $1 : $2 ))
+}
+
+workers=$(min 30 ${NB_WORKERS_ANTARES:-17}) # default to 17 and max is 30
```
MartinBelthle: Did not really follow the subject but why did we choose these values ?

scripts/start.sh
```diff
@@ -5,9 +5,24 @@ set -e
...
+
 if [ -z "$1" ] ; then
   sh $CUR_DIR/pre-start.sh
-  gunicorn --config $BASE_DIR/conf/gunicorn.py --worker-class=uvicorn.workers.UvicornWorker antarest.wsgi:app
+  pids=() # Initialize empty array to store background process IDs
+  for ((i=0; i<workers; i++))
+  do
+    uvicorn antarest.wsgi:app --host 0.0.0.0 --port $((5000 + $i)) --log-level info --timeout-keep-alive 600 &
```
MartinBelthle, line 51: Again not really familiar with the subject but why do we have different ports ? Also shouldn't we lower the tiemout-keep-alive ?

MartinBelthle (commented):
I'll let Sylvain handle this review as he knows the subject better

scripts/start.sh
mabw-rte: - The default `17` is when you don't provide `NB_WORKERS_ANTARES` in your docker-compose env vars
- In order to have a reserved range of ports for `antarest`, we need to have a maximum of reserved ports on your server. It follows that you can't have more workers than ports.

scripts/start.sh
mabw-rte: we want each `uvicorn` to run in a different process with its own port to be considered as a separate worker.

scripts/start.sh
sylvlecl: I would prefer to not hard code here the 17 because it's related to our hardware.

We could use `nproc` to keep the same logic as before (2*nproc + 1), or just throw an error if the variable does not exist.

scripts/start.sh
sylvlecl: keep alive was one day with gunicorn, seems very high compared to standard values but we could keep it for now.

Using different ports is the goal of this PR, it will allow an external component to do the load balancing for us (nginx in our case).

scripts/start.sh
```diff
@@ -5,9 +5,24 @@ set -e
 CUR_DIR=$(cd "$(dirname "$0")" && pwd)
 BASE_DIR=$(dirname "$CUR_DIR")
 
+min() {
```
sylvlecl, line 27: From a global point of view, I think we should keep the possibility to use gunicorn.
It stays a legitimate tool, even if we change our internal deployment to not use it anymore.

This will also allow us to transition more easily between the 2 kind of deployments: we will be able to configure in deployment what kind of load balancing we use.

In order to achieve this, we could add an argument to the script, or an env variable, like `--no-gunicorn` or `--multiple-ports` which would trigger the new behaviour.

Also we need to check if the README and other parts of the developer documentation needs to be updated, as well as the provided `docker-compose.yaml` in the root directory.

scripts/start.sh
mabw-rte: @sylvlecl  
I don't know if the value `(2*nproc + 1)` would be the same on deployment server and `CI/CD` pipeline server. 
So it is better to hard code it.

scripts/start.sh
```diff
@@ -5,9 +5,37 @@ set -e
...
+    for ((i=0; i<workers; i++))
+    do
+      uvicorn antarest.wsgi:app --host 0.0.0.0 --port $((5000 + $i)) --log-level info --timeout-keep-alive 600 &
+      pids+=($!) # Store background process IDs
+    done
+    for pid in ${pids[*]};
+    do
+      wait $pid # Wait for each background process to finish
+    done
 else
```
sylvlecl, line 58: I think a `fi` is missing here, the server fails to start !

docker-compose.yml
```diff
@@ -16,6 +16,7 @@ services:
       - ./resources/deploy/logs:/logs
       - ./resources/deploy/gunicorn.py:/conf/gunicorn.py
       - ./antares-8.2.2-Ubuntu-20.04/bin:/antares_simulator
+    command: ["./scripts/start.sh", "--no-gunicorn"]
```
sylvlecl: Note: here it will be better to stick with the gunicorn version, because we don't provide the nginx config to do the load balancing (and it's ok, let's keep it simple here).

sylvlecl (approved):

PR #2166, opened 2024-10-01
fix(pydantic): allow `str` fields to be populated by `int`

MartinBelthle:
Fix [ANT-2255]

1- Introduces a class `BaseModelInHouse` to allow coercing and use it everywhere in the code
2- Little refactor to avoid circular import
3- Add test case that used to fail: Creation of a thermal cluster with the name 1111

### DIFF
--- a/antarest/core/utils/utils.py
+++ b/antarest/core/utils/utils.py
@@ -26,6 +26,7 @@ from pathlib import Path
 import py7zr
 import redis
 from fastapi import HTTPException
+from pydantic import BaseModel, ConfigDict
 
 from antarest.core.config import RedisConfig
 from antarest.core.exceptions import ShouldNotHappenException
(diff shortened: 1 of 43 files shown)


antarest/core/utils/utils.py
```diff
@@ -24,17 +24,31 @@
...
 
-from antarest.core.config import RedisConfig
 from antarest.core.exceptions import ShouldNotHappenException
 
 logger = logging.getLogger(__name__)
 
 UUID_PATTERN = re.compile("^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$")
 
 
+class BaseModelInHouse(BaseModel):
```
sylvlecl: Proposition: renaming to AntaresBaseModel and move it to `antares/core/serialization`

sylvlecl (dismissed):

PR #2169, opened 2024-10-03
fix(snapshot_cleaning): set `ref_id` to `None` to prevent postgresql crash

maugde:
Fix bug on the clear-snapshot endpoint


sylvlecl:
We should understand why the foreign key issue did not cause an exception with sqlite, and if we can configure it to have the same behaviour as postgresql, otherwise we will have the same problem in the future.

PR #2172, opened 2024-10-07
fix: enable foreign keys for sqlite

sylvlecl:
By default, sqlite does not enforce foreign key constraints, this needs to be enabled explicitly.

This is in particular important to catch those errors in our unit tests, but also for 
desktop mode.

**Note**
I have checked that indeed, tests before [this fix](https://github.com/AntaresSimulatorTeam/AntaREST/pull/2169) did not pass.

### DIFF
--- a/antarest/core/utils/fastapi_sqlalchemy/middleware.py
+++ b/antarest/core/utils/fastapi_sqlalchemy/middleware.py
@@ -1,7 +1,7 @@
 from contextvars import ContextVar, Token
 from typing import Any, Dict, Optional, Type, Union
 
-from sqlalchemy import create_engine  # type: ignore
+from sqlalchemy import create_engine, event  # type: ignore
 from sqlalchemy.engine import Engine  # type: ignore
 from sqlalchemy.engine.url import URL  # type: ignore
 from sqlalchemy.orm import Session, sessionmaker  # type: ignore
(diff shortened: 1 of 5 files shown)


antarest/core/utils/fastapi_sqlalchemy/middleware.py
```diff
@@ -16,6 +16,24 @@
 _session: ContextVar[Optional[Session]] = ContextVar("_session", default=None)
 
 
+def _is_sqlite_connection(dbapi_connection: Any) -> bool:
+    cls = dbapi_connection.__class__
+    full_classe_name = f"{cls.__module__}{cls.__name__}"
+    return "sqlite" in full_classe_name.lower()
```
MartinBelthle: I imagine there's not a way to do this in a cleaner way (check of class instance) :/ ?

MartinBelthle (dismissed):

antarest/core/utils/fastapi_sqlalchemy/middleware.py
sylvlecl: Check of class instance will be more fragile, this is kind of an implementation detail of sqlalchemy driver, I prefer being robust to any implementation change.
It's unlikely that another driver will have sqlite in its name.

MartinBelthle: Ok I see

sylvlecl: I changed the logic to depend on the driver name instead, it should be a little  more reliable than class names.

PR #2173, opened 2024-10-07
feat: increase cleaning snapshot frequency

### LINKED ISSUE #1717: Variant snapshots should be cleaned way more often
## Problem Statement

Currently, the function `clear_snapshot` which removes the `snapshot` directory (it contains the real files of a variant study) is only called in the `auto_archive_service.py` module. 

Therefore, they are only cleaned once a while according to the field `auto_archive_threshold_days` : 60 days by default and 120 days for recette env. !!

## Proposed Solution

Create a new field in the `application.yaml` which concerns only variants with a default value of something like 1 day.

maugde:
PR for the [ANT-1952](https://gopro-tickets.rte-france.com/browse/ANT-1952) ticket.

- add the option `snapshot_lifespan` in both configuration handlers and `CONFIG.md`
- add test related to this operation

### DIFF
--- a/antarest/core/config.py
+++ b/antarest/core/config.py
@@ -155,6 +155,7 @@ class StorageConfig:
     auto_archive_dry_run: bool = False
     auto_archive_sleeping_time: int = 3600
     auto_archive_max_parallel: int = 5
+    variant_snapshot_lifespan_days: int = 1
 
     @classmethod
     def from_dict(cls, data: JSON) -> "StorageConfig":
@@ -184,6 +185,10 @@ class StorageConfig:
             auto_archive_dry_run=data.get("auto_archive_dry_run", defaults.auto_archive_dry_run),
             auto_archive_sleeping_time=data.get("auto_archive_sleeping_time", defaults.auto_archive_sleeping_time),
             auto_archive_max_parallel=data.get("auto_archive_max_parallel", defaults.auto_archive_max_parallel),
+            variant_snapshot_lifespan_days=data.get(
+                "snapshot_variant_lifespan_days",
+                defaults.variant_snapshot_lifespan_days,
+            ),
         )
 
 
--- a/antarest/core/config.py
+++ b/antarest/core/config.py
@@ -155,7 +155,7 @@ class StorageConfig:
     auto_archive_dry_run: bool = False
     auto_archive_sleeping_time: int = 3600
     auto_archive_max_parallel: int = 5
-    variant_snapshot_lifespan_days: int = 1
+    snapshot_lifespan: int = 1
 
     @classmethod
     def from_dict(cls, data: JSON) -> "StorageConfig":
@@ -185,9 +185,9 @@ class StorageConfig:
             auto_archive_dry_run=data.get("auto_archive_dry_run", defaults.auto_archive_dry_run),
             auto_archive_sleeping_time=data.get("auto_archive_sleeping_time", defaults.auto_archive_sleeping_time),
             auto_archive_max_parallel=data.get("auto_archive_max_parallel", defaults.auto_archive_max_parallel),
-            variant_snapshot_lifespan_days=data.get(
+            snapshot_lifespan=data.get(
                 "snapshot_variant_lifespan_days",
-                defaults.variant_snapshot_lifespan_days,
+                defaults.snapshot_lifespan,
             ),
         )
 
--- a/antarest/core/config.py
+++ b/antarest/core/config.py
@@ -155,7 +155,7 @@ class StorageConfig:
     auto_archive_dry_run: bool = False
     auto_archive_sleeping_time: int = 3600
     auto_archive_max_parallel: int = 5
-    snapshot_lifespan: int = 1
+    snapshot_retention_days: int = 1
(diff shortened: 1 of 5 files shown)


antarest/core/config.py
```diff
@@ -155,6 +155,7 @@ class StorageConfig:
     auto_archive_dry_run: bool = False
     auto_archive_sleeping_time: int = 3600
     auto_archive_max_parallel: int = 5
+    snapshot_lifespan: int = 1
```
sylvlecl: lifespan sounds relevant, but I would prefer to use "retention" for consistency sake with the previous evolution on snapshots cleaning.
Also, because here we only have an integer, we don't know what's the unit, so we can postfix it with "days" too: `snapshot_retention_days`

antarest/study/storage/auto_archive_service.py
```diff
@@ -84,10 +84,19 @@ def _try_archive_studies(self) -> None:
                     exc_info=e,
                 )
 
+    def _try_clear_snapshot(self) -> None:
+        # get retention hours from number of days
+        hours_of_lifespan = self.config.storage.snapshot_lifespan * 24
+
+        self.study_service.storage_service.variant_study_service.clear_all_snapshots(
+            datetime.timedelta(hours=hours_of_lifespan)
```
sylvlecl: here you can use `days=...` to avoid the computation above

docs/install/1-CONFIG.md
```diff
@@ -304,6 +304,12 @@ default:
 - **Default value:** 5
 - **Description:** Max auto archival tasks in parallel.
 
+## **snapshot_lifespan**
+
+- **Type:** Integer
```
sylvlecl, line 309: Thanks for the doc :)

sylvlecl (changes requested):
Just renaming suggestion for consistency sake with the cleaning endpoint we added recently

antarest/study/storage/auto_archive_service.py
```diff
@@ -84,10 +84,16 @@ def _try_archive_studies(self) -> None:
...
+    def _try_clear_snapshot(self) -> None:
+        self.study_service.storage_service.variant_study_service.clear_all_snapshots(
+            datetime.timedelta(days=self.config.storage.snapshot_retention_days)
+        )
+
     def _loop(self) -> None:
         while True:
             try:
                 self._try_archive_studies()
+                self._try_clear_snapshot()
```
MartinBelthle: Since you've added this you can change the code in the same file line 45 to 78 to only check for raw studies as variants will be cleaned way more often with your method. And it will simplify the code

antarest/core/config.py
```diff
@@ -184,6 +185,10 @@ def from_dict(cls, data: JSON) -> "StorageConfig":
             auto_archive_dry_run=data.get("auto_archive_dry_run", defaults.auto_archive_dry_run),
             auto_archive_sleeping_time=data.get("auto_archive_sleeping_time", defaults.auto_archive_sleeping_time),
             auto_archive_max_parallel=data.get("auto_archive_max_parallel", defaults.auto_archive_max_parallel),
+            snapshot_retention_days=data.get(
+                "snapshot_variant_lifespan_days",
```
MartinBelthle: This name is not the one you chose so it will never choose a different value than the default one. Plz change this

MartinBelthle (changes requested):
Just little changes but seems good to me

MartinBelthle:
Also don't forget to change the name when you'll merge your PR to respect the commit-lint

antarest/core/config.py
sylvlecl: Actually I see that we don't have unit tests for configuration parsing, maybe we could add some ?

antarest/study/storage/auto_archive_service.py
```diff
@@ -84,6 +85,10 @@ def _try_archive_studies(self) -> None:
                     exc_info=e,
                 )
 
+        self.study_service.storage_service.variant_study_service.clear_all_snapshots(
```
MartinBelthle, line 88: I preferred the old way when you called this inside a separate method but it's not that important.

Also, it's not inside this PR but as you call this function you should change the logs I believe as logging 
`f"Start cleaning all snapshots updated or accessed {self._retention_hours} hours ago."` will log it in a weird way, (see __str__() method of timedelta class.). I think you should do `self._retention_hours.total_seconds() // 3600`. This goes for every log you've introduced

MartinBelthle (changes requested):

antarest/study/storage/auto_archive_service.py
MartinBelthle: Nevermind we have to keep this to archive outputs

antarest/study/storage/auto_archive_service.py
maugde: I'll improve logs on [this branch](https://github.com/AntaresSimulatorTeam/AntaREST/tree/fix/improve-duration-in-logs-of-snapshot-tasks).

sylvlecl (approved):

PR #2175, opened 2024-10-08
feat(link): add update endpoint for link

TheoPascoli:
Add an endpoint to update link


antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -73,11 +104,134 @@ def validate_series(
...
+                UpdateConfig(
+                    target=f"input/links/{area_from}/properties/{area_to}",
+                    data=link_property,
+                    command_context=self.command_context,
+                )
+            )
+        if self.series != other.series:
+            commands.append(
+                ReplaceMatrix(
+                    target=f"@links_series/{area_from}/{area_to}",
```
MartinBelthle, line 118: What is this path ?

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -73,11 +104,134 @@ def validate_series(
...
+                    matrix=strip_matrix_protocol(other.series),
+                    command_context=self.command_context,
+                )
+            )
+        return commands
+
+    def get_inner_matrices(self) -> List[str]:
+        list_matrices = []
+        if self.series:
```
MartinBelthle: You could do a for loop based on `getattr`, see `create_binding_constraints.py` to simplify the code

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -73,11 +104,134 @@ def validate_series(
...
             raise ValueError("Cannot create link on same node")
         return self
 
+    def to_dto(self) -> CommandDTO:
+        args = {
+            "area1": self.area1,
+            "area2": self.area2,
+            "parameters": self.parameters,
+        }
+        if self.series:
```
MartinBelthle: Could use `getattr`

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -73,11 +104,134 @@ def validate_series(
...
+        if self.indirect:
+            assert_this(isinstance(self.indirect, str))
+            list_matrices.append(strip_matrix_protocol(self.indirect))
+        return list_matrices
+
+    def save_series(self, area_from: str, area_to: str, study_data: FileStudy, version: StudyVersion) -> None:
+        self.series = self.series or (self.command_context.generator_matrix_constants.get_link(version=version))
+        self.direct = self.direct or (self.command_context.generator_matrix_constants.get_link_direct())
+        self.indirect = self.indirect or (self.command_context.generator_matrix_constants.get_link_indirect())
+        assert type(self.series) is str
```
MartinBelthle: check with isinstance

antarest/study/storage/variantstudy/model/command/update_link.py
```diff
@@ -0,0 +1,73 @@
...
+            ),
+            {"area_from": area_from, "area_to": area_to},
+        )
+
+    def _apply(self, study_data: FileStudy, *a: str) -> CommandOutput:
+        version = study_data.config.version
+        area_from, area_to = sorted([self.area1, self.area2])
+
+        current_parameters = study_data.tree.get(["input", "links", area_from, "properties", area_to])
+        current_parameters.update(self.parameters or {})
```
MartinBelthle: You shouldn't trust `self.parameters` as users can use the command directly and therefore put anything inside.
I'll suggest you do a roundtrtip around Pydantic objects you've introduced to ensure data quality

antarest/study/storage/variantstudy/model/command/update_link.py
```diff
@@ -0,0 +1,73 @@
...
+
+        study_data.tree.save(current_parameters, ["input", "links", area_from, "properties", area_to])
+
+        output, _ = self._apply_config(study_data.config)
+
+        self.save_series(area_from, area_to, study_data, version)
```
MartinBelthle: Not sure I understood everything but if someone updates a link property without updating a matrix you shouldn't modify them (I think the `save_series` method puts the default value if no value is given and if it does it should change)

antarest/study/storage/variantstudy/model/command/update_link.py
```diff
@@ -0,0 +1,73 @@
...
+from antarest.study.storage.variantstudy.business.utils import strip_matrix_protocol
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput
+from antarest.study.storage.variantstudy.model.command.create_link import AbstractLinkCommand
+from antarest.study.storage.variantstudy.model.command.icommand import MATCH_SIGNATURE_SEPARATOR, ICommand, OutputTuple
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+class UpdateLink(AbstractLinkCommand):
+    """
+    Command used to create a link between two areas.
```
MartinBelthle: Update the doc

antarest/study/web/study_data_blueprint.py
```diff
@@ -198,6 +197,24 @@ def create_link(
         params = RequestParameters(user=current_user)
         return study_service.create_link(uuid, link_creation_info, params)
 
+    @bp.put(
+        "/studies/{uuid}/links",
```
MartinBelthle: I think we should use `link` and not `links` as it's only used for one link.
On the other hand the creation endpoint you can't change is `links` ...
What do you think @sylvlecl ?

antarest/study/web/study_data_blueprint.py
```diff
@@ -198,6 +197,24 @@ def create_link(
...
+        summary="Update a link",
+        response_model=LinkInfoDTOType,
+    )
+    def update_link(
+        uuid: str,
+        link_creation_info: LinkInfoDTOType,
+        current_user: JWTUser = Depends(auth.get_current_user),
+    ) -> t.Any:
+        logger.info(
+            f"Updating link for study {uuid}",
```
MartinBelthle: Add info about which link in the logs

antarest/study/service.py
```diff
@@ -1856,6 +1855,25 @@ def create_link(
...
+        self,
+        uuid: str,
+        link_update_dto: LinkInfoDTOType,
+        params: RequestParameters,
+    ) -> LinkInfoDTOType:
+        study = self.get_study(uuid)
+        assert_permission(params.user, study, StudyPermissionType.WRITE)
+        self._assert_study_unarchived(study)
+        updated_link = self.links.update_link(study, link_update_dto)
+        self.event_bus.push(
```
MartinBelthle, line 1920: We usually don't push this inside the eventbus but perhaps it's a good idea. I think it can flood the bus but not sure

antarest/study/business/link_management.py
```diff
@@ -51,39 +59,119 @@ class LinkManager:
...
+
         execute_or_add_commands(study, file_study, [command], self.storage_service)
-        return LinkInfoDTO(
+
+        return link_data
+
+    def update_link(self, study: RawStudy, link_creation_info: LinkInfoDTOType) -> LinkInfoDTOType:
+        file_study = self.storage_service.get_storage(study).get_raw(study)
+
+        self.check_attributes_coherence(file_study, int(study.version), link_creation_info)
```
MartinBelthle: It would be better to use StudyVersion.parse(study.version)

antarest/study/business/link_management.py
```diff
@@ -137,15 +225,17 @@ def update_links_props(
...
+            new_links_by_ids[(area1, area2)] = updated_link_properties
 
             # Convert the DTO to a configuration object and update the configuration file.
-            properties = LinkProperties(**new_link_dto.model_dump(by_alias=False))
+            properties = LinkProperties(**updated_link_properties.model_dump(by_alias=False))
             path = f"{_ALL_LINKS_PATH}/{area1}/properties/{area2}"
             cmd = UpdateConfig(
                 target=path,
-                data=properties.to_config(),
+                data=properties.to_ini(int(study.version)),
```
MartinBelthle: Use StudyVersion.parse(study.version)

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -0,0 +1,253 @@
...
+@pytest.mark.unit_test
+class TestLink:
+    @pytest.mark.parametrize("study_type", ["raw", "variant"])
+    def test_link_update(self, client: TestClient, user_access_token: str, study_type: str) -> None:
+        client.headers = {"Authorization": f"Bearer {user_access_token}"}  # type: ignore
+
+        preparer = PreparerProxy(client, user_access_token)
+        study_id = preparer.create_study("foo", version=820)
+        if study_type == "variant":
+            study_id = preparer.create_variant(study_id, name="Variant 1")
```
MartinBelthle, line 30: When the study is a variant it could be nice to check the created `update_link` command to ensure it only contains the data you updated and not all the data

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -0,0 +1,253 @@
...
+            "hurdles-cost": False,
+            "link-style": "plain",
+            "link-width": 1.0,
+            "loop-flow": False,
+            "transmission-capacities": "enabled",
+            "use-phase-shifter": False,
+        }
+        assert expected == res.json()
+
+        # Test update link same area
```
MartinBelthle, line 60: It could be nice to check that the endpoint fails when given wrong parameters

MartinBelthle (changes requested):
Little review as I saw that you put the ticket in the  `Review` column in JIRA

antarest/study/storage/variantstudy/model/command/update_link.py
TheoPascoli: It has already been validated in LinkManagement

antarest/study/storage/variantstudy/model/command/create_link.py
TheoPascoli: I'm not sure, it was already here

MartinBelthle: After some research I saw that it's used inside replace matrix (see class AliasDecoder inside utils.py) as an alias to hide the complexity of the versions. Never seen this code before it's pretty unusual but we can keep it. I thought you introduced it

antarest/study/storage/variantstudy/model/command/update_link.py
```diff
@@ -0,0 +1,83 @@
...
+    def _apply(self, study_data: FileStudy, *a: str) -> CommandOutput:
+        version = study_data.config.version
+        area_from, area_to = sorted([self.area1, self.area2])
+
+        try:
+            properties = LinkProperties.model_validate(self.parameters or {}).model_dump(
+                exclude_unset=True, by_alias=True
+            )
+        except ValidationError:
+            raise LinkValidationError("One or more fields are forbidden")
```
MartinBelthle: I'm not sure we should catch this error. Because model_validate will already raise the `ValidationError` which has all info we need and the correct HTTP code I believe. At least you should do this to keep the stacktrace but I prefer not catching:
```python
except ValidationError as e:
     raise LinkValidationError("One or more fields are forbidden") from e
```

antarest/study/storage/variantstudy/model/command/update_link.py
```diff
@@ -0,0 +1,83 @@
...
+
+        current_parameters = study_data.tree.get(["input", "links", area_from, "properties", area_to])
+
+        current_parameters.update(properties or {})
```
MartinBelthle: It's possible to have properties = None after a model_dump ? If not you can remove the `or`

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -0,0 +1,294 @@
...
+                    "hurdles-cost": False,
+                },
+            )
+            assert res.status_code == 200
+
+            res = client.get(f"/v1/studies/{study_id}/commands")
+            commands = res.json()
+            command_args = commands[-1]["args"]
+            assert command_args["parameters"]["hurdles-cost"] == False
+            assert "loop-flow" not in command_args["parameters"]
```
MartinBelthle: You could be more strict and check that the content equals exactly {"parameters": {"hurdles-cost": False}}

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -73,11 +105,128 @@ def validate_series(
...
+                    command_context=self.command_context,
+                )
+            )
+        return commands
+
+    def get_inner_matrices(self) -> List[str]:
+        list_matrices = []
+        for attr in MATRIX_ATTRIBUTES:
+            value = getattr(self, attr, None)
+            if value:
```
MartinBelthle: You can merge the 2 lines above into only one:
`if value := getattr(self, attr, None)`

antarest/study/storage/variantstudy/model/command/update_link.py
```diff
@@ -0,0 +1,83 @@
...
+
+        study_data.tree.save(current_parameters, ["input", "links", area_from, "properties", area_to])
+
+        output, _ = self._apply_config(study_data.config)
+
+        if any([self.series, self.direct, self.indirect]):
+            self.save_series(area_from, area_to, study_data, version)
```
MartinBelthle: It's better cause you solved a part of the issue  I explained but I believe there's still an issue.
If someone just passes the series matrix inside the update you shouldn't modify the 2 others.
And I think entering `save_series` that's what you do
You can add a test to ensure this behaviour is correct

MartinBelthle (changes requested):
Only a few things to say

antarest/study/storage/variantstudy/model/command/update_link.py
TheoPascoli: I removed the catch

antarest/study/service.py
TheoPascoli: I saw this on area updating and other services but I can remove it

antarest/study/web/study_data_blueprint.py
sylvlecl: We should be consistent for all link-related endpoints (can be seen as the "links API"): better to keep `links` here.
However, I think it would make more sense to have either:
- a URL for updating only one link with the link ID in its path:
```studies/{uuid}/links/{link_id}```
- or this URL for updating a list of links

File: antarest/study/web/study_data_blueprint.py [hunk 20]
```diff
@@ -197,6 +197,25 @@ def create_link(
...
 
+    @bp.put(
+        "/studies/{uuid}/links",
+        tags=[APITag.study_data],
+        summary="Update a link",
+        response_model=LinkDTO,
+    )
+    def update_link(
+        uuid: str,
+        link_creation_info: LinkDTO,
```
MartinBelthle: This keyword should change as we are in an update endpoint

antarest/study/web/study_data_blueprint.py
MartinBelthle: Okay. We should probably discuss what's a link id then. Is it `area1%area2` or `area1/area2` or another id ?

antarest/study/service.py
MartinBelthle: Okay why not

antarest/study/business/model/link_model.py
```diff
@@ -98,5 +98,20 @@ class LinkInternal(AntaresBaseModel):
     filter_year_by_year: t.Optional[comma_separated_enum_list] = FILTER_VALUES
 
     def to_dto(self) -> LinkDTO:
-        data = self.model_dump()
-        return LinkDTO(**data)
+        return LinkDTO(
```
MartinBelthle: We should keep the 2 lines with model_dump

antarest/study/business/link_management.py
```diff
@@ -76,6 +78,49 @@ def create_link(self, study: Study, link_creation_dto: LinkDTO) -> LinkDTO:
...
+        updated_link = self.get_ini_link(study, link)
+
+        return updated_link.to_dto()
+
+    def check_link_existence(self, file_study: FileStudy, link: LinkInternal) -> None:
+        area_from, area_to = sorted([link.area1, link.area2])
+        try:
+            file_study.tree.get(["input", "links", area_from, "properties", area_to])
+        except KeyError:
+            raise LinkValidationError(f"The link {area_from} -> {area_to} is not present in the study")
```
MartinBelthle: There's an exception LinkNotFoundError defined inside `xpansion_management.py` that we should move into `exceptions.py` and use it here I believe

antarest/study/business/link_management.py
```diff
@@ -76,6 +78,49 @@ def create_link(self, study: Study, link_creation_dto: LinkDTO) -> LinkDTO:
...
+
+    def get_ini_link(self, study: RawStudy, link: LinkInternal) -> LinkInternal:
+        file_study = self.storage_service.get_storage(study).get_raw(study)
+
+        area_from, area_to = sorted([link.area1, link.area2])
```
MartinBelthle: Shouldn't we do this sorting when creating the LinkDTO object (with pydantic model_validator) to avoid doing this every time ?

antarest/study/business/link_management.py
```diff
@@ -76,6 +78,49 @@ def create_link(self, study: Study, link_creation_dto: LinkDTO) -> LinkDTO:
...
+        link = link_dto.to_internal(StudyVersion.parse(study.version))
+        file_study = self.storage_service.get_storage(study).get_raw(study)
+
+        self.check_link_existence(file_study, link)
+
+        command = UpdateLink(
+            area1=link.area1,
+            area2=link.area2,
+            parameters=link.model_dump(
+                include=link_dto.model_fields_set, exclude={"area1", "area2"}, exclude_none=True
```
MartinBelthle: Why did you use the `include` keyword here ?

antarest/study/business/link_management.py
```diff
@@ -76,6 +78,49 @@ def create_link(self, study: Study, link_creation_dto: LinkDTO) -> LinkDTO:
...
+        area_from, area_to = sorted([link.area1, link.area2])
+        try:
+            link_properties = file_study.tree.get(["input", "links", area_from, "properties", area_to])
+        except KeyError:
+            raise LinkValidationError(f"The link {area_from} -> {area_to} is not present in the study")
+
+        link_properties.update({"area1": area_from, "area2": area_to})
+        updated_link = LinkInternal.model_validate(link_properties)
+
+        return link.model_copy(update=updated_link.model_dump())
```
MartinBelthle: Why don't you simply return the updated_link you got line 120 ?

antarest/study/business/link_management.py
```diff
@@ -76,6 +78,49 @@ def create_link(self, study: Study, link_creation_dto: LinkDTO) -> LinkDTO:
...
+        return updated_link.to_dto()
+
+    def check_link_existence(self, file_study: FileStudy, link: LinkInternal) -> None:
+        area_from, area_to = sorted([link.area1, link.area2])
+        try:
+            file_study.tree.get(["input", "links", area_from, "properties", area_to])
+        except KeyError:
+            raise LinkValidationError(f"The link {area_from} -> {area_to} is not present in the study")
+
+    def get_ini_link(self, study: RawStudy, link: LinkInternal) -> LinkInternal:
```
MartinBelthle: Opinion and really minor but: It could be good to rename this method `get_internal_link` to hide the fact that we get the data inside a ini file as it could change

antarest/study/storage/variantstudy/model/command/update_link.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> OutputTuple:
+        area_from, area_to = sorted([self.area1, self.area2])
+
+        return (
+            CommandOutput(
+                status=True,
+                message=f"Link between '{self.area1}' and '{self.area2}' updated",
+            ),
+            {"area_from": area_from, "area_to": area_to},
```
MartinBelthle: Just a question but where do we use the 2nd part of the OutputTuple ?

antarest/study/storage/variantstudy/model/command/update_link.py
```diff
@@ -0,0 +1,83 @@
...
+        )
+
+    def _apply(self, study_data: FileStudy, listener: t.Optional[ICommandListener] = None) -> CommandOutput:
+        version = study_data.config.version
+        area_from, area_to = sorted([self.area1, self.area2])
+
+        properties = study_data.tree.get(["input", "links", area_from, "properties", area_to])
+
+        new_properties = LinkInternal.model_validate(self.parameters).model_dump(include=self.parameters, by_alias=True)
```
MartinBelthle, line 49: Same question about include here.

antarest/study/storage/variantstudy/model/command/update_link.py
```diff
@@ -0,0 +1,83 @@
...
+        if self.indirect:
+            self.save_indirect(area_from, area_to, study_data, version)
+
+        return output
+
+    def to_dto(self) -> CommandDTO:
+        return super().to_dto()
+
+    def match_signature(self) -> str:
+        return str(
```
MartinBelthle: You could put that inside the AbstractCommand and use it inside the create too

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -57,11 +53,128 @@ def validate_series(
...
+            "area2": self.area2,
+            "parameters": self.parameters,
+        }
+        for attr in MATRIX_ATTRIBUTES:
+            if value := getattr(self, attr, None):
+                args[attr] = strip_matrix_protocol(value)
+        return CommandDTO(action=self.command_name.value, args=args, study_version=self.study_version)
+
+    def match(self, other: ICommand, equal: bool = False) -> bool:
+        if not isinstance(other, CreateLink):
```
MartinBelthle: Should probably be replaced with `if not isinstance(other, self.__class__):` as you're inside AbstractCommand

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -57,11 +53,128 @@ def validate_series(
...
+        return (
+            simple_match
+            and self.parameters == other.parameters
+            and self.series == other.series
+            and self.direct == other.direct
+            and self.indirect == other.indirect
+        )
+
+    def _create_diff(self, other: "ICommand") -> List["ICommand"]:
+        other = cast(CreateLink, other)
```
MartinBelthle: Should be casted inside Abstract I believe to work with UpdateCommand

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -57,11 +53,128 @@ def validate_series(
...
+    def get_inner_matrices(self) -> List[str]:
+        list_matrices = []
+        for attr in MATRIX_ATTRIBUTES:
+            if value := getattr(self, attr, None):
+                assert_this(isinstance(value, str))
+                list_matrices.append(strip_matrix_protocol(value))
+        return list_matrices
+
+    def save_series(self, area_from: str, area_to: str, study_data: FileStudy, version: StudyVersion) -> None:
+        self.series = self.series or (self.command_context.generator_matrix_constants.get_link(version=version))
```
MartinBelthle: You don't have to do this anymore as when you enter this function you're sure you have series.
Same goes for save_direct and save_indirect

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -57,11 +53,128 @@ def validate_series(
...
+        assert isinstance(self.series, str)
+        if version < STUDY_VERSION_8_2:
+            study_data.tree.save(self.series, ["input", "links", area_from, area_to])
+        else:
+            study_data.tree.save(
+                self.series,
+                ["input", "links", area_from, f"{area_to}_parameters"],
+            )
+            study_data.tree.save({}, ["input", "links", area_from, "capacities"])
```
MartinBelthle: Not really sure why you do this, is it to ensure we don't have both matrices writing ? I think you should remove this. 
Same goes for the 2 save_ functions

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -19,6 +20,117 @@
...
+                        "args": {
+                            "area1": area1_id,
+                            "area2": area2_id,
+                            "parameters": {"hurdles-cost": False, "wrong": "parameter"},
+                        },
+                    }
+                ],
+            )
+            assert res.status_code == 500
+            expected = {
```
MartinBelthle: You shouldn't check the entire content as if pydantic changes its error msg we'll have to change this. You could just check our msg: "Unexpected exception occurred when trying to apply command CommandName.UPDATE_LINK: 1 validation error for LinkInternal"

tests/integration/study_data_blueprint/test_link.py
```diff
@@ -19,6 +20,117 @@
...
+                    "hurdlesCost": False,
+                },
+            )
+            assert res.status_code == 200
+
+            res = client.get(f"/v1/studies/{study_id}/commands")
+            commands = res.json()
+            command_args = commands[-1]["args"]
+            assert command_args["parameters"] == {"hurdles_cost": False}
+
```
MartinBelthle, line 118: it could be nice to add tests for matrices as I don't see any.

MartinBelthle (changes requested):

antarest/study/business/link_management.py
TheoPascoli: We must include only the fields that has been specified by the client, otherwise we would replace the existing values by the default ones

antarest/study/storage/variantstudy/model/command/create_link.py
TheoPascoli: I'm not sure why. It was there before. I removed it

antarest/study/business/link_management.py
```diff
@@ -76,6 +78,49 @@ def create_link(self, study: Study, link_creation_dto: LinkDTO) -> LinkDTO:
...
+            raise LinkNotFound(f"The link {area_from} -> {area_to} is not present in the study")
+
+    def get_internal_link(self, study: RawStudy, link: LinkInternal) -> LinkInternal:
+        file_study = self.storage_service.get_storage(study).get_raw(study)
+
+        area_from, area_to = sorted([link.area1, link.area2])
+        try:
+            link_properties = file_study.tree.get(["input", "links", area_from, "properties", area_to])
+        except KeyError:
+            raise LinkValidationError(f"The link {area_from} -> {area_to} is not present in the study")
```
MartinBelthle: You should also raise a LinkNotFound here.

Also, you could perhaps call check_link_existence inside this method to avoid duplicating code (needs a little refactoring to return the properties inside the check method).

antarest/study/business/link_management.py
MartinBelthle: Okay I see I didn't know model_fields_set worked like this, it's nice

antarest/study/business/link_management.py
```diff
@@ -76,6 +78,49 @@ def create_link(self, study: Study, link_creation_dto: LinkDTO) -> LinkDTO:
...
+        )
+
+        execute_or_add_commands(study, file_study, [command], self.storage_service)
+
+        updated_link = self.get_internal_link(study, link)
+
+        return updated_link.to_dto()
+
+    def check_link_existence(self, file_study: FileStudy, link: LinkInternal) -> None:
+        area_from, area_to = sorted([link.area1, link.area2])
```
MartinBelthle: You shouldn't have to sort this anymore as you did it inside the Area pydantic class

Same goes for other parts of the code I'll let you check

antarest/study/storage/variantstudy/model/command/update_link.py
```diff
@@ -0,0 +1,78 @@
...
+#
+# This file is part of the Antares project.
+import typing as t
+
+from antarest.study.business.model.link_model import LinkInternal
+from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfig
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput
+from antarest.study.storage.variantstudy.model.command.create_link import AbstractLinkCommand
+from antarest.study.storage.variantstudy.model.command.icommand import MATCH_SIGNATURE_SEPARATOR, ICommand, OutputTuple
```
MartinBelthle: Remove unused import MATCH_SIGNATURE_SEPARATOR

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -57,11 +53,130 @@ def validate_series(
...
+        else:
+            study_data.tree.save(
+                self.series,
+                ["input", "links", area_from, f"{area_to}_parameters"],
+            )
+
+    def save_direct(self, area_from: str, area_to: str, study_data: FileStudy, version: StudyVersion) -> None:
+        self.direct = self.command_context.generator_matrix_constants.get_link_direct()
+        assert isinstance(self.direct, str)
+        if version >= STUDY_VERSION_8_2:
```
MartinBelthle, line 146: Not really the point of this dev but we should raise an Exception when someone gives "direct" or "indirect" with a study prior to v8.2
This could be validated inside a field_validator now that we have the attribute `study_version` inside the commands.
This way you could remove the version check inside the 2 functions
It could also be good to add tests for this inside your matrix tests

MartinBelthle (changes requested):
Really minor changes, otherwise looks good to me. With the matrix tests we'll be good to go

antarest/study/storage/variantstudy/model/command/create_link.py
MartinBelthle: In fact it could break the compatibility of the R scripts. We should ask them how they call this command before

TheoPascoli: After checking with script team, I removed those conditions and added a model validator

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -57,11 +54,138 @@ def validate_series(
         return validate_matrix(v, new_values) if v is not None else v
 
     @model_validator(mode="after")
-    def validate_areas(self) -> "CreateLink":
+    def validate_areas(self) -> "AbstractLinkCommand":
         if self.area1 == self.area2:
             raise ValueError("Cannot create link on same node")
+
+        if StudyVersion.parse(self.study_version) < STUDY_VERSION_8_2 and (
```
MartinBelthle: Don't need to parse it's already a StudyVersion object

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -57,11 +54,138 @@ def validate_series(
...
+            self.series is not None or self.direct is not None or self.indirect is not None
+        ):
+            raise LinkValidationError(
```
MartinBelthle, line 62: I believe you shouldn't raise an error when series are given as in the code we do handle this case. So I think you should only check the 2 other matrices and put back the tests as they were

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -57,11 +54,138 @@ def validate_series(
...
+    def get_inner_matrices(self) -> List[str]:
+        list_matrices = []
+        for attr in MATRIX_ATTRIBUTES:
+            if value := getattr(self, attr, None):
+                assert_this(isinstance(value, str))
+                list_matrices.append(strip_matrix_protocol(value))
+        return list_matrices
+
+    def save_series(self, area_from: str, area_to: str, study_data: FileStudy, version: StudyVersion) -> None:
+        self.series = self.command_context.generator_matrix_constants.get_link(version=version)
```
MartinBelthle: The 3 save method have the same issue. You replace `self.series` by the default matrix but if someone gives you a matrix you should use the one that was given. You should remove this line

MartinBelthle (changes requested):
Still got little issues

MartinBelthle (dismissed):

antarest/study/business/model/link_model.py
```diff
@@ -78,6 +86,11 @@ def to_internal(self, version: StudyVersion) -> "LinkInternal":
         return LinkInternal(**data)
 
 
+class LinkDtoForUpdate(LinkDTO):
+    area1: SkipJsonSchema[str] = Field("a", exclude=True)
```
MartinBelthle: Weird way to do it. Why did you introduce fields `a` and `b` ?

antarest/study/business/model/link_model.py
```diff
@@ -44,11 +45,18 @@ class Area(AntaresBaseModel):
...
             raise LinkValidationError(f"Cannot create a link that goes from and to the same single area: {self.area1}")
+        area_from, area_to = sorted([self.area1, self.area2])
+        self.area1 = area_from
+        self.area2 = area_to
         return self
 
 
 class LinkDTO(Area):
-    model_config = ConfigDict(alias_generator=to_camel_case, populate_by_name=True, extra="forbid")
+    model_config = ConfigDict(
```
MartinBelthle: Could be in one line

antarest/study/business/model/link_model.py
```diff
@@ -78,6 +86,11 @@ def to_internal(self, version: StudyVersion) -> "LinkInternal":
         return LinkInternal(**data)
 
 
+class LinkDtoForUpdate(LinkDTO):
```
MartinBelthle: Could be DTO instead of Dto to fit with existing class

MartinBelthle (commented):
2 minor comments that don't really need to be adress and 1 question

antarest/study/business/model/link_model.py
MartinBelthle: After testing this allows user to specfiy area1 and area2 inside the body and they we'll be ignored. I don't think we want that. I'd prefer you introduce a class that just has the base attributes and make LinkDTO inherit from this :/

antarest/study/business/model/link_model.py
TheoPascoli: Indeed

antarest/study/business/model/link_model.py
TheoPascoli: I changed it

MartinBelthle (approved):

PR #2179, opened 2024-10-09
style(variant): improve logs

maugde:
Change the display of number of days in logs related to snapshot clearing tasks.

### DIFF
--- a/antarest/study/storage/variantstudy/variant_study_service.py
+++ b/antarest/study/storage/variantstudy/variant_study_service.py
@@ -1110,10 +1110,10 @@ class SnapshotCleanerTask:
                         self._variant_study_service.clear_snapshot(variant)
 
     def run_task(self, notifier: TaskUpdateNotifier) -> TaskResult:
-        msg = f"Start cleaning all snapshots updated or accessed" f" {humanize.naturaltime(self._retention_time)}."
+        msg = f"Start cleaning all snapshots updated or accessed {humanize.naturaltime(self._retention_time)}."
         notifier(msg)
         self._clear_all_snapshots()
-        msg = f"All selected snapshots were successfully cleared."
+        msg = "All selected snapshots were successfully cleared."
         notifier(msg)
         return TaskResult(success=True, message=msg)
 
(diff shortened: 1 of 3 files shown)


antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1072,7 +1072,9 @@ def clear_all_snapshots(self, retention_hours: timedelta, params: t.Optional[Req
         if params is None or (params.user and not params.user.is_site_admin() and not params.user.is_admin_token()):
             raise UserHasNotPermissionError()
 
-        task_name = f"Cleaning all snapshot updated or accessed at least {retention_hours} hours ago."
+        task_name = (
+            f"Cleaning all snapshot updated or accessed at least {retention_hours.total_seconds() // 3600} hours ago."
```
sylvlecl: We could use [humanize](https://pypi.org/project/humanize) so that it's even more readable for high values.
For example, when we have several days it will be better to have it in days than in hours.

For example:
```python
print(humanize.time.precisedelta(datetime.timedelta(days=12, hours=5)))
# --> 12 days and 5 hours
```

Also, we should rename the variable just retention_time because it's not necessarily hours.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -1111,7 +1113,10 @@ def _clear_all_snapshots(self) -> None:
...
 
     def run_task(self, notifier: TaskUpdateNotifier) -> TaskResult:
-        msg = f"Start cleaning all snapshots updated or accessed {self._retention_hours} hours ago."
+        msg = (
+            f"Start cleaning all snapshots updated or accessed"
+            f" {self._retention_hours.total_seconds() // 3600} hours ago."
+        )
         notifier(msg)
         self._clear_all_snapshots()
         msg = f"All selected snapshots were successfully cleared."
```
MartinBelthle: Remove this useless f-string.
Also I prefer the solution that Sylvain proposed it's nice.
You should also apply this change inside the `variant_blueprint` as you introduced the same log there

sylvlecl (approved):

PR #2182, opened 2024-10-09
feat(clusters): convert groups and names to lower case

MartinBelthle:
Will fix [ANT-2311]


antarest/study/storage/rawstudy/model/filesystem/config/cluster.py
```diff
@@ -51,6 +51,10 @@ class ItemProperties(
 
     name: str = Field(description="Cluster name", pattern=r"[a-zA-Z0-9_(),& -]+")
 
+    @field_validator("name", mode="before")
+    def _validate_name(cls, name: str) -> str:
+        return str(name).lower()
```
sylvlecl: is the str conversion really needed ? name is supposed to already be an str

antarest/study/business/areas/thermal_management.py
```diff
@@ -416,7 +416,7 @@ def duplicate_cluster(
         Raises:
             ClusterAlreadyExists: If a cluster with the new name already exists in the area.
         """
-        new_id = transform_name_to_id(new_cluster_name, lower=False)
+        new_id = transform_name_to_id(new_cluster_name)
         lower_new_id = new_id.lower()
```
sylvlecl: I guess it's already in lower case now ?

antarest/study/storage/rawstudy/model/filesystem/config/identifier.py
```diff
@@ -45,7 +45,7 @@ def generate_id(cls, name: str) -> str:
         # Avoid circular imports
         from antarest.study.storage.rawstudy.model.filesystem.config.model import transform_name_to_id
 
-        return transform_name_to_id(name, lower=False)
+        return transform_name_to_id(name)
```
sylvlecl, line 48: It seems to be identical to class `LowerCaseIdentifier` now ?

antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
```diff
@@ -216,6 +216,12 @@ class AbstractBindingConstraintCommand(OptionalProperties, BindingConstraintMatr
 
     coeffs: t.Optional[t.Dict[str, t.List[float]]] = None
 
+    @model_validator(mode="after")
+    def lower_group(self) -> "AbstractBindingConstraintCommand":
```
sylvlecl: it looks like we can achieve the same thing with a type annotation:

```python
class Foo(BaseModel):
    bar: Annotated[str, StringConstraints(to_lower=True)]
```
Shouldn't we try to use this where we can instead of having custom validators ?
We can have an alias for this type of course, like `LowerCaseStr`

antarest/study/storage/variantstudy/model/command/create_cluster.py
```diff
@@ -53,10 +53,17 @@ class CreateCluster(ICommand):
 
     @field_validator("cluster_name", mode="before")
     def validate_cluster_name(cls, val: str) -> str:
-        valid_name = transform_name_to_id(val, lower=False)
-        if valid_name != val:
+        to_return = transform_name_to_id(val)
+        if not to_return:
             raise ValueError("Cluster name must only contains [a-zA-Z0-9],&,-,_,(,) characters")
```
sylvlecl: I think this should be a pydantic ValidationError

antarest/study/storage/variantstudy/model/command/create_cluster.py
```diff
@@ -53,10 +53,17 @@ class CreateCluster(ICommand):
 
     @field_validator("cluster_name", mode="before")
     def validate_cluster_name(cls, val: str) -> str:
```
sylvlecl: I think we can create an annotated type that we can reuse everywhere, instead of duplicating the method in all classes.

https://docs.pydantic.dev/latest/concepts/validators/#annotated-validators

antarest/study/storage/variantstudy/model/command/create_cluster.py
```diff
@@ -53,10 +53,17 @@ class CreateCluster(ICommand):
...
-        valid_name = transform_name_to_id(val, lower=False)
-        if valid_name != val:
+        to_return = transform_name_to_id(val)
+        if not to_return:
             raise ValueError("Cluster name must only contains [a-zA-Z0-9],&,-,_,(,) characters")
-        return val
+        return to_return
+
+    @field_validator("parameters", mode="before")
+    def lower_cluster_group_and_names(cls, params: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
```
sylvlecl: This dict makes it really hard to maintain. We should use a pydantic model instead to make expectations explicit.

antarest/study/storage/variantstudy/model/command/create_renewables_cluster.py
```diff
@@ -48,10 +48,17 @@ class CreateRenewablesCluster(ICommand):
 
     @field_validator("cluster_name")
     def validate_cluster_name(cls, val: str) -> str:
```
sylvlecl: We should reuse an annotated type instead

antarest/study/storage/variantstudy/model/command/create_renewables_cluster.py
```diff
@@ -48,10 +48,17 @@ class CreateRenewablesCluster(ICommand):
...
-        if valid_name != val:
-            raise ValueError("Area name must only contains [a-zA-Z0-9],&,-,_,(,) characters")
-        return val
+        to_return = transform_name_to_id(val)
+        if not to_return:
+            raise ValueError("Cluster name must only contains [a-zA-Z0-9],&,-,_,(,) characters")
+        return to_return
+
+    @field_validator("parameters", mode="before")
+    def lower_cluster_group_and_names(cls, params: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
```
sylvlecl: Same comment, we should have a pydantic model here

antarest/study/storage/variantstudy/model/command/remove_cluster.py
```diff
@@ -39,6 +45,13 @@ class RemoveCluster(ICommand):
     area_id: str
     cluster_id: str
 
+    @field_validator("cluster_id", mode="before")
```
sylvlecl: same, we should re-use the same annotated type

sylvlecl (changes requested):
First set of comments

antarest/study/storage/rawstudy/model/filesystem/config/cluster.py
MartinBelthle: I had the issue in a test case where I named my cluster "111" and it came as an int.

antarest/study/business/areas/thermal_management.py
MartinBelthle: Indeed

antarest/study/storage/rawstudy/model/filesystem/config/identifier.py
MartinBelthle: Indeed, I'll merge the 2 classes

antarest/study/storage/variantstudy/model/command/create_cluster.py
MartinBelthle: Why not

antarest/study/business/table_mode_management.py
```diff
@@ -95,37 +95,43 @@ def __init__(
...
             data = {
-                f"{area1_id} / {area2_id}": link.model_dump(by_alias=True)
+                f"{area1_id} / {area2_id}": link.model_dump(mode="json", by_alias=True)
                 for (area1_id, area2_id), link in links_map.items()
             }
         elif table_type == TableModeType.THERMAL:
             thermals_by_areas = self._thermal_manager.get_all_thermals_props(study)
             data = {
-                f"{area_id} / {cluster_id}": cluster.model_dump(by_alias=True, exclude={"id", "name"})
+                f"{area_id} / {cluster_id.lower()}": cluster.model_dump(
```
sylvlecl: Here we are supposed to have IDs, not names, so we should not need to do this.

We must have a clear idea in which parts of the code we can have names and in which parts of the code we have IDs.
In the second case, we must not handle upper/lower case issues, otherwise we blur the responsibilities.

We need to review all other places where `lower()` or `transform_name_to_id` is used and where it's not relevant anymore, for the same reason.

antarest/study/business/areas/thermal_management.py
```diff
@@ -416,7 +416,7 @@ def duplicate_cluster(
         Raises:
             ClusterAlreadyExists: If a cluster with the new name already exists in the area.
         """
-        new_id = transform_name_to_id(new_cluster_name, lower=False)
```
sylvlecl, line 434: It's weird that we have to do this here inside a "business" method.

From a global point of view, we should choose between one of 2 approaches:

1. either we ensure at the external interfaces of the application that we transform everything to lower-case / to a valid ID, and then in the rest of the code we should not care about it
2. or we allow everywhere in our code to have upper case names, but then we need to make sure in all search methods that we always match based on an ID derived from this. To ensure that we should probably use another class than `str` for those names, otherwise it will be impossible to track and maintain the pplaces where we have a name that need to be transformed into and ID for some processes

I think that in the present evolution, we have chosen option 1, which is indeed simpler to implement.
But then we should only:

1. ensure in the APIs that we transform the names into valid IDs
2. ensure when we read files or commands from database that we transform the names into valid IDs
3. remove everywhere else all processing related to lower case or name transformation

So for example here we should not transform the ID, which should already be valid.
Instead, the valid ID should be created in the API layer, i.e method `duplicate_cluster` in `study_data_blueprint`.
For this, we should use an annotated type there.

antarest/study/storage/variantstudy/model/command/create_cluster.py
```diff
@@ -53,10 +53,17 @@ class CreateCluster(ICommand):
 
     @field_validator("cluster_name", mode="before")
     def validate_cluster_name(cls, val: str) -> str:
-        valid_name = transform_name_to_id(val, lower=False)
-        if valid_name != val:
+        to_return = transform_name_to_id(val)
```
sylvlecl: Its' weird that here we convert the name to a valid ID, whereas in `ItemProperties` (and then in particular in ThermalClusterInput), we have a constraint instead.
I don't understand why, but the constraint does not seem enforced by pydantic, by the way: this does not throw a validation error: `creation = ThermalClusterCreation.model_validate({"name":"clusterè"})`

We should probably make this consistent by transforming the name in ID in item properties class instead ?

antarest/study/service.py
```diff
@@ -1595,15 +1595,17 @@ def apply_commands(
         self, uuid: str, commands: t.List[CommandDTO], params: RequestParameters
     ) -> t.Optional[t.List[str]]:
         study = self.get_study(uuid)
+        parsed_commands: t.List[ICommand] = []
+        # Round trip to do some validation (goes through commands validators) before appending cmd to the variant
```
sylvlecl: Should probably be the responsibility of the variant service to do that validation in the `append_commands` method.
The variant service could have a validation method for commands.

antarest/study/business/binding_constraint_management.py
```diff
@@ -501,7 +501,7 @@ def constraint_model_adapter(constraint: t.Mapping[str, t.Any], study_version: S
             constraint_output["filter_year_by_year"] = _filter_year_by_year
             constraint_output["filter_synthesis"] = _filter_synthesis
         if study_version >= STUDY_VERSION_8_7:
-            constraint_output["group"] = constraint.get("group", DEFAULT_GROUP)
+            constraint_output["group"] = constraint.get("group", DEFAULT_GROUP).lower()
```
sylvlecl: Should be enforced in the `ConstraintOutput870` instead I think

antarest/study/storage/rawstudy/model/filesystem/config/cluster.py
sylvlecl: We may change the signature to Any so that it's more clear to readers

antarest/study/business/areas/renewable_management.py
```diff
@@ -114,7 +114,7 @@ def create_renewable_output(
     cluster_id: str,
     config: t.Mapping[str, t.Any],
 ) -> "RenewableClusterOutput":
-    obj = create_renewable_config(study_version=StudyVersion.parse(study_version), **config, id=cluster_id)
+    obj = create_renewable_config(study_version=StudyVersion.parse(study_version), **config, id=cluster_id.lower())
```
sylvlecl: Should be enforced in the renawable config class instead (actually in ItemProperties)

antarest/study/business/areas/renewable_management.py
```diff
@@ -347,9 +347,8 @@ def duplicate_cluster(
         Raises:
             DuplicateRenewableCluster: If a cluster with the new name already exists in the area.
         """
-        new_id = transform_name_to_id(new_cluster_name, lower=False)
-        lower_new_id = new_id.lower()
-        if any(lower_new_id == cluster.id.lower() for cluster in self.get_clusters(study, area_id)):
+        new_id = transform_name_to_id(new_cluster_name)
```
sylvlecl, line 360: Same as for thermal: should be enforced in the API layer instead.

antarest/study/business/areas/renewable_management.py
```diff
@@ -388,6 +387,7 @@ def update_renewables_props(
         for area_id, update_renewables_by_ids in update_renewables_by_areas.items():
             old_renewables_by_ids = old_renewables_by_areas[area_id]
             for renewable_id, update_cluster in update_renewables_by_ids.items():
+                renewable_id = renewable_id.lower()
```
sylvlecl: In the same idea as other comments:
this should probaby be ensured upfront, in the inputs of that method we should have validated models (for ex. we have RenewableClusterInput objects).

Plus, it's not clear why we just transform it to lower case and not to ID.

sylvlecl (changes requested):
The main point is that we should make clear where the transformation name -> ID should be handled.
I think it should be handled at the "external interfaces" of our code (reading from API, from files, and from database), and **not** elsewhere. Most comments are related to this.

antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
MartinBelthle: I'll try cause if it works it's way better indeed

MartinBelthle:
The issue I'm having is that validating the Union calls model_construct and not model_validate which is a major difference as model_construct is permissive whereas model_validate isn't. 
I'll see if I see something inside Pydantic doc to force the use of model_validate.
Otherwise, I don't really know what to do, I'll keep thinking

MartinBelthle:
I'm currently seeing with Guillaume of the Simulator Team if the group "Other" existed once. Because our example study: STA-mini has this group. So either it existed and I'll have to adapt my code or it never did and I will change our example study.

MartinBelthle:
I asked Alexander what is his opinion on name being the same as ids. Waiting for his answer.

antarest/study/web/study_data_blueprint.py
```diff
@@ -1965,7 +1965,7 @@ def update_renewable_cluster(
         )
         request_params = RequestParameters(user=current_user)
         study = study_service.check_study_access(uuid, StudyPermissionType.WRITE, request_params)
-        return study_service.renewable_manager.update_cluster(study, area_id, cluster_id, cluster_data)
+        return study_service.renewable_manager.update_cluster(study, area_id, cluster_id.lower(), cluster_data)
```
sylvlecl: We can use `LowerCaseStr` in the endpoint definition I think, same for other endpoints.

antarest/study/storage/variantstudy/model/command/remove_st_storage.py
```diff
@@ -43,6 +44,13 @@ class RemoveSTStorage(ICommand):
     area_id: str = Field(description="Area ID", pattern=r"[a-z0-9_(),& -]+")
     storage_id: str = Field(description="Short term storage ID", pattern=r"[a-z0-9_(),& -]+")
 
+    @field_validator("storage_id", mode="before")
+    def validate_cluster_name(cls, val: str) -> str:
```
sylvlecl: I think we should only perform the lower case transformation here, no ?
The field is named ID, we should really have an ID.

Also, this validation seems to supersede the pattern constraint on the field, it seems to make more sense to keep only one of them.

antarest/study/storage/variantstudy/model/command/create_cluster.py
```diff
@@ -199,7 +187,7 @@ def _create_diff(self, other: "ICommand") -> t.List["ICommand"]:
         from antarest.study.storage.variantstudy.model.command.update_config import UpdateConfig
 
         # Series identifiers are in lower case.
-        series_id = transform_name_to_id(self.cluster_name, lower=True)
```
sylvlecl, line 202: I think we could remove the argument "lower" in the function definition, our ids are always in lower case, no ?
lower=False is used almost only in test code

antarest/study/business/areas/renewable_management.py
```diff
@@ -340,7 +340,7 @@ def duplicate_cluster(
         study: Study,
         area_id: str,
         source_id: str,
-        new_cluster_name: str,
+        new_cluster_name: LowerCaseStr,
```
sylvlecl: Actually using the annotated type here has no effect, I think we should remove it from here (and use it in the web layer instead)

antarest/study/storage/rawstudy/model/filesystem/config/field_validators.py
```diff
@@ -87,3 +87,30 @@ def validate_color_rgb(v: t.Any) -> str:
...
+
+    Args:
+        name: The name to convert.
+        lower: The flag used to turn the identifier in lower case.
+    """
+    valid_id = _sub_invalid_chars(" ", name).strip()
+    return valid_id.lower() if lower else valid_id
+
+
+def validate_id_against_name(name: str) -> str:
```
sylvlecl, line 110: This is only used to validate cluster IDs in commands: I think it should be removed ?

antarest/study/storage/variantstudy/model/command/remove_cluster.py
sylvlecl: Marked as resolved but shouldn't we use the LowerCaseStr instead ?

**MB**: Same as previous remark, I think we want to continue validate the id not only with LowerCaseStr

antarest/study/business/binding_constraint_management.py
sylvlecl: What about this comment ? :) Still think it would be more explicit

**MB**: Indeed

antarest/study/business/areas/thermal_management.py
```diff
@@ -414,7 +414,7 @@ def duplicate_cluster(
         study: Study,
         area_id: str,
         source_id: str,
-        new_cluster_name: str,
+        new_cluster_name: LowerCaseStr,
```
sylvlecl: Here too I think it should be removed be cause it does not actually performs the transformation

sylvlecl (changes requested):
Just minor comments, looks good!

antarest/study/storage/variantstudy/model/command/create_cluster.py
MartinBelthle: I kept it for the districts but I'll see if i can put the districts in lower case and remove the lower attribute as it's the only case with lower case to false

MartinBelthle: I saw that disctricts are written in lower case by the simulator so I'll remove the lower=False and the argument `lower`

antarest/study/storage/variantstudy/model/command/remove_st_storage.py
MartinBelthle: Indeed. I'll keep the lower case only as Alexander said he wanted to allow special characters inside the name

antarest/study/storage/rawstudy/model/filesystem/config/field_validators.py
MartinBelthle: I don't think I understand. The name of this function doesn't really make sense but I think we should keep the id validation as an user can use a command directly

antarest/study/business/areas/renewable_management.py
MartinBelthle: Why does this have no effect here ? It's only useful when it's inside an object ?

MartinBelthle: In any case I'll do it

antarest/study/storage/variantstudy/model/command/remove_st_storage.py
MartinBelthle: Same remark, the name should be `validate_id` but I think we should keep it

sylvlecl (approved):

PR #2183, opened 2024-10-10
Feature/poc cache study dao

sylvlecl:
Just demonstrate the introduction of the DAO layer to make easier
to implement a better use of the cache.

Allows to speed up read access for area UI info around 10x.


sylvlecl:
Actual DAO implementation has now been started.

PR #2186, opened 2024-10-14
Fix/fix apidocs redirect

### LINKED ISSUE #1625: Cannot refresh API page on desktop version
## Description

As described in the title, when i want to refresh the API page (/apidoc), it returns a white screen : 

bug_refresh

1. Launch the desktop version of AntaresWeb
2. Click on the API button
3. Refresh your page

## Expected Behavior

I would like my page to refresh correctly

## Environment

v2.14.4

sylvlecl:
Fixes #1625 

Every URL which started with /api, including /apidocs, was identified as a subpath of the API,
and was therefore not redirected to the front page.

The fix is to actually check for "/api/" in the URL.

### DIFF
--- a/antarest/front.py
+++ b/antarest/front.py
@@ -22,7 +22,7 @@ This includes:
 
 import re
 from pathlib import Path
-from typing import Any, Optional, Sequence
+from typing import Any, List, Optional, Sequence
 
 from fastapi import FastAPI
 from starlette.middleware.base import BaseHTTPMiddleware, DispatchFunction, RequestResponseEndpoint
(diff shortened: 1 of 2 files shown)


tests/test_front.py
```diff
@@ -106,6 +106,10 @@ def test_frontend_paths(base_back_app, resources_dir: Path) -> None:
     assert front_route_response.status_code == 200
     assert front_route_response.text == "index"
 
+    front_route_response = client.get("/apidocs")
```
MartinBelthle: Could be changed to `apidoc` to fit with the name we use.

antarest/front.py
```diff
@@ -46,23 +46,38 @@ def __init__(
...
         """
         dispatch = self.dispatch if dispatch is None else dispatch
         super().__init__(app, dispatch)
-        self.known_prefixes = {re.findall(r"/(?:(?!/).)*", p)[0] for p in route_paths if p != "/"}
+
+        if protected_roots is None:
+            protected_roots = []
+        if protected_paths is None:
+            protected_paths = []
+        self.protected_paths = protected_paths
```
MartinBelthle: Could be simplified in 
```python
self.protected_paths = protected_paths or []
protected_roots = protected_roots or []
```

MartinBelthle (dismissed):

tests/test_front.py
MartinBelthle: Already fixed

PR #2191, opened 2024-10-17
feat(tasks): add new endpoint to fetch task progress

MartinBelthle:


antarest/core/interfaces/eventbus.py
```diff
@@ -46,6 +46,7 @@ class EventType(str, Enum):
     WORKER_TASK_STARTED = "WORKER_TASK_STARTED"
     WORKER_TASK_ENDED = "WORKER_TASK_ENDED"
     LAUNCH_PROGRESS = "LAUNCH_PROGRESS"
+    TS_GENERATION_PROGRESS = "TS_GENERATION_PROGRESS"
```
MartinBelthle, line 49: Nothing to do inside this PR, will be used in a next one but doesn't really matter

TheoPascoli (approved):

PR #2194, opened 2024-10-18
feat(ts-gen): display progress bar via websockets

MartinBelthle:


antarest/study/storage/variantstudy/model/command_listener/command_listener.py
```diff
@@ -0,0 +1,33 @@
...
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+
+from abc import ABC, abstractmethod
+
+from antarest.core.serialization import AntaresBaseModel
+
+
+class ICommandListener(ABC, AntaresBaseModel, extra="forbid", arbitrary_types_allowed=True):
```
TheoPascoli: why not set the config with the pydantic way like : 
class Config:
        extra = "forbid"
        arbitrary_types_allowed = True
inside the class ?

antarest/study/storage/variantstudy/model/command_listener/command_listener.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    Interface for all listeners that can be given inside the `apply` method of a command.
+    """
+
+    task_id: str
```
TheoPascoli: If you define a setter for this variable, you can add a underscore like _task_id: str
And also add a property : 
    @property
    def task_id(self) -> str:
        return self._task_id

TheoPascoli (changes requested):

antarest/study/storage/variantstudy/model/command_listener/command_listener.py
MartinBelthle: It's the same thing, I can change it

antarest/study/storage/variantstudy/model/command_listener/command_listener.py
MartinBelthle: Indeed I'll do that

MartinBelthle: In fact Pydantic doesn't handle property decorator really well so I can't validate the model against task_id :/ If you find a way to do so please feel free to provide the way to do it, else I'll keep the old implementation

MartinBelthle:
Still need to add the progress inside DB for the listener and fix the issue on random test I introduced about event buses

antarest/study/storage/variantstudy/model/command_listener/command_listener.py
```diff
@@ -0,0 +1,37 @@
...
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+
+from abc import ABC, abstractmethod
+
+from antarest.core.serialization import AntaresBaseModel
+
+
+class ICommandListener(ABC, AntaresBaseModel):
```
sylvlecl: This should not be a pydantic model, we will not serialize or parse it

antarest/study/service.py
```diff
@@ -213,17 +228,37 @@ def _generate_timeseries(self) -> None:
...
-        self._generate_timeseries()
+        self._generate_timeseries(listener)
         msg = f"Successfully generated thermal timeseries for study '{self._study_id}'"
         notifier(msg)
         return TaskResult(success=True, message=msg)
 
     # Make `ThermalClusterTimeSeriesGeneratorTask` object callable
     __call__ = run_task
 
+    class TsGenerationListener(ICommandListener):
```
sylvlecl: Should be called something else like, this implementation is not specific to timeseries.
Maybe something like EventBusNotifier but it's not great either because it also saves to DB. So maybe just ProgressRecorder or something like that.

antarest/study/service.py
```diff
@@ -284,7 +319,7 @@ def _upgrade_study(self) -> None:
                     file_study = self.storage_service.get_storage(study_to_upgrade).get_raw(study_to_upgrade)
                     file_study.tree.normalize()
 
-    def run_task(self, notifier: TaskUpdateNotifier) -> TaskResult:
+    def run_task(self, notifier: TaskUpdateNotifier, listener: t.Optional[ICommandListener]) -> TaskResult:
```
sylvlecl: We should not need to pass a command listener to all types of tasks, because they don't necessarily need to execute commands.

We need to separate the 2 levels of listeners to be consistent with the task -> commands stack:
- The task listener
- The command listener

Tasks can apply commands, they are higher in the stack.
If we need to have notifications travel up the stack, a command listener can forward to a task listener.

So in practice here, I think we should still have only one argument, either the existing notifier or a TaskListener whhich would have a new "notify_progress" method.

Then, down the stack, when we generate a timeseries, we will create an implementation of ICommandListener which will forward the progress up the stack to the task listener/notifier.

antarest/study/service.py
```diff
@@ -213,17 +228,37 @@ def _generate_timeseries(self) -> None:
...
+        event_bus: IEventBus
+
+        def notify_progress(self, progress: int) -> None:
+            with db():
+                db.session.query(TaskJob).filter(TaskJob.id == self.task_id).update({TaskJob.progress: progress})
+                db.session.commit()
+
+            self.event_bus.push(
+                Event(
+                    type=EventType.TS_GENERATION_PROGRESS,
```
sylvlecl: This should not be an event specific to TS generation, but an event generic to all types of tasks: 
`TASK_PROGRESS`

antarest/study/service.py
```diff
@@ -213,17 +228,37 @@ def _generate_timeseries(self) -> None:
                 )
             )
 
-    def run_task(self, notifier: TaskUpdateNotifier) -> TaskResult:
+    def run_task(self, notifier: TaskUpdateNotifier, listener: t.Optional[ICommandListener]) -> TaskResult:
```
sylvlecl: Se here we should receive still only the task notifier / listener, and in the method we create the command listener implementation which will forward progress to the task listener.

antarest/study/storage/variantstudy/model/command_listener/command_listener.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Interface for all listeners that can be given inside the `apply` method of a command.
+    """
+
+    class Config:
+        extra = "forbid"
+        arbitrary_types_allowed = True
+
+    task_id: str
```
sylvlecl: We should not need this here anymore, once the command listener implementation will just forward to the task listener.

In the task service, we need to make an evolution of the `TaskJobLogRecorder` so that it records the progress to database and pushes progress to event bus, instead of doing it in the `TsGenerationListener` implementation.

antarest/study/service.py
sylvlecl: Finally, we don't need it here:
we should instead make an evolution of `TaskJobLogRecorder` so that it does this for all types of tasks.

Then the command listener implementation we will need to implement is just a forwarding of the progress to the task listener.

sylvlecl (changes requested):
We need to change a little the relation between task notifier and command listener.

The task listener is higher in the stack, and command listener down.
We should not need to have both at the same level. Instead, a command listener can forward events to a task listener.

antarest/core/tasks/service.py
```diff
@@ -94,11 +104,17 @@ def await_task(self, task_id: str, timeout_sec: int = DEFAULT_AWAIT_MAX_TIMEOUT)
...
 
+    def __call__(self, message: str) -> None:
+        return
 
-class TaskJobLogRecorder:
+    def notify_progress(self, progress: int) -> None:
+        return
+
+
+class TaskJobLogRecorder(ITaskNotifier):
```
sylvlecl: We can rename it to TaskLogAndProgressRecorder maybe ? ugly but that's what it does ...

antarest/core/tasks/service.py
```diff
@@ -42,13 +42,23 @@
...
-TaskUpdateNotifier = t.Callable[[str], None]
-Task = t.Callable[[TaskUpdateNotifier], TaskResult]
-
 DEFAULT_AWAIT_MAX_TIMEOUT = 172800  # 48 hours
 """Default timeout for `await_task` in seconds."""
 
 
+class ITaskNotifier(ABC):
+    @abstractmethod
+    def __call__(self, message: str) -> None:
```
sylvlecl: as just discussed, better to use a standard method here, otherwise it's confusing to use this object sometimes as a callable, sometimes as an object which has methods

antarest/core/tasks/service.py
```diff
@@ -138,7 +171,7 @@ def _create_worker_task(
         task_id: str,
         task_type: str,
         task_args: t.Dict[str, t.Union[int, float, bool, str]],
-    ) -> t.Callable[[TaskUpdateNotifier], TaskResult]:
+    ) -> t.Callable[[ITaskNotifier], TaskResult]:
```
sylvlecl: note that we could use `Task` here

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -587,6 +588,7 @@ def generate_task(
         metadata: VariantStudy,
         denormalize: bool = False,
         from_scratch: bool = False,
+        listener: t.Optional[ICommandListener] = None,
```
sylvlecl, line 591: not used anymore

antarest/study/service.py
```diff
@@ -181,6 +182,14 @@ def get_disk_usage(path: t.Union[str, Path]) -> int:
     return total_size
 
 
+class TaskProgressRecorder(ICommandListener):
+    def __init__(self, notifier: ITaskNotifier) -> None:
+        self.notifier = notifier
+
+    def notify_progress(self, progress: int) -> None:
+        return self.notifier.notify_progress(progress)
```
sylvlecl, line 190: Note that in the future, we should probably think of a way of adapting the progress of each command to a global progress of the task (think snapshot generation with X commands).

sylvlecl (changes requested):
Minor stuff

antarest/study/service.py
MartinBelthle: Indeed. We'll refactor this when we'll have to as it's only few lines of code for the moment

sylvlecl (dismissed):

PR #2195, opened 2024-10-21
fix(docker): reduce docker image size

sylvlecl:
PyQT wheel is around 200Mo. Together with the cache which is enabled
in our docker build, it adds up around 400 Mo, whereas it's not used when
not running the desktop version.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - "master"
       - "hotfix/**"
+      - "fix/reduce-docker-image-size"
 
 jobs:
   binary:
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - "master"
       - "hotfix/**"
+      - "fix/reduce-docker-image-size"
```
MartinBelthle: Remove this before merge

.github/workflows/deploy.yml
```diff
@@ -47,8 +48,8 @@ jobs:
       - name: 🐍 Install development dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install pydantic --no-binary pydantic
           pip install -r requirements-dev.txt
+          pip install -r requirements-desktop.txt
```
MartinBelthle: Use this for worker.yml and for mypy

MartinBelthle:
The Desktop app works for Ubuntu 20

MartinBelthle (approved):

PR #2199, opened 2024-10-23
fix(archive): raise Exception when (un)archiving wrong outputs

MartinBelthle:
Fix [ANT-2369]

### DIFF
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -37,6 +37,9 @@ from antarest.core.exceptions import (
     FileDeletionNotAllowed,
     IncorrectPathError,
     NotAManagedStudyException,
+    OutputAlreadyArchived,
+    OutputAlreadyUnarchived,
+    OutputNotFound,
     ReferencedObjectDeletionNotAllowed,
     StudyDeletionNotAllowed,
     StudyNotFoundError,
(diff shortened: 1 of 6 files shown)


antarest/study/service.py
```diff
@@ -2379,6 +2394,12 @@ def unarchive_output(
         assert_permission(params.user, study, StudyPermissionType.READ)
         self._assert_study_unarchived(study)
 
+        output_path = Path(study.path) / "output" / output_id
+        if not is_output_archived(output_path):
```
TheoPascoli, line 2405: not clear to me. Maybe this is better: 
if output_path.exists():
__if not is_output_archived(output_path):
____raise OutputAlreadyUnarchived(output_id)
else:
__raise OutputNotFound(output_id)

MartinBelthle: Yep it's better

TheoPascoli (approved):

antarest/study/service.py
MartinBelthle: In fact it won't work :/ because with the if in this order if an output is unarchived we'll go inside the OutputNotFound (as  if output_path.exists() is False)

PR #2200, opened 2024-10-24
feat(load): add arrow endpoints

TheoPascoli:


antarest/study/web/study_data_blueprint.py
```diff
@@ -524,6 +525,40 @@ def update_inflow_structure(
...
+        logger.info(
+            msg=f"Getting load series data for area {area_id} of study {uuid}",
+            extra={"user": current_user.id},
+        )
+        params = RequestParameters(user=current_user)
+        study = study_service.check_study_access(uuid, StudyPermissionType.READ, params)
+        return study_service.load_manager.get_load_matrix(study, area_id, matrix_format)
+
+    @bp.put(
+        "/{uuid}/{area_id}/load/series",
```
MartinBelthle: lacks `studies` before /{uuid}

antarest/study/web/study_data_blueprint.py
```diff
@@ -524,6 +525,40 @@ def update_inflow_structure(
         study = study_service.check_study_access(uuid, StudyPermissionType.WRITE, params)
         study_service.hydro_manager.update_inflow_structure(study, area_id, values)
 
+    @bp.get(
+        "/{uuid}/{area_id}/load/series",
```
MartinBelthle: Same

requirements.txt
```diff
@@ -38,6 +38,7 @@ pandas~=2.2.3
 paramiko~=3.4.1
 plyer~=2.0.0
 psycopg2-binary~=2.9.9
+pyarrow~=18.0.0
```
MartinBelthle: Perhaps also add pyarrow-stubs inside requirements-dev.txt. I did it inside my old PR probably to shush mypy issues

antarest/study/business/load_management.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if matrix_format == MatrixFormat.JSON:
+            return JSONResponse(content=cast(JSON, matrix_data))
+
+        if matrix_format == MatrixFormat.ARROW:
+            matrix_df = cast(pd.DataFrame, matrix_data)
+            matrix_df.columns = matrix_df.columns.map(str)
+            buffer = io.BytesIO()
+            matrix_df.to_feather(buffer, compression="uncompressed")
+            return Response(content=buffer.getvalue(), media_type="application/octet-stream")
```
MartinBelthle: In my old PR I used `Response(content=output, media_type="application/vnd.apache.arrow.file")` perhaps it's more specific than just bytes ?

antarest/study/business/load_management.py
```diff
@@ -0,0 +1,62 @@
...
+    def get_load_matrix(self, study: Study, area_id: str, matrix_format: MatrixFormat) -> Response:
+        file_study = self.storage_service.get_storage(study).get_raw(study)
+
+        load_path = LOAD_PATH.format(area_id=area_id).split("/")
+        node = file_study.tree.get_node(load_path)
+
+        if not isinstance(node, InputSeriesMatrix):
+            return Response(content="Invalid node type", status_code=400)
+
+        matrix_data = InputSeriesMatrix.parse(node, return_dataframe=(matrix_format == MatrixFormat.ARROW))
```
MartinBelthle, line 40: Not your fault but we lose some time here if we have a matrix registered internally. Because of this code inside `input_series_matrix`:
```python
 if link_path.exists():
    link = link_path.read_text()
    matrix_json = self.context.resolver.resolve(link)
    matrix_json = self.context.resolver.resolve(link, format="json")
    matrix_json = cast(JSON, matrix_json)
    matrix: pd.DataFrame = pd.DataFrame(**matrix_json)
```
so I think we should change this code and probably the code of the resolver to avoid this JSON conversion

antarest/study/service.py
```diff
@@ -365,6 +366,7 @@ def __init__(
         self.adequacy_patch_manager = AdequacyPatchManager(self.storage_service)
         self.advanced_parameters_manager = AdvancedParamsManager(self.storage_service)
         self.hydro_manager = HydroManager(self.storage_service)
+        self.load_manager = LoadManager(self.storage_service)
```
MartinBelthle: Generally speaking for the other matrices, what is your vision here ? For hydro for instance, we want to add the code inside the HydroManager or do we want to create a specific manager on the side ? I personally prefer the 1st option

MartinBelthle (changes requested):

antarest/study/business/load_management.py
```diff
@@ -0,0 +1,60 @@
...
+from antarest.study.storage.storage_service import StudyStorageService
+
+LOAD_PATH = "input/load/series/load_{area_id}"
+
+
+class LoadManager:
+    def __init__(self, storage_service: StudyStorageService) -> None:
+        self.storage_service = storage_service
+
+    def get_load_matrix(self, study: Study, area_id: str) -> Response:
```
sylvlecl: This "business" layer should be independent from web considerations, it should return a dataframe instead.

Formatting to arrow should be done in the web layer.

antarest/study/web/study_data_blueprint.py
```diff
@@ -524,6 +525,41 @@ def update_inflow_structure(
...
+        params = RequestParameters(user=current_user)
+        study = study_service.check_study_access(uuid, StudyPermissionType.READ, params)
+        return study_service.load_manager.get_load_matrix(study, area_id)
+
+    @bp.put(
+        "/studies/{uuid}/{area_id}/load/series",
+        tags=[APITag.study_data],
+        summary="Update load series data",
+    )
+    def update_load_series(
```
sylvlecl: If we want to support arrow input, I don't think we will be able to use a pydantic model (LoadDTO), because they expect JSON, which cannot contain binary data.

So: if we want the same URL to support at the same time JSON and arrow,  we will need  to separate it in 2 functions I think, for example discriminated by the "accept-content" header.
In any case, I think the function which will accept arrow input will need to have something else than a pydantic model as its argument (bytes or something else, I did not look into fastapi support for binary input).

sylvlecl (changes requested):
First comments

antarest/study/web/study_data_blueprint.py
TheoPascoli: Eventually, we decided to only support Arrow format. I'll check to see what is best to handle binary inputs

antarest/study/business/load_management.py
TheoPascoli: Done

PR #2202, opened 2024-10-28
feat(commands): add study_version information inside commands

MartinBelthle:
This PR is the stepping stone for next developments, it doesn't bring a lot in itself.
But thanks to this, when creating a command, we'll know the associated study version. 
It can help us validating data at the creation, thing we only used to do inside the method `apply` as it was the only one knowing the study version.
So in a next development we can type parameters given to command such as parameters for thermal creation :)


alembic/versions/00a9ceb38842_add_study_version_to_commands.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # Gets studies versions
+    study_info = session.query(study_table).filter(study_table.c.id.in_(mapping)).all()
+
+    # Clean orphan commands
+    study_ids = {study_id for study_id, _ in study_info}
+    ids_to_remove = {cmd_id for study_id, cmd_ids in mapping.items() if study_id not in study_ids for cmd_id in cmd_ids}
+    bind.execute(commandblock_table.delete().where(commandblock_table.c.id.in_(ids_to_remove)))
+
+    # Insert new values
```
sylvlecl, line 50: I'm afraid this will take very long to execute on production env, causing errors during the deployment process.
I think we should be able to do this more efficiently in only one statement using a join.

MartinBelthle: Didn't use a join but I believe it will be faster. If you think of another SQL request don't mind putting it here for me to test it

sylvlecl: Ok it looks good!

antarest/study/storage/variantstudy/model/model.py
```diff
@@ -58,21 +60,47 @@ class GenerationResultInfoDTO(AntaresBaseModel):
...
         args: The arguments for the command action.
         version: The version of the command.
+        study_version: The version of the study associated to the command.
     """
 
     id: t.Optional[str] = None
     action: str
     args: t.Union[t.MutableSequence[JSON], JSON]
     version: int = 1
+    study_version: StudyVersion
```
sylvlecl: I think it would be better to serialize it to its string representation, when storing it, otherwise we become very dependent to possible changes to this class (we must guarantee that we will be able to read it again from DB).

We use StudyVersion in quite a lot of pydantic model now, I think we could share an `AnnotatedType` which will parse it on validation and serialize it to string.

sylvlecl (changes requested):
Small comment about how version is stored in serialized form + proposition to share an annotated type.

Otherwise looks good!

antarest/study/storage/variantstudy/model/model.py
MartinBelthle: Okay so to be sure I understand:

1- you prefer having a string here
2- Perform a validation at the creation to ensure it can be parsed as a StudyVersion
3- Serialize it as a string
4- Put 2- and 3- inside a `AnnotatedType` to be use elsewehere, for example inside `StudyMetadataDTO` where I already did these 2 steps (ah it won't work the same as I'm serializing as an int)

Am I right ?

sylvlecl: Yes for 2-3-4 but not what I meant for 1, I think:
with an AnnotatedType, you can still have a `StudyVersion` object inside the class, but say that you want to parse it and serialize it as a string, like:
```python
study_version: t.Annotated[
    StudyVersion,
    BeforeValidator(StudyVersion.parse),
    PlainSerializer(str),
]
```

Then we can define a name for this annotated type to re-use it where relevant (not sure of a good name here, but for example):
```python
StudyVersionStr = t.Annotated[
    StudyVersion,
    BeforeValidator(StudyVersion.parse),
    PlainSerializer(str),
]

...
study_version: StudyVersionStr
# then use it as a StudyVersion everywhere in the rest of the code
```

sylvlecl (dismissed):

PR #2213, opened 2024-11-12
feat(auto_archive_service): increase cleaning snapshot frequency

maugde:
Fix for [ANT-1952]
Update snapshot retention days value in code and documentation.
Fix context and admin rights to execute the task.

### DIFF
--- a/antarest/core/config.py
+++ b/antarest/core/config.py
@@ -155,7 +155,7 @@ class StorageConfig:
     auto_archive_dry_run: bool = False
     auto_archive_sleeping_time: int = 3600
     auto_archive_max_parallel: int = 5
-    snapshot_retention_days: int = 7
+    snapshot_retention_days: int = 1
 
     @classmethod
     def from_dict(cls, data: JSON) -> "StorageConfig":
--- a/antarest/core/config.py
+++ b/antarest/core/config.py
@@ -155,7 +155,7 @@ class StorageConfig:
     auto_archive_dry_run: bool = False
     auto_archive_sleeping_time: int = 3600
     auto_archive_max_parallel: int = 5
-    snapshot_retention_days: int = 1
+    snapshot_retention_days: int = 7
 
     @classmethod
     def from_dict(cls, data: JSON) -> "StorageConfig":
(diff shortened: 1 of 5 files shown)


antarest/study/storage/auto_archive_service.py
```diff
@@ -84,10 +84,11 @@ def _try_archive_studies(self) -> None:
                     f"Failed to auto archive study {study_id}",
                     exc_info=e,
                 )
-
-        self.study_service.storage_service.variant_study_service.clear_all_snapshots(
-            datetime.timedelta(days=self.config.storage.snapshot_retention_days)
-        )
+        with db():
+            self.study_service.storage_service.variant_study_service.clear_all_snapshots(
```
MartinBelthle, line 88: Change signature of `clear_all_snapshots` to always have params (currently there are optional)

antarest/core/config.py
```diff
@@ -155,7 +155,7 @@ class StorageConfig:
     auto_archive_dry_run: bool = False
     auto_archive_sleeping_time: int = 3600
     auto_archive_max_parallel: int = 5
-    snapshot_retention_days: int = 1
+    snapshot_retention_days: int = 7
```
MartinBelthle, line 158: I don't see the change in the doc which is really weird as  I see it inside the last commit.

MartinBelthle (changes requested):
Inside the JIRA ticket it's written: 
Déclenchement du traitement de suppression fréquente des snapshots sur la date de modification et la date du dernier accès. Il faut que les deux dates soient supérieures à 7 pour que le traitement se lance.

I don't see this change in this PR, is this because it was already the case ?
If it wasn't please adapt the code

maugde:
Default `snapshot_retention_days` value must be set à 7 days so the test on `last_access and updated_at should pass without additional checks.

maugde:
> Inside the JIRA ticket it's written: Déclenchement du traitement de suppression fréquente des snapshots sur la date de modification et la date du dernier accès. Il faut que les deux dates soient supérieures à 7 pour que le traitement se lance.
> 
> I don't see this change in this PR, is this because it was already the case ? If it wasn't please adapt the code

Default `snapshot_retention_days` value must be set à 7 days so the test on `last_access` and `updated_at` should pass without additional checks.

MartinBelthle (approved):
Okay for me

PR #2216, opened 2024-11-12
feat(raw): allow folder creation inside `user` folder

MartinBelthle:
Fix [ANT-2072]


antarest/study/service.py
```diff
@@ -2750,3 +2752,31 @@ def delete_file_or_folder(self, study_id: str, path: str, current_user: JWTUser)
...
+        study = self.get_study(study_id)
+        assert_permission(current_user, study, StudyPermissionType.WRITE)
+
+        url = [item for item in path.split("/") if item]
+        if len(url) < 2 or url[0] != "user":
+            raise FolderCreationNotAllowed(f"the given path isn't inside the 'User' folder: {path}")
+        if url[1] == "expansion":
+            raise FolderCreationNotAllowed(f"the given path shouldn't be inside the 'expansion' folder: {path}")
+
+        study_tree = self.storage_service.raw_study_service.get_raw(study, True).tree
```
sylvlecl: Don't we want to support this for variant studies too ?

If not, shouldn't we throw an explicit error if the study is not a raw study ?

If yes, we need to do the editing through a command (probably the existing UpdateConfig)

MartinBelthle: I didn't think about this at all :scream: . I believe we have the same issue in the deletion endpoint just above. I'll test it.
The easy solution will be to forbid it but it doesn't really make sense from an user perspective so I'll think about something.

MartinBelthle:
The issue we have is that we don't have commands to create/delete folders and also to delete a file.
I think if we want to support these endpoints for variants, I have no other choice than creating a command AlterFileSystem or something like this that allows this

antarest/study/service.py
makdeuneuv: Same, I don't think about this use case. I'm for the easy solution from Martin

MartinBelthle: I introduced commands to make the code works for variants too as it makes more sense and Alexander wanted that.
Also, this fix an issue that we had before, creating a file / folder inside a variant `user` folder didn't work

antarest/study/storage/variantstudy/model/command/create_user_resource.py
```diff
@@ -0,0 +1,84 @@
...
+    # ===================
+
+    command_name: CommandName = CommandName.CREATE_USER_RESOURCE
+    version: int = 1
+
+    # Command parameters
+    # ==================
+
+    path: str
+    file: bool
```
sylvlecl: More clear to have an enum than a bool for this

antarest/study/storage/variantstudy/model/command/create_user_resource.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def _apply(self, study_data: FileStudy, listener: t.Optional[ICommandListener] = None) -> CommandOutput:
+        url = [item for item in self.path.split("/") if item]
+        if len(url) < 2 or url[0] != "user":
+            return CommandOutput(status=False, message=f"the given path isn't inside the 'User' folder: {self.path}")
+        if url[1] == "expansion":
+            return CommandOutput(
+                status=False, message=f"the given path shouldn't be inside the 'expansion' folder: {self.path}"
+            )
+
```
sylvlecl: Shouldn't we have something about registered files here ? (maybe handling xpansion case too)

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -528,15 +526,18 @@ def replace_study_file(
         - `uuid`: The UUID of the study.
         - `path`: The path to the data to update. Defaults to "/".
         - `file`: The raw file to be posted (e.g. a CSV file opened in binary mode).
-        - `create_missing`: Flag to indicate whether to create file or parent directories if missing.
+        - `create_missing`: Flag to indicate whether to create file and parent directories if missing.
+        - `is_folder`: When True and `create_missing` is True, creates a folder. Else (default value), it's ignored.
```
sylvlecl: would prefer to have an enum here, type=file or folder, it will be more explicit

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -513,11 +510,12 @@ def edit_study(
     def replace_study_file(
         uuid: str,
         path: str = Param("/", examples=get_path_examples()),  # type: ignore
-        file: bytes = File(...),
+        file: bytes = File(default=b""),
```
sylvlecl: I think it would more make sense to have a really optional argument defaulting to None, not something defaulting to empty

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -528,15 +526,18 @@ def replace_study_file(
...
         """
-        logger.info(
-            f"Uploading new data file at {path} for study {uuid}",
-            extra={"user": current_user.id},
-        )
         path = sanitize_string(path)
         params = RequestParameters(user=current_user)
-        study_service.edit_study(uuid, path, file, params, create_missing=create_missing)
+        if is_folder and create_missing:
+            logger.info(f"Creating folder {path} for study {uuid}", extra={"user": current_user.id})
```
sylvlecl, line 542: Note for another work: it does not make sense to have to explicitly append the user at every call, this should be added at the request level like the request ID.

antarest/study/storage/variantstudy/model/command/create_user_resource.py
```diff
@@ -0,0 +1,84 @@
...
+    path: str
+    file: bool
+
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> t.Tuple[CommandOutput, t.Dict[str, t.Any]]:
+        return CommandOutput(status=True, message="ok"), {}
+
+    def _apply(self, study_data: FileStudy, listener: t.Optional[ICommandListener] = None) -> CommandOutput:
+        url = [item for item in self.path.split("/") if item]
+        if len(url) < 2 or url[0] != "user":
```
sylvlecl: Same comment as for removal: we should have removed the user part upstream.

antarest/study/storage/variantstudy/model/command/remove_user_resource.py
```diff
@@ -0,0 +1,76 @@
...
+    # ==================
+
+    path: str
+
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> t.Tuple[CommandOutput, t.Dict[str, t.Any]]:
+        return CommandOutput(status=True, message="ok"), {}
+
+    def _apply(self, study_data: FileStudy, listener: t.Optional[ICommandListener] = None) -> CommandOutput:
+        url = [item for item in self.path.split("/") if item]
+        if len(url) < 2 or url[0] != "user":
```
sylvlecl: Because this class is supposed to work in the context of the "user" space, I think this path should not include the `user` part, it's redundant.
We should remove this part upstream.

Also, don't we want to forbid also xpansion deletion here ? Or is it handled by "registered files" below ?
I think we should have a common method somewhere to check that a URL is a user-writable URL.

antarest/study/service.py
```diff
@@ -2731,26 +2709,48 @@ def delete_file_or_folder(self, study_id: str, path: str, current_user: JWTUser)
...
         """
-        study = self.get_study(study_id)
-        assert_permission(current_user, study, StudyPermissionType.WRITE)
+        args = {"path": path}
+        self._alter_user_folder(study_id, args, current_user, RemoveUserResource, ResourceDeletionNotAllowed)
 
-        url = [item for item in path.split("/") if item]
-        if len(url) < 2 or url[0] != "user":
-            raise FileDeletionNotAllowed(f"the targeted data isn't inside the 'User' folder: {path}")
+    def create_folder(self, study_id: str, path: str, current_user: JWTUser) -> None:
```
sylvlecl: If only creation in user space is allowed, why not just have a method `create_user_folder` ?

antarest/study/service.py
```diff
@@ -2731,26 +2709,48 @@ def delete_file_or_folder(self, study_id: str, path: str, current_user: JWTUser)
...
+        Raises:
+            FolderCreationNotAllowed: if the path does not comply with the above rules
+        """
+        args = {"path": path, "file": False}
+        self._alter_user_folder(study_id, args, current_user, CreateUserResource, FolderCreationNotAllowed)
+
+    def _alter_user_folder(
+        self,
+        study_id: str,
+        args: t.Dict[str, t.Any],
```
sylvlecl: We should use stronger typing here, the signature is not clear enough.
For example, we could have a small `CreateUserResourceData` class that contains `path` and `file`. This class could be used as the input to `CreateUserResource` creation and inside it.
This si actually a scheme that we could use for all command types, in order to use pydantic models for their associated data.

sylvlecl (changes requested):
A few questions/suggestions.

In particular, if the new commands are dedicated to working inside the user "space", the path should not contain "user" but only the path inside the user space.

antarest/study/web/raw_studies_blueprint.py
MartinBelthle: Indeed. I'll resolve it as i won't tackle it inside this PR

antarest/study/web/raw_studies_blueprint.py
MartinBelthle: Okay but what's the behavior you expect when someone doesn't fill this parameter ? As for empty bytes it will simply erase all content inside an existing file or create an empty one if create_missing=True

antarest/study/storage/variantstudy/model/command/remove_user_resource.py
MartinBelthle: Concerning the deletion it's handled by the `registered_files`.
Indeed I will create a method to use it inside the 2 commands

MartinBelthle: And I removed the user part outside the command

antarest/study/service.py
MartinBelthle: I introduced such classes I'll let you take a look

antarest/study/web/raw_studies_blueprint.py
sylvlecl: I think we should throw a 4XX if we have a `file` argument with a `ResourceType.FOLDER`, or if we have no `file` argument with a `ResourceType.FILE`.
To do those checks we need to differentiate between "empty but present" and "absent".

antarest/study/service.py
sylvlecl: Nice but it would be more clear to use composition instead of inheritance here:
instead of 
```python
class CreateUserResource(ICommand, CreateUserResourceData):
```
let's use
```python
class CreateUserResource(ICommand):
    data: CreateUserResourceData
```

sylvlecl (approved):

PR #2221, opened 2024-11-13
refactor(link,ui-api): update DTO format for link

skamril:
ANT-2312 ANT-1882


webapp/src/redux/ducks/studyMaps.ts
```diff
@@ -167,11 +170,11 @@ export const setLayers = createAction<
...
-  [R.equals("dot"), (): LinkStyle => [[1, 5], "round"]],
-  [R.equals("dash"), (): LinkStyle => [[16, 8], "square"]],
-  [R.equals("dotdash"), (): LinkStyle => [[10, 6, 1, 6], "square"]],
-  [R.T, (): LinkStyle => [[0], "butt"]],
+const makeLinkStyle = R.cond<[TLinkStyle], LinkStyle>([
+  [(v) => v === "dot", () => [[1, 5], "round"]],
+  [(v) => v === "dash", () => [[16, 8], "square"]],
+  [(v) => v === "dotdash", () => [[10, 6, 1, 6], "square"]],
+  [R.T, () => [[0], "butt"]],
 ]);
```
hdinia, line 178: here I would use a simple object lookup:
```typescript
const LINK_STYLES: Record<TLinkStyle, LinkStyle> = {
  dot: [[1, 5], "round"],
  dash: [[16, 8], "square"],
  dotdash: [[10, 6, 1, 6], "square"],
  default: [[0], "butt"]
};

const makeLinkStyle = (style: TLinkStyle) => LINK_STYLES[style] || LINK_STYLES.default;
```

webapp/src/services/api/studies/links/index.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @param params.areaTo - The to area name.
+ * @returns The deleted link id (format: `${areaFromId}%${areaToId}`)
+ */
+export async function deleteLink(params: DeleteLinkParams) {
+  const { studyId, areaFrom, areaTo } = params;
+  const { data } = await client.delete<string>(
+    `/v1/studies/${studyId}/links/${areaFrom}/${areaTo}`,
+  );
+  return data;
+}
```
hdinia, line 50: you can group all API calls into a single `linksApi` object, and destructure parameters directly in function arguments
```typescript
export const linksApi = {
  createLink: async ({ studyId, ...body }: CreateLinkParams) => {
    const { data } = await client.post<LinkDTO>(
      `/v1/studies/${studyId}/links`,
      body
    );
    return data;
  },

  getLinks: async ({ studyId }: { studyId: StudyMetadata["id"] }) => {
    const { data } = await client.get<LinkDTO[]>(
      `/v1/studies/${studyId}/links`
    );
    return data;
  },

  deleteLink: async ({ studyId, areaFrom, areaTo }: DeleteLinkParams) => {
    const { data } = await client.delete<string>(
      `/v1/studies/${studyId}/links/${areaFrom}/${areaTo}`
    );
    return data;
  }
};
```
Usage:
```typescript
import { linksApi } from './api/links/linksApi';

await linksApi.createLink({ studyId, areaFrom, areaTo });
await linksApi.getLinks({ studyId });
await linksApi.deleteLink({ studyId, areaFrom, areaTo });
```

antarest/study/business/link_management.py
```diff
@@ -51,38 +41,39 @@ class LinkManager:
...
-        result = []
+        result: t.List[LinkDTO] = []
+
         for area_id, area in file_study.config.areas.items():
-            links_config: t.Optional[t.Dict[str, t.Any]] = None
-            if with_ui:
-                links_config = file_study.tree.get(["input", "links", area_id, "properties"])
+            links_config = file_study.tree.get(["input", "links", area_id, "properties"])
+
             for link in area.links:
```
MartinBelthle, line 51: It's not really clear for me. I propose something I find clearer but we can keep the code as it is
This
```python
links_config = file_study.tree.get(["input", "links", area_id, "properties"])
for link in area.links:
    link_tree_config: t.Dict[str, t.Any] = links_config[link]
    link_tree_config.update({"area1": area_id, "area2": link})
```
becomes this
```python
links_config = file_study.tree.get(["input", "links", area_id, "properties"])
for area2, link_properties in links_config.items():
    link_properties.update({"area1": area_id, "area2": area2})
```

antarest/study/business/link_management.py
```diff
@@ -51,38 +41,39 @@ class LinkManager:
...
+                link_tree_config.update({"area1": area_id, "area2": link})
+
+                link_internal = LinkInternal.model_validate(link_tree_config)
+
+                result.append(link_internal.to_dto())
 
         return result
 
-    def create_link(self, study: RawStudy, link_creation_info: LinkInfoDTO) -> LinkInfoDTO:
+    def create_link(self, study: RawStudy, link_creation_dto: LinkDTO) -> LinkDTO:
```
MartinBelthle: Why do we have a `RawStudy` here ? It should be just a `Study`

antarest/study/business/model/link_model.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class Area(AntaresBaseModel):
+    area1: str
+    area2: str
+
+    @model_validator(mode="after")
+    def validate_areas(self) -> t.Self:
+        if self.area1 == self.area2:
+            raise LinkValidationError(f"Cannot create link on same area: {self.area1}")
```
MartinBelthle: Perhaps the msg would be clearer with `Cannot create a link that goes from and to the same single area`

hdinia (dismissed):
Looks good to me for the frontend part

antarest/study/business/model/link_model.py
```diff
@@ -0,0 +1,129 @@
...
+    colorb: int = Field(default=DEFAULT_COLOR, ge=0, le=255)
+    colorg: int = Field(default=DEFAULT_COLOR, ge=0, le=255)
+    link_width: float = 1
+    link_style: LinkStyle = LinkStyle.PLAIN
+
+    filter_synthesis: t.Optional[comma_separated_enum_list] = FILTER_VALUES
+    filter_year_by_year: t.Optional[comma_separated_enum_list] = FILTER_VALUES
+
+    def to_internal(self, version: StudyVersion) -> "LinkInternal":
+        if version < STUDY_VERSION_8_2 and (
```
MartinBelthle: could be simplified in 
```python
and {"filter_synthesis", "filter_year_by_year"} & self.model_fields_set:
```
 i believe

antarest/study/business/model/link_model.py
```diff
@@ -0,0 +1,129 @@
...
+            "filter_synthesis" in self.model_fields_set or "filter_year_by_year" in self.model_fields_set
+        ):
+            raise LinkValidationError("Cannot specify a filter value for study's version earlier than v8.2")
+
+        return LinkInternal(
```
MartinBelthle: Couldn't you use self.model_dump() while excluding according to version instead of all this lines ?

antarest/study/business/model/link_model.py
```diff
@@ -0,0 +1,129 @@
...
+    colorr: int = Field(default=DEFAULT_COLOR, ge=0, le=255)
+    colorb: int = Field(default=DEFAULT_COLOR, ge=0, le=255)
+    colorg: int = Field(default=DEFAULT_COLOR, ge=0, le=255)
+    link_width: float = 1
+    link_style: LinkStyle = LinkStyle.PLAIN
+    filter_synthesis: t.Optional[comma_separated_enum_list] = FILTER_VALUES
+    filter_year_by_year: t.Optional[comma_separated_enum_list] = FILTER_VALUES
+
+    def to_dto(self) -> LinkDTO:
+        return LinkDTO(
```
MartinBelthle: Same comment as above

antarest/study/business/table_mode_management.py
```diff
@@ -98,8 +98,9 @@ def _get_table_data_unsafe(self, study: RawStudy, table_type: TableModeType) ->
             data = {area_id: area.model_dump(mode="json", by_alias=True) for area_id, area in areas_map.items()}
         elif table_type == TableModeType.LINK:
             links_map = self._link_manager.get_all_links_props(study)
+            excludes = set() if int(study.version) >= 820 else {"filter_synthesis", "filter_year_by_year"}
```
MartinBelthle: Use study_version instead of int here

antarest/study/business/table_mode_management.py
```diff
@@ -195,8 +196,9 @@ def update_table_data(
         elif table_type == TableModeType.LINK:
             links_map = {tuple(key.split(" / ")): LinkOutput(**values) for key, values in data.items()}
             updated_map = self._link_manager.update_links_props(study, links_map)  # type: ignore
+            excludes = set() if int(study.version) >= 820 else {"filter_synthesis", "filter_year_by_year"}
```
MartinBelthle: Same

antarest/study/storage/rawstudy/model/filesystem/config/links.py
```diff
@@ -66,6 +67,74 @@ class TransmissionCapacity(EnumIgnoreCase):
...
+    DAILY = "daily"
+    WEEKLY = "weekly"
+    MONTHLY = "monthly"
+    ANNUAL = "annual"
+
+
+def validate_filters(
+    filter_value: t.Union[t.List[FilterOption], str], enum_cls: t.Type[FilterOption]
+) -> t.List[FilterOption]:
+    if filter_value is not None and isinstance(filter_value, str):
```
MartinBelthle: Could be simplified in isinstance(filter_value, str) as isinstance(None) always returns False

antarest/study/storage/rawstudy/model/filesystem/config/links.py
```diff
@@ -66,6 +67,74 @@ class TransmissionCapacity(EnumIgnoreCase):
...
+                f"Invalid value(s) in filters: {', '.join(invalid_options)}. "
+                f"Allowed values are: {', '.join(filter_accepted_values)}."
+            )
+
+        return [enum_cls(opt) for opt in options]
+
+    return filter_value
+
+
+def join_with_comma(values: t.List[FilterOption]) -> str:
```
MartinBelthle, line 127: Is it really useful ? Also, it also lowers the values so we could add this inside the function name. 
I think we should remove it as it's only used once

antarest/study/storage/rawstudy/model/filesystem/config/links.py
```diff
@@ -159,11 +228,12 @@ def _validate_colors(cls, values: t.MutableMapping[str, t.Any]) -> t.Mapping[str
         return validate_colors(values)
 
     # noinspection SpellCheckingInspection
-    def to_config(self) -> t.Dict[str, t.Any]:
+    def to_ini(self, version: int) -> t.Dict[str, t.Any]:
```
MartinBelthle: I see no usage of this method so it should be removed.

tests/storage/business/test_arealink_manager.py
```diff
@@ -134,8 +136,16 @@ def test_area_crud(empty_study: FileStudy, matrix_service: SimpleMatrixService):
...
+    study.version = 820
+    link_manager.create_link(
+        study,
+        LinkDTO(
+            area1="test",
+            area2="test2",
+        ),
+    )
     assert empty_study.config.areas["test"].links.get("test2") is not None
+    study.version = -1
```
MartinBelthle: Why don't use simply put the version at 820 at the beginning and don't change it ?

tests/storage/business/test_arealink_manager.py
```diff
@@ -216,7 +227,52 @@ def test_area_crud(empty_study: FileStudy, matrix_service: SimpleMatrixService):
     )
 
     area_manager.create_area(study, AreaCreationDTO(name="test2", type=AreaType.AREA))
-    link_manager.create_link(study, LinkInfoDTO(area1="test", area2="test2"))
+    study.version = 880
```
MartinBelthle: Same here

tests/variantstudy/model/command/test_create_link.py
```diff
@@ -258,15 +250,17 @@ def test_create_diff(command_context: CommandContext):
     other_match = CreateLink(
         area1="foo",
         area2="bar",
-        parameters={"hurdles-cost": "true"},
```
MartinBelthle, line 261: It doesn't work anymore with these values anymore ? If so it's an issue as R scripts will continue calling the command with this writing.

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -211,13 +140,16 @@ def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = N
         output, data = self._apply_config(study_data.config)
         if not output.status:
             return output
+
+        validated_properties = LinkInternal.model_validate(self.parameters).model_dump(
```
MartinBelthle, line 148: I think you should add filters to the excludes here depending on the version as its seems you always write them into the file

MartinBelthle (changes requested):
Several minor comments

antarest/study/business/link_management.py
TheoPascoli: Hello Martin, it does not work that way. I think its really acceptable right now and Im not sure its worth it to spend time trying to make your proposition works

antarest/study/business/link_management.py
TheoPascoli: Yes. You will have to explain to me the differences though so I dont make this mistake again

antarest/study/business/model/link_model.py
TheoPascoli: Sure

antarest/study/storage/rawstudy/model/filesystem/config/links.py
TheoPascoli: I like it to have a method even if its use one time, its clear at first sight the transformation done in the serialization thanks to the method name. We can discuss it

antarest/study/storage/rawstudy/model/filesystem/config/links.py
TheoPascoli: I removed it and everything still works so I guess you're right

tests/storage/business/test_arealink_manager.py
TheoPascoli: Yes

tests/variantstudy/model/command/test_create_link.py
TheoPascoli: I set it back and everything is fine

webapp/src/redux/ducks/studyMaps.ts
skamril: True, I will update

webapp/src/services/api/studies/links/index.ts
skamril: I don't like to force like this, with the standard export you can do as you want:

`import { createLink, getLinks } from "..."`
`import * as linksApi from "..."`

And with an object it's difficult to combine functions.

antarest/study/storage/rawstudy/model/filesystem/config/links.py
MartinBelthle: Why not it doesn't really matter

antarest/study/business/link_management.py
MartinBelthle: RawStudy and VariantStudy both inherits from Study.
- A raw study is a study with an existing filesystem. You can modify it as you want.
- A variantStudy is a study that has a parent study and a list of commands that are stored in DB. It doesn't have a filesystem. When you modify it, it simply adds the corresponding command to the DB.
When you want to get some info about a variant study we generate its snapshot which is the parent filesystem + the DB commands that are applied. Once in a while this snapshot is cleaned to avoid taking some place on the FS.


We rarely specify which type we expect and here it doesn't make sense to make a distinction as we can have both types.

antarest/study/business/link_management.py
MartinBelthle: Ok

antarest/study/business/table_mode_management.py
```diff
@@ -98,8 +98,9 @@ def _get_table_data_unsafe(self, study: RawStudy, table_type: TableModeType) ->
             data = {area_id: area.model_dump(mode="json", by_alias=True) for area_id, area in areas_map.items()}
         elif table_type == TableModeType.LINK:
             links_map = self._link_manager.get_all_links_props(study)
+            excludes = set() if int(study.version) >= STUDY_VERSION_8_2 else {"filter_synthesis", "filter_year_by_year"}
```
MartinBelthle: In the rest of the code we no longer do `int(study.version)`, we do `StudyVersion.parse(study.version)`

antarest/study/business/link_management.py
```diff
@@ -51,38 +41,39 @@ class LinkManager:
     def __init__(self, storage_service: StudyStorageService) -> None:
         self.storage_service = storage_service
 
-    def get_all_links(self, study: RawStudy, with_ui: bool = False) -> t.List[LinkInfoDTO]:
+    def get_all_links(self, study: RawStudy) -> t.List[LinkDTO]:
```
MartinBelthle: I didn't see it last time but same remark, use `Study` instead of `RawStudy`

MartinBelthle (changes requested):
2 minor remarks, otherwise it's good to me.
I'll let Sylvain take a look.

*NB*: When this PR (https://github.com/AntaresSimulatorTeam/AntaREST/pull/2202) will be merge we'll have to change the `parameters` inside the command `CreateLink` to make it a pydantic object and no longer a Dict

MartinBelthle (approved):

sylvlecl (approved):

PR #2226, opened 2024-11-14
feat(ui-settings): allow to change app language

skamril:
ANT-2276


webapp/src/i18n.ts
hdinia, line 1: suggestion: as it grows this file could benefit a small refacto

```typescript
/**
 * Copyright (c) 2024, RTE (https://www.rte-france.com)
 *
 * See AUTHORS.txt
 *
 * This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 *
 * SPDX-License-Identifier: MPL-2.0
 *
 * This file is part of the Antares project.
 */

import i18n from "i18next";
import Backend from "i18next-http-backend";
import LanguageDetector from "i18next-browser-languagedetector";
import { initReactI18next } from "react-i18next";

export const SUPPORTED_LANGUAGES = ["en", "fr"] as const;
export type SupportedLanguage = (typeof SUPPORTED_LANGUAGES)[number];

const backendConfig = {
  loadPath: `${import.meta.env.BASE_URL}locales/{{lng}}/{{ns}}.json?id=${__BUILD_TIMESTAMP__}`,
};

const detectionConfig = {
  convertDetectedLanguage: (lng: string) => lng.split("-")[0], // Remove region code
};

const i18nConfig = {
  supportedLngs: SUPPORTED_LANGUAGES,
  fallbackLng: "en",
  load: "languageOnly" as const,
  cleanCode: true,
  ns: ["main"], // TODO: Add more namespaces https://www.i18next.com/principles/namespaces
  defaultNS: "main",
  backend: backendConfig,
  detection: detectionConfig,
  react: {
    useSuspense: false,
  },
};

// Initialize i18n
// Learn more: https://github.com/i18next
i18n.use(Backend).use(LanguageDetector).use(initReactI18next).init(i18nConfig);

export default i18n;
```

other than that it looks good to me 👍🏼

hdinia (approved):

webapp/src/i18n.ts
skamril: Normally the file will not changed anymore

PR #2228, opened 2024-11-18
feat: add new build directory structure

maugde:
Additional work for the installer

### DIFF
--- a/antarest/desktop/systray_app.py
+++ b/antarest/desktop/systray_app.py
@@ -184,7 +184,7 @@ def run_systray_app(config_file: Path) -> None:
     open_app()
     try:
         systray_app.app.exec_()
-    except: # catch QT error
+    except:  # catch QT error
         pass
     finally:
         server.kill()
--- a/antarest/desktop/systray_app.py
+++ b/antarest/desktop/systray_app.py
@@ -182,5 +182,9 @@ def run_systray_app(config_file: Path) -> None:
     wait_for_server_start()
     notification_popup("Antares Web Server started, you can manage the application within the system tray.")
     open_app()
-    systray_app.app.exec_()
-    server.kill()
+    try:
+        systray_app.app.exec_()
+    except: # catch QT error
+        pass
+    finally:
+        server.kill()
--- a/antarest/desktop/systray_app.py
+++ b/antarest/desktop/systray_app.py
@@ -184,7 +184,6 @@ def run_systray_app(config_file: Path) -> None:
     open_app()
     try:
         systray_app.app.exec_()
-    except:  # catch QT error
-        pass
     finally:
+        # Kill server also on exception, in particular on keyboard interrupt
         server.kill()
--- a/resources/antares-desktop-fs/README.md
+++ b/resources/antares-desktop-fs/README.md
@@ -30,7 +30,8 @@ Run the following command to launch the server:
 
 ## Accessing the Web Server
 
-Once the Antares Web server is running, you can access it using a web browser.
+Once the Antares Web server is running, the default web browser will automatically open to the application home page.
+You can also manually access by following these steps:
 
 1. Open a web browser.
 2. Enter the following URL in the address bar:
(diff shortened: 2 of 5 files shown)


resources/antares-desktop-fs/gunicorn.py
```diff
@@ -0,0 +1,30 @@
+# Reference: https://github.com/benoitc/gunicorn/blob/master/examples/example_config.py
```
sylvlecl: Actually, we don't use gunicorn to run the server in desktop mode:
instead, a single uvicorn worker is started.
So we can remove this file :)

resources/antares-desktop-fs/studies/example_study.zip
sylvlecl, line 1: It's a little annoying to add another 18 Mo to the repo, we need to think of a way to not duplicate it.

Maybe we can just copy it from the other place in the shell script, not great but there is already some processing involved there (we unzip it).

installer
sylvlecl, line 1: I think this update of the submodule is not intended, is it ?

resources/antares-desktop-fs/config.yaml
```diff
@@ -0,0 +1,58 @@
...
+# Serve the API at /api
+api_prefix: "/api"
+
+server:
+  worker_threadpool_size: 12
+  services:
+    - watcher
+
+logging:
+  logfile: ./logs/antarest.log
```
sylvlecl, line 58: Thanks for this !

scripts/package_antares_web.sh
```diff
@@ -15,7 +15,7 @@ ANTARES_SOLVER_VERSION_INT="880"
 SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd -P)
 PROJECT_DIR=$(dirname -- "${SCRIPT_DIR}")
 DIST_DIR="${PROJECT_DIR}/dist/package"
-RESOURCES_DIR="${PROJECT_DIR}/resources"
+RESOURCES_DIR="${PROJECT_DIR}/resources/"
```
sylvlecl: Note that you will have a double slash later when you append a slash: it works but I think it's better to not have it.

sylvlecl (changes requested):
Nice, just a few comments

resources/antares-desktop-fs/studies/example_study.zip
maugde: I would like to put the example_study.zip in a `resources/example_studies` directory since multiple services could read and copy it from one place but also in case of adding other example studies. 
Do we actually know which services use the content of the `resources/deploy` directory or `example_study.zip`?

maugde: Ok maybe I should keep this file in the `resources/deploy` directory and copy it from here since it seems to be used by Docker.

antarest/desktop/systray_app.py
```diff
@@ -182,5 +182,9 @@ def run_systray_app(config_file: Path) -> None:
     wait_for_server_start()
     notification_popup("Antares Web Server started, you can manage the application within the system tray.")
     open_app()
-    systray_app.app.exec_()
-    server.kill()
+    try:
+        systray_app.app.exec_()
+    except:  # catch QT error
```
sylvlecl: I would remove the except so that we still get the stack trace (the exception will be raised after the "finally" clause execution

resources/antares-desktop-fs/README.md
```diff
@@ -30,7 +30,8 @@ Run the following command to launch the server:
 
 ## Accessing the Web Server
 
-Once the Antares Web server is running, you can access it using a web browser.
+Once the Antares Web server is running, the default web browser will automatically open to the application home page.
```
sylvlecl, line 33: Thanks for updating the doc!

sylvlecl:
I just noticed there is a place in the installer where we reference the examples directory, we probably need to update it there too unfortunately.

sylvlecl (approved):

PR #2231, opened 2024-11-19
feat(ts-gen): add failing area and cluster info inside error msg (#2227)

sylvlecl:
Backport on release 2.18

### DIFF
--- a/antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
+++ b/antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
@@ -86,53 +86,57 @@ class GenerateThermalClusterTimeSeries(ICommand):
             # 5- Loop through thermal clusters in alphabetical order
             sorted_thermals = sorted(area.thermals, key=lambda x: x.id)
             for thermal in sorted_thermals:
-                # 6 - Filters out clusters with no generation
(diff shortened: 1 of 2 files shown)


antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -86,53 +86,57 @@ def _build_timeseries(
...
-                # 10- Notify the progress to the notifier
-                generation_performed += 1
-                if listener:
-                    progress = int(100 * generation_performed / total_generations)
-                    listener.notify_progress(progress)
+                    if listener:
+                        progress = int(100 * generation_performed / total_generations)
+                        listener.notify_progress(progress)
+                except Exception as e:
+                    e.args = (f"Area {area_id}, cluster {thermal.id.lower()}: " + e.args[0],)
```
sylvlecl, line 138: @MartinBelthle :
it seems weird to do this, changing an exception after the fact through an opaque ""args" member:
I think it would be more standard to raise a dedicated exception using "from" to still keep the original exception in the stack.

In any case let's not do this in this branch

MartinBelthle: Yes it's a bit odd I agree.
But as you said we should probably do this in another PR if we want to

PR #2236, opened 2024-11-19
feat(ui-commons,ui-api): allow to export matrices to CSV

skamril:
ANT-2407

### DIFF
--- a/webapp/src/services/api/studies/raw/constants.ts
+++ b/webapp/src/services/api/studies/raw/constants.ts
@@ -1,0 +1,21 @@
+/**
+ * Copyright (c) 2024, RTE (https://www.rte-france.com)
+ *
+ * See AUTHORS.txt
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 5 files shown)


webapp/src/services/api/studies/raw/constants.ts
```diff
@@ -0,0 +1,21 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+export const TableExportFormat = {
```
hdinia, line 15: As we said we should use upload/download instead of export/import, and since its about matrices `MatrixDownloadFormat` seems more appropriate here

hdinia (changes requested):

webapp/src/services/api/studies/raw/constants.ts
skamril: Yes but here is the API so I follow the enum name from the backend for maintainability, so when a enum from the backend change it would be easy to update the enum from the front

hdinia (approved):

PR #2239, opened 2024-11-21
feat(api,ui-studies): update study move

skamril:
ANT-2390


antarest/study/service.py
```diff
@@ -1073,11 +1073,15 @@ def copy_task(notifier: ITaskNotifier) -> TaskResult:
...
         return task_or_study_id
 
-    def move_study(self, study_id: str, new_folder: str, params: RequestParameters) -> None:
+    def move_study(self, study_id: str, folder_dest: str, params: RequestParameters) -> None:
         study = self.get_study(study_id)
         assert_permission(params.user, study, StudyPermissionType.WRITE)
         if not is_managed(study):
             raise NotAManagedStudyException(study_id)
+        if folder_dest:
+            new_folder = folder_dest.rstrip("/") + f"/{study.id}"
```
sylvlecl, line 1082: Just a note for later: would probably be best to use `PurePosixPath` objects than just strings to represent our paths.

sylvlecl (commented):
Ok.
I think the behaviour we have chosen before this fix (to append the ID to the path) is not good from the user point of view, but that would be a separate topic.

sylvlecl (dismissed):

webapp/src/components/App/Studies/MoveStudyDialog.tsx
```diff
@@ -13,17 +13,40 @@
...
 import { SubmitHandlerPlus } from "../../common/Form/types";
+import StringFE from "@/components/common/fieldEditors/StringFE";
+import * as R from "ramda";
+import { validatePath } from "@/utils/validation/string";
+
+function formalizePath(
+  path: string | undefined,
+  studyId?: StudyMetadata["id"],
+) {
+  const pathTrimmed = path?.trim();
```
hdinia: `pathTrimmed` -> `trimmedPath`

webapp/src/components/App/Studies/MoveStudyDialog.tsx
```diff
@@ -33,36 +56,35 @@ interface Props extends DialogProps {
...
-        t("studies.error.moveStudy", { study: study.name }),
-        e as Error,
-      );
-    }
+    onClose();
+
+    enqueueSnackbar(
+      t("studies.success.moveStudy", {
+        study: study.name,
+        path: data.values.path,
```
hdinia: you named it `folder` in the i18n key (`path` -> `folder`)
or you can rename the key if its more accurate

webapp/src/services/api/study.ts
```diff
@@ -163,11 +163,10 @@ export const copyStudy = async (
...
 };
 
-export const moveStudy = async (sid: string, folder: string): Promise<void> => {
-  const folderWithId = trimCharsStart("/", `${folder.trim()}/${sid}`);
-  await client.put(
-    `/v1/studies/${sid}/move?folder_dest=${encodeURIComponent(folderWithId)}`,
-  );
+export const moveStudy = async (sid: string, folder: string) => {
+  await client.put(`/v1/studies/${sid}/move`, null, {
+    params: { folder_dest: folder },
```
hdinia, line 168: not sure it's useful but maybe you should keep encoding here to avoid any potential issue
`params: { folder_dest: encodeURIComponent(folder) }`

hdinia: I know it's legacy code, but it would be nice if you can rename, `sid` to `studyId`

webapp/src/utils/validation/string.ts
hdinia, line 1: it would be nice to have some unit tests, but optional

webapp/src/utils/validation/string.ts
```diff
@@ -183,3 +207,78 @@ function generatePattern(
...
+    return t("form.field.path.startWithSlashNotAllowed");
+  }
+
+  if (!allowToEndWithSlash && path.endsWith("/")) {
+    return t("form.field.path.endWithSlashNotAllowed");
+  }
+
+  if (
+    path
+      .replace(/^\/|\/$/g, "") // Remove first "/" and last "/" if present
```
hdinia: this regex is triggering warning in sonar (see sonar job)

hdinia (changes requested):

webapp/public/locales/fr/main.json
```diff
@@ -631,11 +634,10 @@
   "studies.error.createStudy": "Erreur lors de la création de l'étude {{studyname}}",
   "studies.success.saveData": "Donnée sauvegardée avec succès",
   "studies.success.scanFolder": "L'analyse du dossier a commencé",
-  "studies.success.moveStudy": "L'étude {{study}} a été déplacée avec succès vers {{folder}}",
+  "studies.success.moveStudy": "L'étude \"{{study}}\" a été déplacée avec succès vers \"{{folder}}\"",
```
hdinia, line 637: remove backslash, use singlequotes

webapp/src/services/api/study.ts
skamril: Axios encode automatically with `params` parameter.

webapp/src/utils/validation/string.ts
skamril: I need to learn the subject

webapp/src/components/App/Studies/MoveStudyDialog.tsx
skamril: ok

webapp/src/services/api/study.ts
hdinia: true

webapp/src/utils/validation/string.ts
hdinia: ok

webapp/src/components/App/Studies/MoveStudyDialog.tsx
skamril: Changed 'folder' to 'path' in the FR key

skamril: Ok for the empty path

hdinia: in addition I think you should handle the case where path is empty to adapt the UI message and not display ""

webapp/src/utils/validation/string.ts
skamril: What are the fix so? I have no idea of this issue.

hdinia: To fix it you can use native string methods (`startsWith`, `endsWith`, `split`) which are more secure but this will make the code longer to achieve the same thing.. Don't seem a major security issue to me i don't know why Sonar raises errors for this 
maybe you can try something like this:
```typescript
path = path.startsWith('/') ? path.slice(1) : path;
path = path.endsWith('/') ? path.slice(0, -1) : path;
```
or 
```typescript
path = path.replace(/^\//, "").replace(/\/$/, "");
```

seems like the OR (|) operator is problematic

hdinia (approved):

PR #2240, opened 2024-11-21
feat(ui-api, digest): enhance `DigestDialog` and add dedicated endpoint for digest UI

MartinBelthle:
Fix [ANT-2449] for the back-end part

**Endpoint:** GET "/private/studies/{study_id}/outputs/{output_id}/digest-ui"

**ResponseModel:**
```python
class DigestMatrixUI(AntaresBaseModel):
    rowHeaders: t.List[str]
    columns: t.List[str]
    data: t.List[t.List[str]]
    grouped_columns: bool


class DigestUI(AntaresBaseModel):
    area: DigestMatrixUI
    districts: DigestMatrixUI
    flow_linear: DigestMatrixUI
    flow_quadratic: DigestMatrixUI
```


antarest/study/storage/rawstudy/model/filesystem/root/output/simulation/mode/mcall/digest.py
```diff
@@ -0,0 +1,131 @@
...
+class DigestMatrixDTO(AntaresBaseModel):
+    columns: t.List[t.Union[str, t.List[str]]]
+    data: t.List[t.List[t.Any]]
+    groupedColumns: bool
+
+
+class DigestDTO(AntaresBaseModel):
+    area: DigestMatrixDTO
+    districts: DigestMatrixDTO
+    flowLinear: DigestMatrixDTO
```
sylvlecl: We should stick with kebab_case for class fields (I am surprised CI does not check this)

antarest/study/storage/rawstudy/model/filesystem/root/output/simulation/mode/mcall/digest.py
```diff
@@ -0,0 +1,131 @@
...
+
+from antarest.core.model import JSON
+from antarest.core.serialization import AntaresBaseModel
+from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfig
+from antarest.study.storage.rawstudy.model.filesystem.context import ContextServer
+from antarest.study.storage.rawstudy.model.filesystem.root.output.simulation.mode.mcall.synthesis import OutputSynthesis
+
+
+class DigestMatrixDTO(AntaresBaseModel):
+    columns: t.List[t.Union[str, t.List[str]]]
```
sylvlecl, line 27: The concepts of columns and rows are really related to the display: in an API, we should have more self-defined and business-related names, like `areas` for row names and `variables` for columns names. Otherwise, we don't know what the data represents.

In the same order of idea, instead of a list of string for the headers we could have a small class containing variable name, unit, and aggregation type.

We can keep this if we consider that it's really only meant to be used for the display, not by an API client, but in that case maybe we could rename the DigestDTO and the endpoint so that it's really identified as ui-related ?

antarest/study/storage/rawstudy/model/filesystem/root/output/simulation/mode/mcall/digest.py
sylvlecl: For the flows data, I am not really not sure a matrix will be the best representation (in the data, not in the front), because by nature it will be very sparse.

Basically we have 2 values per-link, it seems to me it would make more sense to have a structure such as:
```json
[
    {
        "from": "fr",
        "to": "be",
        "direct_flow": 100,
        "opposite_flow": -100,
        "direct_quad_flow": 20,
        "opposite_quad_flow": -20,
    },
    ...
]
```

antarest/study/storage/rawstudy/model/filesystem/root/output/simulation/mode/mcall/digest.py
```diff
@@ -0,0 +1,131 @@
...
+import pandas as pd
+
+from antarest.core.model import JSON
+from antarest.core.serialization import AntaresBaseModel
+from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfig
+from antarest.study.storage.rawstudy.model.filesystem.context import ContextServer
+from antarest.study.storage.rawstudy.model.filesystem.root.output.simulation.mode.mcall.synthesis import OutputSynthesis
+
+
+class DigestMatrixDTO(AntaresBaseModel):
```
sylvlecl: DTOs are really front-facing classes which should be close to the web layer, not deep inside the filesystem layer.
We should at least rename them here to not have DTO in the name.

antarest/study/web/studies_blueprint.py
```diff
@@ -815,6 +816,26 @@ def unarchive_output(
         )
         return content
 
+    @bp.get(
+        "/private/studies/{study_id}/outputs/{output_id}/digest",
+        tags=[APITag.study_outputs],
+        summary="Get an output digest file",
+        response_model=DigestDTO,
```
sylvlecl: I don't think this is needed ?

antarest/study/web/studies_blueprint.py
```diff
@@ -815,6 +816,26 @@ def unarchive_output(
         )
         return content
 
+    @bp.get(
+        "/private/studies/{study_id}/outputs/{output_id}/digest",
```
sylvlecl: If we consider it's really a UI-related endpoint (with already formatted tables instead of business related objects), we should probably make it clear in the path, like  "/private/studies/{study_id}/outputs/{output_id}/digest-ui"

sylvlecl (commented):
My interrogations:
we should chose between 2 things for this new endpoint:
- is it only UI related ? in this case we could stick with the "poor" table representation, but we should make it more obvious in the path and in the class names
- is it meant to become an API endpoint ? in that case we need to have a more specific model with better names and probably another structure (not tables everywhere)

MartinBelthle:
After discussing with Sylvain, we decided that this endpoint will be strictly used by the front-end so we''l had `-ui` at the end.
If someone wants to use it via API we'll design a new endpoint with the structure Sylvain proposed as it's indeed better for an API user

antarest/study/web/studies_blueprint.py
MartinBelthle: You talk about the response model ? If so indeed i think it's not needed but why would I remove it ?

antarest/study/storage/rawstudy/model/filesystem/root/output/simulation/mode/mcall/digest.py
MartinBelthle: Okay, I renamed it DigestUI to fit with what we said. But how do you picture the class not being here ?

MartinBelthle:
I'll have to add tests in the back-end when we'll be sure the format is right.

hdinia:
> After discussing with Sylvain, we decided that this endpoint will be strictly used by the front-end so we''l had `-ui` at the end. If someone wants to use it via API we'll design a new endpoint with the structure Sylvain proposed as it's indeed better for an API user

I get the intent of marking UI-only endpoints 👍🏼 , but `-ui` as a suffix breaks REST conventions where URLs represent resources, not their usage. A prefix like `/v1/ui/` would be preferable for UI related routes

antarest/study/storage/rawstudy/model/filesystem/root/output/simulation/mode/mcall/digest.py
```diff
@@ -0,0 +1,132 @@
...
+from antarest.core.model import JSON
+from antarest.core.serialization import AntaresBaseModel
+from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfig
+from antarest.study.storage.rawstudy.model.filesystem.context import ContextServer
+from antarest.study.storage.rawstudy.model.filesystem.root.output.simulation.mode.mcall.synthesis import OutputSynthesis
+
+
+class DigestMatrixUI(AntaresBaseModel):
+    columns: t.List[t.Union[str, t.List[str]]]
+    data: t.List[t.List[t.Any]]
```
hdinia: should not be `data: t.List[t.List[str]]` ?

MartinBelthle: I changed it to str

webapp/src/components/common/dialogs/DigestDialog/DigestMatrix.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { GridOff } from "@mui/icons-material";
+import { useTranslation } from "react-i18next";
+
+const isGroupedColumns = (
+  columns: string[] | string[][],
+): columns is string[][] => {
+  return Array.isArray(columns[0]);
+};
+
+const isColumns = (columns: string[] | string[][]): columns is string[] => {
```
skamril: Opinion: Not necessary to create this one, it's more clear to use `isGroupedColumns` every time for me

webapp/src/components/common/dialogs/DigestDialog/DigestTabs.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+interface DigestTabsProps {
+  matrices: DigestData;
+}
+
+export function DigestTabs({ matrices }: DigestTabsProps) {
+  const tabItems = [
+    {
+      label: "Area",
+      content: matrices.area && <DigestMatrix matrix={matrices.area} />,
```
skamril, line 28: It's not better to hide tabs not used?

```typescript
const tabItems = [
  matrices.area && { label: "Area", ... },
  ....
].filter(Boolean)
```

webapp/src/components/common/dialogs/DigestDialog/DigestTabs.tsx
```diff
@@ -0,0 +1,59 @@
...
+      ),
+    },
+  ];
+
+  ////////////////////////////////////////////////////////////////
+  // JSX
+  ////////////////////////////////////////////////////////////////
+
+  return (
+    <Box sx={{ height: "100%", display: "flex", flexDirection: "column" }}>
```
skamril: Tip: in sx prop `height: "100%"` is same as `height: 1`

webapp/src/components/common/dialogs/DigestDialog/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+      title="Digest"
+      okButtonText={t("global.close")}
+      fullScreen
+      sx={{ m: 5 }}
+    >
+      <UsePromiseCond
+        response={digestRes}
+        ifPending={() => <Skeleton sx={{ height: 1, transform: "none" }} />}
+        ifFulfilled={(matrices) =>
+          matrices.data && <DigestTabs matrices={matrices.data} />
```
skamril: No message for empty matrice?

skamril (changes requested):

webapp/src/components/common/dialogs/DigestDialog/DigestMatrix.tsx
hdinia: both are necessary because depending on the matrix we can have groups or not

webapp/src/components/common/dialogs/DigestDialog/DigestTabs.tsx
hdinia: it might be better yes, I have to check why I added this check I dont remember

webapp/src/components/common/dialogs/DigestDialog/DigestTabs.tsx
hdinia: updated, but I find it not intuitive you might think its set to 1px if you dont know

webapp/src/components/common/dialogs/DigestDialog/index.tsx
hdinia: It's already handled by DigestMatrix, for more granular control for each tab individually (some matrices can be empty while others not)

webapp/src/components/common/dialogs/DigestDialog/DigestMatrix.tsx
skamril: `isColumns` only call `isGroupedColumns`, and its name is misleading:

`isColumns(COLUMNS_WIHTOUT_GROUPS)` => `false`, but in facts they are columns.

So `isGroupedColumns` is sufficient, and more clear:

`customColumns: isColumns(matrix.columns) ? matrix.columns : undefined,`

to

`customColumns: isGroupedColumns(matrix.columns) ? undefined : matrix.columns,`

webapp/src/components/common/dialogs/DigestDialog/types.ts
```diff
@@ -0,0 +1,26 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+export interface DigestMatrixType {
```
skamril: You can rename it `DigestMatrix`

hdinia: DigestMatrix is already taken by the component, I will rename it something else as the suffix "Type" for non-enums should be avoided, `DigestMatrixData` or `DigestMatrixConfig`

skamril (approved):

TheoPascoli (approved):

PR #2241, opened 2024-11-21
feat(ui-studies): allow to move an archived study

skamril:
ANT-2391

### DIFF
--- a/webapp/src/components/App/Studies/StudyCard/ActionsMenu.tsx
+++ b/webapp/src/components/App/Studies/StudyCard/ActionsMenu.tsx
@@ -1,0 +1,194 @@
+/**
+ * Copyright (c) 2024, RTE (https://www.rte-france.com)
+ *
+ * See AUTHORS.txt
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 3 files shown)


webapp/src/components/App/Studies/StudyCard.tsx
```diff
@@ -450,20 +470,23 @@ const StudyCard = memo((props: Props) => {
...
-                  sx={{
-                    color: "action.active",
-                    width: "24px",
-                    height: "24px",
-                  }}
-                />
-              </ListItemIcon>
-              <ListItemText>{t("global.unarchive")}</ListItemText>
-            </MenuItem>
+            <>
```
hdinia: Avoid using a Fragment on Menu component, you can replace it by a div or Box. 

Ideally we should refactor the menu items into their own component, as an array, then loop over, but since this component will be redone its a waste of time

see: 
- https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Menu/Menu.js#L134
- https://github.com/mui/material-ui/issues/16181

their is another existing Fragment in the Menu component below your added code, could you please also remove it ?

hdinia (changes requested):

hdinia (dismissed):

PR #2244, opened 2024-11-23
feat(raw-api): add an endpoint to retrieve files in their original format

mabw-rte:
Context:
Users found it a little inconvenient to work with some files (retrieved using the `raw` API) e.g. dealing with a `.json` file while it should be a `.ini` file as it is in its original format

Solution:
Create an new endpoint to retrieve files in their original format

### DIFF
--- a/antarest/core/exceptions.py
+++ b/antarest/core/exceptions.py
@@ -693,11 +693,6 @@ class PathIsAFolderError(HTTPException):
         super().__init__(HTTPStatus.UNPROCESSABLE_ENTITY, message)
 
 
-class InvalidContentError(HTTPException):
-    def __init__(self, message: str) -> None:
-        super().__init__(HTTPStatus.UNPROCESSABLE_ENTITY, message)
-
-
 class WorkspaceNotFound(HTTPException):
     """
     This will be raised when we try to load a workspace that does not exist
--- a/antarest/core/exceptions.py
+++ b/antarest/core/exceptions.py
@@ -711,15 +711,6 @@ class BadArchiveContent(Exception):
         super().__init__(message)
 
 
-class StudyNotArchived(HTTPException):
-    """
-    Exception raised when the study is supposed to be archived but is not.
-    """
-
-    def __init__(self, message: str = "") -> None:
-        super().__init__(HTTPStatus.UNPROCESSABLE_ENTITY, message)
-
-
 class FolderNotFoundInWorkspace(HTTPException):
     """
     This will be raised when we try to load a folder that does not exist
--- a/antarest/core/exceptions.py
+++ b/antarest/core/exceptions.py
@@ -688,6 +688,16 @@ class ChildNotFoundError(HTTPException):
         super().__init__(HTTPStatus.NOT_FOUND, message)
 
 
+class PathIsAFolderError(HTTPException):
+    def __init__(self, message: str) -> None:
+        super().__init__(HTTPStatus.UNPROCESSABLE_ENTITY, message)
+
+
+class InvalidContentError(HTTPException):
+    def __init__(self, message: str) -> None:
+        super().__init__(HTTPStatus.UNPROCESSABLE_ENTITY, message)
+
+
 class WorkspaceNotFound(HTTPException):
     """
     This will be raised when we try to load a workspace that does not exist
@@ -704,6 +714,15 @@ class BadArchiveContent(Exception):
 
     def __init__(self, message: str = "Unsupported archive format") -> None:
         super().__init__(message)
+
+
+class StudyNotArchived(HTTPException):
+    """
+    Exception raised when the study is supposed to be archived but is not.
+    """
+
+    def __init__(self, message: str = "") -> None:
+        super().__init__(HTTPStatus.UNPROCESSABLE_ENTITY, message)
 
 
 class FolderNotFoundInWorkspace(HTTPException):
--- a/antarest/core/utils/archives.py
+++ b/antarest/core/utils/archives.py
@@ -145,6 +145,30 @@ def extract_file_to_tmp_dir(archive_path: Path, inside_archive_path: Path) -> t.
     return path, tmp_dir
 
 
+def read_original_file_in_archive(archive_path: Path, posix_path: str) -> bytes:
+    """
+    Read a file from an archive.
+
+    Args:
+        archive_path: the path to the archive file.
+        posix_path: path to the file inside the archive.
+
+    Returns:
+        The content of the file as `bytes`.
+    """
+
+    if archive_path.suffix == ArchiveFormat.ZIP:
+        with zipfile.ZipFile(archive_path) as zip_obj:
+            with zip_obj.open(posix_path) as f:
+                return f.read()
+    elif archive_path.suffix == ArchiveFormat.SEVEN_ZIP:
+        with py7zr.SevenZipFile(archive_path, mode="r") as szf:
+            output: bytes = szf.read([posix_path])[posix_path].read()
+            return output
+    else:
+        raise ValueError(f"Unsupported {archive_path.suffix} archive format for {archive_path}")
+
+
 def read_file_from_archive(archive_path: Path, posix_path: str) -> str:
     """
     Read a file from an archive.
@@ -157,16 +181,7 @@ def read_file_from_archive(archive_path: Path, posix_path: str) -> str:
         The content of the file as a string.
     """
 
-    if archive_path.suffix == ArchiveFormat.ZIP:
-        with zipfile.ZipFile(archive_path) as zip_obj:
-            with zip_obj.open(posix_path) as f:
-                return f.read().decode("utf-8")
-    elif archive_path.suffix == ArchiveFormat.SEVEN_ZIP:
-        with py7zr.SevenZipFile(archive_path, mode="r") as szf:
-            file_text: str = szf.read([posix_path])[posix_path].read().decode("utf-8")
-            return file_text
-    else:
-        raise ValueError(f"Unsupported {archive_path.suffix} archive format for {archive_path}")
+    return read_original_file_in_archive(archive_path, posix_path).decode("utf-8")
 
 
 def extract_lines_from_archive(root: Path, posix_path: str) -> t.List[str]:
--- a/antarest/study/common/studystorage.py
+++ b/antarest/study/common/studystorage.py
@@ -17,18 +17,12 @@ from pathlib import Path
 from antarest.core.exceptions import StudyNotFoundError
 from antarest.core.model import JSON
 from antarest.core.requests import RequestParameters
-from antarest.core.serialization import AntaresBaseModel
 from antarest.study.model import Study, StudyMetadataDTO, StudyMetadataPatchDTO, StudySimResultDTO
 from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfigDTO
 from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.rawstudy.model.filesystem.inode import OriginalFile
 
 T = t.TypeVar("T", bound=Study)
-
-
-class OriginalFile(AntaresBaseModel):
-    suffix: str
-    content: bytes
-    filename: str
 
 
 class IStudyStorageService(ABC, t.Generic[T]):
--- a/antarest/study/common/studystorage.py
+++ b/antarest/study/common/studystorage.py
@@ -17,11 +17,18 @@ from pathlib import Path
 from antarest.core.exceptions import StudyNotFoundError
 from antarest.core.model import JSON
 from antarest.core.requests import RequestParameters
+from antarest.core.serialization import AntaresBaseModel
 from antarest.study.model import Study, StudyMetadataDTO, StudyMetadataPatchDTO, StudySimResultDTO
 from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfigDTO
 from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
 
 T = t.TypeVar("T", bound=Study)
+
+
+class OriginalFile(AntaresBaseModel):
+    suffix: str
+    content: bytes
+    filename: str
 
 
 class IStudyStorageService(ABC, t.Generic[T]):
@@ -53,6 +60,23 @@ class IStudyStorageService(ABC, t.Generic[T]):
             formatted: indicate if raw files must be parsed and formatted
 
         Returns: study data formatted in json
+
+        """
+
+    @abstractmethod
+    def get_file(
+        self,
+        metadata: T,
+        url: str = "",
+    ) -> OriginalFile:
+        """
+        Entry point to fetch for a specific file inside a study folder
+
+        Args:
+            metadata: study
+            url: path data inside study to reach the file
+
+        Returns: study file content and extension
 
         """
 
--- a/antarest/study/storage/abstract_storage_service.py
+++ b/antarest/study/storage/abstract_storage_service.py
@@ -19,22 +19,11 @@ from pathlib import Path
 from uuid import uuid4
 
 from antarest.core.config import Config
-from antarest.core.exceptions import (
-    BadOutputError,
-    PathIsAFolderError,
-    ShouldNotHappenException,
-    StudyOutputNotFoundError,
-)
+from antarest.core.exceptions import BadOutputError, StudyOutputNotFoundError
 from antarest.core.interfaces.cache import CacheConstants, ICache
 from antarest.core.model import JSON, PublicMode
 from antarest.core.serialization import from_json
-from antarest.core.utils.archives import (
-    ArchiveFormat,
-    archive_dir,
-    extract_archive,
-    read_original_file_in_archive,
-    unzip,
-)
+from antarest.core.utils.archives import ArchiveFormat, archive_dir, extract_archive, unzip
 from antarest.core.utils.utils import StopWatch
 from antarest.login.model import GroupDTO
 from antarest.study.common.studystorage import IStudyStorageService, OriginalFile, T
@@ -54,69 +43,10 @@ from antarest.study.storage.patch_service import PatchService
 from antarest.study.storage.rawstudy.model.filesystem.config.files import get_playlist
 from antarest.study.storage.rawstudy.model.filesystem.config.model import Simulation
 from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy, StudyFactory
-from antarest.study.storage.rawstudy.model.filesystem.folder_node import FolderNode
-from antarest.study.storage.rawstudy.model.filesystem.inode import INode
-from antarest.study.storage.rawstudy.model.filesystem.lazy_node import LazyNode
-from antarest.study.storage.rawstudy.model.filesystem.matrix.input_series_matrix import InputSeriesMatrix
 from antarest.study.storage.rawstudy.model.helpers import FileStudyHelpers
 from antarest.study.storage.utils import extract_output_name, fix_study_root, remove_from_cache
 
 logger = logging.getLogger(__name__)
-
-
-def _assert_not_folder_node(file_node: INode[t.Any, t.Any, t.Any]) -> None:
-    if isinstance(file_node, FolderNode):
-        raise PathIsAFolderError("Node is a folder node.")
-
-
-def _parse_file_from_link(file_node: INode[t.Any, t.Any, t.Any], link: str) -> bytes:
-    if not isinstance(file_node, InputSeriesMatrix):
-        raise ShouldNotHappenException(f"Node {file_node.config.path} has a link but is not an InputSeriesMatrix.")
-    with tempfile.TemporaryDirectory() as tmp_dir:
-        # target path to save the file
-        target_path = Path(tmp_dir) / file_node.config.path.name
-        # use `.tsv` as suffix for the file
-        target_path = target_path.with_suffix(".tsv")
-        # retrieve the file content
-        file_node.parse(return_dataframe=True).to_csv(  # type: ignore
-            target_path,
-            sep="\t",
-            index=False,
-            header=True,
-            float_format="%.6f",
-        )
-        output = target_path.read_bytes()
-    return output
-
-
-def _extract_original_file_from_unarchived_study(
-    file_node: INode[t.Any, t.Any, t.Any],
-) -> OriginalFile:
-    _assert_not_folder_node(file_node)
-    if isinstance(file_node, LazyNode) and file_node.get_link_path().is_file():
-        return OriginalFile(
-            suffix=".tsv",
-            content=_parse_file_from_link(file_node, file_node.get_link_path().read_text()),
-            filename=file_node.config.path.with_suffix(".tsv").name,
-        )
-    else:
-        return OriginalFile(
-            suffix=f"{file_node.config.path.suffix}",
-            content=file_node.config.path.read_bytes(),
-            filename=file_node.config.path.name,
-        )
-
-
-def _extract_original_file_from_archived_study(
-    file_node: INode[t.Any, t.Any, t.Any], archive_path: Path
-) -> OriginalFile:
-    _assert_not_folder_node(file_node)
-    relative_path_inside_archive = str(file_node.get_relative_path_inside_archive())
-    return OriginalFile(
-        suffix=f"{file_node.config.path.suffix}",
-        content=read_original_file_in_archive(archive_path, relative_path_inside_archive),
-        filename=file_node.config.path.name,
-    )
 
 
 class AbstractStorageService(IStudyStorageService[T], ABC):
@@ -261,10 +191,8 @@ class AbstractStorageService(IStudyStorageService[T], ABC):
 
         file_node = study.tree.get_node(parts)
 
-        if file_node.config.archive_path:
-            return _extract_original_file_from_archived_study(file_node, file_node.config.archive_path)
-        else:
-            return _extract_original_file_from_unarchived_study(file_node)
+        content, suffix, filename = file_node.get_file_content()
+        return OriginalFile(content=content, suffix=suffix, filename=filename)
 
     def get_study_sim_result(
         self,
--- a/antarest/study/storage/abstract_storage_service.py
+++ b/antarest/study/storage/abstract_storage_service.py
@@ -18,17 +18,26 @@ from abc import ABC
 from pathlib import Path
 from uuid import uuid4
 
-import py7zr
-
 from antarest.core.config import Config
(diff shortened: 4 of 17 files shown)


antarest/study/common/studystorage.py
```diff
@@ -24,6 +26,12 @@
 T = t.TypeVar("T", bound=Study)
 
 
+class OriginalFile(BaseModel):
```
MartinBelthle: Use AntaresBaseModel instead of BaseModel here

antarest/core/utils/archives.py
```diff
@@ -9,6 +9,7 @@
 # SPDX-License-Identifier: MPL-2.0
 #
 # This file is part of the Antares project.
+import io
```
MartinBelthle: Remove unused import

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -185,6 +190,54 @@ def get_study(
...
+        parameters = RequestParameters(user=current_user)
+        original_file = study_service.get_file(uuid, path, params=parameters)
+        filename = original_file.filename
+        output = original_file.content
+        suffix = original_file.suffix
+        headers = {
+            "Content-Disposition": f"attachment; filename={filename}",
+        }
+
+        if isinstance(output, bytes):
```
MartinBelthle: You should remove this check as you typed OrginalFile output to be bytes

antarest/study/storage/abstract_storage_service.py
```diff
@@ -171,6 +241,33 @@ def get(
...
+        Returns: a file content with its extension and name
+
+        """
+        self._check_study_exists(metadata)
+        study = self.get_raw(metadata, use_cache)
+        parts = [item for item in url.split("/") if item]
+
+        file_node = study.tree.get_node(parts)
+
+        if file_node.config.archive_path:
```
MartinBelthle: Not reliable. Replace by `if metadata.archived:`

antarest/study/storage/abstract_storage_service.py
```diff
@@ -45,12 +50,77 @@
...
+from antarest.study.storage.rawstudy.model.filesystem.inode import INode
+from antarest.study.storage.rawstudy.model.filesystem.lazy_node import LazyNode
+from antarest.study.storage.rawstudy.model.filesystem.matrix.input_series_matrix import InputSeriesMatrix
 from antarest.study.storage.rawstudy.model.helpers import FileStudyHelpers
 from antarest.study.storage.utils import extract_output_name, fix_study_root, remove_from_cache
 
 logger = logging.getLogger(__name__)
 
 
+def _assert_not_folder_node(file_node: INode[t.Any, t.Any, t.Any]) -> None:
```
MartinBelthle: You could do this inside the get_file method before calling the extract_xxx to simplify the code.

Also I don't really like the name of this exception, it could be better to rename it something like:
PathIsAFolderError with a msg like f"The given path {file_node.config.path} points to a folder and not a file"

antarest/study/storage/abstract_storage_service.py
```diff
@@ -45,12 +50,77 @@
...
+    # retrieve the file content
+    file_node.parse_from_link(link).dropna(how="any", axis=1).to_csv(
+        target_path,
+        sep="\t",
+        index=True,
+        header=True,
+        float_format="%.6f",
+    )
+    output = target_path.read_bytes()
+    tmp_dir.cleanup()
```
MartinBelthle: Instead of doing a cleanup you could do a with tempfile ... that will do the clean up for you. Because if you fail before you won't clean it.

antarest/study/storage/abstract_storage_service.py
```diff
@@ -45,12 +50,77 @@
...
+def _parse_file_from_link(file_node: INode[t.Any, t.Any, t.Any], link: str) -> bytes:
+    if not isinstance(file_node, InputSeriesMatrix):
+        raise ShouldNotHappenException(f"Node {file_node.config.path} has a link but is not an InputSeriesMatrix.")
+    tmp_dir = tempfile.TemporaryDirectory()
+    # target path to save the file
+    target_path = Path(tmp_dir.name) / file_node.config.path.name
+    # use `.tsv` as suffix for the file
+    target_path = target_path.with_suffix(".tsv")
+    # retrieve the file content
+    file_node.parse_from_link(link).dropna(how="any", axis=1).to_csv(
```
MartinBelthle: I think you shouldn't use this method. You can just use parse(return_dataframe=True) as it will do the same thing and it avoids creating new entry points inside the node.

You can also revert what you did inside input_series_matrix as the method will no longer be use

antarest/study/storage/abstract_storage_service.py
```diff
@@ -45,12 +50,77 @@
...
+            content=_parse_file_from_link(file_node, file_node.get_link_path().read_text()),
+            filename=file_node.config.path.with_suffix(".tsv").name,
+        )
+    elif file_node.config.path.is_file():
+        return OriginalFile(
+            suffix=f"{file_node.config.path.suffix}",
+            content=file_node.config.path.read_bytes(),
+            filename=file_node.config.path.name,
+        )
+    else:
```
MartinBelthle: I don't believe this else brings something

antarest/study/storage/abstract_storage_service.py
```diff
@@ -45,12 +50,77 @@
...
+
+def _extract_original_file_from_unarchived_study(
+    file_node: INode[t.Any, t.Any, t.Any], metadata: T, url: str
+) -> OriginalFile:
+    _assert_not_folder_node(file_node)
+    if isinstance(file_node, LazyNode) and file_node.get_link_path().is_file():
+        return OriginalFile(
+            suffix=".tsv",
+            content=_parse_file_from_link(file_node, file_node.get_link_path().read_text()),
+            filename=file_node.config.path.with_suffix(".tsv").name,
```
MartinBelthle: I think what you want here is stem instead of name. Same for line 103.

antarest/study/storage/abstract_storage_service.py
```diff
@@ -45,12 +50,77 @@
...
+) -> OriginalFile:
+    _assert_not_folder_node(file_node)
+    relative_path_inside_archive = str(file_node.get_relative_path_inside_archive())
+    if file_or_dir_is_in_archive(archive_path, relative_path_inside_archive):
+        return OriginalFile(
+            suffix=f"{file_node.config.path.suffix}",
+            content=read_original_file_in_archive(archive_path, relative_path_inside_archive),
+            filename=file_node.config.path.name,
+        )
+    else:
```
MartinBelthle: Same here it makes the code less readable I think

antarest/study/storage/rawstudy/model/filesystem/inode.py
```diff
@@ -124,6 +124,16 @@ def denormalize(self) -> None:
         """
         raise NotImplementedError()
 
+    def get_relative_path_inside_archive(self) -> Path:
+        archive_path = self.config.archive_path
+        if archive_path:
+            return self.config.path.relative_to(archive_path.parent / self.config.study_id)
+        else:
+            raise ShouldNotHappenException(
```
MartinBelthle: You should raise another Exception as this one is really for cases that shouldn't happen. Here it can happen if someone misuses the method.

antarest/study/storage/abstract_storage_service.py
```diff
@@ -45,12 +50,77 @@
...
+        raise ShouldNotHappenException(f"Path to url={url} in study {metadata.id} is not a file.")
+
+
+def _extract_original_file_from_archived_study(
+    file_node: INode[t.Any, t.Any, t.Any], metadata: T, url: str, archive_path: Path
+) -> OriginalFile:
+    _assert_not_folder_node(file_node)
+    relative_path_inside_archive = str(file_node.get_relative_path_inside_archive())
+    if file_or_dir_is_in_archive(archive_path, relative_path_inside_archive):
```
MartinBelthle: You shouldn't need to do this as the method get_relative_path_inside_archive would raise an issue if it wasn't the case

antarest/study/storage/abstract_storage_service.py
```diff
@@ -45,12 +50,77 @@
...
+    tmp_dir = tempfile.TemporaryDirectory()
+    # target path to save the file
+    target_path = Path(tmp_dir.name) / file_node.config.path.name
+    # use `.tsv` as suffix for the file
+    target_path = target_path.with_suffix(".tsv")
+    # retrieve the file content
+    file_node.parse_from_link(link).dropna(how="any", axis=1).to_csv(
+        target_path,
+        sep="\t",
+        index=True,
```
MartinBelthle: You shouldn't save index and headers when exporting as the raw format doesn't have any

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
```diff
@@ -237,115 +237,315 @@ def test_get_study(
...
+        # Then, use the API to download the files from the "user/folder" directory
+        user_folder_dir = study_dir.joinpath("user/folder")
+        for file_path in user_folder_dir.glob("*.*"):
+            rel_path = file_path.relative_to(study_dir).as_posix()
+            res = client.get(original_file_url, params={"path": rel_path})
+            assert res.status_code == 200, res.json()
+
+            if file_path.suffix == ".txt.link":
+                # special case for JSON files
+                actual = res.json()
```
MartinBelthle: You cannot reach this code as you're inside the user folder

MartinBelthle (changes requested):
One test you wrote seems to fail on Windows

antarest/study/storage/abstract_storage_service.py
```diff
@@ -45,12 +50,77 @@
...
+    output = target_path.read_bytes()
+    tmp_dir.cleanup()
+    return output
+
+
+def _extract_original_file_from_unarchived_study(
+    file_node: INode[t.Any, t.Any, t.Any], metadata: T, url: str
+) -> OriginalFile:
+    _assert_not_folder_node(file_node)
+    if isinstance(file_node, LazyNode) and file_node.get_link_path().is_file():
```
sylvlecl: The service layer must not know about the concept of "link", this must be handled inside the node implementation, by a way or another.
(but I think it's more or less a duplicate of the comment of Martin above about using the `parse(return_dataframe=True)` method)

sylvlecl (changes requested):
Looks good, however a comment about "links" handling which should not go into the service layer.

antarest/study/storage/abstract_storage_service.py
MartinBelthle: Actually what you did is more reliable as metadata.archived works for study but not for outputs. We'll keep the code as you wrote it.

antarest/study/storage/abstract_storage_service.py
mabw-rte: This is basically what I was going to do but the functions `extract_xxx` should only accept a `file node`, and there is no class that abstracts this concept. 
So In the end you should raise a `folder Error` in each `extract_xxx` or create a type that excludes `non-file-nodes`.
The first option is better.

antarest/study/storage/abstract_storage_service.py
mabw-rte: I prefer the `name` so the suffix is preserved in `filename` as it is returned in the headers

antarest/study/storage/abstract_storage_service.py
mabw-rte: We need to isolate this specific case but if it is moved into `inode.py` that would cause an `import loop` as the case correspond to a `LazyNode` which inherits from `Inode`.

Kinda of stuck

sylvlecl: Then we need to have an abstract method in `INode` and an implementation in `LazyNode`.
The `if isinstance(file_node, LazyNode)` also is a clear hint that inheritance is relevant here. For example, we can have
```python
class INode:
    def get_content() -> bytes:
        return self.config.path.read_bytes()
```
```python
class LazyNode:
    def get_content() -> bytes:
        # do stuff to handle links
```

sylvlecl: Also, `FolderNode` implementation should raise an exception.

antarest/study/storage/rawstudy/model/filesystem/folder_node.py
```diff
@@ -216,3 +216,6 @@ def extract_child(self, children: TREE, url: t.List[str]) -> t.Tuple[t.List[str]
             if not isinstance(children[name], child_class):
                 raise FilterError("Filter selection has different classes")
         return names, sub_url
+
+    def get_file_content(self) -> t.Tuple[bytes, str, str]:
+        raise PathIsAFolderError("Node is a folder node.")
```
MartinBelthle: You could add the self.config.path inside the error msg

antarest/study/storage/rawstudy/model/filesystem/inode.py
```diff
@@ -124,6 +124,35 @@ def denormalize(self) -> None:
...
+        if self.config.archive_path:
+            return (
+                read_original_file_in_archive(self.config.archive_path, str(self.get_relative_path_inside_archive())),
+                suffix,
+                filename,
+            )
+        else:
+            return self.config.path.read_bytes(), suffix, filename
+
+    def get_relative_path_inside_archive(self) -> Path:
```
MartinBelthle: I think you should give as an argument to this method the `archive_path`. This way you ensure we don't go inside the else and you can remove this Exception

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -146,6 +147,42 @@ def get_lazy_content(
...
+        suffix = self.config.path.suffix
+        filename = self.config.path.name
+        if self.config.archive_path:
+            return (
+                read_original_file_in_archive(self.config.archive_path, str(self.get_relative_path_inside_archive())),
+                suffix,
+                filename,
+            )
+        elif self.get_link_path().is_file():
+            with tempfile.TemporaryDirectory() as tmp_dir:
```
MartinBelthle: As I said last time, you can use a buffer instead of using a TemporaryDirectory I think. It would make the code easier and it would be faster I believe as you'll do everything in memory

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -146,6 +147,42 @@ def get_lazy_content(
...
+        """
+        suffix = self.config.path.suffix
+        filename = self.config.path.name
+        if self.config.archive_path:
+            return (
+                read_original_file_in_archive(self.config.archive_path, str(self.get_relative_path_inside_archive())),
+                suffix,
+                filename,
+            )
+        elif self.get_link_path().is_file():
```
MartinBelthle: I think you can move this method inside InputSeriesMatrix as it's the only type of LazyNode which has links.

antarest/core/exceptions.py
```diff
@@ -688,6 +688,16 @@ def __init__(self, message: str) -> None:
         super().__init__(HTTPStatus.NOT_FOUND, message)
 
 
+class PathIsAFolderError(HTTPException):
+    def __init__(self, message: str) -> None:
+        super().__init__(HTTPStatus.UNPROCESSABLE_ENTITY, message)
+
+
+class InvalidContentError(HTTPException):
```
MartinBelthle: You don't use this Exception anymore, remove it

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
```diff
@@ -297,206 +298,405 @@ def test_get_study(self, client: TestClient, user_access_token: str, internal_st
...
+
+        shutil.copytree(
+            ASSETS_DIR.joinpath("user"),
+            study_dir.joinpath("user"),
+            dirs_exist_ok=True,
+        )
+
+        # Then, use the API to download the files from the "user/folder" directory
+        user_folder_dir = study_dir.joinpath("user/folder")
+        for file_path in user_folder_dir.glob("*.*"):
```
MartinBelthle, line 546: Don't see the point of this part, you can remove it IMO

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
```diff
@@ -297,206 +298,405 @@ def test_get_study(self, client: TestClient, user_access_token: str, internal_st
...
+            "exception": "ChildNotFoundError",
+        }
+
+        # If you try to retrieve a folder, we should get an Error 422
+        res = client.get(original_file_url, params={"path": "user/folder"})
+        assert res.status_code == 422, res.json()
+        assert "Node is a folder node." in res.json()["description"]
+        assert res.json()["exception"] == "PathIsAFolderError"
+
+    def test_retrieve_original_file_from_archive(self, client: TestClient, user_access_token: str) -> None:
```
MartinBelthle: I think it could be better to parametrize this test, archived and unarchived.

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
```diff
@@ -297,206 +298,405 @@ def test_get_study(self, client: TestClient, user_access_token: str, internal_st
...
+
+        # retrieve the binding constraint `bc_1` less matrix
+        rel_path = "input/bindingconstraints/bc_1_lt"
+        res = client.get(
+            f"v1/studies/{study_id}/raw/original-file",
+            params={"path": rel_path},
+        )
+        res.raise_for_status()
+
+        # skip the test if the OS is Windows
```
MartinBelthle: You shouldn't skip this test on windows

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -146,6 +147,42 @@ def get_lazy_content(
...
+                # target path to save the file
+                target_path = Path(tmp_dir) / self.config.path.name
+                # use `.tsv` as suffix for the file
+                target_path = target_path.with_suffix(".tsv")
+                # retrieve the file content
+                self.parse(return_dataframe=True).to_csv(  # type: ignore
+                    target_path,
+                    sep="\t",
+                    index=False,
+                    header=True,
```
MartinBelthle: You shouldn't have headers

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
```diff
@@ -297,206 +298,405 @@ def test_get_study(self, client: TestClient, user_access_token: str, internal_st
...
-    # retrieve a `study.antares` file from inside the archive
-    rel_path = "study"
-    res = client.get(
-        f"/v1/studies/{study_id}/raw",
-        params={"path": rel_path, "formatted": True},
-    )
-    assert res.status_code == 200
+        res.raise_for_status()
+
+        # Create a cluster in the first area
```
MartinBelthle: You don't need to create the cluster and the area, it's just misleading as it doesn't bring anything. You can just create the BC like this and it still works: 
```python
bc_obj = {
    "name": "bc_1",
    "enabled": True,
    "time_step": "daily",
    "operator": "less",
}
res = client.post(f"/v1/studies/{study_id}/bindingconstraints", json=bc_obj)
res.raise_for_status()
```

MartinBelthle (changes requested):
It seems nice just a few comments

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
mabw-rte: it is the API test, so you retrieve a file via the API and verify that the bytes content match with what you have on disk

mabw-rte: you do this for many files so you know there is no a specific extension that does not work

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
mabw-rte: for the unarchived case it is done in the previous review and I would like to follow the same logic as we had with `raw` API before:
- tests for unarchived study with many file types
- specific test for an archived study

tests/integration/raw_studies_blueprint/test_fetch_raw_data.py
mabw-rte: I thought we already talked about this and agreed to ignore it for Windows as it fails for a link in an archive.

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -186,6 +190,51 @@ def get_study(
         json_response = to_json(output)
         return Response(content=json_response, media_type="application/json")
 
+    @bp.get(
+        "/studies/{uuid}/raw/original-file",
+        tags=[APITag.study_raw_data],
+        summary="Retrieve Raw file from a Study folder in its original format",
```
sylvlecl, line 197: We should document that this is not 100% true:
all original files are not kept, because we can transform them to store them (matrix store).

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -186,6 +190,51 @@ def get_study(
...
+    )
+    def get_study_file(
+        uuid: str,
+        path: str = Param("/", examples=get_path_examples()),  # type: ignore
```
sylvlecl: You should use `Query` instead of `Param`, should get rid of the type ignore.
Same elsewhere in the file, actually. `Param` seems to be a parent of Query, Headers etc, but we really want a query param here, not something else. I guess it's not exposed in `fastapi` top module for this reason.

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -186,6 +190,51 @@ def get_study(
...
+        output = original_file.content
+        suffix = original_file.suffix
+        headers = {
+            "Content-Disposition": f"attachment; filename={filename}",
+        }
+
+        # Guess the suffix form the filename suffix
+        content_type, _ = CONTENT_TYPES.get(suffix, (None, None))
+        if content_type:
+            return StreamingResponse(
```
sylvlecl: It's not useful to use a streaming response here, we already have the whole output in memory, so I would advise to use a plain `Response` instead for now, to avoid being misleading.
It would make sense if we got an iterable or a file object out of the `get_file` method, which would be better for large files, actually ! to avoid to load the whole file in memory.

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -186,6 +190,51 @@ def get_study(
...
+                io.BytesIO(output),
+                media_type=content_type,
+                headers=headers,
+            )
+        else:
+            # Unknown content types are considered binary,
+            # because it's better to avoid raising an exception.
+            return Response(content=output, media_type="application/octet-stream", headers=headers)
```
sylvlecl: only media type differs, we can merge it with the previous statement.

antarest/study/storage/rawstudy/model/filesystem/inode.py
```diff
@@ -124,6 +124,32 @@ def denormalize(self) -> None:
         """
         raise NotImplementedError()
 
+    def get_file_content(self) -> Tuple[bytes, str, str]:
```
sylvlecl: Why not a class instead of unnamed tuple ? I would be more clear and avoid the need for documentation copied from place to place

antarest/study/common/studystorage.py
```diff
@@ -17,13 +17,20 @@
...
 from antarest.core.requests import RequestParameters
+from antarest.core.serialization import AntaresBaseModel
 from antarest.study.model import Study, StudyMetadataDTO, StudyMetadataPatchDTO, StudySimResultDTO
 from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfigDTO
 from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
 
 T = t.TypeVar("T", bound=Study)
 
 
+class OriginalFile(AntaresBaseModel):
```
sylvlecl: More a thought, not a must-do: for such simple classes that are never serialized, I would propose to favor plain dataclasses. I think it makes the intent more clear: it's just a simple struct used in-memory.

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
```diff
@@ -131,3 +132,37 @@ def copy_file(self, target: str) -> None:
...
+                read_original_file_in_archive(
+                    self.config.archive_path, self.get_relative_path_inside_archive(self.config.archive_path)
+                ),
+                suffix,
+                filename,
+            )
+        elif self.get_link_path().is_file():
+            target_path = self.config.path.with_suffix(".txt")
+            buffer = io.StringIO()
+            self.parse(return_dataframe=True).to_csv(  # type: ignore
```
sylvlecl: Not necessarily in this PR, but we should remove this `return_dataframe` arg and have 2 separate methods instead. We should not need to add type ignores for this.

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
```diff
@@ -131,3 +132,37 @@ def copy_file(self, target: str) -> None:
...
+            return (
+                read_original_file_in_archive(
+                    self.config.archive_path, self.get_relative_path_inside_archive(self.config.archive_path)
+                ),
+                suffix,
+                filename,
+            )
+        elif self.get_link_path().is_file():
+            target_path = self.config.path.with_suffix(".txt")
+            buffer = io.StringIO()
```
sylvlecl: It seems we can provide a byteIO, this will avoid having to encode the result ourself in a second time (see .encode() below)

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
sylvlecl: Note that we have a `dump_dataframe` in `matrix.py` that was created to avoid code duplication where we need to re-use the same arguments of `to_csv` everywhere.
I think we should re-use this method (probably adapting it a bit to allow for an IO object).

sylvlecl (changes requested):
A few implementation details

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
MartinBelthle: I won't do it here as it's used inside other parts of the code

antarest/study/web/raw_studies_blueprint.py
MartinBelthle: I didn't find the right wording so I didn't document it

sylvlecl (approved):

MartinBelthle (approved):

PR #2248, opened 2024-11-26
Feat/poc add dao layer

TheoPascoli:
POC on link handling to test a new architecture. While currently limited to links, it can be extended in the future.

### Refactoring of DAO Layer

- Implemented `CompositeLinkDAO`, which coordinates interactions between:
  - **`LinkFromCacheDAO`**: Handles cached data.
  - **`LinkFromStorageDAO`**: Manages data in persistent storage.
- Both DAOs (`LinkFromCacheDAO` and `LinkFromStorageDAO`) now implement the common `LinkDAO` interface, ensuring a unified API for link management.

---

### Updated API and Service Layers

- API now exclusively works with `LinkDTO`, removing any dependencies on internal models.
- The service layer transforms `LinkDTO` objects into internal models when interacting with the DAOs, ensuring modularity.

---

### Improved Architecture

- Adopted a composition-based approach in `CompositeLinkDAO` for delegating tasks to cache and storage layers.
- Enhanced separation of concerns and flexibility for future extensions.

### Architecture Diagram

```mermaid
graph TD
    A[API<br>Receives requests<br>and returns DTOs]
    A --> B[Service<br>Business logic,<br>delegates to DAO]
    B --> C[Composite DAO<br>Inherits from LinkDAO<br>Delegates tasks to:<br>- LinkFromCacheDAO<br>- LinkFromStorageDAO]
    C --> D[LinkFromCacheDAO<br>Implements LinkDAO<br>Interacts with the caching system]
    C --> E[LinkFromStorageDAO<br>Implements LinkDAO<br>Interacts with persistent storage]

### DIFF
--- a/antarest/study/business/link/CompositeLinkDAO.py
+++ b/antarest/study/business/link/CompositeLinkDAO.py
@@ -1,0 +1,93 @@
+# Copyright (c) 2024, RTE (https://www.rte-france.com)
+#
+# See AUTHORS.txt
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+
+from typing import List
+from antarest.study.business.link.LinkDAO import LinkDAO
+from antarest.study.business.model.link_model import LinkDTO
+from antarest.study.model import Study
+
+
+class CompositeLinkDAO(LinkDAO):
+    """
+    CompositeLinkDAO acts as a composite data access object (DAO) for managing study links.
+    It delegates operations to two underlying DAOs: one for cache and one for persistent storage.
+    This class ensures that the cache and storage are kept in sync during operations.
+
+    Attributes:
+        cache_dao (LinkDAO): The DAO responsible for managing links in the cache.
+        storage_dao (LinkDAO): The DAO responsible for managing links in persistent storage.
+    """
+
+    def __init__(self, cache_dao: LinkDAO, storage_dao: LinkDAO):
+        """
+        Initializes the CompositeLinkDAO with a cache DAO and a storage DAO.
+
+        Args:
+            cache_dao (LinkDAO): DAO for managing links in the cache.
+            storage_dao (LinkDAO): DAO for managing links in persistent storage.
+        """
+        self.cache_dao = cache_dao
+        self.storage_dao = storage_dao
+
+    def get_all_links(self, study: Study) -> List[LinkDTO]:
+        """
+        Retrieves all links for a given study.
+
+        This method first tries to retrieve links from the cache. If the cache is empty,
+        it fetches the links from the persistent storage, updates the cache with the retrieved links,
+        and then returns the list of links.
+
+        Args:
+            study (Study): The study for which to retrieve the links.
+
+        Returns:
+            List[LinkDTO]: A list of all links associated with the study.
+        """
+        links = self.cache_dao.get_all_links(study)
+        if not links:
+            links = self.storage_dao.get_all_links(study)
+            for link in links:
+                self.cache_dao.create_link(study, link)
+        return links
+
+    def create_link(self, study: Study, link_dto: LinkDTO) -> LinkDTO:
+        """
+        Creates a new link for a given study.
+
+        This method adds the link to both the persistent storage and the cache
+        to ensure that they remain consistent.
+
+        Args:
+            study (Study): The target study.
+            link_dto (LinkDTO): The link to be added.
+
+        Returns:
+            LinkDTO: The link that was added.
+        """
+        self.storage_dao.create_link(study, link_dto)
+        self.cache_dao.create_link(study, link_dto)
+        return link_dto
+
+    def delete_link(self, study: Study, area1_id: str, area2_id: str) -> None:
+        """
+        Deletes a specific link for a given study.
+
+        This method removes the link from both the persistent storage and the cache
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: antarest/study/business/link/CompositeLinkDAO.py
Comment (sylvlecl) [inline antarest/study/business/link/CompositeLinkDAO.py:1]: A general comment:
I think we should isolate all the DAO layer in its own package (I propose `antarest.study.dao`).

This DAO layer must be completely independent from the "business" layer which will use it.
The dependency must go only from the business layer to the DAO layer, not the other way around, so we should not mix them in the same package.

File: antarest/study/business/link/CompositeLinkDAO.py [hunk 1]
```diff
@@ -0,0 +1,94 @@
...
+        and then returns the list of links.
+
+        Args:
+            study (Study): The study for which to retrieve the links.
+
+        Returns:
+            List[LinkDTO]: A list of all links associated with the study.
+        """
+        links = self.cache_dao.get_all_links(study)
+        if not links:
```
sylvlecl, line 57: There is a difficulty with using the `LinkDAO` interface for the cache:
we cannot really know if the data is not in cache, or if it's just that the data is an empty list.

We should have an `Optional` instead, but we don't want to introduce those optionals in the interface just for the need of the cache.
So we are left with 2 options I think:
- either have a specific interface for the cache, with optionals
- or use the approach of the other PoC branch where the `CacheDAO` implementation can check itself if the data is in the cache or not

antarest/study/business/link/LinkFromCacheDAO.py
```diff
@@ -0,0 +1,124 @@
...
+class LinkFromCacheDAO(LinkDAO):
+    """
+    LinkFromCacheDAO is responsible for managing study links in the cache.
+    It provides methods to retrieve, create, and delete links stored in the cache.
+
+    Attributes:
+        redis (ICache): The cache interface used for storing and retrieving links.
+    """
+
+    def __init__(self, redis: ICache):
```
sylvlecl, line 31: We should not name it redis, the cache implementation is not always based on redis

sylvlecl (commented):
First quick comments

antarest/study/business/link/CompositeLinkDAO.py
sylvlecl: Or maybe a third alternative !

- the cache implementation could throw a `CacheMissError`, that we will catch here 

It will rely on a quite "hidden" convention but why not ! Relying on exceptions is quite pythonic from what I know ...

antarest/study/business/link/LinkFromStorageDAO.py
```diff
@@ -0,0 +1,121 @@
...
+            area2_id (str): The ID of the target area of the link.
+        """
+        file_study = self.storage_service.get_storage(study).get_raw(study)
+        command = RemoveLink(
+            area1=area1_id,
+            area2=area2_id,
+            command_context=self.storage_service.variant_study_service.command_factory.command_context,
+            study_version=file_study.config.version,
+        )
+        execute_or_add_commands(study, file_study, [command], self.storage_service)
```
sylvlecl, line 121: In the end the DAO layer should not use the commands, it's the opposite:
the commands should use the DAO layer.

Otherwise, we will not be able to change the underlying storage without changing the commands implementation.

antarest/study/business/link/CompositeLinkDAO.py
```diff
@@ -0,0 +1,94 @@
...
+
+        Args:
+            study (Study): The target study.
+            link_dto (LinkDTO): The link to be added.
+
+        Returns:
+            LinkDTO: The link that was added.
+        """
+        self.storage_dao.create_link(study, link_dto)
+        self.cache_dao.create_link(study, link_dto)
```
sylvlecl, line 78: I am not sure it's a good idea that the cache implementation has to implement the modifications methods:
i can be error prone because we may need to implement duplicate logic between the "storage" implementation and the "cache" implementation, and ensure they are consistent.

In general, we populate a cache from what we read from the underlying storage, so we could just do nothing with the cache here except invalidate it (remove the link from the cache). Then on the next read it would be put in the cache again.
There will be a small performance loss but it's probably not too bad, at least in a first implementation, compared to the risk of error we introduce by duplicating the modification logic

antarest/study/business/link/CompositeLinkDAO.py
```diff
@@ -0,0 +1,94 @@
...
+            study (Study): The study for which to retrieve the links.
+
+        Returns:
+            List[LinkDTO]: A list of all links associated with the study.
+        """
+        links = self.cache_dao.get_all_links(study)
+        if not links:
+            links = self.storage_dao.get_all_links(study)
+            for link in links:
+                self.cache_dao.create_link(study, link)
```
sylvlecl, line 60: I don't think we should use the create_link method here, it's not the same thing to actually create a link and to just put an existing link in the cache.

sylvlecl (commented):
A few more comments, about the cache population logic.

antarest/study/business/link/CompositeLinkDAO.py
```diff
@@ -0,0 +1,94 @@
...
+        Initializes the CompositeLinkDAO with a cache DAO and a storage DAO.
+
+        Args:
+            cache_dao (LinkDAO): DAO for managing links in the cache.
+            storage_dao (LinkDAO): DAO for managing links in persistent storage.
+        """
+        self.cache_dao = cache_dao
+        self.storage_dao = storage_dao
+
+    def get_all_links(self, study: Study) -> List[LinkDTO]:
```
sylvlecl, line 42: about naming: I think we should reserve DTO for the web layer. The DAO layer should have its own suffix, for example Data.

PR #2249, opened 2024-11-26
fix(outputs): allow reading inside archive + output with `.` in the name

MartinBelthle:
Fix [ANT-2476]

### DIFF
--- a/antarest/study/storage/rawstudy/model/filesystem/inode.py
+++ b/antarest/study/storage/rawstudy/model/filesystem/inode.py
@@ -14,7 +14,7 @@ from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Dict, Generic, List, Optional, Tuple, TypeVar
 
-from antarest.core.exceptions import ShouldNotHappenException, WritingInsideZippedFileException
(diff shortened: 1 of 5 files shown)


antarest/study/storage/rawstudy/model/filesystem/inode.py
```diff
@@ -148,13 +148,9 @@ def _extract_file_to_tmp_dir(
...
             the tmp_dir object which MUST be cleared after use of the file
         """
-        if self.config.archive_path is None:
-            raise ShouldNotHappenException()
-        inside_archive_path = self.config.path.relative_to(self.config.archive_path.parent / self.config.study_id)
-        if self.config.archive_path:
-            return extract_file_to_tmp_dir(self.config.archive_path, inside_archive_path)
-        else:
-            raise ShouldNotHappenException()
+        assert self.config.archive_path is not None
```
TheoPascoli: I suggest that you use a if condition instead of assert

TheoPascoli (changes requested):

TheoPascoli (approved):

PR #2251, opened 2024-11-27
feat(launcher): allow local launcher to work with xpress

MartinBelthle:
Will fix [ANT-2472]

Introduces few changes:
- allow user to launch a simulation with xpress and presolve in local (tested by Alexander and it works)
- use simulator 8.8.11 instead of 8.8.5
- Don't import failed outputs inside the study
- separate logs in stdout and stderr as it was done with the slurm launcher
- Persist the logs if the simulation fails (by introducing a local_workspace)


antarest/launcher/adapters/local_launcher/local_launcher.py
```diff
@@ -89,59 +88,60 @@ def run_study(
...
+
+            simulator_args, environment_variables = self.parse_launcher_options(launcher_parameters)
+            new_args = [str(antares_solver_path)] + simulator_args + [str(export_path)]
+
+            std_out_file = study_path / "output" / "logs" / f"{job_id}-err.log"
+            with open(std_out_file, "w") as err_file:
+                process = subprocess.Popen(
+                    new_args,
+                    env=environment_variables,
+                    stdout=subprocess.PIPE,
```
sylvlecl: I think we should log both to files, there is no reason to have this asymmetry which complicates the code.
Also, keeping the stdout log in memory uses memory for no added value.

antarest/launcher/adapters/local_launcher/local_launcher.py
```diff
@@ -89,59 +88,60 @@ def run_study(
...
-                universal_newlines=True,
-                encoding="utf-8",
-            )
-            self.job_id_to_study_id[str(uuid)] = (
+            self.callbacks.export_study(job_id, study_uuid, export_path, launcher_parameters)
+
+            simulator_args, environment_variables = self.parse_launcher_options(launcher_parameters)
+            new_args = [str(antares_solver_path)] + simulator_args + [str(export_path)]
+
+            std_out_file = study_path / "output" / "logs" / f"{job_id}-err.log"
```
sylvlecl: should be named err, not out

antarest/launcher/adapters/local_launcher/local_launcher.py
```diff
@@ -89,59 +88,60 @@ def run_study(
...
         study_uuid: str,
-        uuid: UUID,
+        study_path: Path,
+        job_id: str,
         launcher_parameters: LauncherParametersDTO,
     ) -> None:
         end = False
 
+        # create study logs
+        logs_path = study_path / "output" / "logs"
```
sylvlecl: As discussed:
to better separate responsibilities, like in the slurm launcher, we'd better not log straight to the study path, and remove it from the arguments.

We can have the same behaviour as in slurm launcher:
- while the computation is running or when it has failed, logs are in a directory managed by this service
- when computation succeeds, logs are copied to the actual output of the study

antarest/launcher/adapters/local_launcher/local_launcher.py
```diff
@@ -168,33 +168,53 @@ def stop_reading_output() -> bool:
...
                 str(e),
                 None,
             )
         finally:
-            logger.info(f"Removing launch {uuid} export path at {tmp_path}")
+            logger.info(f"Removing launch {job_id} export path at {tmp_path}")
             end = True
             shutil.rmtree(tmp_path)
 
+    def parse_launcher_options(self, launcher_parameters: LauncherParametersDTO) -> Tuple[List[str], Dict[str, Any]]:
```
sylvlecl: better make it private

antarest/launcher/adapters/local_launcher/local_launcher.py
```diff
@@ -168,33 +168,53 @@ def stop_reading_output() -> bool:
...
             output_id: Optional[str] = None
-            try:
-                output_id = self.callbacks.import_output(str(uuid), export_path / "output", {})
-            except Exception as e:
-                logger.error(
-                    f"Failed to import output for study {study_uuid} located at {export_path}",
-                    exc_info=e,
-                )
-            del self.job_id_to_study_id[str(uuid)]
+            if process.returncode == 0:
```
sylvlecl, line 138: In slurm laucher we still call the import output when it has failed, we need to understand if it makes sense

sylvlecl (commented):
Main comments are about log management, as discussed

antarest/launcher/adapters/local_launcher/local_launcher.py
MartinBelthle: From my understanding, if the simulator fails, the output is not retrieved from calin, therefore the `_import_output` fails. But with the local launcher the output exists and therefore it can be retrieved. I think we should keep the code as it is, else i would have to touch code shared with the slurm part

sylvlecl: Ok. Indeed if we still wanted to import the output, we should have some kind of status to say that it's a "failed output".

I wonder if we should not also delete the import output in slurm implementation, to clarify the behaviour ?

sylvlecl (commented):
Looks very good!

We need to do a few things for the desktop version:
 - adapt the config in resources/antares-desktop-fs
 - add the new empty directory in resources/antares-desktop-fs
 - we also need to write code in the installer to do the same thing to upgrade existing installations ...

sylvlecl (approved):

PR #2252, opened 2024-11-27
feat(commands): add creation timestamp and user name inside commands

maugde:
Resolve [ANT-1705](https://gopro-tickets.rte-france.com/browse/ANT-1705)


alembic/versions/a8e9ee2cd4be_add_user_and_updated_date_info_to_.py
```diff
@@ -0,0 +1,77 @@
...
+# constraint name
+USER_ID_FKEY = 'commandblock_user_id_fk'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('commandblock', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('user_id', sa.Integer(), nullable=True))
+        batch_op.add_column(sa.Column('updated_at', sa.DateTime(), nullable=True))
+        batch_op.alter_column('study_version',
```
MartinBelthle: Don't understand why you did this alteration

alembic/versions/a8e9ee2cd4be_add_user_and_updated_date_info_to_.py
```diff
@@ -0,0 +1,77 @@
...
+               existing_type=sa.VARCHAR(length=36),
+               nullable=True,
+               existing_server_default=sa.text("'fake_version'"))
+        batch_op.create_foreign_key(USER_ID_FKEY, 'identities', ['user_id'], ['id'], ondelete='CASCADE')
+
+    with op.batch_alter_table('group_metadata', schema=None) as batch_op:
```
MartinBelthle: Should be removed, same for other tables that have nothing to do with this dev. same for downgrade method

MartinBelthle (changes requested):
I talked with Alexander and put the conclusion inside the JIRA ticket but basically wehave to do the conversion in the back from user_id to user name. Also ,we have to do it cleverly meaning if we retrieve 500 commands don't do 500 requests to have the user name but rather one for each individual user

antarest/study/storage/variantstudy/model/dbmodel.py
```diff
@@ -66,6 +66,8 @@ class CommandBlock(Base):  # type: ignore
     version: int = Column(Integer)
     args: str = Column(String())
     study_version: str = Column(String(36))
+    user_id: int = Column(Integer, ForeignKey("identities.id", ondelete="CASCADE"), nullable=True)
```
MartinBelthle: I think we should put set_null instead of cascade as I think we don't want to remove the command if the user is deleted but rather remove the info on who did it. But it's a case that never happens (we never delete user) so it probably doesn't really matter

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -106,6 +107,17 @@ def __init__(
...
+    @staticmethod
+    def _get_user_name_from_id(user_id: int) -> str:
+        """
+        Utility method that retrieves a user's name based on their id.
+        Args:
+            user_id: user id (user must exist)
+        Returns: String representing the user's name
+        """
+        user_obj = db.session.query(Identity).get(user_id)
+        return t.cast(str, user_obj.name)
```
MartinBelthle: Why do you have to cast to str ?

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -119,7 +131,9 @@ def get_command(self, study_id: str, command_id: str, params: RequestParameters)
 
         try:
             index = [command.id for command in study.commands].index(command_id)  # Maybe add Try catch for this
```
MartinBelthle: Not your code but remove comment as there's already a try catch

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -119,7 +131,9 @@ def get_command(self, study_id: str, command_id: str, params: RequestParameters)
...
-            return t.cast(CommandDTOAPI, study.commands[index].to_dto().to_api())
+            command = study.commands[index]
+            user_name = self._get_user_name_from_id(command.user_id)
```
MartinBelthle: command.user_id could be None. If so don't call the DB just put None as the user_name

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -132,7 +146,16 @@ def get_commands(self, study_id: str, params: RequestParameters) -> t.List[Comma
...
         study = self._get_variant_study(study_id, params)
-        return [command.to_dto().to_api() for command in study.commands]
+
+        id_to_name: t.Dict[int, str] = {}
+        command_list = []
+
+        for command in study.commands:
+            if command.user_id and command.user_id not in id_to_name.keys():
+                user_name = self._get_user_name_from_id(command.user_id)
+                id_to_name[command.user_id] = str(user_name)
```
MartinBelthle: Same about the cast to str

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -201,6 +224,13 @@ def append_commands(
         command_objs = self._check_commands_validity(study_id, commands)
         validated_commands = transform_command_to_dto(command_objs, commands)
         first_index = len(study.commands)
+
+        # type check in case user is None
+        if not params.user or not params.user.id:
```
MartinBelthle: I belive you can't reach this code as you check above that the user has the right to edit the study and if the user is None we crash here. I'll let you check but I think you can remove this if

MartinBelthle (changes requested):
The code seems okay but you'll have to add integration tests.
- Tests with 2 users altering the same study, you retrieve their names.
- Tests that one user appending two commands only call once the DB for user name
- Tests that it works with bots (add the test inside test_integration_token_end_to_end)

antarest/study/storage/variantstudy/variant_study_service.py
maugde: Mypy cannot know the type of user_obj.name. Maybe I have to do it another way

antarest/study/storage/variantstudy/snapshot_generator.py
```diff
@@ -67,7 +67,7 @@ def __init__(
     def generate_snapshot(
         self,
         variant_study_id: str,
-        jwt_user: JWTUser,
+        owner: JWTUser,
```
MartinBelthle: Why did you change this name ? I think it was better before as it's a JWTUser class. Also I believe any user can reach this code, it doesn't have to be the study owner

tests/integration/test_integration_token_end_to_end.py
```diff
@@ -173,6 +174,15 @@ def test_nominal_case_of_an_api_user(client: TestClient, admin_access_token: str
     res = client.post(f"/v1/studies/{variant_id}/commands", headers=bot_headers, json=commands)
     assert res.status_code == 200
 
+    # check if the author's name and date of update are retrieved with commands created by a bot
+    commands_res = client.get(f"/v1/studies/{variant_id}/commands", headers=bot_headers)
+    assert commands_res.status_code == 200
+    assert commands_res.json()
+
+    for command in commands_res.json():
+        assert command.get("user_name") in ["admin", "admin_bot"]
```
MartinBelthle: Could you check strictly that each command has the correct user name ?
Also minor but I prefer to assert "updated_at" is inside the command than doing a get as I find it less readable but not really important

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -118,8 +130,10 @@ def get_command(self, study_id: str, command_id: str, params: RequestParameters)
         study = self._get_variant_study(study_id, params)
 
         try:
-            index = [command.id for command in study.commands].index(command_id)  # Maybe add Try catch for this
-            return t.cast(CommandDTOAPI, study.commands[index].to_dto().to_api())
+            index = [command.id for command in study.commands].index(command_id)
+            command = study.commands[index]
+            user_name = self._get_user_name_from_id(command.user_id) if command.user_id else None
+            return t.cast(CommandDTOAPI, command.to_dto().to_api(user_name))
```
MartinBelthle: Instead of casting, you can type: `command: CommandBlock = study.commands[index]` and mypy will understand the code so you won't have to type

tests/variantstudy/model/test_variant_model.py
```diff
@@ -190,3 +198,104 @@ def test_commands_service(
...
+    @pytest.mark.parametrize("root_study_id", [False], indirect=True)
+    @with_db_context
+    def test_command_same_author(
+        self,
+        jwt_user: JWTUser,
+        variant_study_service: VariantStudyService,
+        root_study_id: str,
+    ):
+        """
+        Test if multiple commands of the same user have the same author
```
MartinBelthle: This test doc is not really clear :/ 
I would prefer something more verbose with bullet points and the same structure as the test above.

tests/variantstudy/model/test_variant_model.py
```diff
@@ -190,3 +198,104 @@ def test_commands_service(
...
+                CommandDTO(
+                    action="update_comments",
+                    args={"comments": f"new comment {index}"},
+                    study_version=StudyVersion.parse(variant_study.version),
+                )
+            )
+
+        variant_study_service.append_commands(variant_study.id, commands, params=owner_params)
+
+        with patch.object(
```
MartinBelthle: It works well but it doesn't really check that we call the DB once, it checks we call the method. If you can change this to really change the DB it would be nice. Otherwise it's not really a big deal

MartinBelthle (changes requested):
Minor reviews, it seems almost done :)

tests/variantstudy/model/test_variant_model.py
```diff
@@ -190,3 +199,126 @@ def test_commands_service(
...
+    @pytest.mark.parametrize("root_study_id", [True], indirect=True)
+    @with_db_context
+    def test_command_several_authors(
+        self,
+        jwt_user: JWTUser,
+        variant_study_service: VariantStudyService,
+        root_study_id: str,
+    ):
+        """
+        Test two different users that are authors of two different commands of the same variant
```
MartinBelthle: typo: of -> on

tests/variantstudy/model/test_variant_model.py
```diff
@@ -190,3 +199,126 @@ def test_commands_service(
...
+        Set up:
+            Retrieve the user that will be the owner of the study and variant
+            Create a second user
+            Create a study and a variant study
+            Create a command created by each user
```
MartinBelthle: Not really clear: Each user creates a command is clearer IMO

tests/variantstudy/model/test_variant_model.py
```diff
@@ -190,3 +199,126 @@ def test_commands_service(
...
+                    args={"comments": f"new comment {index}"},
+                    study_version=StudyVersion.parse(variant_study.version),
+                )
+            )
+
+        variant_study_service.append_commands(variant_study.id, commands, params=owner_params)
+
+        nb_queries_before = nb_queries  # store initial state
+
+        with patch.object(
```
MartinBelthle: Now that you introduced the orm listener I think you can remove this as it's an implementation detail

tests/variantstudy/model/test_variant_model.py
```diff
@@ -190,3 +199,126 @@ def test_commands_service(
...
+            Make the user generates five commands on the newly created variant
+        Test:
+            Each time a command is retrieved, the database must be accessed only if
+            the author of the currently retrieved command is not already known during
+            the process
+        """
+        nb_queries = 0  # Store number of orm queries to database
+
+        # Watch orm events and update `nb_queries`
+        @event.listens_for(db.session, "do_orm_execute")
```
MartinBelthle, line 291: Nice :)

tests/integration/test_integration_token_end_to_end.py
```diff
@@ -173,6 +174,23 @@ def test_nominal_case_of_an_api_user(client: TestClient, admin_access_token: str
     res = client.post(f"/v1/studies/{variant_id}/commands", headers=bot_headers, json=commands)
     assert res.status_code == 200
 
+    # Check if the author's name and date of update are retrieved with commands created by a bot
+    commands_res = client.get(f"/v1/studies/{variant_id}/commands", headers=bot_headers)
+    assert commands_res.status_code == 200
+    assert commands_res.json()
```
MartinBelthle: This check doesn't really make sense you can remove it

tests/integration/test_integration_token_end_to_end.py
```diff
@@ -173,6 +174,23 @@ def test_nominal_case_of_an_api_user(client: TestClient, admin_access_token: str
...
+
+    for command in commands_res.json():
+        # Some commands, such as those that modify study configurations, are run by admin user
+        # Thus the `user_name` for such type of command will be the admin's name
```
MartinBelthle, line 182: This is really weird. I'll check why we have this. Ah okay I see why. It's because inside `execute_or_add_commands` we use admin user to add a command ... It's a problem. We're gonna tackle this in a next PR I discussed with Sylvain about this issue, you'll do a 2nd dev after this to fix it :) (see JIRA ticket 2543)


On the other hand, it made me realize that there are some places where you didn't put the updated_at and the user_id 
- line 273 and 916 in variant_study_service.

MartinBelthle (changes requested):

MartinBelthle (approved):

PR #2256, opened 2024-12-02
fix(links): fix a bug that occurred when updating links via table-mode

TheoPascoli:
Resolve a bug that occurs while updating a link in the table mode view, with some refactoring.


antarest/study/business/link_management.py
```diff
@@ -11,35 +11,24 @@
...
 from antarest.study.model import RawStudy, Study
-from antarest.study.storage.rawstudy.model.filesystem.config.links import LinkProperties
 from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
 from antarest.study.storage.storage_service import StudyStorageService
 from antarest.study.storage.variantstudy.model.command.create_link import CreateLink
 from antarest.study.storage.variantstudy.model.command.remove_link import RemoveLink
-from antarest.study.storage.variantstudy.model.command.update_config import UpdateConfig
 from antarest.study.storage.variantstudy.model.command.update_link import UpdateLink
 
 _ALL_LINKS_PATH = "input/links"
```
MartinBelthle: You don't use this constant you can remove it.

antarest/study/business/link_management.py
```diff
@@ -99,25 +99,21 @@ def update_link(self, study: RawStudy, area_from: str, area_to: str, link_update
...
-            raise LinkNotFound(f"The link {link.area1} -> {link.area2} is not present in the study")
-
-    def get_internal_link(self, study: RawStudy, link: LinkInternal) -> LinkInternal:
-        file_study = self.storage_service.get_storage(study).get_raw(study)
-
-        link_properties = self.get_link_if_exists(file_study, link)
-
-        link_properties.update({"area1": link.area1, "area2": link.area2})
-        updated_link = LinkInternal.model_validate(link_properties)
+    def update_all_links(
```
MartinBelthle: I woudn't call it `update_all_links` as you don't necessary update all links, you can just update some of them. So `update_links` seems better

tests/integration/study_data_blueprint/test_table_mode.py
```diff
@@ -188,25 +189,176 @@ def test_lifecycle__nominal(
...
+            "linkWidth",
+            "loopFlow",
+            "transmissionCapacities",
+            "usePhaseShifter",
         }
 
-        res = client.put(
-            f"/v1/studies/{internal_study_id}/table-mode/links",
-            json={
+        if study_version < STUDY_VERSION_8_2:
```
MartinBelthle: Why do you test it only after v8.2 ? Is it because of the filterSynthesis/yearByYear attributes  ? If so, you could just adapt the expected response according to the version but perform this test for all versions

tests/integration/study_data_blueprint/test_table_mode.py
```diff
@@ -188,25 +189,176 @@ def test_lifecycle__nominal(
...
+                    "transmissionCapacities": "enabled",
+                    "usePhaseShifter": False,
+                },
+            }
+
+            # asserts actual equals expected without the non-updated link.
+            actual = res.json()
+            expected_result = copy.deepcopy(expected_links)
+            del expected_result["de / it"]
+            assert actual == expected_result
```
MartinBelthle: This test file is already really long (1000+ lines), could you store the JSON inside a variable instead of copying it everywhere as it makes the test hard to read

MartinBelthle (changes requested):
The bug is fixed but I don't see the test case inside the tests you've added, could you add it please ? (just a get /links inside the table-mode to ensure we didn't break the links)

antarest/study/business/link_management.py
TheoPascoli: Yes

antarest/study/business/link_management.py
TheoPascoli: Indeed

tests/integration/study_data_blueprint/test_table_mode.py
TheoPascoli: Ok, I changed the logic here

tests/integration/study_data_blueprint/test_table_mode.py
TheoPascoli: I extracted the JSON input and the expected links into another method that determines the desired output based on the version. While it is not shorter in terms of line count, it is definitely easier to read

TheoPascoli:
> The bug is fixed but I don't see the test case inside the tests you've added, could you add it please ? (just a get /links inside the table-mode to ensure we didn't break the links)

There is a GET request right after the PUT request to ensure that the link has been modified and returns the correct values

tests/integration/study_data_blueprint/test_table_mode.py
```diff
@@ -188,128 +190,39 @@ def test_lifecycle__nominal(
...
-                    "comments": "Link from France to Italie",
-                    "assetType": "DC",  # case-insensitive
-                },
-            },
-        )
+        # Test links
+
+        json_input, expected_links = get_table_mode_link_test_data(StudyVersion.parse(study_version))
+
+        # Envoi de la requête PUT
```
MartinBelthle: English plz :). Same for the next comment

tests/integration/study_data_blueprint/test_table_mode.py
```diff
@@ -188,128 +190,39 @@ def test_lifecycle__nominal(
...
+        # Validation partielle
         actual = res.json()
-        expected_result = copy.deepcopy(expected_links)
-        del expected_result["de / it"]
-        assert actual == expected_result
+        expected_partial = copy.deepcopy(expected_links)
+        del expected_partial["de / it"]
+        assert actual == expected_partial
 
+        # Requête GET pour validation complète
```
MartinBelthle: What I meant by the test-case was GET /links as the endpoint which is called inside the map not inside the table-mode as it was this call that used to fail. And unless I'm missing something I don't see it

tests/integration/study_data_blueprint/test_table_mode.py
```diff
@@ -945,3 +858,207 @@ def test_table_type_aliases(client: TestClient, user_access_token: str) -> None:
...
+        }
+    else:
+        json_input = {
+            "de / fr": {
+                "assetType": "ac",
+                "colorb": 100,
+                "colorg": 150,
+                "colorr": 200,
+                "displayComments": False,
+                "filterSynthesis": "hourly, daily, weekly, annual",
```
MartinBelthle: Couldn't you just add these keys ?
json_input = ...
if version > 82:
   add keys.

It would be shorter and clearer

Same for the rest of the method

MartinBelthle (changes requested):
Minor changes only inside tests

tests/integration/study_data_blueprint/test_table_mode.py
TheoPascoli: Unfortunately, the GET /links endpoint formats the link results differently because the result classes are distinct. As a result, defining the expected values would require adding a significant amount of JSON text to this method, which I believe is unnecessary

tests/integration/study_data_blueprint/test_table_mode.py
TheoPascoli: Ok done

tests/integration/study_data_blueprint/test_table_mode.py
MartinBelthle: Yep indeed it's unnecesary I meant just test that we receive a 200 and not a 422

MartinBelthle (changes requested):
Okay for me if you add the little test case. 
But you have a lot of things in this PR that are already inside dev try fixing it :)

TheoPascoli:
> Okay for me if you add the little test case. But you have a lot of things in this PR that are already inside dev try fixing it :)

It's clean now

MartinBelthle (approved):

PR #2259, opened 2024-12-02
fix(events): avoid slow processing of events

sylvlecl:
- events are now retrieved in batch from redis, instead of 1 by 1
- the event loop is only put at rest when it has nothing to process any more


sylvlecl:
I could check in dev environment that it fixes the issue:
- doing a GET synthesis on a variant study with 500 commands triggers a light generation with lots of events
- triggering a timeseries generation just after

--> works as expected after the fix, and gets stuck before

PR #2260, opened 2024-12-02
feat(ui-debug): add unsupported files handling

hdinia:
ANT-2362 => handle unsupported files display
ANT-2529 => display output matrices as raw text


webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
```diff
@@ -83,21 +83,44 @@ export function isFolder(treeData: TreeData): treeData is TreeFolder {
...
+    if (treeData.startsWith("file://")) {
+      const supportedTextExtensions = [".txt", ".log", ".csv", ".tsv", ".ini"];
+
+      // Check if the file ends with any of the supported extensions
+      if (supportedTextExtensions.some((ext) => treeData.endsWith(ext))) {
+        return "text";
+      }
+
+      // Any other extension with file:// prefix is unsupported
+      return "unsupported";
```
skamril: Remove this return and replace `return "text";` at the end by `return "unsupported"`

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
```diff
@@ -83,21 +83,44 @@ export function isFolder(treeData: TreeData): treeData is TreeFolder {
...
+
+    // Handle regular files with file:// prefix
+    // All files except matrices and json-formatted content use this prefix
+    // We filter to only allow extensions that can be properly displayed (.txt, .log, .csv, .tsv, .ini)
+    // Other extensions (like .RDS or .xlsx) are marked as unsupported since they can't be shown in the UI
+    if (treeData.startsWith("file://")) {
+      const supportedTextExtensions = [".txt", ".log", ".csv", ".tsv", ".ini"];
+
+      // Check if the file ends with any of the supported extensions
+      if (supportedTextExtensions.some((ext) => treeData.endsWith(ext))) {
```
skamril: By security `ext.toLowerCase()`

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
```diff
@@ -0,0 +1,75 @@
...
+  return (
+    <Flex>
+      <Menubar>
+        <Filename>{filename}</Filename>
+        <UploadFileButton
+          studyId={studyId}
+          path={filePath}
+          accept={{ "text/plain": [".txt"] }}
+          onUploadSuccessful={handleUploadSuccessful}
+        />
```
hdinia: upload does not make sense for this type of files in my opinion, I added it because it's currently enabled on dev, but not sure we should have it, what do you think ?

webapp/public/locales/en/main.json
```diff
@@ -252,6 +252,7 @@
   "study.bindingconstraints": "Binding Constraints",
   "study.debug": "Debug",
   "study.debug.file.image": "Image file",
```
skamril: Remove image key

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  ////////////////////////////////////////////////////////////////
+  // JSX
+  ////////////////////////////////////////////////////////////////
+
+  return (
+    <Flex>
+      <Menubar>
+        <Filename>{filename}</Filename>
+        <UploadFileButton
```
skamril: `{canEdit && ...`

webapp/public/locales/fr/main.json
```diff
@@ -253,6 +253,7 @@
   "study.debug": "Debug",
   "study.debug.file.image": "Fichier image",
```
skamril: Remove image key

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
```diff
@@ -111,24 +123,31 @@ function Text({ studyId, filePath, filename, canEdit }: DataCompProps) {
...
-              customStyle={{
-                margin: 0,
-                padding: theme.spacing(2),
-                borderRadius: theme.shape.borderRadius,
-                fontSize: theme.typography.body2.fontSize,
-              }}
-              {...getSyntaxProps(text)}
+            <DownloadButton
+              onClick={handleDownload}
+              disabled={isEmptyContent(text)}
```
skamril: For me it's not necessary to disable download for this

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
```diff
@@ -111,24 +123,31 @@ function Text({ studyId, filePath, filename, canEdit }: DataCompProps) {
...
             />
-          </Box>
+          </Menubar>
+          {isEmptyContent(text) ? (
+            <EmptyView icon={GridOffIcon} title={t("study.results.noData")} />
```
skamril, line 128: It can be confused to display a message for an empty file (no data can be undestand as there is not file at all), display nothing for me (or ask Alexander)

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  ////////////////////////////////////////////////////////////////
+  // Event Handlers
+  ////////////////////////////////////////////////////////////////
+
+  const handleDownload = () => {
+    if (res.data) {
+      downloadFile(
+        res.data,
+        filename.endsWith(".txt") ? filename : `${filename}.txt`,
```
skamril: Use extension of the file

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
```diff
@@ -0,0 +1,75 @@
...
+  ////////////////////////////////////////////////////////////////
+
+  return (
+    <Flex>
+      <Menubar>
+        <Filename>{filename}</Filename>
+        <UploadFileButton
+          studyId={studyId}
+          path={filePath}
+          accept={{ "text/plain": [".txt"] }}
```
skamril: Use type from the file

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
skamril: For you can keep it

webapp/src/components/App/Singlestudy/explore/Debug/Data/index.tsx
```diff
@@ -12,44 +12,40 @@
  * This file is part of the Antares project.
  */
 
+import { ComponentType } from "react";
```
skamril, line 15: Use `React.ComponentType` without import

skamril (changes requested):

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
skamril: It's good for me

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
hdinia: if its empty you should not be able to download it

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
hdinia: why add `canEdit`, it's supposed to display nothing anyway ?

webapp/public/locales/en/main.json
hdinia: thanks forgot to remove them

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
hdinia: its not needed, as I want to match the exact extension and exclude everything else

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
hdinia: thanks, will see that and test if it behaves like I wanted

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
hdinia: ok I let it

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
hdinia: before:
image



after the changes:
image


it makes sense to handle this state, I think its fine 
you're right about the message that can be maybe more explicit like "empty file content" but does not make a huge difference in my opinion

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
skamril: Upload is only allowed on certain folders

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
skamril: yes but you can have ".TXT" or ".CSV" and it's valid

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
skamril: yes, maybe the space are used I don't know, I wouldn't add a restriction for this

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
skamril: Better to ask Alexander to prevent a feedback

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
skamril: Even if the user cannot show the file, he can dowload it and update it

hdinia: The users must still be able to download files (for the upload I am not sure)

skamril: Upload is allowed only when the file is editable, this why we have to use `canRead` property

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
```diff
@@ -0,0 +1,71 @@
...
+  ////////////////////////////////////////////////////////////////
+
+  return (
+    <Flex>
+      <Menubar>
+        <Filename>{filename}</Filename>
+        <UploadFileButton
+          studyId={studyId}
+          path={filePath}
+          onUploadSuccessful={handleUploadSuccessful}
```
skamril: You removed the `accept` property

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
hdinia: This is not correct, **import + export (+file edit) should always be available on all files** as requested by users, even on the "user" folder

Also the code you mention is confusing, the purpose of the issue you addressed with `canEdit` function is to prevent the deletion on the "user/expansion" folder. (see ANT-2074)

```typescript
/**
 * Checks if a study's file can be edited.
 *
 * @param study  - The study where the file is located.
 * @param filePath - The path of the file.
 * @returns True if the file can be edited, false otherwise.
 */
export function canEditFile(study: StudyMetadata, filePath: string): boolean {
  return (
    !study.archived &&
    (filePath === "user" || filePath.startsWith("user/")) &&
    // To remove when Xpansion tool configuration will be moved to "input/expansion" directory
    !(filePath === "user/expansion" || filePath.startsWith("user/expansion/"))
  );
}
```

1- this should come from the api and not handled by a frontend manipulation, we should ask for a boolean flag that indicate if the file or folder is protected
2- the function name and functionality does not reflect its purpose 

I am updating this to reflect the requirements

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
hdinia: **update after changes:** this change is problematic for some special files that does not have the file:// prefix but should be displayed as text (like the `list` file in `input/areas`) its a very rare case but we need to keep the text fallback as a default case, I am reverting the change you asked

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
hdinia: you are right about this it should not be disabled mybad

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
hdinia: **Update:** you are not wrong but the current debug view does not allow to edit text files, when it will be the case we should set the text editor to a fixed height and allow edition even if the file is empty. Currently it does not make sense to display an empty block

webapp/src/components/App/Singlestudy/explore/Debug/Data/Unsupported.tsx
hdinia: We should not limit the type of file that can be uploaded, as we don't know at that moment the exact list of all types of files and extensions we can upload. And also we don't have a predictable way to get the extension of the current file from the API, we have it in the path and we have to manipulate it to get the extension.
What you recommand for this ?

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
```diff
@@ -49,17 +50,37 @@ export interface DataCompProps extends FileInfo {
...
+  fileType: string;
+}
+
 ////////////////////////////////////////////////////////////////
-// File Info
+// Utils
 ////////////////////////////////////////////////////////////////
 
+const URL_SCHEMES = {
+  MATRIX: ["matrix://", "matrixfile://"],
```
skamril, line 63: By convention, lower case are used for object keys, even for a constant

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
```diff
@@ -83,26 +104,31 @@ export function isFolder(treeData: TreeData): treeData is TreeFolder {
...
-    if (treeData.startsWith("file://") && treeData.endsWith(".ico")) {
-      return "image";
-    }
+
+    // Handle regular files with file:// prefix
+    // All files except matrices and json-formatted content use this prefix
+    // We filter to only allow extensions that can be properly displayed (.txt, .log, .csv, .tsv, .ini)
+    // Other extensions (like .RDS or .xlsx) are marked as unsupported since they can't be shown in the UI
+    return treeData.startsWith(URL_SCHEMES.FILE) &&
+      SUPPORTED_EXTENSIONS.some((ext) => treeData.endsWith(ext))
```
skamril: Like I had already said to you, use `treeData.toLowerCase().endsWith(ext)`, because extension are case insensitive, `.txt` and `.TXT` are both valid, an user can import file with uppercase extension!

skamril: Be careful, file without extension are text

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
skamril: Even if the file is not editable, we have to display the content, even if there is only line breaks. To change this behavior by display a specific message, we need a spec. This is why I asked you to tell Alexander about it.

And spaces and line breaks != no data.

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
```diff
@@ -115,7 +145,138 @@ export function canEditFile(study: StudyMetadata, filePath: string): boolean {
...
 }
+
+/**
+ * Checks if a file path is within the output folder
+ *
+ * @param path - The file path to check
+ * @returns Whether the path is in the output folder
+ */
+export function isOutputFolder(path: string): boolean {
+  return path.startsWith("output/");
```
skamril, line 162: "ouput" value has been forgotten:

```typescript
return path.split("/")[0] === "output";
```

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
```diff
@@ -113,22 +121,34 @@ function Text({ studyId, filePath, filename, canEdit }: DataCompProps) {
...
+                  paddingRight: theme.spacing(3),
+                }}
+                customStyle={{
+                  margin: 0,
+                  padding: theme.spacing(2),
+                  borderRadius: theme.shape.borderRadius,
+                  fontSize: theme.typography.body2.fontSize,
+                  flex: 1,
+                }}
+                {...getSyntaxProps(parseContent(text, { filePath, fileType }))}
```
skamril: Better to parse in the usePromiseWithSnackbarError to make it once

skamril (changes requested):

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
hdinia: it's intended

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
skamril: Or rename `isInOuputFolder`

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
skamril: Sorry I thought that txt files without extension were possible.

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
hdinia: For this type of constant object where both the object name and its keys represent fixed values or schemes, it's more conventional to have both in UPPER_SNAKE_CASE, it's not a regular enum-like object see: https://google.github.io/styleguide/tsguide.html#naming

skamril: Your link said like me lol https://google.github.io/styleguide/tsguide.html#identifiers-constants

```typescript
const URL_SCHEMES = {
  matrix: ...,
  ...
}
```

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
hdinia: it's intended I'll rename it to better reflect the purpose

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
hdinia: at this point their no possible error

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
hdinia: mybad, its fixed

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
skamril: Yes the error it's for the API call, but I said to had directly after the call to memoized the value.

```typescript
() => getStudyData<string>(studyId, filePath).then(text => parseContent(text, { filePath, fileType })):
...
deps: [studyId, filePath, fileType],
```

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
hdinia: I know its confusing, the style guide example uses lowercase because they're dealing with actual unit names rather than symbolic constants.

The difference is that:

In URL_SCHEMES, the keys (MATRIX, JSON, FILE) are symbolic identifiers representing categories/types
In UNIT_SUFFIXES, the keys ('milliseconds', 'seconds') are the actual string names of time units

look at "Rules by identifier type" section:
says:  "CONSTANT_CASE 	global constant values, including enum values. See [Constants](https://google.github.io/styleguide/tsguide.html#identifiers-constants) below."

hdinia: we should create a ticket and spend time to make our set of rules, especially for this kind of conventions

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
hdinia: ok I will see that

webapp/src/components/App/Singlestudy/explore/Debug/utils.ts
hdinia: for a more general example:

1. When keys are symbolic constants (use UPPER_SNAKE_CASE):
```typescript
const ERROR_CODES = {
    NOT_FOUND: 404,
    SERVER_ERROR: 500,
    UNAUTHORIZED: 401
};
```
Here, NOT_FOUND is a symbolic name representing the concept of a "not found" error. The key itself is a symbol/identifier.

2. When keys are actual VALUES/NAMES (use lowercase):
```typescript
const MIME_TYPES = {
    'image/jpeg': '.jpg',
    'image/png': '.png',
    'text/html': '.html'
};
```
Here, 'image/jpeg' is the actual MIME type string that would be used in HTTP headers.


Think of it this way:
- If you're using the key as a reference/symbol (like URL_SCHEMES.MATRIX), use UPPER_SNAKE_CASE
- If you're using the key as a literal value (like MIME_TYPES['image/jpeg']), use lowercase

skamril: Alright! I didn't know this subtlety.

webapp/src/components/App/Singlestudy/explore/Debug/Data/Text.tsx
hdinia: indeed its good like this, done

skamril (approved):

PR #2262, opened 2024-12-04
fix(ui-bc): use `matrixindex` timesteps for row display

hdinia:


MartinBelthle:
Don't know if the code is good but from an user perspective it works well :)

skamril (approved):

PR #2263, opened 2024-12-06
fix(tasks): fix frozen task with load balanced pgpool

sylvlecl:
This work aims at fixing issues encountered with load balanced pgpool,
where a commit is not always instantly visible in another session (when
read from replica).

Fixes are:
- restore a workaround which consists in re-trying the read until success
- ensure exceptions occurring in the task execution thread are 
  ALL caught

Also removing a technical debt: 
- remove optional session from task job repo which was just used in tests.
  It did not work with code which uses both the repo and also the session
  from the singleton db.session.

### DIFF
--- a/antarest/core/tasks/repository.py
+++ b/antarest/core/tasks/repository.py
@@ -27,15 +27,6 @@ class TaskJobRepository:
     Database connector to manage Tasks/Jobs entities.
     """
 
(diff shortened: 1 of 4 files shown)


MartinBelthle (commented):
Okay for me once the installer modifications are removed from this PR

MartinBelthle (approved):

PR #2265, opened 2024-12-11
feat(ui-cmd): add command details on variants commands panel

hdinia:
ANT-2536

### DIFF
--- a/webapp/src/components/App/Singlestudy/Commands/Edition/DraggableCommands/CommandListItem/CommandDetails.tsx
+++ b/webapp/src/components/App/Singlestudy/Commands/Edition/DraggableCommands/CommandListItem/CommandDetails.tsx
@@ -1,0 +1,83 @@
+/**
+ * Copyright (c) 2024, RTE (https://www.rte-france.com)
+ *
+ * See AUTHORS.txt
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+import { Fragment } from "react";
+import { Box, Typography, Divider, Tooltip } from "@mui/material";
+import { CommandItem } from "../../commandTypes";
+import { PersonOutlineOutlined, UpdateOutlined } from "@mui/icons-material";
+import { format, formatDistanceToNow, parseISO } from "date-fns";
+import { enUS, fr } from "date-fns/locale";
+import { useTranslation } from "react-i18next";
+import { getCurrentLanguage } from "@/utils/i18nUtils";
+
+interface Props {
+  item: CommandItem;
+}
+
+const formatDate = (dateStr: string) => {
+  const utcDate = parseISO(dateStr + "Z");
+  return format(utcDate, "dd/MM/yyyy HH:mm");
+};
+
+const formatDateWithLocale = (dateStr: string) => {
+  const date = parseISO(dateStr);
+  const lang = getCurrentLanguage();
+  const locale = lang.startsWith("fr") ? fr : enUS;
+
+  return formatDistanceToNow(date, {
+    addSuffix: true,
+    locale,
+  });
+};
+
+function CommandDetails({ item }: Props) {
+  const [t] = useTranslation();
+  const details = [
+    {
+      icon: <PersonOutlineOutlined sx={{ fontSize: 20 }} />,
+      text: item.user || t("global.unknown"),
+    },
+    {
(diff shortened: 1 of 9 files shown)


MartinBelthle:
3 remarks:

1) For some reason, when clicking on the button to see the commands list, the front-end does 4 calls (2 times the same 2 calls) and I think it should only perform the 2 first. For long lists of commands it could be a little time consuming

2) I tested on my local env and even if the back-end sends the info 
`[{"id":"74403a00-e2a8-454a-aec2-e3770ded7866",
"action":"create_area","args": "area_name":"area_1"},"version":1,"user_name":"admin","updated_at":"2024-12-11T16:15:34.281725"}]`, I see Unknown and unknown for user_name and the date :thinking:  

3) What is the 3rd info displayed in the front ? It's written unknown but i don't really know what's supposed to be displayed

hdinia:
> 3 remarks:
> 
>     1. For some reason, when clicking on the button to see the commands list, the front-end does 4 calls (2 times the same 2 calls) and I think it should only perform the 2 first. For long lists of commands it could be a little time consuming
> 
>     2. I tested on my local env and even if the back-end sends the info
>        `[{"id":"74403a00-e2a8-454a-aec2-e3770ded7866", "action":"create_area","args": "area_name":"area_1"},"version":1,"user_name":"admin","updated_at":"2024-12-11T16:15:34.281725"}]`, I see Unknown and unknown for user_name and the date 🤔
> 
>     3. What is the 3rd info displayed in the front ? It's written unknown but i don't really know what's supposed to be displayed

Thanks for the feedback. Let me clarify each point:

- Those duplicate API calls should not be due to my recent changes, it's legacy code that's pretty messy. We should definitely rework this view at some point.

- About the "Unknown" values, it's intentional that we show "Unknown" as a default state when fields are empty. This maintains UI consistency and avoids having empty/blank spaces in the interface. While you're seeing this now, this case should actually become pretty rare over time, new studies will always have a user and timestamps associated with them. You think we should  display nothing instead ?

- The third field is just showing the last update time in a human-readable format. Like "1 hour ago"

MartinBelthle:
Reply to Hatim last comment:

1- Indeed it's legacy i just wanted to point it out as you're modifying the view. But if it's too complex to refactor we'll keep it this way

2- I agree that we're supposed to see "Unknown" if the back returns empty values but for new commands you'll have the info and they should be displayed

3- The third field isn't a duplicate of the 2nd ?

webapp/src/components/App/Singlestudy/Commands/Edition/DraggableCommands/CommandListItem/CommandDetails.tsx
```diff
@@ -0,0 +1,103 @@
...
+    },
+    {
+      icon: <ScheduleOutlined sx={{ fontSize: 20 }} />,
+      text: argsData.created ? formatDate(argsData.created) : "Unknown",
+      label: t("global.date"),
+    },
+    {
+      icon: <UpdateOutlined sx={{ fontSize: 20 }} />,
+      text: argsData.lastsave
+        ? formatDateWithLocale(argsData.lastsave, i18n.language)
```
skamril: Use `getCurrentLanguage` from i18nUtils. Then you can remove the second parameter and call it directly in `formatDateWithLocale`.

From the doc: 

> If you need the primary used language depending on your configuration (supportedLngs, load) you will prefer using i18next.languages[0].

webapp/src/components/App/Singlestudy/Commands/Edition/DraggableCommands/CommandListItem/CommandDetails.tsx
```diff
@@ -0,0 +1,103 @@
...
+  });
+};
+
+function CommandDetails({ item }: Props) {
+  const [t, i18n] = useTranslation();
+
+  const args = item.args as CommandArgsDTO;
+  const argsData = args?.data?.antares || {};
+
+  const details = [
```
skamril, line 46: Use `t("global.unknown")` for all labels with `"Unknown"`

skamril (changes requested):

webapp/src/components/App/Singlestudy/Commands/Edition/DraggableCommands/CommandListItem/CommandDetails.tsx
hdinia: thanks, done

webapp/src/components/App/Singlestudy/Commands/Edition/DraggableCommands/CommandListItem/CommandDetails.tsx
hdinia: done

MartinBelthle:
I tested on my local env and the bullet points 2 and 3 are resolved IMO

skamril (approved):

PR #2276, opened 2024-12-19
feat: add dao layer

TheoPascoli:
This PR is a second iteration of a previous one #2248 , which I'm about to close because it is too complex to modify.

I am implementing a DAO (Data Access Object) layer between the business layer (manager) and the storage layer. Additionally, I have implemented a caching logic to accelerate data retrieval, reducing the overhead of fetching frequently accessed data directly from the storage layer.

The goal is to create a clean separation of concerns, allowing the business logic to interact with the data through well-defined interfaces. This approach will make it easier to:

- Support additional storage systems (e.g., PostgreSQL, Redis, or others) by simply extending the DAO implementations
- Centralize and encapsulate data retrieval and manipulation logic, reducing duplication and improving maintainability
- Using caching to enhance performance for read-heavy operations

One of the next steps is to reverse the current behavior of the commands. Instead of having the StorageDAO call the commands, I want the commands to directly invoke the DAOs. This inversion will improve modularity and make the system more flexible to extend or modify in the future.

If you have suggestions or see potential improvements, feel free to share them


antarest/study/business/link_management.py
```diff
@@ -11,123 +11,35 @@
...
-from antarest.study.storage.variantstudy.model.command.create_link import CreateLink
-from antarest.study.storage.variantstudy.model.command.remove_link import RemoveLink
-from antarest.study.storage.variantstudy.model.command.update_link import UpdateLink
 
 
 class LinkManager:
-    def __init__(self, storage_service: StudyStorageService) -> None:
-        self.storage_service = storage_service
+    def __init__(self, dao_factory: DAOFactory) -> None:
+        self.composite_dao = dao_factory.create_link_dao()
```
sylvlecl, line 23: It's important that we keep an abstract class DAO here, because the manager layer should not know what implementation of DAO it's using.

Typically, for unit tests we could use an in-memory DAO implementation which does not know anything about caching or the study file format or a databse

antarest/study/dao/link/composite_link_dao.py
```diff
@@ -0,0 +1,57 @@
...
+
+from antarest.core.interfaces.cache import ICache
+from antarest.study.business.model.link_model import LinkBaseDTO, LinkDTO
+from antarest.study.dao.link.cache_link_dao import CacheLinkDao
+from antarest.study.dao.link.storage_link_dao import StorageLinkDao
+from antarest.study.model import Study
+from antarest.study.storage.storage_service import StudyStorageService
+
+
+class CompositeLinkDAO:
```
sylvlecl, line 23: See other comment:
I think the composite DAO et the storage DAO should both implement the same interface (abstract class), so they can be used interchangeably.

The cache "DAO" is better separated because actually it's not really a DAO (it has different methods).

antarest/study/dao/link/storage_link_dao.py
```diff
@@ -0,0 +1,127 @@
...
+from antarest.study.storage.variantstudy.model.command.create_link import CreateLink
+from antarest.study.storage.variantstudy.model.command.remove_link import RemoveLink
+from antarest.study.storage.variantstudy.model.command.update_link import UpdateLink
+
+
+class StorageLinkDao:
+    def __init__(self, storage_service: StudyStorageService):
+        self.storage_service = storage_service
+
+    def create_link(self, study: Study, link_dto: LinkDTO) -> LinkDTO:
```
sylvlecl, line 31: After more thought about the design of our "layers":
I think we should try to have business domain classes with plain names like `Link`, `Area`, etc.
The DAO will be in charge of implementing "CRUD" operations for those objects.
We should keep "DTO" naming for pure web layer objects.
However, it's something we can change later on, I think.

sylvlecl (commented):
Apologies for not having given a look earlier ...!
A few comments that we can discuss

PR #2278, opened 2024-12-20
fix(raw): replace NaN values by 'NaN' inside output matrices

MartinBelthle:


MartinBelthle:
Some output matrices have NaN values. Currently we try to replace them by None but it does nothing (output_series_matrix file) and we serialize NaN values as NaN (with pydanitic type adapter). But JavaScript cannot seems to parse the JSON and so the front-end has to do some post-processing to display those matrices. 

This PR replaces the NaN by "NaN" for the JSON to be JavaScript parsable. It works well but we're not too sure we should do it so I'll mark this PR as draft

sylvlecl:
As discussed on Friday: I think it's better to close this and keep the current behaviour, if that's OK with @hdinia 
The correct handling of NaNs by the application needs some special handling on the javascript side anyway to display it the way we want (empty cell), so it looks OK to benefit from the feature from pydantic to mark those values.

Replacing them with strings will have the drawback that it could hurt performances on the python side, because numerical numpy arrays will be converted to object arrays, which are more heavy weight.

sylvlecl:
Closing this, does not seem to be an issue.

PR #2279, opened 2024-12-20
fix(list_dir): check permissions and consider workspace filters config

MartinBelthle:
Will fix back-end issues inside [ANT-615]


antarest/study/storage/explorer_service.py
```diff
@@ -40,8 +40,17 @@ def list_dir(
...
+        try:
+            children = list(directory_path.iterdir())
+        except PermissionError:
+            children = []  # we don't want to try to read folders we can't access
+        for child in children:
+            if (
+                child.is_dir()
+                and not is_study_folder(child)
+                and not should_ignore_folder_for_scan(child)
+                and not child.name.startswith((".", "$"))
```
smailio: the name check for hidden file doesn't work on windows, does it?

MartinBelthle: I removed this code to use the workspace filters to mutualize the code

antarest/study/storage/utils.py
```diff
@@ -476,4 +477,8 @@ def should_ignore_folder_for_scan(path: Path) -> bool:
         logger.info(f"TS generation temporary folder found. Will skip further scan of folder {path}")
         return True
 
-    return False
+    return not (
+        path.is_dir()
+        and any([re.search(regex, path.name) for regex in filter_in])
```
smailio: do you need to use a list here instead of just the generator ? like this  `any(re.search(regex, path.name) for regex in filter_in)`

PR #2281, opened 2025-01-06
fix(link): add comment attribute in link that was missing and causing an 422 error

TheoPascoli:
"comment" fields can exist in Links, therefore we need it in the model so pydantic does not raise an error


MartinBelthle (changes requested):
Could you add a test where we have this value inside the .ini file and the GET /links called inside the map succeeds ?

MartinBelthle (approved):
Okay for me. just add 'link' in the PR name to make it more explicit

PR #2282, opened 2025-01-06
feat(commands): re-use the cache more often

MartinBelthle:
Wait for the PR for lower case to be merged as we have too many issues otherwise


antarest/study/storage/variantstudy/model/command/icommand.py
```diff
@@ -184,6 +184,14 @@ def get_inner_matrices(self) -> t.List[str]:
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def can_update_study_config(self) -> bool:
```
sylvlecl, line 188: maybe `supports_study_config_update` ?

antarest/study/business/utils.py
```diff
@@ -40,12 +42,23 @@ def execute_or_add_commands(
 ) -> None:
     if isinstance(study, RawStudy):
         executed_commands: t.MutableSequence[ICommand] = []
+        should_invalidate_cache = False
         for command in commands:
+            if not command.can_update_study_config():
```
sylvlecl, line 47: Reading the code, I am thinking that maybe it would be even better to have the "should_invalidate" in the command output instead.
It seems to make more sense, because it's the command application which implements and knows if the config has been updated or not, so it would make sense to return this information with the command application result.

sylvlecl (commented):
Nice, only minor comments but looks good already

antarest/study/business/utils.py
sylvlecl: Note that we'll need to adapt study interfaces now ...

sylvlecl:
Another comment:
as we have seen recently, the snapshot generator today assumes that all commands update the config correctly, which is wrong.

We need to fix the snapshot generator so that it does not update the config in cache when it should not.

MartinBelthle:
Duplicata of the more recent PR https://github.com/AntaresSimulatorTeam/AntaREST/pull/2609

PR #2283, opened 2025-01-06
chore(lower-case): read and write cluster and bc groups in lower case

MartinBelthle:


antarest/study/web/study_data_blueprint.py
```diff
@@ -1955,7 +1955,7 @@ def create_renewable_cluster(
     def update_renewable_cluster(
         uuid: str,
         area_id: str,
-        cluster_id: str,
+        cluster_id: LowerCaseStr,
```
sylvlecl: I don't think we should be case insensitive here: we want the user to use the ID, not the name (same in other endpoints).

For areas for example, I don't think we are case-insensitive ?

antarest/study/storage/rawstudy/model/filesystem/config/field_validators.py
```diff
@@ -87,3 +87,28 @@ def validate_color_rgb(v: t.Any) -> str:
...
+
+    Valid characters are `[a-zA-Z0-9_(),& -]` (including space).
+
+    Args:
+        name: The name to convert.
+    """
+    return _sub_invalid_chars(" ", name).strip().lower()
+
+
+def validate_id_against_name(name: str) -> str:
```
sylvlecl: Doesn't seem used

antarest/study/storage/rawstudy/model/filesystem/config/identifier.py
```diff
@@ -45,9 +43,9 @@ def generate_id(cls, name: str) -> str:
...
         """
         # Avoid circular imports
-        from antarest.study.storage.rawstudy.model.filesystem.config.model import transform_name_to_id
+        from antarest.study.storage.rawstudy.model.filesystem.config.field_validators import transform_name_to_id
 
-        return transform_name_to_id(name, lower=False)
+        return transform_name_to_id(name)
 
     @model_validator(mode="before")
     def validate_id(cls, values: t.MutableMapping[str, t.Any]) -> t.Mapping[str, t.Any]:
```
sylvlecl, line 36: looks like generate_id is never used outside of this class, it should be removed

antarest/study/storage/rawstudy/model/filesystem/ini_file_node.py
```diff
@@ -325,3 +325,51 @@ def _validate_param(
                 if raising:
                     raise ValueError(msg)
                 errors.append(msg)
+
+    def get_lowered_content(
```
sylvlecl: we'd better start those protected methods with _

antarest/study/storage/rawstudy/model/filesystem/ini_file_node.py
```diff
@@ -325,3 +325,51 @@ def _validate_param(
...
+        self, url: t.Optional[t.List[str]] = None, depth: int = -1, expanded: bool = False
+    ) -> SUB_JSON:
+        output = self._get(url, depth, expanded, get_node=False)
+        assert not isinstance(output, INode)
+        if depth <= -1 and expanded:
+            return output
+
+        # We need to lower the group attribute and the cluster ids
```
sylvlecl: Here we inject some knowledge (group should be lowercase) about specific subclasses of this class, it's not a very good separation of responsibilities.

A better design would be to specify as arguments of the method how stuff should be read or written.
For example, we could have a dict:
```python
parsing_methods = {
    (None, "group"): lambda value: str(value).lower()
}
```
Meaning that all values associated to any key at the 1st level (None) and "group" key at the 2nd level should use the provided parsing method.

Actually this should probably be something we provide to he ini reader.

This would allow, in the future, to also fix how we parse integers, because we will be able to specify that we want to parse them as strings, instead of letting the reader try to guess the type.

antarest/study/storage/rawstudy/model/filesystem/ini_file_node.py
```diff
@@ -325,3 +325,51 @@ def _validate_param(
...
+                    output[new_key]["group"] = str(output[new_key]["group"]).lower()
+        elif len(url) == 1:
+            assert isinstance(output, dict)
+            if "group" in output:
+                output["group"] = str(output["group"]).lower()
+        elif len(url) == 2 and url[1] == "group":
+            output = str(output).lower()
+        return output
+
+    def save_lowered_content(self, data: SUB_JSON, url: t.List[str]) -> None:
```
sylvlecl: same comments, we could provide the serializers as arguments

antarest/study/storage/rawstudy/model/filesystem/root/input/renewables/clusters.py
```diff
@@ -30,12 +33,22 @@ def __init__(
             "group": str,
             "enabled": bool,
             "unitcount": int,
-            "nomialcapacity": 0,
+            "nomialcapacity": float,
```
sylvlecl: looks like there is a typo in nominal capacity :/

sylvlecl (commented):
A few comments: for the low level parsing, I think we should have a more configurable method, in particular so that we can re-use it later for the "int to string" issue, but also to have a better separation of concerns: the parsing function should not know by itself how to parse particular fields, it shuld be an argument of the function.

antarest/study/storage/rawstudy/model/filesystem/config/identifier.py
MartinBelthle: The doc says it was used to avoid circular imports but now that i put the method elsewhere there's no need to keep it

antarest/study/storage/rawstudy/model/filesystem/ini_file_node.py
MartinBelthle: I agree with you for the get.
But IMO the only save we'll ever need is this one and it's already generic.
So I think i should just rename the method and change the comments i wrote inside but keep the code as it is.
I'll also have to call the get with the right parsing method.
I'll commit this change and you can give me your opinion on it

sylvlecl:
Closing, will be superseded by #2333

PR #2284, opened 2025-01-07
chore(commands): fix user name display in commands

maugde:
fix [ANT-2543](https://gopro-tickets.rte-france.com/browse/ANT-2543)

### DIFF
--- a/antarest/login/utils.py
+++ b/antarest/login/utils.py
@@ -10,6 +10,7 @@
 #
 # This file is part of the Antares project.
 
+import contextlib
 import typing as t
 from contextvars import ContextVar
 from typing import Optional
@@ -17,31 +18,50 @@ from typing import Optional
 from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
 from starlette.requests import Request
 from starlette.responses import Response
+from starlette.types import ASGIApp
 from typing_extensions import override
 
 from antarest.core.jwt import JWTUser
-from antarest.core.serialization import from_json
 from antarest.fastapi_jwt_auth import AuthJWT
+from antarest.fastapi_jwt_auth.exceptions import AuthJWTException
+from antarest.login.auth import Auth
 
-_current_user: ContextVar[t.Optional[AuthJWT]] = ContextVar("_current_user", default=None)
+_current_user: ContextVar[t.Optional[JWTUser]] = ContextVar("_current_user", default=None)
 
 
 class CurrentUserMiddleware(BaseHTTPMiddleware):
+    def __init__(self, app: ASGIApp, auth: Auth) -> None:
+        super().__init__(app)
+        self.auth = auth
+
     @override
     async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -> Response:
         global _current_user
-        auth_jwt = AuthJWT(Request(request.scope))
-        _current_user.set(auth_jwt)
+        auth_jwt = AuthJWT(request)
+
+        try:
+            jwt_user = self.auth.get_current_user(auth_jwt)  # fail when no cookies provided
+        except AuthJWTException:
+            # fall into this block in testing context
+            jwt_user = None
+
+        _current_user.set(jwt_user)
 
         response = await call_next(request)
         return response
 
 
 def get_current_user() -> Optional[JWTUser]:
-    auth_jwt = _current_user.get()
-    if auth_jwt:
-        json_data = from_json(auth_jwt.get_jwt_subject())
-        current_user = JWTUser.model_validate(json_data)
-    else:
-        current_user = None
+    current_user = _current_user.get()
     return current_user
+
+@contextlib.contextmanager
+def current_user_context(token: JWTUser) -> JWTUser:
+    global _current_user
+    jwt_user = token
+    _current_user.set(jwt_user)
+
+    try:
+        yield _current_user.get()
+    finally:
+        _current_user.set(None)
--- a/antarest/login/utils.py
+++ b/antarest/login/utils.py
@@ -55,8 +55,9 @@ def get_current_user() -> Optional[JWTUser]:
     current_user = _current_user.get()
     return current_user
 
+
 @contextlib.contextmanager
-def current_user_context(token: JWTUser) -> JWTUser:
+def current_user_context(token: JWTUser) -> t.Generator[JWTUser|None]:
     global _current_user
     jwt_user = token
     _current_user.set(jwt_user)
--- a/antarest/login/utils.py
+++ b/antarest/login/utils.py
@@ -57,7 +57,7 @@ def get_current_user() -> Optional[JWTUser]:
 
 
 @contextlib.contextmanager
-def current_user_context(token: JWTUser) -> t.Generator[JWTUser|None]:
+def current_user_context(token: JWTUser) -> t.Iterator[JWTUser|None]:
     global _current_user
     jwt_user = token
     _current_user.set(jwt_user)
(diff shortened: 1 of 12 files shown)


antarest/study/service.py
```diff
@@ -229,11 +229,13 @@ def __init__(
         repository: StudyMetadataRepository,
         storage_service: StudyStorageService,
         event_bus: IEventBus,
+        jwt_user: JWTUser,
```
sylvlecl: Actually we should have a more generic mechanism in the `TaskJobService` so that every time a task is submitted, we fill the `_current_user` variable for the thread which will execute it.
This way, we will be able to always use the information from this context var, instead of having 2 different mechanisms.
Also, this will ensure that the user is available for all types of tasks, not only for TS generations tasks.

In practice:
- [here](https://github.com/AntaresSimulatorTeam/AntaREST/blob/345b0a7df31e056f6e1f27dbca1cb03c12eb6845/antarest/core/tasks/service.py#L306) we can get the current user and give to the `_run_task` method as an argument
- in the `_run_task` method, the first thing we should do is to set the value of `_current_user`. We need to pay attention to remove this value at the end of the method, so the best option is probably to create a small "context manager" for this

antarest/study/business/utils.py
```diff
@@ -37,7 +38,11 @@ def execute_or_add_commands(
     commands: t.Sequence[ICommand],
     storage_service: StudyStorageService,
     listener: t.Optional[ICommandListener] = None,
+    jwt_user: t.Optional[JWTUser] = None,
 ) -> None:
+    # get current user if not in session, otherwise get session user
+    current_user = get_current_user() or jwt_user
```
sylvlecl: Here, if the task service correctly fills the current user in the task execution thread, we should not be able to rely only on `get_current_user()`

sylvlecl (changes requested):
I think we should have a slightly more generic approach so that all tasks fill the `_current_user` context var, see comments.
Later, this will ensure that all tasks have the current user available, not only TS generation tasks. This will also simplify the implementation of `execute_or_add_commands`

tests/storage/business/test_arealink_manager.py
```diff
@@ -122,205 +122,210 @@ def test_area_crud(empty_study: FileStudy, matrix_service: SimpleMatrixService):
...
-
-    link_manager.create_link(
-        study,
-        LinkDTO(
-            area1="test",
-            area2="test2",
-        ),
-    )
-    assert empty_study.config.areas["test"].links.get("test2") is not None
+    with patch("antarest.study.business.utils.get_current_user", return_value=DEFAULT_ADMIN_USER):
```
sylvlecl: Note that if we create a context manager to set the current user (in task service), we will be more difficulte to maintain

antarest/login/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    return current_user
+
+
+@contextlib.contextmanager
+def current_user_context(token: JWTUser) -> t.Iterator[JWTUser | None]:
+    global _current_user
+    jwt_user = token
+    _current_user.set(jwt_user)
+
+    try:
```
sylvlecl: I don't think the try/finally is useful here ?

antarest/login/utils.py
```diff
@@ -0,0 +1,68 @@
...
+        global _current_user
+        auth_jwt = AuthJWT(request)
+
+        try:
+            jwt_user = self.auth.get_current_user(auth_jwt)  # fail when no cookies provided
+        except AuthJWTException:
+            # fall into this block in testing context
+            jwt_user = None
+
+        _current_user.set(jwt_user)
```
sylvlecl, line 48: I think it's better here to use the context manager you have created below, in particular to ensure that we reset the user to `None` at the end of the request:
```python
    return await call_next(request)
```

antarest/login/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+def get_current_user() -> Optional[JWTUser]:
+    current_user = _current_user.get()
+    return current_user
+
+
+@contextlib.contextmanager
+def current_user_context(token: JWTUser) -> t.Iterator[JWTUser | None]:
+    global _current_user
+    jwt_user = token
```
sylvlecl: we can remove that intermediate variable

antarest/login/utils.py
```diff
@@ -0,0 +1,68 @@
...
+        super().__init__(app)
+        self.auth = auth
+
+    @override
+    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -> Response:
+        global _current_user
+        auth_jwt = AuthJWT(request)
+
+        try:
+            jwt_user = self.auth.get_current_user(auth_jwt)  # fail when no cookies provided
```
sylvlecl: I did not think about it too much before:
here we introduce a small technical debt because we perform 2x the authentication (I mean transforming the token into a user object), for the same request (once here, and another time when we add the jwt user in endpoints dependencies).

It blurs a little the responsibilities of who is actually doing the authentication, I think we should add a `TODO` comment here to mention this issue that we should try to resorb in the future.
Ideally the authentication should be done in one place, and the "result" (the user) re-used in the other places. I am not sure yet of the best way to do this, so we could keep it this way for this PR.

antarest/login/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+class CurrentUserMiddleware(BaseHTTPMiddleware):
+    def __init__(self, app: ASGIApp, auth: Auth) -> None:
+        super().__init__(app)
+        self.auth = auth
+
+    @override
+    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -> Response:
+        global _current_user
+        auth_jwt = AuthJWT(request)
```
sylvlecl: I think this should go into the try/except, because we can have an exception if the input token is malformed.

antarest/login/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+        try:
+            jwt_user = self.auth.get_current_user(auth_jwt)  # fail when no cookies provided
+        except AuthJWTException:
```
sylvlecl: Here I think we should catch all exceptions, in case any error occurs during the parsing of the user.

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -212,12 +212,13 @@ def test_generate_task(
             study_version=study_version,
         )
 
-        execute_or_add_commands(
-            variant_study,
-            file_study,
-            commands=[create_area_fr, create_st_storage],
-            storage_service=study_storage_service,
-        )
+        with patch("antarest.study.business.utils.get_current_user", return_value=DEFAULT_ADMIN_USER):
```
sylvlecl: I think we can replace it with the new `current_user_context` method, it will be more reliable.

tests/core/test_tasks.py
```diff
@@ -476,6 +476,12 @@ def test_ts_generation_task(
     raw_study_path = tmp_path / "study"
 
     regular_user = User(id=99, name="regular")
+    jwt_user = Mock(
```
sylvlecl, line 477: It does not seem to be used

sylvlecl, line 477: But! it would be nice to add unit test to check that the transfer of the user to the task works correctly:
- submit a task to the task service that will just get the current user
- check that it indeed got the right user (the one that submitted the task)

sylvlecl (changes requested):
The new implementation looks good!
Still some last comments, in particular I think we should add a TODO to clarify more the place where the parsing of the token should be done (middleware or fastapi "dependency").

sylvlecl (approved):
Thanks for the additional unit test!

TheoPascoli (approved):
Hello Mauranne, great work ! :+1:

PR #2286, opened 2025-01-08
fix(matrix): return default empty matrix even when called with formatted=False

MartinBelthle:
Will fix [ANT-2611]


TheoPascoli (dismissed):
That looks good

PR #2289, opened 2025-01-10
feat(ui-tablemode,ui-playlist): replace Handsontable by Glide Data Grid

skamril:
ANT-1217


webapp/src/components/App/Singlestudy/explore/TableModeList/index.tsx
```diff
@@ -84,7 +85,25 @@ function TableModeList() {
...
+  const handleDeleteClick = () => {
+    if (selectedTemplate) {
+      setDialog({
+        type: "delete",
+        templateId: selectedTemplate.id,
+      });
+    }
+  };
+
+  const handleDelete = () => {
```
hdinia, line 106: you can improve the naming here, (subject + action), to be more clear about their purpose, for example:
```ts
const handleTemplateEdit = () => { ... }
const handleTemplateDelete = () => { ... }
const handleTemplateDeleteConfirm = () => { ... }
```

webapp/index.html
```diff
@@ -17,6 +17,8 @@
   </head>
   <body>
     <div id="root"></div>
+    <!-- https://github.com/glideapps/glide-data-grid/issues/795#issuecomment-1787348408 -->
+    <div id="portal" style="position: fixed; left: 0; top: 0; z-index: 9999;"></div>
```
hdinia, line 21: It resolves the portal issue thanks 👍🏼

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
```diff
@@ -44,51 +45,68 @@ interface Props {
 function ScenarioPlaylistDialog(props: Props) {
   const { study, open, onClose } = props;
   const { t } = useTranslation();
-  const tableRef = useRef({} as HotTableClass);
+  const dataGridApiRef = useRef<DataGridFormApi<PlaylistData>>(null!);
```
hdinia: avoid the non-null assertion operator (`null!`) as it bypass TS checks, directly check for null values in the concerned part where you use this (even if you have to repeat, it's safer)

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
```diff
@@ -44,51 +45,68 @@ interface Props {
...
-    if (!api) {
-      return;
-    }
-
-    const changes: Array<[number, string, boolean]> = api
-      .getDataAtProp("status")
-      .map((status, index) => [index, "status", fn(status)]);
-
-    api.setDataAtRowProp(changes);
+    const { data, setData } = dataGridApiRef.current;
```
hdinia: you can add a check for the ref before that line 
```ts
   if (!dataGridApiRef.current) {
      return;
    }
```

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
```diff
@@ -44,51 +45,68 @@ interface Props {
...
+  const [isDirty, setIsDirty] = useState(false);
+  const closeAction = useConfirm();
   const res = usePromise(() => getPlaylist(study.id), [study.id]);
 
+  const columns = useMemo(() => {
+    return [
+      {
+        id: "status",
+        title: t("global.status"),
+        kind: GridCellKind.Boolean,
```
hdinia: property `kind` does not exists on Glide's `GridColumn` type, there must be a typing error on `DataGridFrom` columns prop, in addition you can add any random property it doesn't even raise an error
(also you re-use `GridCellKind` in the context of columns it's confusing)

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
```diff
@@ -44,51 +45,68 @@ interface Props {
...
 
   const handleResetWeights = () => {
-    const api = tableRef.current.hotInstance as Handsontable;
-
-    api.setDataAtRowProp(
-      api.rowIndexMapper
-        .getIndexesSequence()
-        .map((rowIndex) => [rowIndex, "weight", DEFAULT_WEIGHT]),
-    );
+    const { data, setData } = dataGridApiRef.current;
```
hdinia: same for the ref check

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
```diff
@@ -44,51 +45,68 @@ interface Props {
...
+
+    onClose();
+  };
+
+  const handleFormStateChange: DataGridFormProps["onStateChange"] = (
+    formState,
+  ) => {
+    setIsSubmitting(formState.isSubmitting);
+    setIsDirty(formState.isDirty);
   };
```
hdinia: you can import and use `DataGridFormState`, that seems more accurate type here:

```ts
  const handleFormStateChange = (formState: DataGridFormState) => {
    setIsSubmitting(formState.isSubmitting);
    setIsDirty(formState.isDirty);
  };
```

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
```diff
@@ -99,18 +117,25 @@ function ScenarioPlaylistDialog(props: Props) {
     <BasicDialog
       title={t("study.configuration.general.mcScenarioPlaylist")}
       open={open}
-      onClose={onClose}
-      actions={<Button onClick={onClose}>{t("global.close")}</Button>}
-      // TODO: add `maxHeight` and `fullHeight` in BasicDialog`
```
hdinia, line 104: did you address this todo ?

webapp/src/components/common/DataGrid.tsx
```diff
@@ -0,0 +1,325 @@
...
+}
+
+export interface DataGridProps
+  extends Omit<
+    DataEditorProps,
+    "rowMarkers" | "onGridSelectionChange" | "gridSelection"
+  > {
+  rowMarkers?: RowMarkerOptionsPlus["kind"] | RowMarkerOptionsPlus;
+  enableColumnResize?: boolean;
+}
```
hdinia: while I understand the purpose of this, I do not recommend overwriting base library type like this (private type)

is it the only way to achieve this functionality ?

you can try this, using composition instead of extension:

```ts
interface CustomRowMarkers {
  kind: "string" | "clickable-string";
  getTitle: (rowIndex: number) => string;
}

interface DataGridProps
  extends Omit<
    DataEditorProps,
    "rowMarkers" | "onGridSelectionChange" | "gridSelection"
  > {
  rowMarkers?: DataEditorProps["rowMarkers"] | CustomRowMarkers;
  enableColumnResize?: boolean;
}
```

then define a type guard 

```ts

function isCustomRowMarkers(
  markers: DataEditorProps["rowMarkers"] | CustomRowMarkers,
): markers is CustomRowMarkers {
  if (!markers || typeof markers === "string") {
    return false;
  }

  return markers.kind === "string" || markers.kind === "clickable-string";
}

```

and use it to replace this
```ts
const isRowMarkersString =
    rowMarkersOpts.kind === "string" ||
    rowMarkersOpts.kind === "clickable-string";
```
to

```ts
  const customRowMarkers = isCustomRowMarkers(rowMarkersOpts);

```
an update the rest like:
```ts
const title = customRowMarkers
            ? rowMarkersOpts.getTitle(rowIndex)
            : `Row ${rowIndex + 1}`;
```

webapp/src/components/common/DataGrid.tsx
```diff
@@ -0,0 +1,325 @@
...
+    onColumnResizeStart,
+    onColumnResizeEnd,
+    enableColumnResize = true,
+    freezeColumns,
+    ...rest
+  } = props;
+
+  const rowMarkersOpts =
+    typeof rowMarkers === "string" ? { kind: rowMarkers } : rowMarkers;
+  const isRowMarkersString =
```
hdinia: consider adding a space between multi-lines variables blocks, for readability

webapp/src/components/common/DataGrid.tsx
```diff
@@ -0,0 +1,325 @@
...
+
+  return (
+    <DataEditor
+      groupHeaderHeight={30}
+      headerHeight={30}
+      rowHeight={30}
+      smoothScrollX
+      smoothScrollY
+      overscrollX={2}
+      overscrollY={2}
```
hdinia, line 318: are you sure about adding an overscroll ?

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+> {
+  defaultData: TFieldValues;
+  columns: ReadonlyArray<GridColumn & { id: keyof TFieldValues[number] }>;
+  rowMarkers?: DataGridProps["rowMarkers"];
+  allowedFillDirections?: FillHandleDirection;
+  enableColumnResize?: boolean;
+  onSubmit?: (
+    data: SubmitHandlerPlus<TFieldValues>,
+    event?: React.BaseSyntheticEvent,
+  ) => void | Promise<SubmitReturnValue>;
```
hdinia, line 73: - how onSubmit can be synchronous ?
- `onSubmit` prop is currently optional -> a form is meant to always be submittable no ?

think it should be
```ts
onSubmit: (
    data: SubmitHandlerPlus<TData>,
    event?: React.BaseSyntheticEvent,
  ) => Promise<TData> | Promise<void>;
  ```

File: webapp/src/hooks/useCloseFormSecurity.ts [hunk 13]
```diff
@@ -0,0 +1,33 @@
...
+
+import usePrompt from "./usePrompt";
+import { useTranslation } from "react-i18next";
+
+export interface UseFormCloseProtectionParams {
+  isSubmitting: boolean;
+  isDirty: boolean;
+}
+
+function useFormCloseProtection({
```
hdinia, line 23: "Close" is ambiguous, it could refer to closing a modal, closing a browser tab, as this hook protect against unsaved changes when navigating away I suggest to rename it to reflect this purpose
 `useFormCloseProtection` -> `useNavigationGuard`

optional: you could also make it more generic to handle different scenarios beyond just forms (you are using it in the Matrix component for example, which is not strictly a Form), and be more flexible (handle multiple blocking scenarios with different messages and priorities).

quick example:
```ts

export interface NavigationGuardParams {
  shouldBlock: boolean;
  message?: string;
  priority?: number;
}

 function useNavigationGuard(...guards: NavigationGuardParams[]) {

  // sort guards by priority if specified
  const activeGuard = guards
    .filter(guard => guard.shouldBlock)
    .sort((a, b) => (b.priority ?? 0) - (a.priority ?? 0))
    [0];

  if (activeGuard) {
    usePrompt(
      activeGuard.message ?? t("navigation.defaultWarning"),
      activeGuard.shouldBlock
    );
  }
}

// usage

 useNavigationGuard(
    { // form
      shouldBlock: isSubmitting,
      message: t("form.submit.inProgress"),
      priority: 2
    },
    { //file upload
    shouldBlock: isUploading,
    message: t("upload.inProgress"),
    priority: 1
  }
  );
  ```

File: webapp/src/components/common/DataGrid.tsx [hunk 14]
```diff
@@ -0,0 +1,325 @@
...
+        if (newSelection.columns.hasIndex(0)) {
+          // TODO find have the rows length to select all the rows like other row markers
+          // setSelection({
+          //   ...newSelection,
+          //   columns: CompactSelection.empty(),
+          //   rows: CompactSelection.fromSingleSelection([
+          //     0,
+          //     // rowsLength
+          //   ]),
+          // });
```
hdinia: did you forgot to implement this ?

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
hdinia: for `dataGridApiRef` & `DataGridFormApi` any reason to add "Api" in the name here ? this represents the form's state and controls, maybe you can name it like `DataGridFormHandle` or `DataGridFormControl`.
`dataGridApiRef` -> `dataGridFormRef` or `formRef`

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
```diff
@@ -133,19 +158,28 @@ function ScenarioPlaylistDialog(props: Props) {
...
+            <DataGridForm
+              defaultData={defaultValues}
+              columns={columns}
+              rowMarkers={{
+                kind: "clickable-string",
+                getTitle: (index) => `MC Year ${index + 1}`,
               }}
+              onSubmit={handleSubmit}
+              onStateChange={handleFormStateChange}
+              apiRef={dataGridApiRef}
```
hdinia: `handleRef={dataGridFormRef}` or `formRef={dataGridFormRef}`

webapp/src/components/common/DataGrid.tsx
```diff
@@ -0,0 +1,325 @@
...
+  CompactSelection,
+  DataEditor,
+  EditListItem,
+  GridCellKind,
+  GridSelection,
+  type DataEditorProps,
+} from "@glideapps/glide-data-grid";
+import "@glideapps/glide-data-grid/dist/index.css";
+import { useCallback, useEffect, useState } from "react";
+import type { RowMarkerOptions } from "node_modules/@glideapps/glide-data-grid/dist/dts/data-editor/data-editor";
```
hdinia: do not import from node_modules, you can use `DataEditorProps["rowMarkers"]`

webapp/src/components/common/DataGrid.tsx
```diff
@@ -0,0 +1,325 @@
...
+
+  const ifNotRowNameColumn = (
+    colIndex: number,
+    fn: (colIndex: number) => void,
+  ) => {
+    return ifElseRowNameColumn(colIndex, voidFn, fn);
+  };
+
+  ////////////////////////////////////////////////////////////////
+  // Callbacks
```
hdinia: remove "Callbacks" comment block, place `getCellContentWrapper` under utils block, or refactor it into its own custom hook

webapp/src/components/common/DataGrid.tsx
```diff
@@ -0,0 +1,325 @@
...
+        (adjustedColIndex) => {
+          return getCellContent([adjustedColIndex, rowIndex]);
+        },
+      );
+    },
+    [getCellContent, isRowMarkersString],
+  );
+
+  ////////////////////////////////////////////////////////////////
+  // Edition
```
hdinia, line 160: remove all event handlers related comments block (Edition, Resize, Selection etc) and put the handlers under the regular "Event Handlers" block

webapp/src/components/common/DataGrid.tsx
```diff
@@ -0,0 +1,325 @@
...
+      smoothScrollY
+      overscrollX={2}
+      overscrollY={2}
+      width="100%"
+      theme={darkTheme}
+      {...rest}
+      columns={columns}
+      rowMarkers={
+        isRowMarkersString ? "none" : (rowMarkersOpts as RowMarkerOptions)
+      }
```
hdinia: with previous adjustments, you can update to: `rowMarkers={customRowMarkers ? "none" : rowMarkersOpts}`

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+  apiRef,
+}: DataGridFormProps<TFieldValues>) {
+  const { t } = useTranslation();
+  const enqueueErrorSnackbar = useEnqueueErrorSnackbar();
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const [savedData, setSavedData] = useState(defaultData);
+  const [{ present: data }, { set: setData, undo, redo, canUndo, canRedo }] =
+    useUndo(defaultData);
+
+  const isDirty = savedData !== data;
```
hdinia, line 107: direct comparison between savedData and data, will produce state issues, instead use a deep value comparison.

use : ` const isDirty = !R.equals(savedData, data);`

note: undo/redo, creates new object references even if the values are identical to the initial state.

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+          ),
+        };
+      }
+
+      return acc;
+    }, {} as DeepPartial<TFieldValues>);
+  };
+
+  ////////////////////////////////////////////////////////////////
+  // Callbacks
```
hdinia: for comment blocks same remark as `DataGrid` component, keep it like the rest of the app. if needed prefer dedicated files for custom logic, custom hooks etc

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+        setSavedData(data);
+        onSubmitSuccessful?.(dataArg, submitRes);
+      })
+      .catch((err) => {
+        enqueueErrorSnackbar(t("form.submit.error"), err);
+      })
+      .finally(() => {
+        setIsSubmitting(false);
+      });
+  };
```
hdinia: - rename `handleSubmit`
- remove `onSumbit` check as it should not be optional 
- use async/await, and error utils
- add a check for result before `onSubmitSuccessful` callback

```ts
  const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
    event.preventDefault();

    const payload = {
      values: data,
      dirtyValues: getDirtyValues(),
    };

    try {
      setIsSubmitting(true);
      const result = await onSubmit(payload, event);
      setSavedData(data);

      if (result) {
        onSubmitSuccessful?.(payload, result);
      }
    } catch (err) {
      enqueueErrorSnackbar(t("form.submit.error"), toError(err));
    } finally {
      setIsSubmitting(false);
    }
  };
  ```

File: webapp/src/components/common/DataGrid.tsx [hunk 23]
```diff
@@ -0,0 +1,325 @@
...
+      groupHeaderHeight={30}
+      headerHeight={30}
+      rowHeight={30}
+      smoothScrollX
+      smoothScrollY
+      overscrollX={2}
+      overscrollY={2}
+      width="100%"
+      theme={darkTheme}
+      {...rest}
```
hdinia, line 321: put `{...rest}` on top or bottom

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+        className="DataGridForm__Footer"
+        sx={{
+          display: "flex",
+          flexDirection: "column",
+          gap: 1.5,
+          mt: 1.5,
+        }}
+      >
+        <Divider flexItem />
+        <Box className="DataGridForm__Footer__Actions" sx={{ display: "flex" }}>
```
hdinia: same remark for custom css class

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+            getTitle: (index) => rowNames[index],
+          }
+        }
+        fillHandle
+        allowedFillDirections={allowedFillDirections}
+        enableColumnResize={enableColumnResize}
+        getCellsForSelection
+      />
+      <Box
+        className="DataGridForm__Footer"
```
hdinia: custom css class seems not necessary, consider removing them (not sure about it since I haven't tested)

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+          >
+            {t("global.save")}
+          </LoadingButton>
+          <Divider sx={{ mx: 2 }} orientation="vertical" flexItem />
+          <Tooltip title={t("global.undo")}>
+            <span>
+              <IconButton
+                size="small"
+                onClick={undo}
+                disabled={!canUndo || isSubmitting}
```
hdinia, line 323: use color="primary", the default one is not visible enough (same for the redo button)

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+import { LoadingButton } from "@mui/lab";
+import { mergeSxProp } from "@/utils/muiUtils";
+import * as R from "ramda";
+import { SubmitHandlerPlus } from "./Form/types";
+import useEnqueueErrorSnackbar from "@/hooks/useEnqueueErrorSnackbar";
+import useFormCloseProtection from "@/hooks/useCloseFormSecurity";
+import { useUpdateEffect } from "react-use";
+
+type GridFieldValuesByRow = Record<
+  IdType,
```
hdinia: you're using string | number (`IdType`), do you have a case where the id is a number?

webapp/src/components/common/TableMode.tsx
```diff
@@ -25,47 +25,64 @@ import {
...
   const { t } = useTranslation();
+  const [gridColumns, setGridColumns] = useState<
+    DataGridFormProps<TableData>["columns"]
+  >([]);
+  const columnsDep = columns.join(",");
 
   const res = usePromise(
     () => getTableMode({ studyId, tableType: type, columns }),
-    [studyId, type, columns.join(",")],
+    [studyId, type, columnsDep],
```
hdinia, line 57: `columnsDep` does unnecessary string creation on every render, why not memoizing the columns in the parent ?

webapp/src/components/common/TableMode.tsx
```diff
@@ -25,47 +25,64 @@ import {
...
-      }
+    if (rowNames.length === 0) {
+      setGridColumns([]);
+      return;
+    }
 
-      const data = res.data!;
-      const dataRowKeys = Object.keys(data[dataKeys[0]]);
+    const data = res.data!;
+    const columnNames = Object.keys(data[rowNames[0]]);
```
hdinia: avoid the non-null assertion (!). even if the access is safe here because you check rowNames.length, keep the code explicit: 
- it does not bypass TS
- keep the code clear for future maintainers that  won't have to trace through the logic to understand why the ! is safe
- more explicit about our assumptions
```ts
    const data = res.data || {};
    const firstRow = Object.keys(data)[0];

    if (!firstRow) {
      setGridColumns([]);
      return;
    }

    const columnNames = Object.keys(data[firstRow]);
```

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+import { useTranslation } from "react-i18next";
+import { LoadingButton } from "@mui/lab";
+import { mergeSxProp } from "@/utils/muiUtils";
+import * as R from "ramda";
+import { SubmitHandlerPlus } from "./Form/types";
+import useEnqueueErrorSnackbar from "@/hooks/useEnqueueErrorSnackbar";
+import useFormCloseProtection from "@/hooks/useCloseFormSecurity";
+import { useUpdateEffect } from "react-use";
+
+type GridFieldValuesByRow = Record<
```
hdinia: `GridFieldValuesByRow` -> `GridData`

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+  IdType,
+  Record<string, string | boolean | number>
+>;
+
+export interface DataGridFormState {
+  isDirty: boolean;
+  isSubmitting: boolean;
+}
+
+export interface DataGridFormApi<TFieldValues extends GridFieldValuesByRow> {
```
hdinia: when using generics please simplify the naming or it make it hard to understand, consider:
`TFieldValues` -> `TData` 
`GridFieldValuesByRow ` -> `GridData`

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+  data: TFieldValues;
+  setData: Actions<TFieldValues>["set"];
+  formState: DataGridFormState;
+}
+
+export interface DataGridFormProps<
+  TFieldValues extends GridFieldValuesByRow = GridFieldValuesByRow,
+  SubmitReturnValue = unknown,
```
hdinia, line 63: you can remove it as it's unused and overwrite the promise return type to be "unknown"

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+  allowedFillDirections?: FillHandleDirection;
+  enableColumnResize?: boolean;
+  onSubmit?: (
+    data: SubmitHandlerPlus<TFieldValues>,
+    event?: React.BaseSyntheticEvent,
+  ) => void | Promise<SubmitReturnValue>;
+  onSubmitSuccessful?: (
+    data: SubmitHandlerPlus<TFieldValues>,
+    submitResult: SubmitReturnValue,
+  ) => void;
```
hdinia:
```ts
 onSubmitSuccessful?: (
    data: SubmitHandlerPlus<TData>,
    submitResult?: TData,
  ) => void;
```

hdinia (changes requested):

webapp/src/components/App/Singlestudy/explore/TableModeList/index.tsx
skamril: I add `Click` because it's not the action launch here but only the handle to call the action.
This is why I have `handleDeleteClick` to call the action, and `handleDelete` that make the action.

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
skamril: I know for the `null!`, but when we are **SURE** that the `null` value is impossible it's not a problem.
Without that I have to add condition without any sense.

cf. https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/hooks/#useref

skamril: `Api` is valid here why you don't like it ?
Also it keeps consistency with `Form` component that have `apiRef` property.

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
skamril: I respond above

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
skamril: Yes it's an error, I removed `kind`

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
skamril: I respond above

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
skamril: It is similar but ok

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
skamril: It's not necessary (I wrote it)

webapp/src/components/common/DataGrid.tsx
skamril: What is issue? I create a new component with extra rowMarkers, I did not update the core library.
With your code you cannot set `rowMarkers="string"`, but without this point it's the same principle.

webapp/src/components/common/DataGrid.tsx
skamril: I extends Prettier with new size in another PR.

webapp/src/components/common/DataGrid.tsx
skamril: 0 it's not good in UI

webapp/src/components/common/DataGridForm.tsx
skamril: I remove the optional, but there is no reason to remove the possibility to make a sync submit.

webapp/src/hooks/useCloseFormSecurity.ts
skamril: I see `useFormCloseProtection` more clear than `useNavigationGuard` (it's a generic name).
It protects against closing a browser tab or the browser when we are submitting or there is change not saved.

I use it in the Matrix component because you used the same code as the form with the same message (`form.XXX`), but you can make a specific one for this component. But for me when you have `isSubmitting` in the component  is not a problem to use this hook, it is look like a form (submit, undo/redo).

webapp/src/components/common/DataGrid.tsx
skamril: No I did not find a way to have the rows length yet

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
skamril: I respond below

webapp/src/components/common/DataGrid.tsx
skamril: It's an `import type`, it will be removed on the build. `DataEditorProps["rowMarkers"]` includes string list.

webapp/src/components/common/DataGrid.tsx
skamril: I rename the block "Content" (Content, Edition, Resize, Selection)

webapp/src/components/common/DataGrid.tsx
skamril: "Event Handlers" is the base, but when you have a lot of handlers like this component it's more clear to split more specifically. And `getCellContent` is not an event handler.

webapp/src/components/common/DataGrid.tsx
skamril: I respond below

webapp/src/components/common/DataGridForm.tsx
skamril: No there is no issue, shallow equal works well. Undo/redo keeps the ref (it works like Redux).

webapp/src/components/common/DataGridForm.tsx
skamril: I rename it "Content", a custom hook for this simple code is overkill.

webapp/src/components/common/DataGridForm.tsx
skamril: - rename handleSubmit => OK
- remove onSumbit check as it should not be optional => OK
- use async/await, and error utils => async/wait here is only a preference but ok
- add a check for result before onSubmitSuccessful callback => no the result can be empty, `onSubmitSuccessful` must be call if the submit succeeded

webapp/src/components/common/DataGrid.tsx
skamril: No, it's done on purpose. Only the properties  above `{...rest}` can be overridden.

webapp/src/components/common/DataGridForm.tsx
skamril: Done

webapp/src/components/common/DataGridForm.tsx
skamril: "primary" make it yellow it's ugly. For me is visible, this is the same for the Form component and there was no feedback for that. And the color will change with the new theme.

webapp/src/components/common/DataGridForm.tsx
skamril: Yes playlist

webapp/src/components/common/TableMode.tsx
skamril: Because `columns.join(",")` for arrays of this size is insignifiant, there is no reason to force `useMemo` for this (and maybe is heavier)

webapp/src/components/common/TableMode.tsx
skamril: Done

webapp/src/components/common/DataGridForm.tsx
skamril: Yes I left the old system

webapp/src/components/common/DataGridForm.tsx
skamril: Yes I left the old system

webapp/src/components/common/DataGridForm.tsx
skamril: No it is used to check the type of the `submitResult` on `onSubmitSuccessful`. Without that the type would be `any`.  With that is based on the `onSubmit` function you give.

webapp/src/components/common/DataGridForm.tsx
skamril: No `submitResult` can be any value, `TData` is already present in `data.values`. It's depend on the API return value. And `onSubmit` can return a specific object also to display a message in `onSubmitSuccessful` for example.

Look at `CreateVariantDialog` to see an example.

webapp/src/components/common/DataGridForm.tsx
skamril: Ok

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
skamril: No typing error in `DataGridFrom`, extra properties are not considered as an error in TS

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
hdinia: ok I get it for you in this context it looks fine, but someone touching this code later and that is not aware of this can introduce a runtime error.

from the link you provide: _"Note that you are opting out of type safety here - you will have a runtime error if you forget to assign the ref to an element in the render, or if the ref-ed element is conditionally rendered."_

you should never make assumptions in your code, particularly if you dont work alone on the codebase. better to repeat 2 conditions and be safe, I maintain my point.

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
hdinia: ok

webapp/src/components/common/DataGrid.tsx
hdinia: sure I didn't meant you overwritten the core library sorry if It was not clear
the issue is:
you're using a type that is not exposed by the public API (import from node_modules) + you overwrite this type (the property `kind`) instead of declaring a custom one for your needs and compose it with the existing type, this is fragile and more complex. It's only types their is no major risk of breaking change, but we should avoid this anyway

I spent time writing a full solution, if you cannot see the benefit let's move on

webapp/src/components/common/DataGrid.tsx
hdinia: sure it will be better for readability and resolve this

webapp/src/components/common/DataGrid.tsx
hdinia: alright

webapp/src/components/common/DataGridForm.tsx
hdinia: I dont get it, in what case you could need to make a sync submit ? usually a form calls an Api to update data

webapp/src/hooks/useCloseFormSecurity.ts
hdinia: it was a suggestion here, but ok

webapp/src/components/common/DataGrid.tsx
hdinia: those rules for comments block are coming from you, you should stick to them if you asks others to do so

webapp/src/components/common/DataGrid.tsx
hdinia: not talking about `getCellContent` I mentioned event handlers.

you make sure these stay standard on all PR's, but now you introduce random names I dont get it
I am not gonna enforce this, but we should work on rules asap and stick to them

webapp/src/components/common/DataGridForm.tsx
hdinia: check it and you will see:
using !== for object comparison only checks reference equality, not deep equality of the objects' content
the reference of the object can change and not the content, this will cause to make the form dirty in a state that shloud not

webapp/src/components/common/DataGridForm.tsx
hdinia: alright thanks 👍🏼

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
skamril: It's like the `as` operator, if you want to remove the non-null assertion you have to remove the `as` also because you will have a runtime error if the type assertion is false.

webapp/src/components/common/DataGrid.tsx
hdinia: true, mybad

webapp/src/components/common/DataGridForm.tsx
hdinia: are you sure ? I see only strings, look at this response (from Playlist dialog):

```json
{"1":{"status":false,"weight":1.0}}
```

webapp/src/components/common/TableMode.tsx
hdinia: ok

webapp/src/components/common/DataGridForm.tsx
hdinia: hum seems weird to me, "result" imply its the returned value of the submit handler, which is suppose to return data or nothing. but I will check it more in depth

webapp/src/components/common/DataGridForm.tsx
skamril: Maybe it can be used to changed a state only or optimistic change, the code will be the same, so why remove `void`?

webapp/src/components/common/DataGrid.tsx
hdinia: you should never do that (except when the package documentation explicitly tells to do so) even if it's just a type, prefer the public API

webapp/src/components/common/DataGrid.tsx
hdinia: ok

webapp/src/components/common/DataGrid.tsx
skamril: I did like this years ago it's not new, I didn't write a doc so how do you know I didn't follow the rules?

The goal is to organize the code, if there is a lot of event handlers and I put them all inside, there is no organization.

This is not a rule that can be followed to the letter in all cases, sometimes the developer can be pragmatic, it is not like an ESLint rule.

webapp/src/components/common/DataGrid.tsx
skamril: I respond below... Make one comment for the same file next time please

webapp/src/components/common/DataGridForm.tsx
skamril: I know what `!==` do thank you. Do you test in a browser and find an issue? 
Because I already respond about this.

webapp/src/components/common/DataGridForm.tsx
skamril: image

Look at the API

webapp/src/components/common/DataGridForm.tsx
hdinia: so you should know that relying on reference equality can be fragile and hard to predict when working with non-primitive types

edit: I tested on the UI and found an issue, to reproduce update a value, then rollback by hand -> the form state stays dirty, you can sumbit and you should not be able to. the issue does not seem related to undo/redo as I thought mybad

webapp/src/components/common/DataGridForm.tsx
hdinia: I dont know I was asking you for what case you added it 😄  if no particular reason I think it should stay async because a form usually calls an api so the submit callback should be async

webapp/src/components/common/DataGrid.tsx
hdinia: you didn't write a doc, but you check for this to maintain consistency across the codebase on code reviews, if  different comment block are added, you enforce the regular comments blocks "Events handlers" "Utils" "Jsx" in components. So why not sticking to the structure you often ask to respect ?
-> those comments should not be random, stick to overall structure or lets discuss this and create a fixed doc that all should follow

webapp/src/components/common/DataGrid.tsx
hdinia: mybad confused with DataGridForm

webapp/src/components/common/DataGridForm.tsx
hdinia: ok I'll check that again

webapp/src/components/common/DataGridForm.tsx
hdinia: final take on that one: seems not a major bug, however if later on this component is extended and new features rely on dirty state this can cause hard to predict bugs. please make sure to document this issue
-> we should also test the performance implications of a deep compare for large dataset, and use it if performance is not affected

webapp/src/components/common/DataGridForm.tsx
hdinia: ok

webapp/src/components/common/DataGridForm.tsx
hdinia: this is just a visual representation don't rely on it. here the first level key (`id`) is a string
-> check it using `typeof`

webapp/src/components/App/Singlestudy/explore/TableModeList/index.tsx
hdinia: ok

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
hdinia: `null!` is more difficult to understand reading the code, you have to follow the data flow and component lifecycle to ensure their is no risk . even if the current implementation guarantees non-null values, future changes could break this assumption. 
your reason for avoiding explicit checks is just to prevent repeating 2 conditions, that tradeoff isn't worth the potential maintenance and safety issues it introduces

example: 
```ts
  useEffect(() => {
    // this assume dataGridApiRef.current exists because of null!
    // runs before DataGridForm mounts -> runtime error without TS error
    const currentData = dataGridApiRef.current.data;
    console.log(currentData);
  }, []);
  ```
  
  someone could introduce this runtime error without knowing

skamril: So I will add this rule https://typescript-eslint.io/rules/no-non-null-assertion/ in https://github.com/AntaresSimulatorTeam/AntaREST/pull/2294, to prevent the same conversation in another PR

webapp/src/components/common/DataGrid.tsx
skamril: Why?

webapp/src/components/common/DataGrid.tsx
skamril: You can have "Event Handlers" and also "Resize Handlers", "Edition Handlers"... for example if "Event Handlers" is too big, don't overcomplicate it, the important thing is to organize the component

webapp/src/components/common/DataGridForm.tsx
skamril: You can do `obj[1]`, but I discovered that all keys are converted in string in JS https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors#property_names

I will correct the return type from the method API also who indicate `number`.

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
hdinia: fine looks good to me thanks 👍🏼

webapp/src/components/common/DataGrid.tsx
hdinia: This is considered an anti-pattern, you can search why it's a bad practice but it's obvious
here few reasons:

- Direct imports from node_modules/*/dist bypass the package's entry points and module resolution system 
- Internal types can change between minor versions (x.Y.z) since they follow semver for runtime behavior, not type definitions
- The location and structure of .d.ts files isn't guaranteed to be stable
- Internal types often reference other internal types, creating a chain of unstable dependencies
- If Glide maintainers refactors internally, they might move RowMarkerOptions to a different file, or split it into smaller types, renames properties etc..

webapp/src/components/common/DataGrid.tsx
hdinia: ok mybad this is not the place to discuss such details, we will work on this when we do the frontend contribution rules

webapp/src/components/common/DataGridForm.tsx
hdinia: indeed that's what my comment is about, thanks for the link

webapp/src/components/common/DataGrid.tsx
skamril: You can check the change

webapp/src/components/common/DataGridForm.tsx
skamril: I add a comment

hdinia (approved):
Looks good to me, thanks for the changes 👍🏼

PR #2290, opened 2025-01-10
feat(variant): add reason when variant generation fails

MartinBelthle:


antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -971,20 +971,28 @@ def _safe_generation(self, metadata: VariantStudy, timeout: int = DEFAULT_AWAIT_
...
+            if not result.result:
+                raise ValueError("No task result")
+            if result.result.success:
                 # OK, the study has been generated
                 return
-            raise ValueError("No task result or result failed")
+            # The variant generation failed, we have to raise a clear exception.
+            error_msg = result.result.message
+            stripped_msg = error_msg.removeprefix(
+                f"Task {task_id} failed: Unhandled exception 417: Failed to generate variant study {metadata.id}"
```
sylvlecl: It's a little fragile to do this, if we change one letter in the base message, this will not work anymore: do you think we could do something more robust?

tests/storage/business/test_variant_study_service.py
```diff
@@ -104,7 +101,7 @@ def task_status(*args):
             yield t
 
     study_service.task_service.status_task.side_effect = task_status()
-    with pytest.raises(VariantGenerationError, match="Error while generating study2.py"):
+    with pytest.raises(VariantGenerationError, match="Error while generating variant study2.py"):
```
sylvlecl: We can improve the test by adding an error message in the result

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -971,20 +971,28 @@ def _safe_generation(self, metadata: VariantStudy, timeout: int = DEFAULT_AWAIT_
...
             # Raise a REQUEST_TIMEOUT error (408)
             logger.error(f"⚡ Timeout while generating variant study {metadata.id}", exc_info=e)
-            raise VariantGenerationTimeoutError(f"Timeout while generating {metadata.id}") from None
+            raise VariantGenerationTimeoutError(f"Timeout while generating variant {metadata.id}") from None
 
         except Exception as e:
             # raise a EXPECTATION_FAILED error (417)
             logger.error(f"⚡ Fail to generate variant study {metadata.id}", exc_info=e)
-            raise VariantGenerationError(f"Error while generating {metadata.id}") from None
+            raise VariantGenerationError(f"Error while generating variant {metadata.id} {e.args[0]}") from None
```
sylvlecl: Isn't `{e}` enough instead of {e.args[0]} ?

sylvlecl (changes requested):
minor comment + we should adapt unit test

TheoPascoli (commented):
Hello Martin, that looks good ! 
I have a few questions about the "remove prefix" : why do we need it ? And is there a unit test for it ? This looks like something that can lead to error in the future, we should make sure that there is no better way to do it, and in that case, to fully test it

MartinBelthle:
> Hello Martin, that looks good ! I have a few questions about the "remove prefix" : why do we need it ? And is there a unit test for it ? This looks like something that can lead to error in the future, we should make sure that there is no better way to do it, and in that case, to fully test it

I used remove prefix to remove duplicated information and for the user to have a clearer message. Also, this method cannot fail, see doc: ` If the string starts with the prefix string, return string[len(prefix):]. Otherwise, return a copy of the original string:` so I don't think it's risky. Also, the remove prefix isn't tested I'll try to add an integration test

TheoPascoli (approved):

PR #2293, opened 2025-01-13
fix(disk-usage): suppress exceptions

MartinBelthle:
Fix [ANT-2567]

### DIFF
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -186,12 +186,13 @@ def get_disk_usage(path: t.Union[str, Path]) -> int:
     if is_archive_format(path.suffix.lower()):
         return os.path.getsize(path)
     total_size = 0
-    with os.scandir(path) as it:
-        for entry in it:
-            if entry.is_file():
-                total_size += entry.stat().st_size
-            elif entry.is_dir():
-                total_size += get_disk_usage(path=str(entry.path))
+    with contextlib.suppress(FileNotFoundError, PermissionError):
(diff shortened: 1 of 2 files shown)


antarest/study/service.py
```diff
@@ -186,12 +186,13 @@ def get_disk_usage(path: t.Union[str, Path]) -> int:
     if is_archive_format(path.suffix.lower()):
         return os.path.getsize(path)
     total_size = 0
-    with os.scandir(path) as it:
-        for entry in it:
-            if entry.is_file():
-                total_size += entry.stat().st_size
-            elif entry.is_dir():
-                total_size += get_disk_usage(path=str(entry.path))
+    with contextlib.suppress(FileNotFoundError, PermissionError):
```
sylvlecl, line 189: We should do it inside the loop otherwise we will skip information from valid files that come after the "invalid" one

MartinBelthle: I thought contextlib could continue if an exception was encountered but I misunderstood the doc

smailio: contextlib.suppress acts like try except pass. You do it both for the dir and for each file so you should be fine.  

```python
import os 
from contextlib import suppress

l = []
with suppress(FileNotFoundError):
    for i in range(10):
        l += [f"{i}"]
        if i == 5:
            os.remove('somefile.tmp')

print(' '.join(l)) # this will run even if an exception was raised above and will print 1 2 3 4 5
```

PR #2294, opened 2025-01-14
chore(ui): update Node version, ESLint and Prettier

skamril:
ANT-2231


webapp/eslint.config.js
```diff
@@ -0,0 +1,139 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+import globals from "globals";
+import eslint from "@eslint/js";
+import tseslint from "typescript-eslint";
+import reactPlugin from "eslint-plugin-react";
+import reactHookPlugin from "eslint-plugin-react-hooks";
```
hdinia, line 19: look like their is an issue with types declaration files for react hook plugin and license header plugin, can you check it ?

webapp/eslint.config.js
```diff
@@ -0,0 +1,139 @@
...
+    ignores: ["dist/*", "license-header.js"],
+  },
+  eslint.configs.recommended,
+  ...tseslint.configs.recommended,
+  ...tseslint.configs.stylistic,
+  {
+    ...reactPlugin.configs.flat.recommended,
+    settings: {
+      react: {
+        version: "detect",
```
hdinia, line 38: you can use the current project version "18.3"

webapp/eslint.config.js
```diff
@@ -0,0 +1,139 @@
...
+              message:
+                'Import icon directly `import [NAME]Icon "@mui/icons-material/[NAME]"` (safest for bundle size).',
+              allowTypeImports: true,
+            },
+            {
+              name: "lodash",
+              message:
+                'Import method directly `import [METHOD] from "lodash/[METHOD]"` (safest for bundle size).',
+              allowTypeImports: true,
+            },
```
hdinia: This is not necessary: modern bundlers like Rollup have built-in tree-shaking that optimize production bundle size

webapp/eslint.config.js
```diff
@@ -0,0 +1,139 @@
...
+          ],
+          patterns: [
+            {
+              group: ["react"],
+              importNamePattern:
+                "^(React|Function|Ref|Mutable|CSS|Component|Props|Form)|(Event|Handler|Attributes)$",
+              message: "Use `React.TYPE` instead of import it.",
+            },
+          ],
```
hdinia: you should discuss such rules additions with other maintainers of the app before enforcing them, but it's ok for me I can see the benefit

small typo in the message, you can update it to this for example: `message: "Use React.TYPE namespace (e.g., React.FC, React.ComponentType) instead of importing types directly from 'react'.",`

webapp/eslint.config.js
```diff
@@ -0,0 +1,139 @@
...
+      sourceType: "module",
+      globals: {
+        ...globals.browser,
+        ...globals.es2022,
+      },
+      // parserOptions: {
+      //   project: ["tsconfig.json", "tsconfig.node.json"],
+      //   tsconfigRootDir: import.meta.dirname,
+      // },
+    },
```
hdinia: you forgot to uncomment this ?

webapp/eslint.config.js
```diff
@@ -0,0 +1,139 @@
...
+          ],
+        },
+      ],
+      curly: "error",
+      "jsdoc/no-defaults": "off",
+      "jsdoc/require-jsdoc": "off",
+      "jsdoc/require-hyphen-before-param-description": "warn",
+      "jsdoc/tag-lines": ["warn", "any", { startLines: 1 }], // Expected 1 line after block description
+      "license-header/header": ["error", "license-header.js"],
+      "no-console": "warn",
```
hdinia, line 101:
```ts

      'no-console': ['error', { allow: ['warn', 'error', 'debug'] }]
    
```

ban .log, keep the ones we might need in production (error, warn etc)

webapp/eslint.config.js
```diff
@@ -0,0 +1,139 @@
...
+          // Function declarations are hoisted, so it’s safe.
+          functions: false,
+        },
+      ],
+      "react-refresh/only-export-components": ["warn", { allowConstantExport: true }],
+      "react/hook-use-state": "error",
+      "react/prop-types": "off",
+      "react/self-closing-comp": "error",
+      "require-await": "warn", // TODO: switch to "error" when the quantity of warning will be low
+    },
```
hdinia, line 128: Here's some rules we might need to add in addition:
```ts

// Prevents non-null assertions
"@typescript-eslint/no-non-null-assertion": "error"
"@typescript-eslint/no-non-null-asserted-optional-chain": "error"
"@typescript-eslint/no-non-null-asserted-nullish-coalescing": "error"

// Prevents redundant type annotations
"@typescript-eslint/no-inferrable-types": "error"


// Prevents unnecessary boolean comparisons
"@typescript-eslint/no-unnecessary-boolean-literal-compare": "error"

// Prevents unnecessary type assertions
"@typescript-eslint/no-unnecessary-type-assertion": "error"

```

webapp/eslint.config.js
```diff
@@ -0,0 +1,139 @@
...
+      "@typescript-eslint/no-unused-vars": ["error", { args: "none", ignoreRestSiblings: true }],
+      camelcase: [
+        "error",
+        {
+          properties: "never", // TODO: remove when server responses are camel case
+          allow: [
+            "MRT_", // For material-react-table
+          ],
+        },
+      ],
```
hdinia, line 94: you can instead make use of @typescript-eslint/naming-convention which is more powerful and TypeScript-aware.

hdinia: example:
```ts
"@typescript-eslint/naming-convention": [ 
        "error",
        {
          selector: "default",
          format: ["camelCase"],
          leadingUnderscore: "allow",
          trailingUnderscore: "allow",
        },
        {
          selector: "variable",
          format: ["camelCase", "UPPER_CASE", "PascalCase"],
          leadingUnderscore: "allow",
          trailingUnderscore: "allow",
          filter: { regex: "^MRT_", match: false }
        },
        {
          selector: "typeLike",
          format: ["PascalCase"],
        },
        {
          selector: "property",
          format: null, // Allows any format for properties
          leadingUnderscore: "allow",
          trailingUnderscore: "allow",
        }
      ],
      ```

**Review by hdinia — CHANGES_REQUESTED** [review]

File: webapp/eslint.config.js (hunk 1 above)
Comment (skamril) [reply webapp/eslint.config.js:19]: No issue, what problem are you encountering?

File: webapp/eslint.config.js (hunk 2 above)
Comment (skamril) [reply webapp/eslint.config.js:38]: `detect` is better, we don't need to change this file when React is updated

File: webapp/eslint.config.js (hunk 7 above)
Comment (skamril) [reply webapp/eslint.config.js:128]: @typescript-eslint/no-non-null-assertion in another PR, some other are already in @typescript-eslint/recommended

File: webapp/eslint.config.js (hunk 8 above)
Comment (skamril) [reply webapp/eslint.config.js:94]: The current rule is easy to read and works well, it's not worth it to complicated it.
I don't understand what you code does.

File: webapp/eslint.config.js (hunk 6 above)
Comment (skamril) [reply webapp/eslint.config.js:101]: You add a Vite setting to remove all console API:

```
esbuild: {
  // Remove logs safely when building production bundle
  // https://esbuild.github.io/api/#pure
  pure: mode === "production" ? ["console"] : [],
},
```

File: webapp/eslint.config.js (hunk 5 above)
Comment (skamril) [reply webapp/eslint.config.js]: It's needed to use ESLint type checking, I comment it for the moment

File: webapp/eslint.config.js (hunk 3 above)
Comment (skamril) [reply webapp/eslint.config.js]: Not safe with lodash (there is a plugin for it, but for the minimal use that we make with lodash this rule is sufficient), and for "@mui/icons-material" I think it's good https://mui.com/material-ui/guides/minimizing-bundle-size/#when-and-how-to-use-tree-shaking

Comment (hdinia) [reply webapp/eslint.config.js]: from the doc that you link: "The following instructions are only needed if you want to optimize your development startup times or if you are using an older bundler that doesn't support tree-shaking."

Rollup and modern bundlers are built to optimize this for production bundles. We dont care for dev, so their no point making imports verboose for this, please remove it

File: webapp/eslint.config.js (hunk 1 above)
Comment (hdinia) [reply webapp/eslint.config.js:19]: yes, types declarations files are missing, but it's negligible for this (just to let you know in case you missed it)

File: webapp/eslint.config.js (hunk 2 above)
Comment (hdinia) [reply webapp/eslint.config.js:38]: ok

File: webapp/eslint.config.js (hunk 6 above)
Comment (hdinia) [reply webapp/eslint.config.js:101]: true, its maybe too radical. some log can be useful no ? did you ever check if this is effective in prod env ?

File: webapp/eslint.config.js (hunk 7 above)
Comment (hdinia) [reply webapp/eslint.config.js:128]: we should add it now that we touch the config, later it will be never done, but as you prefer. if don't do it now we should trace it by a TODO or a ticket maybe

File: webapp/eslint.config.js (hunk 4 above)
Comment (skamril) [reply webapp/eslint.config.js:90]: For the more impactful rules I talk about them upstream, but for this one the PR is sufficient.

File: webapp/eslint.config.js (hunk 8 above)
Comment (hdinia) [reply webapp/eslint.config.js:94]: It's the same but using the typescript-eslint plugin, it's better for TS and we have more granular control of the settings, like we can set prefix: "T" for types for example. It's not complicating but if you have trouble making it work don't bother

File: webapp/eslint.config.js (hunk 5 above)
Comment (hdinia) [reply webapp/eslint.config.js]: I don't understand, this is needed for some TS rules so why it has to be commented ? It causes an issue with ESlint ?

File: webapp/eslint.config.js (hunk 3 above)
Comment (hdinia) [reply webapp/eslint.config.js]: might be true for lodash I didn't try it, but we should ditch this lib anyway.
we can see the bundle (if tree-shaking is well performed) with that: https://github.com/btd/rollup-plugin-visualizer

do you want to test it ?

File: webapp/.prettierrc.json [hunk 9]
```diff
@@ -1,3 +1,4 @@
 {
-  "endOfLine": "auto"
+  "endOfLine": "auto",
+  "printWidth": 100
```
hdinia, line 3: 🥳🎉🎉👍

webapp/eslint.config.js
skamril: I share the link to say that it's good for "@mui/icons-material", tree-shaking is supported. I will remove the rule for this package.

webapp/eslint.config.js
skamril: https://typescript-eslint.io/getting-started/typed-linting/

webapp/eslint.config.js
skamril: "debug" (already installed) can be used for logging instead https://github.com/debug-js/debug

webapp/eslint.config.js
skamril: I already make the code for @typescript-eslint/no-non-null-assertion, this made me change code logics, so I didn't include it in this PR, which is just modifications without consequences.

webapp/eslint.config.js
skamril: You can make a new PR to add it if you want and test if it's work

webapp/eslint.config.js
hdinia: ok sorry did not understand this, you should remove it for lodash also, was we rarely use it and we should remove it in the future. The real issue here are not imports, but the outdated nature of lodash that does not use ES modules. 

Also if you want to make this effective you have to replace all existing imports (don't bother with that)

hdinia: Our bundle size is very large: index-DCu2zW0N.js: 6.3MB (1.8MB gzipped)

image


If you want to optimize it:
- replace all lodash occurence and remove the package
- remove handsontable
- remove jsoneditor 
- remove moment

optimize the rest using code splitting

webapp/eslint.config.js
hdinia: still not understand, why not using the same as from the doc ?

webapp/eslint.config.js
skamril: `tseslint.configs.recommendedTypeChecked`? Hard to remove all current errors in the app

webapp/eslint.config.js
skamril: `tseslint.configs.recommendedTypeChecked`? Hard to remove all current errors in the app with it.
I remove the comment.

hdinia (approved):

PR #2295, opened 2025-01-15
feat(bc): add constraint duplication endpoint

MartinBelthle:


antarest/study/web/study_data_blueprint.py
```diff
@@ -1340,6 +1340,27 @@ def create_binding_constraint(
...
 
+    @bp.post(
+        "/studies/{uuid}/bindingconstraints/{binding_constraint_id}",
+        tags=[APITag.study_data],
+        summary="Duplicates a given binding constraint",
+    )
+    def duplicate_binding_constraint(
+        uuid: str,
+        binding_constraint_id: str,
+        new_name: str,
```
TheoPascoli: Can you replace new_name with new_constraint_name to have the same name as in the duplicate_binding_constraint method ?
Do we need to validate this field ? Like adding a max length

antarest/study/business/binding_constraint_management.py
```diff
@@ -795,6 +795,75 @@ def create_binding_constraint(
...
+        # Checks if the new constraint already exists
+        new_constraint_id = transform_name_to_id(new_constraint_name)
+        existing_constraints = self.get_binding_constraints(study)
+        if new_constraint_id in {bc.id for bc in existing_constraints}:
+            raise DuplicateConstraintName(
+                f"A binding constraint with the same name already exists: {new_constraint_name}."
+            )
+
+        # Retrieval of the source constraint properties
+        source_constraint = next(iter(bc for bc in existing_constraints if bc.id == source_id), None)
```
TheoPascoli, line 823: This code is correct, but it can be improved. It runs in linear time O(n), whereas a dict would find it in constant time O(1). If the list is long it is better to have a dict, it is not completely necessary though as we dont have millions of bc.

antarest/study/business/binding_constraint_management.py
```diff
@@ -795,6 +795,75 @@ def create_binding_constraint(
...
+                args[command_attribute] = matrix
+
+        # Creates and applies constraint
+        command = CreateBindingConstraint(**args)
+        execute_or_add_commands(study, file_study, [command], self.storage_service)
+
+        # Returns the new constraint
+        source_constraint.name = new_constraint_name
+        source_constraint.id = new_constraint_id
+        return source_constraint
```
TheoPascoli, line 865: For this method, we could split the concerns instead of having everything in one method. Like one method for validating and getting the constraint, one to handle matrix, and on to create and execute the command, something like that

TheoPascoli (changes requested):
Hello Martin, great work overall, I've left a few comments for improvements. Let me know if you have questions

antarest/study/business/binding_constraint_management.py
MartinBelthle: I agree. But i don't see how I can have the dict, as the `get_binding_constraints` method returns a list :/
I could change the method to return a dict. But i don't think it belongs inside this PR. 
Do you have something in mind ?

antarest/study/web/study_data_blueprint.py
MartinBelthle: I can rename it. 
Concerning validation, we don't do it currently at the creation so I don't think we should implement it for the duplication

antarest/study/business/binding_constraint_management.py
MartinBelthle: I tried to do the command creation inside a separated generic method to use inside creation and duplication but for creation I need the variable `new_constraint` I introduce in the middle of the method for the method response ... So I gave up

TheoPascoli (dismissed):

PR #2298, opened 2025-01-16
feat: add an endpoint to allow multiple deletion of binding constrain…

TheoPascoli:
…t at once


antarest/study/business/binding_constraint_management.py
```diff
@@ -589,6 +592,21 @@ def terms_to_coeffs(terms: t.Sequence[ConstraintTerm]) -> t.Dict[str, t.List[flo
                     coeffs[term.id].append(term.offset)
         return coeffs
 
+    def check_binding_constraints_exists(self, study: Study, bc_ids: t.List[str]) -> None:
+        storage_service = self.storage_service.get_storage(study)
+        file_study = storage_service.get_raw(study)
+        existing_constraints = file_study.tree.get(["input", "bindingconstraints", "bindingconstraints"])
+
+        if not existing_constraints:
```
MartinBelthle: Don't think you should add this if.
Existing constraints can only be an empty dict and if so `existing_constraints.values()` will be an empty list and `missing_bc_ids` will be your entire list. 
So i think you should remove it

antarest/study/business/binding_constraint_management.py
```diff
@@ -1018,6 +1036,31 @@ def remove_binding_constraint(self, study: Study, binding_constraint_id: str) ->
         )
         execute_or_add_commands(study, file_study, [command], self.storage_service)
 
+    def remove_multiple_binding_constraints(self, study: Study, binding_constraints_ids: t.List[str]) -> None:
+        """
+        Removes multiple binding constraints from a study.
+
+        Args:
+            study: The study from which to remove the constraint.
+            binding_constraints_ids: The ID sof the binding constraints to remove.
```
MartinBelthle: little typo

antarest/study/web/study_data_blueprint.py
```diff
@@ -1378,6 +1378,25 @@ def delete_binding_constraint(
         study = study_service.check_study_access(uuid, StudyPermissionType.WRITE, params)
         return study_service.binding_constraint_manager.remove_binding_constraint(study, binding_constraint_id)
 
+    @bp.delete(
+        "/studies/{uuid}/bindingconstraints",
+        tags=[APITag.study_data],
+        summary="Delete multiple binding constraint",
```
MartinBelthle: constraintS ?

tests/variantstudy/test_command_factory.py
```diff
@@ -159,6 +159,16 @@
     CommandDTO(
         action=CommandName.REMOVE_BINDING_CONSTRAINT.value, args=[{"id": "id"}], study_version=STUDY_VERSION_8_8
     ),
+    CommandDTO(
```
MartinBelthle, line 162: Why did you add 2 DTOs that are almost the same ?

tests/integration/study_data_blueprint/test_binding_constraints.py
```diff
@@ -972,6 +972,31 @@ def test_for_version_870(self, client: TestClient, user_access_token: str, study
...
+        # Delete multiple binding constraint
+
+        preparer.create_binding_constraint(study_id, name="bc1", group="grp1", **args)
+        preparer.create_binding_constraint(study_id, name="bc2", group="grp2", **args)
+
+        binding_constraints_list = preparer.get_binding_constraints(study_id)
+        assert len(binding_constraints_list) == 4
+
+        groups = preparer.get_binding_constraints_groups(study_id)
+        assert len(groups) == 4
```
MartinBelthle: I'm perhaps missing something but why are you testing the amount of bc groups ?

tests/integration/study_data_blueprint/test_binding_constraints.py
```diff
@@ -972,6 +972,31 @@ def test_for_version_870(self, client: TestClient, user_access_token: str, study
...
+        res = client.request(
+            "DELETE",
+            f"/v1/studies/{study_id}/bindingconstraints",
+            json=["bc1", "bc2"],
+        )
+        assert res.status_code == 200, res.json()
+
+        # Asserts that the deletion worked
+        binding_constraints_list = preparer.get_binding_constraints(study_id)
+        assert len(binding_constraints_list) == 2
```
MartinBelthle, line 991: Could be more precise and check that the 2 left are not the one you removed

antarest/study/storage/variantstudy/model/command/remove_multiple_binding_constraints.py
```diff
@@ -0,0 +1,112 @@
...
+
+class RemoveMultipleBindingConstraints(ICommand):
+    """
+    Command used to remove multiple binding constraints at once.
+    """
+
+    command_name: CommandName = CommandName.REMOVE_MULTIPLE_BINDING_CONSTRAINTS
+    version: int = 1
+
+    # Properties of the `REMOVE_MULTIPLE_BINDING_CONSTRAINT` command:
```
MartinBelthle: typo: constraintS

antarest/study/storage/variantstudy/model/command/remove_multiple_binding_constraints.py
```diff
@@ -0,0 +1,112 @@
...
+    ids: t.List[str]
+
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> OutputTuple:
+        # If at least one bc is missing in the database, we raise an error
```
MartinBelthle, line 40: You're doing the same thing twice it seems. You can call `_apply_config` at the start of the `apply` to avoid having duplicated code.

antarest/study/storage/variantstudy/model/command/remove_multiple_binding_constraints.py
```diff
@@ -0,0 +1,112 @@
...
+
+        binding_constraints = study_data.tree.get(["input", "bindingconstraints", "bindingconstraints"])
+
+        old_groups = {bd.get("group", DEFAULT_GROUP).lower() for bd in binding_constraints.values()}
+
+        deleted_binding_constraints = {}
+
+        for key in list(binding_constraints.keys()):
+            if binding_constraints[key].get("id") in self.ids:
+                deleted_binding_constraints[key] = binding_constraints.pop(key)
```
MartinBelthle: I think you should store the bc_ids in a list instead of a dict as you don't use its keys afterwards it seems (keys are 0, 1, 2 and are useless)

antarest/study/storage/variantstudy/model/command/remove_multiple_binding_constraints.py
```diff
@@ -0,0 +1,112 @@
...
+        study_data.tree.save(
+            binding_constraints,
+            ["input", "bindingconstraints", "bindingconstraints"],
+        )
+
+        for bc in deleted_binding_constraints:
+            if study_data.config.version < STUDY_VERSION_8_7:
+                study_data.tree.delete(["input", "bindingconstraints", deleted_binding_constraints[bc].get("id")])
+            else:
+                existing_files = study_data.tree.get(["input", "bindingconstraints"], depth=1)
```
MartinBelthle: You could do this get once before your loop instead of doing this for every constraint to delete

MartinBelthle (changes requested):
Minor reviews, otherwise looks good

tests/variantstudy/test_command_factory.py
TheoPascoli: I copied what was done right above with the REMOVE_BINDING_CONSTRAINT DTO

tests/integration/study_data_blueprint/test_binding_constraints.py
TheoPascoli: Because I delete the groups at the end of the bc removing with remove_bc_from_scenario_builder. Maybe I'm missing something

tests/variantstudy/test_command_factory.py
MartinBelthle: Okay

tests/integration/study_data_blueprint/test_binding_constraints.py
MartinBelthle: Okay I see. In fact, what this function achieves is removing the groups inside the scenario builder which is a separated file "scenariobuilder.dat". You could test that you're doing it well but I'm not sure your scenario builder is filled in the first place so you'll have to fill it manually beforehand

MartinBelthle: And so you can remove your tests on groups as they don't check what we want

TheoPascoli:
> Minor reviews, otherwise looks good

Thank you for your remarks ! I did what you suggested

antarest/study/storage/variantstudy/model/command/remove_multiple_binding_constraints.py
```diff
@@ -0,0 +1,108 @@
...
+                for term in ["lt", "gt", "eq"]:
+                    matrix_id = f"{bc.get('id')}_{term}"
+                    if matrix_id in existing_files:
+                        study_data.tree.delete(["input", "bindingconstraints", matrix_id])
+
+        new_groups = {bd.get("group", DEFAULT_GROUP).lower() for bd in binding_constraints.values()}
+        removed_groups = old_groups - new_groups
+        remove_bc_from_scenario_builder(study_data, removed_groups)
+
+        return self._apply_config(study_data.config)[0]
```
MartinBelthle: Now that you call this method at the beginning of the `apply` you could just store the result in a variable to return here instead of calling the method twice

antarest/study/storage/variantstudy/model/command/remove_multiple_binding_constraints.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: t.Optional[ICommandListener] = None) -> CommandOutput:
+        self._apply_config(study_data.config)
+
+        binding_constraints = study_data.tree.get(["input", "bindingconstraints", "bindingconstraints"])
+
+        old_groups = {bd.get("group", DEFAULT_GROUP).lower() for bd in binding_constraints.values()}
+
+        deleted_binding_constraints = []
```
MartinBelthle, line 58: Minor: I was thinking and as you checked inside `apply_config` that the given ids fit, you could just use them instead of introducing a new list ?

MartinBelthle (changes requested):
2 minor comments + the last one on the group tests

MartinBelthle (approved):

PR #2301, opened 2025-01-20
fix(ui-studies): display 'default' workspace even if empty

smailio:


webapp/src/components/App/Studies/StudyTree/utils.ts
```diff
@@ -23,9 +23,18 @@ import { StudyTreeNode, NonStudyFolderDTO } from "./types";
...
 export function buildStudyTree(studies: StudyMetadata[]) {
+  // It is important to initialize the root node with the default workspace as a child
+  // Otherwise we won't see the default workspace if no study has a path (which only
+  // happens when a user moves a study to another folder)
   const tree: StudyTreeNode = {
     name: "root",
-    children: [],
+    children: [
+      {
+        name: "default",
```
skamril, line 33: `default` is not necessarily the name of the folder. It's based on you configuration.

image

Am I wrong?

smailio: It's hard coded in the backend so it should work (should also work for the windows desktop version).

skamril (approved):

PR #2302, opened 2025-01-20
feat(ts-gen): add links ts-generation info inside DB

MartinBelthle:
Will fix [ANT-2651]

**What does this PR achieves**:

- Creates 2 new tables inside the DB to store several information
     - 1 table to store for each study, how many years should be generated for the links TS-generation
     - 1 table to store the generation parameters for each link of each study
- Add the TS generation parameters inside the response of each link endpoint
     - At the creation, the user can specify these parameters
     - At the update, the user can specify these parameters
     - These parameters can be visualized inside the table-mode
- Handling of the variants
    - When we invalidate the cache of a variant, the 2 tables are also cleared
    - When a variant snapshot is generated, parents information inside DB are copied to ensure the information is up to date


alembic/versions/dadad513f1b6_create_links_ts_generation_tables.py
```diff
@@ -0,0 +1,58 @@
...
+        sa.Column("modulation", sa.String(), nullable=True),
+        sa.Column("unit_count", sa.Integer(), nullable=False, server_default="1"),
+        sa.Column("nominal_capacity", sa.Float(), nullable=False, server_default="0"),
+        sa.Column("law_planned", sa.Enum('uniform', 'geometric', name='lawplanned'), nullable=False),
+        sa.Column("law_forced", sa.Enum('uniform', 'geometric', name='lawforced'), nullable=False),
+        sa.Column("volatility_planned", sa.String(), nullable=False, server_default="0"),
+        sa.Column("volatility_forced", sa.String(), nullable=False, server_default="0"),
+        sa.Column("force_no_generation", sa.Boolean(), nullable=False, server_default=default_boolean),
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.ForeignKeyConstraint(
```
sylvlecl, line 48: don't forget to add an index on this, otherwise it will get slow when we have more studies

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -445,6 +453,12 @@ def clear_snapshot(self, variant_study: Study) -> None:
         self.invalidate_cache(variant_study, invalidate_self_snapshot=True)
         shutil.rmtree(self.get_study_path(variant_study), ignore_errors=True)
 
+        # Removes TS-generation related information from the database
```
sylvlecl: should go into `invalidate_cache`, in particular it needs to be applied to descendants

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -644,6 +658,7 @@ def create_variant_study(self, uuid: str, name: str, params: RequestParameters)
             additional_data=additional_data,
         )
         self.repository.save(variant_study)
+        self._copy_parent_ts_generation_info(study.id, new_id)
```
sylvlecl: as discussed, should go at the start of snapshot generation, it's part of the same "copy parent" operation that we do at the beginning of generation

sylvlecl (commented):
a few comments

PR #2303, opened 2025-01-20
feat(ts-gen)!: replace legacy endpoints with new ones

MartinBelthle:
ANT-2675


webapp/src/services/api/studies/timeseries/index.ts
```diff
@@ -26,3 +27,7 @@ export async function generateTimeSeries(params: { studyId: StudyMetadata["id"]
   const { data } = await client.put<string>(`/v1/studies/${params.studyId}/timeseries/generate`);
   return data;
 }
+
+export async function setTimeSeriesConfig<T>({ studyId, values }: SetTimeSeriesConfigParams<T>) {
+  await client.put(`v1/studies/${studyId}/timeseries/config`, values);
```
hdinia: minor: missing leading /

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/Fields.tsx
```diff
@@ -59,29 +60,31 @@ function Fields() {
...
-                size="small"
-                disabled={isReadyMade}
-                rules={{ validate: validateNumber({ min: 1 }) }}
-                sx={{ width: 110 }}
-              />
-            </TableCell>
-          </TableRow>
+          {Object.values(TSType).map((type) => (
+            <TableRow key={type}>
+              <TableCell sx={{ fontWeight: "bold" }}>{capitalize(type)}</TableCell>
```
hdinia: is this an old code ?
tip: avoid using lodash for such simple thing. you can add an enum-like object with i18n keys and map it to the type (e.g. `TimeSeriesLabels`)

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/Fields.tsx
```diff
@@ -59,29 +60,31 @@ function Fields() {
...
+                  size="small"
+                />
+              </TableCell>
+              <TableCell align="center">
+                <NumberFE
+                  name={`${type}.number` as const}
+                  control={control}
+                  size="small"
+                  disabled={formValues[type].stochasticTsStatus === false}
+                  rules={{ validate: validateNumber({ min: 1 }) }}
```
hdinia: question: their is no max in specs ? what if a user submit with a very large number

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
```diff
@@ -12,76 +12,19 @@
...
-import type { DeepPartial } from "react-hook-form";
-import type { StudyMetadata } from "../../../../../../common/types";
-import client from "../../../../../../services/api/client";
+import { TSType } from "@/services/api/studies/timeseries/constants";
+import type { TSTypeConfig, TSConfigDTO, TTSType } from "@/services/api/studies/timeseries/types";
 
-////////////////////////////////////////////////////////////////
-// Enums
-////////////////////////////////////////////////////////////////
+export type TSConfigValues = Record<TTSType, TSTypeConfig & { stochasticTsStatus: boolean }>;
```
hdinia: put in types file under service/.../.../timeseries (it's also related to api service)

webapp/src/services/api/studies/timeseries/types.ts
```diff
@@ -0,0 +1,32 @@
...
+  number: number;
+}
+
+export type TSConfigDTO = Record<TTSType, TSTypeConfig>;
+
+export interface SetTimeSeriesConfigParams<T> {
+  studyId: StudyMetadata["id"];
+  // Extra fields not allowed by the API
+  values: DeepPartial<F.Exact<T, TSConfigDTO>>;
+}
```
hdinia: Few remarks about the types and constants
- the API contract and Form UI state are mixed 
cf: `type TSConfigValues = Record<TTSType, TSTypeConfig & { stochasticTsStatus: boolean }>;`
-> split to one for the Form state + one for the API contract

- some types are overly complex (usage of generics seems not appropriate here), same for type utilities from TS toolbelt cf:
```ts
export interface SetTimeSeriesConfigParams<T> {
  studyId: StudyMetadata["id"];
  // Extra fields not allowed by the API
  values: DeepPartial<F.Exact<T, TSConfigDTO>>;
}
```
-> please consider simplifying 

- the naming especially the usage of the "T" prefix + abbreviations reduce clarity cf: `TTSType`


recommandations: 
- remove abbreviations and prefix (you can define a `type` and an `enum-like` with the same declaration name)
- split concerns 

```ts
import { TimeSeriesType } from "./constants";

interface TimeSeriesConfig {
  number: number;
}

export interface TimeSeriesFormData extends TimeSeriesConfig {
  isStochastic: boolean;
}

export type TimeSeriesType = (typeof TimeSeriesType)[keyof typeof TimeSeriesType];
export type TimeSeriesFormValues = Record<TimeSeriesType, TimeSeriesFormData>;
export type TimeSeriesDTO = Record<TimeSeriesType, TimeSeriesConfig>;

export interface UpdateGenerationConfigParams {
  studyId: string;
  data: TimeSeriesDTO;
}

```

hdinia: in contants.ts

```ts
export const TimeSeriesType = {
  THERMAL: "thermal",
} as const;
```

webapp/src/services/api/studies/timeseries/index.ts
```diff
@@ -26,3 +27,7 @@ export async function generateTimeSeries(params: { studyId: StudyMetadata["id"]
   const { data } = await client.put<string>(`/v1/studies/${params.studyId}/timeseries/generate`);
   return data;
 }
+
+export async function setTimeSeriesConfig<T>({ studyId, values }: SetTimeSeriesConfigParams<T>) {
```
hdinia: `setTimeSeriesConfig` -> `updateGenerationConfig`

question: is `generateTimeSeries` responsible for creating the initial generation config, or something else ?

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
```diff
@@ -12,76 +12,19 @@
...
-  },
-};
-
-////////////////////////////////////////////////////////////////
-// Functions
-////////////////////////////////////////////////////////////////
+export const DEFAULT_VALUES = Object.values(TSType).reduce((acc, type) => {
+  acc[type] = { number: 1, stochasticTsStatus: false };
+  return acc;
+}, {} as TSConfigValues);
```
hdinia: minor:
- improper use of uppercase constant notation (it's scoped to the Form)
- consider naming the accumulator when using a reduce, it make it easier to read. (opinion: the mutation of the acc not necessary here)

```ts
export const defaultValues = Object.values(TimeSeriesType).reduce(
  (values, type) => ({
    ...values,
    [type]: {
      number: 1,
      isStochastic: false,
    },
  }),
  {} as TimeSeriesFormValues,
);
```

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/Fields.tsx
```diff
@@ -16,16 +16,17 @@ import { Table, TableBody, TableCell, TableContainer, TableHead, TableRow } from
...
+import type { TSConfigValues } from "./utils";
+import { TSType } from "@/services/api/studies/timeseries/constants";
 
 const borderStyle = "1px solid rgba(255, 255, 255, 0.12)";
 
 function Fields() {
-  const { control, watch } = useFormContextPlus<TSFormFields>();
-  const isReadyMade = watch("thermal.stochasticTsStatus") === false;
+  const { control, watch } = useFormContextPlus<TSConfigValues>();
+  const formValues = watch();
```
hdinia: tip: `useWatch` with specific fields prevents full form re-renders (not an issue here)
https://react-hook-form.com/docs/usewatch

hdinia (changes requested):

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/Fields.tsx
MartinBelthle: There's not really a max number. It's just an int that should be superior or equal to 1

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
```diff
@@ -12,76 +12,19 @@
...
-export function setTimeSeriesFormFields(
-  studyId: StudyMetadata["id"],
-  values: DeepPartial<TSFormFields>,
-): Promise<void> {
-  return client.put(`v1/studies/${studyId}/config/timeseries/form`, values);
+export function toConfigDTO(data: TSConfigValues) {
+  return Object.entries(data).reduce((acc, [key, { stochasticTsStatus, ...config }]) => {
+    acc[key as TTSType] = config;
+    return acc;
+  }, {} as TSConfigDTO);
```
hdinia: minor:
the usage of the adapter pattern is good, but consider few improvements:
- this should live with api methods and be private to them (put in service file)
- the naming is too generic "toConfig" is broad

```ts
function formValuesToDTO(formValues: TimeSeriesFormValues): TimeSeriesDTO {
  return Object.entries(formValues).reduce((request, [type, values]) => {
    const { isStochastic, ...dto } = values;

    return {
      ...request,
      [type]: dto,
    };
  }, {} as TimeSeriesDTO);
}
```

best but optional: 
you can use a facade function and expose methods for the service
```ts
// Private adapter
function formValuesToDTO(formValues: TimeSeriesFormValues): TimeSeriesDTO {
  return Object.entries(formValues).reduce((request, [type, values]) => {
    const { isStochastic, ...dto } = values;
    return {
      ...request,
      [type]: dto,
    };
  }, {} as TimeSeriesDTO);
}

export const timeSeriesService = {
  async createGenerationConfig(studyId: StudyMetadata["id"]) {
    const { data } = await client.put<string>(`/v1/studies/${studyId}/timeseries/generate`);
    return data;
  },

  async updateGenerationConfig(studyId: StudyMetadata["id"], formValues: TimeSeriesFormValues) {
    const data = formValuesToDTO(formValues);
    await client.put(`v1/studies/${studyId}/timeseries/config`, data);
  },
};

//usage
await timeSeriesService.createGenerationConfig(studyId);
await timeSeriesService.updateGenerationConfig(studyId, formValues);
```

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/Fields.tsx
hdinia: ok thanks I was wondering if a very large number for the generation could cause issues

webapp/src/services/api/studies/timeseries/index.ts
skamril: Ok

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/Fields.tsx
skamril: It's difficult to specified fields in my case (I read it in a loop). 
I will create a component for this field and use`useWatch` in it. Thanks.

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/Fields.tsx
skamril: It loops over `TSType` which is a enum like. Importing specific method from lodash (and not the entire library) it's not an issue for the bundle size, it's equivalent to create my own util. 
But it's an old code and I think it's better to add a translation, so I will add the i18n keys for that.

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
skamril: No it's specific to the form, 'stochasticTsStatus' doesn't exist in the API anymore.

webapp/src/services/api/studies/timeseries/index.ts
skamril: I prefer `set` over `update`, because it replaces all the config and not making partial modifications.
`generateTimeSeries` only read the config and create TS matrices.

webapp/src/services/api/studies/timeseries/types.ts
skamril: > the API contract and Form UI state are mixed
> cf: type TSConfigValues = Record<TTSType, TSTypeConfig & { stochasticTsStatus: boolean }>;
> -> split to one for the Form state + one for the API contract

What's the problem to mix them? I already split them: `TTSType` and `TSTypeConfig` are in the "api" folder, and `TSConfigValues` in the form folder because it's specific to it (I answered already above).

> some types are overly complex (usage of generics seems not appropriate here), same for type utilities from TS toolbelt cf:

It's not 'complex' for no reason (I wrote a comment), the PUT endpoint throw an error when we add extra fields in the config, but without the use of `F.Exact`, TypeScript doesn't complain about that. In my form I have extra fields, and this 'complexity' allow to show the error directly in the IDE, I know that I have to omit them before.

> the naming especially the usage of the "T" prefix + abbreviations reduce clarity cf: TTSType

We already talk about that, it's a known convention to have the 'T' prefix in certain case. 
You can see in you example that you have two `TimeSeriesType`: an enum-like and a type, it's confusing.
The type is based on the enum-like, so I kept the name of it, with a 'T' prefix, to prevent duplicate name.

For the abbreviations, like I said, I keep the same names as the backend types in the "api" folder, to simplify the search inside the code and the maintainability. But if @MartinBelthle can change the backend name, I will change them also.

skamril: > ```
> import { TimeSeriesType } from "./constants";
> 
> interface TimeSeriesConfig {
>   number: number;
> }
> 
> export interface TimeSeriesFormData extends TimeSeriesConfig {
>   isStochastic: boolean;
> }
> 
> export type TimeSeriesType = (typeof TimeSeriesType)[keyof typeof TimeSeriesType];
> export type TimeSeriesFormValues = Record<TimeSeriesType, TimeSeriesFormData>;
> export type TimeSeriesDTO = Record<TimeSeriesType, TimeSeriesConfig>;
> 
> export interface UpdateGenerationConfigParams {
>   studyId: string;
>   data: TimeSeriesDTO;
> }
> ```

* `TimeSeriesFormData` is not an API type.
* More simple to read `O.UnionOf<typeof TimeSeriesType>` rather than `(typeof TimeSeriesType)[keyof typeof TimeSeriesType];`. I don't know why you don't like it.
* `TimeSeriesFormValues` is not an API type
* `TimeSeriesDTO` is not precise, this the DTO of the config, this is why "Config" is added, without that we can think about matrices. The global config includes the specific config for each type, this is why we have: `TSTypeConfig`.

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
skamril: > this should live with api methods and be private to them (put in service file)

No, it's specific to the form, `stochasticTsStatus` is not in the API anymore.

> the naming is too generic "toConfig" is broad

It's not `toConfig` but `toConfigDTO`. It is only include in the TS form folder (not a shared method), the function argument is typed, so it's clear in this context. It's not necessary to be too verbose here.

> you can use a facade function and expose methods for the service

`/v1/studies/{studyId}/timeseries/generate` generate TS matrices, not a config.
Otherwise, I don't see the benefits of this encapsulation compared to the direct export of methods (in addition to maintaining consistency in the code).

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
skamril: > improper use of uppercase constant notation (it's scoped to the Form)

Ok

> consider naming the accumulator when using a reduce, it make it easier to read. (opinion: the mutation of the acc not necessary here)

We use `acc` almost everywhere (for me it's not a problem, because it is often small codes like here, but if you want more precise name you have to add a new regex in ESLint rule 'no-param-reassign').

Then, I prefer to use mutation with `reduce` every time, as it is safe with an initial value and, in all cases, more efficient than the spread operator. So why complicate things just for a code styling preference? This approach is simpler than mutating only when the object is large (what is the limit?...), especially since its size can sometimes change.

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/Fields.tsx
hdinia: thanks for the change, yes or a simple util is preferable to load lodash `String.prototype.charAt(0).toUppercase()`
but here the keys like you added are best

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
hdinia: you're point is valid, it's local to the form, but with the other changes (moving the adapter to the service layer) it becomes questionnable. I let you decide both are fine but it's best to localize all types in my opinion

webapp/src/services/api/studies/timeseries/index.ts
hdinia: ok

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
hdinia: >We use acc almost everywhere

I think we should reconsider this, sometimes it makes it hard to follow. but not a big issue we can see this later.

(Sure about the mutation, but be careful it's not to use every time because it can make the code harder to debug and read, so you have to consider the tradeoff depending on the context.)

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
hdinia: >No, it's specific to the form, stochasticTsStatus is not in the API anymore.

I'm not talking about types here, but just the usage you do with that adapter, you use it in the Form, which is not the appropriate place ( adapters live at service layer ) your Form should not handle concerns about the API contract.

>It's not toConfig but toConfigDTO. It is only include in the TS form folder (not a shared method), the function argument is typed, so it's clear in this context. It's not necessary to be too verbose here.

you can be converting anything to config DTO here, adapter naming (what you adapt + the target of the transformation) here you convert form values to DTO so simply `formValuesToDTO`. if you want to be verboose (which is not bad) you could name it `formValuesToTimeSeriesConfigDTO`.

put it in the service file, and use it here not in the Form

you can look more about the adapter pattern here: [Adapter pattern](https://sendgrid.com/en-us/blog/using-the-adapter-design-pattern-with-react)

webapp/src/services/api/studies/timeseries/types.ts
hdinia: >What's the problem to mix them? I already split them: TTSType and TSTypeConfig are in the "api" folder, and TSConfigValues in the form folder because it's specific to it (I answered already above).

my comment was unclear sorry, I mean that when looking the types this one looks like the DTO (because of the name `TSConfigValues`) rather than the form state values. but since it's local to the form you are correct

-> you're point is valid, consider renaming it to reflect it's purpose

>It's not 'complex' for no reason (I wrote a comment), the PUT endpoint throw an error when we add extra fields in the config

not I don't like the syntax, but it adds nothing more. TypeScript provides everything you need out of the box here. The only exception is for complex utility types that are too messy to write in TS.

when it becomes so complex for such a simple type I have a problem with it, see:
`  values: DeepPartial<F.Exact<T, TSConfigDTO>>;` 
look at the type signature
```ts
(parameter) values: DeepPartial<T extends TSConfigDTO ? T extends Narrowable ? T : { [K in keyof T]: K extends "thermal" ? Exact<T[K], TSConfigDTO[K]> : never; } : TSConfigDTO>
```
it's barely understandable, and not necessary because the enum object dictates already the form fields

>We already talk about that, it's a known convention to have the 'T' prefix in certain case.

-> The only case where it's justified is for function parameters, for the rest it's redundant (like the "I" prefix for Interfaces). Same for suffix Type I used it's useless we have to drop both

You should drop the addition of "T", what info does it give you if it's added to every single type name? Of course it's a type, in the places where it's used it can only be a type.

With that logic, why wouldn't you do the same thing for variables? 
`const userVariable: User = ...` or with  prefix `cont Vuser: User = ... `

❌ const user: UserType = ...  or  const user: TUser = ...    
✅ const user: User = ...

I let you do your own research on that topic if you dont believe me 

>For the abbreviations, like I said, I keep the same names as the backend types in the "api" folder, to simplify the search inside the code and the maintainability.

-> this is a non-sense.. backend types has their own context, maybe for DTO's and even you should see the frontend as an isolated layer, and not rely on external API's to name your types.

hdinia: for your second comment on types, please focus on the main points rather than the details:
- organisation
- naming
- removing the over-complexity

for the rest you can adapt to the context if you judge `TimeSeriesConfigDTO` is better than `TimeSeriesDTO` I have no problem with that

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
```diff
@@ -12,76 +12,19 @@
...
-  thermal: {
-    stochasticTsStatus: false,
-    number: 1,
-  },
-};
-
-////////////////////////////////////////////////////////////////
-// Functions
-////////////////////////////////////////////////////////////////
+export const defaultValues = Object.values(TSType).reduce((acc, type) => {
```
hdinia: I don't see anything wrong with putting it in the component file at top. but as you prefer

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/Fields/TypeConfigFields.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { useWatch } from "react-hook-form";
+
+interface Props {
+  type: TTSType;
+}
+
+function TypeConfigFields({ type }: Props) {
+  const { control } = useFormContextPlus<TSConfigValues>();
+
+  const isEnable = useWatch({
```
hdinia: `isEnabled`

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
```diff
@@ -15,15 +15,15 @@
 import { TSType } from "@/services/api/studies/timeseries/constants";
 import type { TSTypeConfig, TSConfigDTO, TTSType } from "@/services/api/studies/timeseries/types";
 
-export type TSConfigValues = Record<TTSType, TSTypeConfig & { stochasticTsStatus: boolean }>;
+export type TSConfigValues = Record<TTSType, TSTypeConfig & { enable: boolean }>;
```
hdinia: `enabled` (same for the other occurrences to not repeat comments)

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
skamril: > because it can make the code harder to debug and read, so you have to consider the tradeoff depending on the context.

Mutation in `reduce` is to be banned only when there is no initial value, to prevent side effects.

webapp/src/services/api/studies/timeseries/types.ts
skamril: > look at the type signature

Why are you talking about the type signature? I use `F.Exact` so the complexity is not visible.

> it's barely understandable, and not necessary because the enum object dictates already the form fields

Without `F.Exact`:

`setTimeSeriesConfig(CONFIG_WITH_EXTRA)` => No TS error, but the API throw an error

With `F.Exact`:

`setTimeSeriesConfig(CONFIG_WITH_EXTRA)` => TS throw an error

>  The only case where it's justified is for function parameters, for the rest it's redundant (like the "I" prefix for Interfaces). Same for suffix Type I used it's useless we have to drop both

I didn't say that adding a prefix is a general rule, I said it can be used to prevent duplicate name.

>  I let you do your own research on that topic if you dont believe me

This is a subjective point, you can find several different opinions.
The important point is that the `T` prefix is ​​not unknown among typescript developers, and it's not used only for function parameters. In addition, I have already used it for the same problem of name duplication, so the best approach is to maintain consistency and not change my mind constantly.

But to avoid constantly coming back to this point, I can change with this naming: `MyEnum` and `MyEnumValue`.
But keeping the same name is to be banned.

sylvlecl (dismissed):

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
skamril: Yes but I create a file to share the type, so I add it too

webapp/src/components/App/Singlestudy/explore/Configuration/TimeSeriesManagement/utils.ts
skamril: It's for the field name, so it's good

hdinia (dismissed):

hdinia (approved):

PR #2304, opened 2025-01-20
ci(github): add PR title lint job

skamril:


MartinBelthle (commented):
Seems good. I just don't understand the change inside commitlint.js
Also it's a bit odd to use the gh action of a random dude but doesn't really matter

skamril:
> Seems good. I just don't understand the change inside commitlint.js Also it's a bit odd to use the gh action of a random dude but doesn't really matter

I convert `commitlint.config.js` to a CommonJS syntax, because ES Modules is not supported by the new action.
The action doesn't use any permissions, so it's not a problem for the security.

sylvlecl (approved):

PR #2305, opened 2025-01-20
fix(matrix): allow odd matrix format when importing

MartinBelthle:
The csv sniffer is not reliable and can detect the wrong delimiter. So we can't trust it.
We have to check ourselves. It's probably a bit longer but at least we don't have failures anymore

### DIFF
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -13,7 +13,6 @@
 import base64
 import collections
 import contextlib
-import csv
 import http
 import io
 import logging
@@ -25,6 +24,7 @@ from pathlib import Path, PurePosixPath
 from uuid import uuid4
 
 import numpy as np
+import numpy.typing as npt
 import pandas as pd
 from antares.study.version import StudyVersion
 from fastapi import HTTPException, UploadFile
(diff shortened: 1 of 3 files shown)


antarest/study/service.py
```diff
@@ -197,6 +197,20 @@ def get_disk_usage(path: t.Union[str, Path]) -> int:
...
+    str_data = data.decode("utf-8")
+    if not str_data:
+        return np.zeros(shape=(0, 0))
+    for delimiter in [",", ";", "\t"]:
+        with contextlib.suppress(Exception):
+            df = pd.read_csv(io.BytesIO(data), delimiter=delimiter, header=None).replace(",", ".", regex=True)
+            df = df.dropna(axis=1, how="all")  # We want to remove columns full of NaN at the import
+            matrix = df.to_numpy(dtype=np.float64)
+            return matrix
+    raise ValueError("Could not import the matrix")
```
TheoPascoli: Do we have more informations to make the error message more explicit ?

antarest/study/service.py
```diff
@@ -197,6 +197,20 @@ def get_disk_usage(path: t.Union[str, Path]) -> int:
...
 
 
+def _imports_matrix_from_bytes(data: bytes) -> npt.NDArray[np.float64]:
+    """Tries to convert bytes to a numpy array when importing a matrix"""
+    str_data = data.decode("utf-8")
+    if not str_data:
+        return np.zeros(shape=(0, 0))
+    for delimiter in [",", ";", "\t"]:
+        with contextlib.suppress(Exception):
+            df = pd.read_csv(io.BytesIO(data), delimiter=delimiter, header=None).replace(",", ".", regex=True)
```
TheoPascoli, line 208: is there a way to load the matrix only once ? Maybe we can do something to "detect" the delimiter to prevent three openings

TheoPascoli (commented):
Looks good, just two questions

antarest/study/service.py
TheoPascoli: I just realised that it is exactly what was done before ...

MartinBelthle: Yep indeed. And it's because it wasn't reliable that I had to do this `for` loop

antarest/study/service.py
MartinBelthle: Unfortunately, inside the calling method, I don't even know on which study I'm applied

MartinBelthle: But I can perhaps catch it higher in the code to raise a different error. For the moment we have a HTTP 500 saying Unexpected server error: Could not import the matrix

MartinBelthle: I added a home-made exception to raise a 422 with a clearer exception name  (+ a test). So it's slightly better

TheoPascoli: Ok thanks !

TheoPascoli (approved):

PR #2310, opened 2025-01-22
Draft: custom parsing, serialization and matching in INI file nodes

sylvlecl:
Trying to rationalize a little on top of https://github.com/AntaresSimulatorTeam/AntaREST/pull/2283 :+1: 

- introduce custom parsers and serializers in ini readers and writers
- introduce a custom way of matching sections in INI file nodes ...

For cluster files, we should have cluster **names** as section names, to be consistent with simulator parsing.
However, in past versions we have stored versions with either ID or upper case IDs in those section names.
Therefore we need to have a strategy to handle compatibility for those studies. When matching, we should be able to convert to ID and perform matching on ID, since they need to be unique anyway.


sylvlecl:
Closing in favor of more targetted PRs.

PR #2313, opened 2025-01-23
feat(ui-matrix): add matrix selection stats

hdinia:


webapp/src/components/common/DataGrid.tsx
```diff
@@ -38,9 +38,12 @@ type RowMarkers =
 type RowMarkersOptions = Exclude<RowMarkers, string>;
```
skamril, line 38: You only need to do this:

```typescript
export interface DataGridProps extends Omit<DataEditorProps, "rowMarkers" | "gridSelection"> {
  rowMarkers?: RowMarkers;
  enableColumnResize?: boolean;
}

// ...

const handleGridSelectionChange = (newSelection: GridSelection) => {      
   // ...

   setSelection(newSelection);

   onGridSelectionChange?.(newSelection);
  };
```

You doesn't need to add `gridSelection` in props.

webapp/src/components/common/Matrix/components/MatrixStats/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+    count: number;
+    sum: number;
+    average: number;
+    min: number;
+    max: number;
+  } | null;
+}
+
+function MatrixStats({ stats }: MatrixStatsProps) {
+  if (!stats) {
```
skamril, line 29: Better to add all logic here:

```typescript

interface MatrixStatsProps {
  data: number[][];
  gridSelection: GridSelection;
  gridToData: (item: Item) => Item | null;
}

function MatrixStats({ data, gridSelection, gridToData }: MatrixStatsProps) {
  const stats = useSelectionStats({
    data,
    selection: gridSelection,
    gridToData,
  });

  // ...
  
}
```

skamril (changes requested):

webapp/src/components/common/DataGrid.tsx
```diff
@@ -241,16 +242,7 @@ function DataGrid(props: DataGridProps) {
         // Prevent selecting the row marker column
         if (newSelection.columns.hasIndex(0)) {
           // TODO find a way to have the rows length to select all the rows like other row markers
-          // setSelection({
```
skamril: Keep this comments, is linked with the TODO

webapp/src/components/common/DataGrid.tsx
hdinia: done thanks

webapp/src/components/common/Matrix/components/MatrixStats/index.tsx
hdinia: this has pros and cons, at that moment I prefer to keep it a presentational component

webapp/src/components/common/DataGrid.tsx
hdinia: This comment look like a personal note rather than something worth documenting, so it should be removed as it's unused, and adds unnecessary clutter.

skamril (approved):

PR #2316, opened 2025-01-27
feat(study): normalize study path

smailio:
The front end can't handle windows path, the desktop version on windows won't display the study hierarchy correctly. 

The front end relies on the folder field of Study. This PR adds a validator to that field that normalize any path, including windows path to a posix path.


tests/storage/business/test_raw_study_service.py
```diff
@@ -66,9 +66,6 @@ def test_get(tmp_path: str, project_path) -> None:
     data = {"titi": 43}
     sub_route = "settings"
 
-    path = path_study / "settings"
-    key = "titi"
-
```
smailio, line 71: These variables were unused

tests/storage/business/test_raw_study_service.py
```diff
@@ -100,7 +97,6 @@ def test_get_cache(tmp_path: str) -> None:
     path_study.mkdir()
     (path_study / "settings").mkdir()
     (path_study / "study.antares").touch()
-    path = path_study / "settings"
```
smailio, line 100: This was unused

antarest/study/model.py
```diff
@@ -288,6 +289,23 @@ def __eq__(self, other: t.Any) -> bool:
...
+        return normalize_path(folder)
+
+
+def normalize_path(path: str) -> str:
+    """
+    Turns any path including a windows path (with \ separator) to a posix path (with / separator).
+    """
+    if not path:
+        return path
+    windows_path = PureWindowsPath(path)
```
MartinBelthle: Why not instanciating simply a PurePath ? I wonder what this achieves when given an unix formatted path

antarest/study/model.py
```diff
@@ -288,6 +289,23 @@ def __eq__(self, other: t.Any) -> bool:
...
 
+    @validates("folder")
+    def validate_folder(self, key, folder) -> str:
+        """
+        We want to store the path in posix format in the database, even on windows.
+        """
+        return normalize_path(folder)
+
+
+def normalize_path(path: str) -> str:
```
MartinBelthle: The folder attribute being Nullable we should probably type the arg as an optional here. Same for the validation method. And that would justify the if right after

MartinBelthle (changes requested):
Before merging we should generate a Desktop version to give to Alexander for him to try it

antarest/study/model.py
smailio: Yeah you're right I thought it would work in all cases but if I give it a posix path with back slash in it, it produce an unexpected behavior.

antarest/study/model.py
```diff
@@ -288,6 +289,23 @@ def __eq__(self, other: t.Any) -> bool:
     def to_json_summary(self) -> t.Any:
         return {"id": self.id, "name": self.name}
 
+    @validates("folder")  # type: ignore
```
smailio, line 292: sqlachemu doesn't provide a type for this decorator

antarest/study/model.py
```diff
@@ -288,6 +289,23 @@ def __eq__(self, other: t.Any) -> bool:
...
+
+
+def normalize_path(path: t.Optional[str]) -> t.Optional[str]:
+    """
+    Turns any path including a windows path (with \ separator) to a posix path (with / separator).
+    """
+    if not path:
+        return path
+    pure_path = PurePath(path)
+    return pure_path.as_posix()
```
MartinBelthle, line 307: Minor: all this could be simplfied in one line:
`return PurePath(path).as_posix() if path else path`

MartinBelthle (approved):

PR #2317, opened 2025-01-27
fix: fix a bug that would occurs after deleting multiple bc

TheoPascoli:
We need to reset the index so it always starts from zero

### DIFF
--- a/antarest/study/storage/variantstudy/model/command/remove_multiple_binding_constraints.py
+++ b/antarest/study/storage/variantstudy/model/command/remove_multiple_binding_constraints.py
@@ -61,8 +61,11 @@ class RemoveMultipleBindingConstraints(ICommand):
             if binding_constraints[key].get("id") in self.ids:
                 deleted_binding_constraints.append(binding_constraints.pop(key))
 
+        # BC dict should start at index 0
+        new_binding_constraints = {str(i): value for i, value in enumerate(binding_constraints.values())}
+
         study_data.tree.save(
-            binding_constraints,
+            new_binding_constraints,
             ["input", "bindingconstraints", "bindingconstraints"],
         )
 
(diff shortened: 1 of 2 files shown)


antarest/study/storage/variantstudy/model/command/remove_multiple_binding_constraints.py
```diff
@@ -61,8 +61,11 @@ def _apply(self, study_data: FileStudy, listener: t.Optional[ICommandListener] =
             if binding_constraints[key].get("id") in self.ids:
                 deleted_binding_constraints.append(binding_constraints.pop(key))
 
+        # BC dict should start at index 0
+        new_binding_constraints = {str(i): value for i, value in enumerate(binding_constraints.values())}
```
MartinBelthle, line 65: Could you add a test to reproduce Alvin bug ?

MartinBelthle (changes requested):

antarest/study/storage/variantstudy/model/command/remove_multiple_binding_constraints.py
TheoPascoli: It does not seem to be easy to check the dict id from the tests

MartinBelthle: From what I understand, we saved keys not starting at 0 inside the ini. So perhaps we could check that after deleting the constraint corresponding to the key 0 we still have the key 0 inside the ini (via the GET /raw inside integration tests or just the node / file path in an unit test)

TheoPascoli: Indeed, I test it with the raw endpoint

MartinBelthle (approved):

PR #2320, opened 2025-01-28
fix(ui-tablemode): adjust the size of the column with the initial data

skamril:
ANT-1217

### DIFF
--- a/webapp/src/components/common/DataGrid.tsx
+++ b/webapp/src/components/common/DataGrid.tsx
@@ -19,18 +19,21 @@ import {
   type EditListItem,
   type GridSelection,
   type DataEditorProps,
+  type GridCell,
 } from "@glideapps/glide-data-grid";
 import "@glideapps/glide-data-grid/dist/index.css";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useState } from "react";
 import { voidFn } from "@/utils/fnUtils";
 import { darkTheme } from "./Matrix/styles";
+import { useUpdateEffect } from "react-use";
 
 interface StringRowMarkerOptions {
   kind: "string" | "clickable-string";
   getTitle?: (rowIndex: number) => string;
+  width?: number;
 }
 
-type RowMarkers =
+export type RowMarkers =
   | NonNullable<DataEditorProps["rowMarkers"]>
   | StringRowMarkerOptions["kind"]
   | StringRowMarkerOptions;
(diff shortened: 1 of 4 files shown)


webapp/src/utils/domUtils.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+export function measureTextWidth(text: string, font?: string) {
+  const canvas = document.createElement("canvas");
+  const context = canvas.getContext("2d");
+  if (context) {
+    context.font = font || getCssFont();
+    return context.measureText(text).width;
+  }
+  return 0;
+}
```
hdinia: update `measureTextWidth` to use a factory function instead:

```ts
/**
 * Uses `canvas.measureText` to compute and return the width of the specified text,
 * using the specified canvas font if defined (use font from the "body" otherwise).
 *
 * @see https://stackoverflow.com/a/21015393
 *
 * @returns The width of the text in pixels.
 */
export function createTextMeasurer() {
  const canvas = document.createElement("canvas");
  const context = canvas.getContext("2d");
  const font = getFont();

  if (!context) {
    return (text: string) => 0;
  }

  context.font = font;

  return (text: string) => context.measureText(text).width;
}
```

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -107,16 +107,49 @@ function DataGridForm<TData extends Data>({
...
+            ? column.width
+            : Math.max(
+                measureTextWidth(column.title),
+                ...rowNames.map((rowName) =>
+                  measureTextWidth(defaultData[rowName][column.id].toString()),
+                ),
+              ),
+      })),
+    [columns, defaultData, rowNames],
+  );
```
hdinia, line 128: here a canvas is re-created at each iteration in the Math.max(), it's expensive

```ts
  const columnsWithAdjustedSize = useMemo(() => {
    const measureText = createTextMeasurer();

    return columns.map((column) => {
      if ("width" in column) {
        return column;
      }

      const maxWidth = Math.max(
        measureText(column.title),
        ...rowNames.map((rowName) => measureText(defaultData[rowName][column.id].toString())),
      );

      return { ...column, width: maxWidth };
    });
  }, [columns, defaultData, rowNames]);
  ```
  
  see comment about `domUtils.ts` file for the `createTextMeasurer()` util using the factory function pattern to avoid re-creation of canvas

File: webapp/src/components/common/DataGridForm.tsx [hunk 3]
```diff
@@ -107,16 +107,49 @@ function DataGridForm<TData extends Data>({
...
+
+  const rowMarkers = useMemo<RowMarkers>(
+    () =>
+      rowMarkersFromProps || {
+        kind: "clickable-string",
+        getTitle: (index) => rowNames[index],
+        width: Math.max(...rowNames.map((name) => measureTextWidth(name))),
+      },
+    [rowMarkersFromProps, rowNames],
+  );
```
hdinia, line 140:
```ts
const rowMarkers = useMemo<RowMarkers>(() => {
    if (rowMarkersFromProps) {
      return rowMarkersFromProps;
    }

    const measureText = createTextMeasurer();
    const maxWidth = Math.max(...rowNames.map(measureText));

    return {
      kind: "clickable-string",
      getTitle: (index) => rowNames[index],
      width: maxWidth,
    };
  }, [rowMarkersFromProps, rowNames]);
```

webapp/src/components/common/DataGrid.tsx
```diff
@@ -78,24 +80,26 @@ function DataGrid(props: DataGridProps) {
   // Add a column for the "string" row markers if needed
   useEffect(() => {
     setColumns(
-      isStringRowMarkers ? [{ id: "", title: "" }, ...columnsFromProps] : columnsFromProps,
+      isStringRowMarkers
+        ? [{ id: "", title: "", width: rowMarkersOptions.width }, ...columnsFromProps]
+        : columnsFromProps,
     );
-  }, [columnsFromProps, isStringRowMarkers]);
+  }, [columnsFromProps, isStringRowMarkers, rowMarkersOptions.width]);
```
hdinia: here about the useEffect, not sure about it but seems not necessary no ?

```ts
  const columns = useMemo(() => {
    if (isStringRowMarkers) {
      return [{ id: "", title: "", width: rowMarkersOptions.width }, ...columnsFromProps];
    }

    return columnsFromProps;
  }, [columnsFromProps, isStringRowMarkers, rowMarkersOptions.width]);
  ```

hdinia (changes requested):

webapp/src/utils/domUtils.ts
skamril: You forget `font` argument, otherwise I create the canvas element outside, it's enough

webapp/src/components/common/DataGrid.tsx
skamril: `setColumns` is used for columns resizing

PR #2322, opened 2025-01-29
feat(area): move area command creation

TheoPascoli:
add a command to move an area on the map. Replace multiple UpdateConfig commands


antarest/study/storage/variantstudy/model/command/move_area.py
```diff
@@ -0,0 +1,96 @@
...
+    # ===================
+
+    command_name: CommandName = CommandName.MOVE_AREA
+    version: int = 1
+
+    # Command parameters
+    # ==================
+
+    area_name: str
+    new_area_parameters: t.Dict[str, t.Any]
```
MartinBelthle: I think we should use a pydantic object here (UpdateAreaUi seems good)

antarest/study/storage/variantstudy/model/command/move_area.py
```diff
@@ -0,0 +1,96 @@
...
+    layer: str
+
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> t.Tuple[CommandOutput, t.Dict[str, t.Any]]:
+        return (
```
MartinBelthle: Style: All in one line ?

antarest/study/storage/variantstudy/model/command/move_area.py
```diff
@@ -0,0 +1,96 @@
...
+            CommandOutput(
+                status=True,
+                message=f"area '{self.area_name}' updated",
```
MartinBelthle: The message could specify we updated it's `ui` parameters

antarest/study/storage/variantstudy/model/command/move_area.py
```diff
@@ -0,0 +1,96 @@
...
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput
+from antarest.study.storage.variantstudy.model.command.icommand import MATCH_SIGNATURE_SEPARATOR, ICommand
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+class MoveArea(ICommand):
+    """
+    Command used to move an area in the study.
```
MartinBelthle: Could be nice to say that moving means moving it inside the map or updating its ui parameters

MartinBelthle (changes requested):
We could add unit test for this command as we only have one currently.

antarest/study/storage/variantstudy/model/command/move_area.py
TheoPascoli: I cannot import UpdateAreaUi because of circular imports. However, I can extract all the area objects to a model package (like the links)

antarest/study/storage/variantstudy/model/command/move_area.py
```diff
@@ -24,7 +24,7 @@
 
 class MoveArea(ICommand):
     """
-    Command used to move an area in the study.
+    Command used to move an area inside the map and to update its UI.
```
maugde: Thanks for updating the comments

maugde (commented):
It seems good to me

antarest/study/business/area_management.py
```diff
@@ -10,262 +10,43 @@
...
-from antarest.study.business.all_optional_meta import all_optional_model, camel_case_model
+from antarest.study.business.model.area_model import (
+    AreaCreationDTO,
+    AreaInfoDTO,
+    AreaOutput,
+    AreaType,
+    ClusterInfoDTO,
+    LayerInfoDTO,
+    UpdateAreaUi,
+    _get_area_layers,
```
MartinBelthle: I think we should keep these privates methods inside area_management

antarest/study/business/model/area_model.py
```diff
@@ -0,0 +1,247 @@
...
+    id: str
+    name: str
+    areas: t.List[str]
+
+
+class UpdateAreaUi(AntaresBaseModel, extra="forbid", populate_by_name=True):
+    """
+    DTO for updating area UI
+
+    Usage:
```
MartinBelthle: Should we keep this doc ? Seems really useless to me and makes the file harder to read.

MartinBelthle (changes requested):
Could be nice to add a unit test on the command

MartinBelthle (commented):
I encountered a "bug" which was already present before your dev. When updating the color of an area inside the layer1, it also brings back its position to the layer 0 position. I don't think we should do that.

We should investigate and if we can change this, fix it and add a test inside the test_area.py file you introduced

TheoPascoli:
We can merge this. A ticket adressing the bug will be created

TheoPascoli:
We can merge this. I will create a ticket adressing the bug that might comes from the front

antarest/study/business/model/area_model.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class AreaCreationDTO(AntaresBaseModel):
+    name: str
+    type: AreaType
+    metadata: t.Optional[PatchArea] = None
+    set: t.Optional[t.List[str]] = None
+
+
+# review: is this class necessary?
```
maugde: Is it still wip ?

TheoPascoli: I removed the comment because it is still used. Thanks for pointing out

maugde (dismissed):

antarest/study/business/area_management.py
```diff
@@ -680,73 +505,17 @@ def update_area_metadata(
         )
 
     def update_area_ui(self, study: Study, area_id: str, area_ui: UpdateAreaUi, layer: str = "0") -> None:
```
MartinBelthle: You can remove the ` = "0"` as we already did this transformation inside the blueprint

antarest/study/storage/variantstudy/model/command/update_area_ui.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> t.Tuple[CommandOutput, t.Dict[str, t.Any]]:
+        return CommandOutput(status=True, message=f"area '{self.area_id}' UI updated"), {}
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: t.Optional[ICommandListener] = None) -> CommandOutput:
+        current_area = study_data.tree.get(["input", "areas", self.area_id, "ui"])
+
+        if self.layer == "0":
```
MartinBelthle, line 52: This function doesn't work correctly. We're not using the `layer_xxx ` attributes of the class UpdateAreaUi
Here's an example of what the front sends when changing the color of an area inside a layer.
```
{"action":"update_area_ui",
"args":{
"area_id":"area_1",
"area_ui":{
"x":-15,
"y":194,
"color_rgb":[171,255,0],
"layer_x":{"0":-15,"1":-257},
"layer_y":{"0":194,"1":-125},
"layer_color":{"0":"255, 0, 131","1":"230, 108, 44"}},
"layer":"1"
}
}
```
The attributes x and y always correspond to the layer 0 info but the color_rgb corresponds to the laye we're moving. It doesn't make any sense. We'll have to fix it inside another PR that deals with both the front and this little method

PR #2323, opened 2025-01-29
fix(commands): fix error on command notifications

maugde:
fix [ANT-2511](https://gopro-tickets.rte-france.com/browse/ANT-2511)


antarest/study/storage/variantstudy/variant_command_generator.py
```diff
@@ -80,6 +88,10 @@ def _generate(
                 )
                 logger.error(output.message, exc_info=e)
 
+            if not command_block_dict.get(cmd.command_id):
```
smailio: Can't you use `cmd.command_id in command_block_dict` ? 

Also this means that you can meet several commands with the same id in the list ? Then the order matters to count correctly the `command_block_index` no ?

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -432,3 +447,115 @@ def utcnow(cls) -> datetime.datetime:
...
+        simple_matrix_service,
+        patch_service,
+    ):
+        """
+        Set Up:
+            Create a raw study, create a variant study, add some commands to it.
+        Test:
+            - Check if the generated commands does not return an error with the notifier.
+        """
+        # variant_study_service.generator = VariantCommandGenerator(variant_study_service.study_factory)
```
smailio: delete comment

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -432,3 +447,115 @@ def utcnow(cls) -> datetime.datetime:
...
+        db.session.add(user)
+        db.session.commit()
+        jwt_user = Mock(spec=JWTUser, id=user.id, impersonator=user.id, is_site_admin=Mock(return_value=True))
+        params = Mock(spec=RequestParameters, user=jwt_user)
+
+        group = Group(id="test", name="test")
+        db.session.add(group)
+        db.session.commit()
+
+        path = tmp_path / "testing-study"
```
smailio: maybe add a comment before each test input to describe what's specefic/ relevant about them in this test ?

antarest/study/storage/variantstudy/variant_command_generator.py
```diff
@@ -62,12 +63,19 @@ def _generate(
         study_id = "-" if metadata is None else metadata.id
 
         # flatten the list of commands
-        all_commands = [command for command_batch in commands for command in command_batch]
-
+        # the result here is a list of all commands
+        # note that has a list of arguments is added as a command too to the final command list
```
smailio: a word is missing

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -432,3 +447,120 @@ def utcnow(cls) -> datetime.datetime:
...
+            new_area = study_service.create_area(variant_study.id, area_creation_dto, params)
+            study_service.general_manager.set_field_values(variant_study, general_form_fields)
+
+        # at this point, the command must have a total of 3 command block and more regarding all commands
+        assert len(variant_study_service.get_commands(variant_study.id, params)) == 3
+
+        # the call to generate_study_config must not add
+        # an error log such as "fail to notify command ..."
+        with caplog.at_level(logging.ERROR, logger=logger.name):
+            result = variant_study_service.generate_study_config(variant_study.id, params)
```
MartinBelthle: You don't use the `result` variable, you can remove it

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -432,3 +447,120 @@ def utcnow(cls) -> datetime.datetime:
...
+            selection_mode=True,
+            year_by_year=True,
+            simulation_synthesis=False,
+            mc_scenario=False,
+            filtering=False,
+            geographic_trimming=False,
+            thematic_trimming=False,
+        )
+        with current_user_context(token=jwt_user):
+            new_area = study_service.create_area(variant_study.id, area_creation_dto, params)
```
MartinBelthle: Same, you can remove the variable `new_area`

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -432,3 +447,120 @@ def utcnow(cls) -> datetime.datetime:
...
+        ## create an area
+        area_manager = AreaManager(Mock(spec=StudyStorageService), Mock(spec=StudyMetadataRepository))
+
+        area_creation_dto = AreaCreationDTO(
+            name="area_test",
+            type=AreaType.AREA,
+        )
+
+        ## add a command with multiple sub commands
+        general_form_fields = GeneralFormFields(
```
MartinBelthle: Don't think you need to fill the whole object, i reproduced the behavior with just this:
`general_form_fields = GeneralFormFields(mode=Mode.ECONOMY,
            nb_years=3)`

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -432,3 +447,120 @@ def utcnow(cls) -> datetime.datetime:
...
+        user = User(id=32, name="Tester")
+        db.session.add(user)
+        db.session.commit()
+
+        # Create a JWT to simulate the user permissions on the studies
+        jwt_user = Mock(spec=JWTUser, id=user.id, impersonator=user.id, is_site_admin=Mock(return_value=True))
+        params = Mock(spec=RequestParameters, user=jwt_user)
+
+        # Add a group that links the user and the studies
+        group = Group(id="test", name="test")
```
MartinBelthle: You don't have to introduce a group

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -432,3 +447,120 @@ def utcnow(cls) -> datetime.datetime:
...
+            config=Mock(spec=Config),
+        )
+
+        # add some commands
+        ## add comments
+        comments_dto = CommentsDto(comments="add some comments")
+        study_service.edit_comments(variant_study.id, comments_dto, params)
+
+        ## create an area
+        area_manager = AreaManager(Mock(spec=StudyStorageService), Mock(spec=StudyMetadataRepository))
```
MartinBelthle: You can remove this line as you don't use it

antarest/study/storage/variantstudy/variant_command_generator.py
```diff
@@ -90,7 +103,7 @@ def _generate(
             results.details.append(detail)
 
             if notifier:
-                notifier(index - 1, output.status, output.message)
+                notifier(command_block_dict[cmd.command_id], output.status, output.message)
```
MartinBelthle: Perhaps I'm missing something, but from my understanding if you have 3 "macro" commands and 20 "micro" ones, you're still gonna notify 20 times. As we just want to notify 3 times, i think there's an issue here

MartinBelthle (changes requested):

tests/study/storage/variantstudy/test_variant_study_service.py
```diff
@@ -429,6 +443,101 @@ def utcnow(cls) -> datetime.datetime:
...
+            name="area_test",
+            type=AreaType.AREA,
+        )
+
+        ## add a command with multiple sub commands
+        # we give only required fields and ignore mypy type check
+        general_form_fields = GeneralFormFields(
+            mode=Mode.ECONOMY,
+            nb_years=3,
+        )  # type: ignore
```
MartinBelthle: You can remove this `type: ignore` and put this in one line

antarest/study/storage/variantstudy/variant_command_generator.py
```diff
@@ -61,13 +62,21 @@ def _generate(
         logger.info("Applying commands")
         study_id = "-" if metadata is None else metadata.id
 
-        # flatten the list of commands
-        all_commands = [command for command_batch in commands for command in command_batch]
+        # Flatten the list of commands
+        # The result here is a list of all commands
+        # Note that each command can have a list of arguments (sub commands) that are added as commands
```
MartinBelthle: Minor: This msg is not really clear to me. I think the other comment you wrote after this is enough and you can keep the comment as it was

antarest/study/storage/variantstudy/variant_command_generator.py
```diff
@@ -61,13 +62,21 @@ def _generate(
...
+        # so len(all_commands) >= len(commands)
+        all_commands: List[ICommand] = list(itertools.chain.from_iterable(commands))
 
         # Prepare the stopwatch
         cmd_notifier = CmdNotifier(study_id, len(all_commands))
         stopwatch.reset_current()
 
+        # since we need the commands without their sub commands ONLY for the notifier, we'll use
+        # a dictionary to associate main commands id with a boolean value
+        command_block_dict: Dict[Optional[uuid.UUID], int] = {}
```
MartinBelthle: You could just use a set here I think and check if the value is inside the set. Also it would simplify your if as you don't have to check as the `.add` function works even if the value is already in the set

MartinBelthle (changes requested):

antarest/study/storage/variantstudy/variant_command_generator.py
maugde: The set wouldn't have any information about the command index nor if the command was already used by notifier. My current proposition isn't correct too. I'm thinking of a dict[uuid, tuple[bool, int]]. But it's perhaps too complex in this situation

antarest/study/storage/variantstudy/__init__.py
```diff
@@ -9,3 +9,6 @@
 # SPDX-License-Identifier: MPL-2.0
 #
 # This file is part of the Antares project.
+import typing as t
+
+CommandNotifier = t.Callable[[int, bool, str], None]
```
sylvlecl: Great, thanks for this

antarest/study/storage/variantstudy/variant_command_generator.py
```diff
@@ -89,8 +102,9 @@ def _generate(
             }
             results.details.append(detail)
 
-            if notifier:
-                notifier(index - 1, output.status, output.message)
+            if notifier and cmd.command_id not in command_block_set:
+                notifier(command_block_dict[cmd.command_id], output.status, output.message)
```
sylvlecl: @maugde @MartinBelthle : 
Actually it's weird to notify only on the first command of a block.
In particular, we will notify with the status of this first command, but not for subsequent commands:
so for example if the first command succeeds and the second one fails, it will be invisible from the listener.

In the end I would propose to change the scope of this work:
I think the notification mechanism is actually not used (it just sends an event to the event bus, but this event type is never used for now. I cannot see a usage in the near future.

--> I propose to simply delete this notifier and the associated event type

sylvlecl (approved):
Thanks!

PR #2325, opened 2025-01-30
fix(study-tree-ui): fix folder loading

smailio:
Add item loading mechanic. Currently the study tree display the folders from the initial study tree, and then display folder loaded dynamically if any. This is annonying for the user as a folder can move at the time he's clicking on it. 

Replace onClickNode by onItemExpansionToggle. The onClick is only applied to the text part of the tree item and not the arrow part. This is confusing for the user as the subfolders aren't loaded when the user click on an arrow, and the user is stuck with a loading message.

Disable scan on default folder as this makes no sense. The front end call the api with invalid arguments which just returns an error. 

Add a useUpdateEffect in  StudyTree/index.tsx, this is done to synchronize with updates from the backend. We also need to add new merge functions to handle this case.


webapp/src/components/App/Studies/StudyTree/StudyTreeNode.tsx
```diff
@@ -12,51 +12,48 @@
...
   studyTreeNode,
   parentId,
+  itemsLoading,
   onNodeClick,
 }: StudyTreeNodeProps) {
-  const isLoadingFolder = studyTreeNode.hasChildren && studyTreeNode.children.length === 0;
   const id = parentId ? `${parentId}/${studyTreeNode.name}` : studyTreeNode.name;
-
+  const isLoadingFolder = itemsLoading.includes(id);
+  const hasUnloadedChildern = studyTreeNode.hasChildren && studyTreeNode.children.length === 0;
```
hdinia, line 29: 🕵🏼 typo `hasUnloadedChildren`

webapp/src/components/App/Studies/StudiesList/index.tsx
```diff
@@ -87,6 +88,9 @@ function StudiesList(props: StudiesListProps) {
   const [selectionMode, setSelectionMode] = useState(false);
   const [confirmFolderScan, setConfirmFolderScan] = useState(false);
   const [isRecursiveScan, setIsRecursiveScan] = useState(false);
+  const isInDefaultWorkspace = !!folder && folder.startsWith(DEFAULT_WORKSPACE_PREFIX);
+  const isRootFolder = folder === ROOT_FOLDER_NAME;
+  const scanDisabled: boolean = isInDefaultWorkspace || isRootFolder;
```
hdinia, line 93: the type is inferred by TS compiler you dont need the `:boolean` assertion here

webapp/src/components/App/Studies/StudyTree/StudyTreeNode.tsx
```diff
@@ -12,51 +12,48 @@
...
   const sortedChildren = useMemo(
-    () => R.sortBy(R.prop("name"), studyTreeNode.children),
+    () => R.sortBy(R.compose(R.toLower, R.prop("name")), studyTreeNode.children),
     [studyTreeNode.children],
   );
 
-  if (isLoadingFolder) {
+  // Either the user clicked on the folder and we need to show the folder is loading
+  // Or the explorer api says that this folder has children so we need to load at least one element
+  // so the arrow to explanse the element is displayed which indicate to the user that this is a folder
```
hdinia, line 37: // Show loading indicator either when the folder is actively loading
  // or when it has children that haven't been loaded yet

webapp/src/components/App/Studies/StudyTree/utils.ts
```diff
@@ -275,3 +275,47 @@ export async function fetchAndInsertWorkspaces(studyTree: StudyTreeNode): Promis
...
+  right: StudyTreeNode[],
+): Array<[StudyTreeNode, StudyTreeNode]> {
+  return left.reduce<Array<[StudyTreeNode, StudyTreeNode]>>((acc, leftNode) => {
+    const matchedRightNode = right.find((rightNode) => rightNode.name === leftNode.name);
+    if (matchedRightNode) {
+      acc.push([{ ...leftNode }, { ...matchedRightNode }]);
+    }
+    return acc;
+  }, []);
+}
```
hdinia, line 321: The code works fine, but the docs and naming could be a bit reworked. Consider giving these a second pass to improve clarity:


```typescript
/**
 * Merges two study trees, preserving data from the left tree while updating 
 * with the structure of the right tree.
 *
 * @param leftTree - The source tree containing data to preserve
 * @param rightTree - The target tree providing the updated structure
 * @returns A new tree with the data from the left tree merged into the right tree structure
 */
export function mergeDeepRightStudyTree(leftTree: StudyTreeNode, rightTree: StudyTreeNode): StudyTreeNode {
  const nodesOnlyInLeft = leftTree.children.filter(
    (leftNode) => !rightTree.children.some((rightNode) => leftNode.name === rightNode.name)
  );
  
  const nodesOnlyInRight = rightTree.children.filter(
    (rightNode) => !leftTree.children.some((leftNode) => leftNode.name === rightNode.name)
  );
  
  const matchedNodePairs = innerJoin(leftTree.children, rightTree.children);
  const mergedMatchedNodes = matchedNodePairs.map(
    ([leftNode, rightNode]) => mergeDeepRightStudyTree(leftNode, rightNode)
  );
  
  const mergedChildren = [...nodesOnlyInLeft, ...mergedMatchedNodes, ...nodesOnlyInRight];
  
  return {
    ...rightTree,
    children: mergedChildren,
  };
}

/**
 * Joins two arrays of study tree nodes based on matching node names.
 *
 * @param leftNodes - Array of nodes from the left tree
 * @param rightNodes - Array of nodes from the right tree
 * @returns Array of matched node pairs [leftNode, rightNode]
 */
export function innerJoin(
  leftNodes: StudyTreeNode[],
  rightNodes: StudyTreeNode[],
): Array<[StudyTreeNode, StudyTreeNode]> {
  return leftNodes.reduce<Array<[StudyTreeNode, StudyTreeNode]>>((pairs, leftNode) => {
    const matchedRightNode = rightNodes.find((rightNode) => rightNode.name === leftNode.name);

    if (matchedRightNode) {
      pairs.push([{ ...leftNode }, { ...matchedRightNode }]);
    }

    return pairs;
  }, []);
}
```


to go further we could do the same for the tests about this code 
```typescript
test("mergeDeepRightStudyTree should correctly merge two trees with overlapping and unique nodes", () => {
  // Arrange
  const sourceTree: StudyTreeNode = {
    name: "root",
    path: "/",
    children: [
      { name: "A", path: "/A", children: [] },
      { name: "B", path: "/B", children: [] },
    ],
  };
  const targetTree: StudyTreeNode = {
    name: "root",
    path: "/",
    children: [
      { name: "A", path: "/A1", children: [] }, // Same name but different path
      { name: "C", path: "/C", children: [] },  // Unique to targetTree
    ],
  };

  // Act
  const mergedTree = mergeDeepRightStudyTree(sourceTree, targetTree);
  
  // Assert
  expect(mergedTree.children).toHaveLength(3);
  
  // Check that all expected nodes exist in the merged tree
  expect(mergedTree.children.some(child => child.name === "A")).toBeTruthy();
  expect(mergedTree.children.some(child => child.name === "B")).toBeTruthy();
  expect(mergedTree.children.some(child => child.name === "C")).toBeTruthy();
  
  // Verify that node A has the path from the target tree (right-side priority)
  const nodeA = mergedTree.children.find(child => child.name === "A");
  expect(nodeA?.path).toBe("/A1");
});
```

webapp/src/components/App/Studies/StudyTree/index.tsx
```diff
@@ -106,15 +118,29 @@ function StudyTree() {
...
   ////////////////////////////////////////////////////////////////
   // Event Handlers
   ////////////////////////////////////////////////////////////////
 
-  const handleTreeItemClick = async (itemId: string, studyTreeNode: StudyTreeNode) => {
+  // we need to handle both the expand event and the onClick event
+  // because the onClick isn't triggered when user click on arrow
+  // Also the expanse event isn't triggered when the item doesn't have any subfolder
+  // but we stil want to apply the filter on the clicked folder
+  const handleItemExpansionToggle = async (
```
hdinia, line 132: I suppose either the `async` is not needed, or the `await` missing in front of `updateTree`

hdinia (changes requested):

PR #2326, opened 2025-02-03
chore(commands): remove obsolete variant-related features

sylvlecl:
Remove  of unused command-related features:
 - the possibility to build a diff between variants
 - the possibility to build a list of commands to build a study from scratch
 - the possibility to revert commands

Reasons:
- Those features are not used, and probably full of bugs.
- their future usage in this shape is very unlikely.
  For example, for the diff feature:
  if we want a diff between studies, we should make it a separate feature from commands with its own diff model.
  Having a diff as commands does not seem very useful:
  we already have the target study, what will we use the output commands for ?
  If we want to apply commands on another study, we don't need a diff for this, the user just
  needs to structure its studies for this and copy/paste the commands he wants to copy.
- it hurts our capacity to maintain, refactor, and make evolutions in commands implementation


MartinBelthle (dismissed):
Minor: We can remove the parts of the code where we mentionned the cli.py file:
- `pyproject.toml` inside omit for coverage
- `AntaresWebLinux/Windows.spec`
- one line inside `architecture/introduction.md`

MartinBelthle (approved):

PR #2329, opened 2025-02-05
fix(explorer): normalize folder path

smailio:
Since we converted folder to Posix path Study we also need to convert paths returned by the explorer api.


MartinBelthle (changes requested):
Could be nice to add a test. Otherwise looks good to me

smailio:
> Could be nice to add a test. Otherwise looks good to me

I think it's not that relevant. To test this I'd have to patch pathlib dir so it returns a Windows path, then I could check this path is indeed converted to a Posix paths. But this case is already covered by the github action that runs tests on windows no ?

MartinBelthle:
> > Could be nice to add a test. Otherwise looks good to me
> 
> I think it's not that relevant. To test this I'd have to patch pathlib dir so it returns a Windows path, then I could check this path is indeed converted to a Posix paths. But this case is already covered by the github action that runs tests on windows no ?

Not sure to understand. From my understanding you changed the NonStudyFolder `path` attribute to always return a Unix path. So I think we should add a test that ensure no matter the OS we'll find a a UNix path (test that would fail inside the windows CI before your dev)

MartinBelthle (dismissed):

smailio:
> > > Could be nice to add a test. Otherwise looks good to me
> > 
> > 
> > I think it's not that relevant. To test this I'd have to patch pathlib dir so it returns a Windows path, then I could check this path is indeed converted to a Posix paths. But this case is already covered by the github action that runs tests on windows no ?
> 
> Not sure to understand. From my understanding you changed the NonStudyFolder `path` attribute to always return a Unix path. So I think we should add a test that ensure no matter the OS we'll find a a UNix path (test that would fail inside the windows CI before your dev)

Yeah you're right. 

I meant the case where the NonStudyFolder take as an input a WindowsPath is covered by the windows tests, but it's true that we need to have a test that would fail in the windows CI before this dev.

PR #2330, opened 2025-02-05
fix(ui-common): bump MUI to v6.4.3 to resolve Select list bug causing crashes

hdinia:


webapp/src/components/common/SelectMulti.tsx
```diff
@@ -82,7 +81,6 @@ function SelectMulti(props: Props) {
         multiple
         value={data}
         variant="filled"
-        placeholder={placeholder}
         onChange={handleChange}
         renderValue={tagsMode === true ? chipRender : checkboxRender}
       >
```
skamril, line 86: Keep `placeholder` prop:

```typescript
  >
    {placeholder && (
      <MenuItem disabled value="">
        <em>{placeholder}</em>
      </MenuItem>
    )}
    {list.map(({ id, name }) => (
```

cf. https://mui.com/material-ui/react-select/#placeholder

webapp/src/components/common/fieldEditors/SelectFE.tsx
```diff
@@ -85,7 +85,7 @@ function SelectFE(props: SelectFEProps) {
       fullWidth={fullWidth}
     >
       <InputLabel id={labelId}>{label}</InputLabel>
-      <Select {...selectProps} label={label} labelId={labelId}>
+      <Select {...selectProps} label={label} labelId={labelId} value={selectProps.value ?? ""}>
```
skamril: This is not necessary, and `defaultValue` cannot be used anymore with it

skamril (changes requested):

webapp/src/components/common/fieldEditors/SelectFE.tsx
hdinia: to turn off the warning

webapp/src/components/common/SelectMulti.tsx
hdinia: this prop does not exists anymore

skamril: I know, this is why the doc indicates to add the code I share to you to support placeholder

webapp/src/components/common/fieldEditors/SelectFE.tsx
skamril: Which warning?

skamril (approved):

PR #2331, opened 2025-02-06
build(all): bump pydantic and linting packages

TheoPascoli:
Upgrade pydantic to 2.9 because it introduces performances improvement. I pushed it to 2.10 because it does not contain breaking changes.
I also upgraded mypy, black and isort because why not

### DIFF
--- a/antarest/core/filetransfer/service.py
+++ b/antarest/core/filetransfer/service.py
@@ -79,9 +79,11 @@ class FileTransferManager:
                 Event(
                     type=EventType.DOWNLOAD_CREATED,
                     payload=download.to_dto(),
-                    permissions=PermissionInfo(owner=owner.impersonator)
-                    if owner
-                    else PermissionInfo(public_mode=PublicMode.READ),
+                    permissions=(
+                        PermissionInfo(owner=owner.impersonator)
+                        if owner
+                        else PermissionInfo(public_mode=PublicMode.READ)
+                    ),
                 )
             )
         return download
@@ -98,9 +100,11 @@ class FileTransferManager:
                 Event(
                     type=EventType.DOWNLOAD_READY,
                     payload=download.to_dto(),
-                    permissions=PermissionInfo(owner=download.owner)
-                    if download.owner
-                    else PermissionInfo(public_mode=PublicMode.READ),
+                    permissions=(
+                        PermissionInfo(owner=download.owner)
+                        if download.owner
+                        else PermissionInfo(public_mode=PublicMode.READ)
+                    ),
                 )
             )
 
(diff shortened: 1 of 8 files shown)


antarest/core/filetransfer/service.py
```diff
@@ -79,9 +79,11 @@ def request_download(
                 Event(
                     type=EventType.DOWNLOAD_CREATED,
                     payload=download.to_dto(),
-                    permissions=PermissionInfo(owner=owner.impersonator)
-                    if owner
-                    else PermissionInfo(public_mode=PublicMode.READ),
+                    permissions=(
```
MartinBelthle, line 82: Why does your IDE change this ?

requirements.txt
```diff
@@ -17,7 +17,7 @@ antares-timeseries-generation==0.1.7
 # and **manage their versions** for better control and to avoid unnecessary dependencies.
 fastapi~=0.110.3
 uvicorn[standard]~=0.30.6
-pydantic~=2.8.2
+pydantic~=2.10.0
```
MartinBelthle, line 20: In the title of the PR you said you'd use pydantic 2.9. You changed your mind ?

tests/variantstudy/model/command/test_remove_st_storage.py
```diff
@@ -81,16 +81,19 @@ def test_init__invalid_storage_id(self, recent_study: FileStudy, command_context
                 storage_id="?%$$",  # bad name
                 study_version=STUDY_VERSION_8_8,
             )
-        assert ctx.value.errors() == [
+        expected_error = [
```
MartinBelthle: Couldn't you just change the url to put 2.10 ?

TheoPascoli: I can, I feel like it is just postponing the problem. If we need the URL in the error message, I can indeed just change the url

antarest/core/filetransfer/service.py
TheoPascoli: Black did it

tests/variantstudy/model/command/test_remove_st_storage.py
MartinBelthle: Indeed it's postponing for the next time we're gonna bump the lib but I think it's a small issue and it makes the test clearer IMO

antarest/core/filetransfer/service.py
MartinBelthle: Ah okay

MartinBelthle (dismissed):

MartinBelthle (approved):

PR #2333, opened 2025-02-06
chore(group): enforce lower case for item groups

sylvlecl:
More limited version of https://github.com/AntaresSimulatorTeam/AntaREST/pull/2283 
which will affect only groups, not cluster names and identifiers.

For v8.6 studies, short term storage groups are serialized to title case, in order
to comply with the stricter solver format at the time. 

Some improvements along the way:
- clarify versioning of commands:
  serialization format is versioned, not the commands themselves
- add and test some annotated type `ItemName` and `AreaId` that will be re-usable
- remove `group` from parent class `ItemProperties` which did not belong there


antarest/study/business/areas/st_storage_management.py
```diff
@@ -86,10 +89,11 @@ def validate_name(cls, name: t.Optional[str]) -> str:
             raise ValueError("'name' must not be empty")
         return name
 
-    # noinspection PyUnusedLocal
-    def to_config(self, study_version: StudyVersion) -> STStorageConfigType:
-        values = self.model_dump(mode="json", by_alias=False, exclude_none=True)
-        return create_st_storage_config(study_version=study_version, **values)
+
+def _to_properties(version: StudyVersion, creation: STStorageCreation) -> STStoragePropertiesType:
```
MartinBelthle: Why didn't you put this as a method of STStorageCreation ? Because you did a method for RenewableCreation and ThermalCreation so we should probably harmonize

antarest/study/storage/rawstudy/model/filesystem/config/field_validators.py
```diff
@@ -9,12 +9,35 @@
...
 
 
+_VALID_NAME_PATTERN = re.compile(r"[a-zA-Z0-9_(),& -]+")
+
+
+def _validate_item_name(name: Any) -> str:
+    if isinstance(name, int):
+        name = str(name)
+    if not isinstance(name, str):
+        ValueError(f"Invalid name '{name}'.")
```
MartinBelthle: Sonar Issue: We should add a `raise` here :)

antarest/study/storage/rawstudy/model/filesystem/root/input/st_storage/clusters/area/list.py
```diff
@@ -10,10 +10,15 @@
 #
 # This file is part of the Antares project.
 
+from antarest.study.storage.rawstudy.ini_reader import LOWER_CASE_PARSER, IniReader, any_section_option_matcher
+from antarest.study.storage.rawstudy.ini_writer import LOWER_CASE_SERIALIZER, IniWriter
 from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfig
 from antarest.study.storage.rawstudy.model.filesystem.context import ContextServer
 from antarest.study.storage.rawstudy.model.filesystem.ini_file_node import IniFileNode
 
+_VALUE_PARSERS = {any_section_option_matcher("group"): LOWER_CASE_PARSER}
```
MartinBelthle, line 22: Question: Should we put these parsers / serializers in a util file instead of declarating them every time ?

tests/integration/study_data_blueprint/test_renewable.py
```diff
@@ -117,7 +117,7 @@ def test_lifecycle(
         fr_solar_pv_props = {
             **DEFAULT_PROPERTIES,
             "name": fr_solar_pv,
-            "group": "Solar PV",
+            "group": "solar pv",
```
MartinBelthle: I think here we should keep the MAJ to ensure we can create a cluster with a group in the wrong format.
We also have to change the check after to see that we return the value lowered

antarest/study/storage/variantstudy/model/command/create_cluster.py
```diff
@@ -42,50 +46,49 @@ class CreateCluster(ICommand):
...
+    @property
+    def cluster_name(self) -> str:
+        return self.parameters.name
+
+    @model_validator(mode="before")
+    @classmethod
+    def validate_model(cls, values: t.Dict[str, t.Any], info: ValidationInfo) -> t.Dict[str, t.Any]:
+        # Validate parameters
+        if isinstance(values["parameters"], dict):
+            parameters = copy.deepcopy(values["parameters"])
```
MartinBelthle: Don't think we need to make a copy here

antarest/study/storage/variantstudy/model/command/create_cluster.py
```diff
@@ -42,50 +46,49 @@ class CreateCluster(ICommand):
...
+            if info.context and info.context.version == 1:
+                parameters["name"] = values["cluster_name"]
```
MartinBelthle: Could be done in one line: `parameters["name"] = values.pop("cluster_name")` as pop returns the value

tests/variantstudy/test_utils.py
```diff
@@ -42,3 +44,24 @@ def test_aggregate_commands(command_context: CommandContext):
 
     command_dto_list = transform_command_to_dto(command_list, command_ref_list)
     assert len(command_dto_list) == 4
+
+
+def test_aggregate_commands_with_version(command_context: CommandContext):
```
MartinBelthle: Could be nice to test that 2 commands with different versions are not aggregated

antarest/study/storage/variantstudy/command_factory.py
```diff
@@ -109,13 +116,15 @@ def _to_single_command(
         """Convert a single CommandDTO to ICommand."""
         if action in COMMAND_MAPPING:
             command_class = COMMAND_MAPPING[action]
-            return command_class(  # type: ignore
-                **args,
-                command_context=self.command_context,
-                version=version,
-                command_id=command_id,
-                study_version=study_version,
+            data = copy.deepcopy(args)
```
MartinBelthle, line 120: Same here, don't think you need a copy

antarest/study/storage/variantstudy/model/command/create_renewables_cluster.py
```diff
@@ -39,21 +38,33 @@ class CreateRenewablesCluster(ICommand):
...
+    @property
+    def cluster_name(self) -> str:
+        return self.parameters.name
+
+    @model_validator(mode="before")
+    @classmethod
+    def validate_model(cls, values: t.Dict[str, t.Any], info: ValidationInfo) -> t.Dict[str, t.Any]:
+        # Validate parameters
+        if isinstance(values["parameters"], dict):
+            parameters = copy.deepcopy(values["parameters"])
```
MartinBelthle: Same remark on the copy and the one-line

antarest/study/storage/variantstudy/model/command/create_st_storage.py
```diff
@@ -54,13 +59,18 @@ class CreateSTStorage(ICommand):
...
+    @override
+    def version(self) -> int:
+        return 2
 
     # Command parameters
     # ==================
 
-    area_id: str = Field(description="Area ID", pattern=r"[a-z0-9_(),& -]+")
-    parameters: STStorageConfigType
+    # TODO SL: why lower case here and not in cluster / renewables ?
```
MartinBelthle: We can remove this TODO

antarest/study/storage/variantstudy/model/command/create_st_storage.py
```diff
@@ -54,13 +59,18 @@ class CreateSTStorage(ICommand):
...
+    area_id: LowerCaseStr = Field(description="Area ID", pattern=r"[a-z0-9_(),& -]+")
```
MartinBelthle: Do we want to keep the LowerCaseStr here ?

antarest/study/storage/variantstudy/model/command/create_st_storage.py
```diff
@@ -85,13 +95,24 @@ class CreateSTStorage(ICommand):
...
     @property
     def storage_name(self) -> str:
         """The label representing the name of the storage for the user."""
         return self.parameters.name
 
+    @model_validator(mode="before")
+    @classmethod
+    def validate_model(cls, values: t.Dict[str, t.Any], info: ValidationInfo) -> t.Dict[str, t.Any]:
+        if isinstance(values["parameters"], dict):
+            parameters = copy.deepcopy(values["parameters"])
```
MartinBelthle: Same comment about the copy and the one-line

tests/integration/study_data_blueprint/test_st_storage.py
```diff
@@ -361,7 +361,7 @@ def test_lifecycle__nominal(
         grand_maison = "Grand'Maison"
         grand_maison_properties = {
             "name": grand_maison,
-            "group": "PSP_closed",
+            "group": "psp_closed",
```
MartinBelthle: Same. We could keep the old writing

tests/integration/test_integration_token_end_to_end.py
```diff
@@ -116,15 +116,15 @@ def test_nominal_case_of_an_api_user(client: TestClient, admin_access_token: str
                 "parameters": {
                     "group": "Gas",
                     "marginal-cost": 98,
-                    "unitCount": 1,
```
MartinBelthle, line 119: Did you change it because it doesn't work anymore ? If so, it could be an issue as it's kind of a breaking change (even though the scripts don't use this writing)

tests/storage/repository/filesystem/special_node/test_lower_case_nodes.py
```diff
@@ -0,0 +1,118 @@
...
+    file_path = study_path / "test.ini"
+    file_path.touch()
+    node = BindingConstraintsIni(
+        context=Mock(),
+        config=FileStudyTreeConfig(study_path=study_path, path=file_path, version=-1, study_id="id"),
+    )
+
+    data = {"0": {"name": "BC_1", "group": "GRP_1"}}
+    node.save(data)
+    # Asserts the data is saved correctly
```
MartinBelthle: Could be nice to add the same test as above concerning the save of a new group specifically

tests/variantstudy/model/command/test_create_cluster.py
```diff
@@ -177,28 +176,63 @@ def test_to_dto(self, command_context: CommandContext):
...
         dto = command.to_dto()
         assert dto.model_dump() == {
             "action": "create_cluster",
             "args": {
                 "area_id": "foo",
-                "cluster_name": "Cluster1",
-                "parameters": {"group": "Nuclear", "nominalcapacity": 2400, "unitcount": 2},
-                "prepro": prepro_id,
-                "modulation": modulation_id,
+                "modulation": "acc60faf9e0c47c77ec61c0d522382676102d8c1de1eb4aac1c5bb3a11682ff8",
```
MartinBelthle: Why did you change the prepro and modulation ids ? I think it was better before. Also this hash could be random  no ?

tests/variantstudy/test_command_factory.py
```diff
@@ -27,401 +27,712 @@
...
+        None,
+        id="remove_area2",
+    ),
+    pytest.param(
+        CommandDTO(
+            action=CommandName.CREATE_DISTRICT.value,
+            args={
                 "name": "id",
-                "base_filter": "add-all",
+                "filter_items": ["a"],
```
MartinBelthle, line 83: Why did this change ?

MartinBelthle (changes requested):
As we discussed with Florian, we should do something in the run method in case the study is in version 8.6 or 8.7 and the user didn't specified a solver higher or equal to 8.8
Could be done in another PR but it would mean that this doesn't work on it's own

antarest/study/storage/rawstudy/model/filesystem/root/input/st_storage/clusters/area/list.py
sylvlecl: I think it's better to leave them here, each type of file must be responsible for its own parsing.
It would be an error to put this in a util file just to avoid a few duplication, because the format of each of the 3 files are not tied together and could evolve separately. Actually strictly speaking they are already different since we don't expect the same values in each one of those files, although we don't check it here, we could.

antarest/study/business/areas/st_storage_management.py
sylvlecl: Moved it back to the class

antarest/study/storage/variantstudy/command_factory.py
sylvlecl: Here I prefer to keep it, otherwise we will mutate the input DTO which is a weird side-effect. The arguments of the function will be modified by the function.

antarest/study/storage/variantstudy/model/command/create_st_storage.py
sylvlecl: Actually I left it because it is still relevant, just removed my name from it :)

antarest/study/storage/variantstudy/model/command/create_st_storage.py
sylvlecl: You're right, we should not change this in that PR, I will revert it and remove the TODO

antarest/study/storage/variantstudy/model/command/create_st_storage.py
sylvlecl: Removed finally, see other comment

antarest/study/storage/variantstudy/model/command/create_st_storage.py
sylvlecl: I created an `AreaId` type used in the 2 other commands too, which also fixes the pattern constraint:
the pattern constraint does not work because it does not perform a full match, I added matching start and end of expression.

tests/integration/study_data_blueprint/test_renewable.py
sylvlecl: Done

tests/integration/study_data_blueprint/test_st_storage.py
sylvlecl: Done

tests/integration/test_integration_token_end_to_end.py
sylvlecl: Actually on dev branch it just does not work, you get errors when the variant is generated, but we just don't check for them.
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for ThermalConfig
unitCount
  Extra inputs are not permitted [type=extra_forbidden, input_value=1, input_type=int]
```

Now the input is validated as soon as the command is posted, that's why the test does not pass any more.

tests/variantstudy/test_command_factory.py
sylvlecl: It did not, it corresponds to line 46 before the PR (single district creation, not list creation).

tests/variantstudy/model/command/test_create_cluster.py
sylvlecl: I reverted this change.

tests/variantstudy/test_utils.py
sylvlecl: Actually I removed this behaviour, it did not make sense because commands are not versioned, only DTOs are.
To make it clear, I removed `version` from the `ICommand` public API, and replaced it with a `_SERIALIZATION_VERSION` final field in implementations where it's needed.

We cannot have DTOs in different versions in this aggregation method, because they are all created from the same version of the code, so the test does not make sense anymore, I removed it.

tests/storage/repository/filesystem/special_node/test_lower_case_nodes.py
sylvlecl: Added. I also adapted the implementation and the test sor that we also *write* the group in lower case.

antarest/study/storage/rawstudy/model/filesystem/config/field_validators.py
```diff
@@ -9,12 +9,36 @@
...
 
-from typing import Any, List, Mapping, MutableMapping
+from pydantic import BeforeValidator, Field
+
+from antarest.study.storage.rawstudy.model.filesystem.config.identifier import transform_name_to_id
 
 _ALL_FILTERING = ["hourly", "daily", "weekly", "monthly", "annual"]
 
 
+_VALID_NAME_PATTERN = re.compile(r"[a-zA-Z0-9_(),& -]+")
```
MartinBelthle: It seems you're not using this regex anywhere

antarest/study/storage/rawstudy/model/filesystem/config/field_validators.py
```diff
@@ -9,12 +9,36 @@
...
+    if not transform_name_to_id(name):
+        raise ValueError(f"Invalid name '{name}'.")
+    return name
+
+
+# Type to be used for item names, will raise an error if name
+# does not comply with antares-simulator limitations.
+ItemName = Annotated[str, BeforeValidator(_validate_item_name)]
+
+AreaId = Annotated[str, Field(description="Area ID", pattern=r"^[a-z0-9_(),& -]+$")]
```
MartinBelthle, line 40: Not sure to understand why this pattern is different from the pattern used inside `transform_name_to_id`. Perhaps add a comment

antarest/study/storage/rawstudy/model/filesystem/config/st_storage.py
```diff
@@ -160,7 +163,29 @@ class STStorage880Config(STStorage880Properties, LowerCaseIdentifier):
 
 # NOTE: In the following Union, it is important to place the older version first,
 # because otherwise, creating a short term storage always creates a v8.8 one.
-STStorageConfigType = STStorageConfig | STStorage880Config
+STStorageConfigType = Union[STStorageConfig, STStorage880Config]
```
MartinBelthle: You can replace `Union` by `|` to fit with a PR Theo merged recently

antarest/study/storage/rawstudy/model/filesystem/config/thermal.py
```diff
@@ -408,7 +412,8 @@ class Thermal870Config(Thermal870Properties, IgnoreCaseIdentifier):
 
 # NOTE: In the following Union, it is important to place the most specific type first,
 # because the type matching generally occurs sequentially from left to right within the union.
-ThermalConfigType = Thermal870Config | Thermal860Config | ThermalConfig
+ThermalConfigType = Union[Thermal870Config, Thermal860Config, ThermalConfig]
```
MartinBelthle: Same comment about Union

antarest/study/storage/variantstudy/business/utils.py
```diff
@@ -103,38 +103,58 @@ def decode(alias: str, study: FileStudy) -> str:
 
 def transform_command_to_dto(
     commands: Sequence[ICommand],
-    ref_commands: Optional[Sequence[CommandDTO]] = None,
+    ref_command_dtos: Optional[Sequence[CommandDTO]] = None,
     force_aggregate: bool = False,
 ) -> List[CommandDTO]:
+    """
```
MartinBelthle, line 109: Nice to put a comment, I barely understand your "refactoring"

antarest/study/storage/variantstudy/command_factory.py
MartinBelthle: Okay

tests/variantstudy/test_utils.py
```diff
@@ -9,10 +9,13 @@
 # SPDX-License-Identifier: MPL-2.0
 #
 # This file is part of the Antares project.
+import pytest
```
MartinBelthle: I think these imports are useless

antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
```diff
@@ -239,9 +241,7 @@ def to_dto(self) -> CommandDTO:
             if matrix_attr is not None:
                 args[matrix_name] = matrix_service.get_matrix_id(matrix_attr)
 
-        return CommandDTO(
-            action=self.command_name.value, args=args, version=self.version, study_version=self.study_version
-        )
+        return CommandDTO(action=self.command_name.value, args=args, version=1, study_version=self.study_version)
```
MartinBelthle: Perhaps instead of putting version=1 we should declare it as you did in `create_cluster`

MartinBelthle (changes requested):

antarest/study/storage/rawstudy/model/filesystem/config/field_validators.py
sylvlecl: This pattern is meant to be used as a full match, hence the `^` and `$`. Otherwise, the constraint does not do anything, because it partially matches more or less any string.
The characters inside brackets are valid characters for IDs.

In transform_name_to_id, the `^` is used to **negate** the characters in the bracket: we want to find characters that are **different** from valid characters.

sylvlecl: Note: I've added some unit tests in test_field_validators to check the behaviour (before, the annotation in ST storage creation did not work properly)

antarest/study/storage/variantstudy/model/command/create_binding_constraint.py
sylvlecl: Done

MartinBelthle (approved):

PR #2334, opened 2025-02-07
fix(scan): fix workspaces can't overlap nor be renamed

smailio:
fix ANT-2768 and ANT-2773


sylvlecl (commented):
Let's wait until we decide what to do exactly:
- we should probably forbid nested workspaces in configuration
- we could "hide" (not return them in studies listing) studies belonging to worskpaces that are not configured anymore
- I think in desktop version, we should just forbid external workspaces, to me, they are not adapted to this use case. They raise too much questions in such unstable and user-defined environment.

sylvlecl (dismissed):

sylvlecl (approved):

PR #2335, opened 2025-02-07
feat(ui): update MUI theme

skamril:


webapp/src/components/App/Singlestudy/explore/Modelization/Map/style.ts
```diff
@@ -17,7 +17,7 @@ import mapbackground from "../../../../../../assets/img/mapbackground.png";
 import { getTextColor, type RGB } from "./utils";
 
 ////////////////////////////////////////////////////////////////
-// Map(index.tsx)
+// Map(ThermalForm.tsx)
```
hdinia: to be removed

webapp/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "antares-web",
   "version": "2.19.0",
+  "homepage": "https://github.com/AntaresSimulatorTeam/AntaREST",
```
hdinia: any reason to add this ? If it's not used by any build or deployment process you can remove it

webapp/public/locales/en/main.json
```diff
@@ -100,20 +103,26 @@
...
   "documentation.title": "Documentation",
   "github.title": "Github",
   "recentStudy.title": "Recent study",
   "studies.title": "Studies",
-  "logout.title": "Logout",
   "settings.title": "Settings",
   "data.title": "Data",
   "dialog.title.confirmation": "Confirmation",
-  "dialog.message.logout": "Are you sure you want to logout?",
+  "dialog.message.signOut": "Are you sure you want to logout?",
```
hdinia: you can also update the key content (logout vs sign out)

webapp/public/locales/en/main.json
```diff
@@ -11,7 +12,9 @@
   "global.unarchive": "Unarchive",
   "global.export": "Export",
   "global.export.error": "Export failed",
+  "global.filterAction": "Filter",
```
hdinia, line 15: you can remove this one and use`"global.filter"`

webapp/src/components/App/Container/MenuItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+      if (isExternalLink) {
+        window.open(link, "_blank");
+      } else {
+        navigate(link);
+      }
+      return;
+    }
+
+    onClick?.();
+  };
```
hdinia: use an early return
```ts
const handleClick = () => {
  if (!link) {
    onClick?.();
    return;
  }
  
  isExternalLink ? window.open(link, "_blank") : navigate(link);
};
```

webapp/src/components/App/Container/MenuItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+  isMenuOpen: boolean;
+  icon: React.ReactNode;
+  onClick?: VoidFunction;
+  link?: string;
+}
+
+function MenuItem({ title, isMenuOpen, icon, onClick, link }: Props) {
+  const navigate = useNavigate();
+  const location = useLocation();
+  const isExternalLink = link?.startsWith("http");
```
hdinia, line 30: "https" not needed ?

webapp/src/components/App/Container/MenuItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+  // JSX
+  ////////////////////////////////////////////////////////////////
+
+  return (
+    <Tooltip title={isMenuOpen ? "" : title} placement="right-end">
+      <ListItem disablePadding onClick={handleClick}>
+        <ListItemButton
+          sx={[{ minHeight: 48, px: 2.5, justifyContent: isMenuOpen ? "initial" : "center" }]}
+          selected={isSelected}
+        >
```
hdinia: add accessibility attributes 
```tsx
  role="menuitem"
  aria-label={title}
```

webapp/src/components/App/Container/MenuItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  ////////////////////////////////////////////////////////////////
+  // JSX
+  ////////////////////////////////////////////////////////////////
+
+  return (
+    <Tooltip title={isMenuOpen ? "" : title} placement="right-end">
+      <ListItem disablePadding onClick={handleClick}>
+        <ListItemButton
+          sx={[{ minHeight: 48, px: 2.5, justifyContent: isMenuOpen ? "initial" : "center" }]}
```
hdinia: remove array from sx prop

webapp/src/components/App/Container/MenuItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const isSelected = link === location.pathname;
+
+  ////////////////////////////////////////////////////////////////
+  // Event Handlers
+  ////////////////////////////////////////////////////////////////
+
+  const handleClick = () => {
+    if (link) {
+      if (isExternalLink) {
+        window.open(link, "_blank");
```
hdinia:
```ts 
window.open(link, "_blank", "noopener,noreferrer"); 
```

webapp/src/components/App/Container/TaskIcon.tsx
```diff
@@ -0,0 +1,42 @@
...
+      color="primary"
+      sx={{
+        ".MuiBadge-badge": {
+          right: -4,
+          top: 11,
+          border: `2px solid`,
+          borderColor: "background.paper",
+          padding: "0 4px",
+        },
+      }}
```
hdinia, line 35: optional:
maybe you can add the custom styling in the theme file using `variants` (creates a new variant while preserving default behavior), see [MUI variants](https://mui.com/material-ui/customization/theme-components/?srsltid=AfmBOor4ZC6Yc0-0N0m36FRcVa5zEiVYkOwf9lpH5-aVSa1bKHCphoq7#variants)
And make it a global component, if we could use it elsewhere

webapp/src/components/common/dialogs/DigestDialog/DigestTabs.tsx
hdinia, line 1: revert this change the check is needed here as individual matrices can be empty

webapp/src/hooks/useThemeColorScheme.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Gets the return value from `useColorScheme` hook with additional properties.
+ *
+ * @returns An object.
+ */
+function useThemeColorScheme() {
+  const colorScheme = useColorScheme();
+  const { mode, systemMode } = colorScheme;
+  const isDarkMode = mode === "dark" || systemMode === "dark";
+  const isLightMode = !isDarkMode;
```
hdinia: you can define this in the return and remove this line ` isLightMode: !isDarkMode,`

hdinia (changes requested):

webapp/package.json
skamril: It's a valid attribute for the package.json https://docs.npmjs.com/cli/v11/configuring-npm/package-json#homepage

I used it in About page in Settings.

webapp/public/locales/en/main.json
skamril: "global.filter" already exist, in fact I add it for the french translation, because 'Filtrer' and 'Filtre" in French is translated by 'Filter' in English, there is no distinction.

webapp/src/components/App/Container/MenuItem.tsx
skamril: I don't like to use `!link`, I prefer `if link => use link`, faster to understand. 

It doesn't deserve a comment, we don't force the use of ternary, but I make this change:

```typescript
if (link) {
  return isExternalLink ? window.open(link, "_blank") : navigate(link);
}
onClick?.();
```

webapp/src/components/App/Container/MenuItem.tsx
skamril: "http" includes "https"

webapp/src/components/App/Container/MenuItem.tsx
skamril: We don't use ARIA in the project, so we have to think about that and make a ticket, if necessary, for the all app.

webapp/package.json
hdinia: homepage is used for the app production url not the repo, to point out the repo their is a `repository` property
see: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#repository

and for your use case, you should just define the url in a global constant instead of importing it from the package.json

webapp/public/locales/en/main.json
hdinia: okay I thought they had the same label on the UI

webapp/src/components/App/Container/MenuItem.tsx
hdinia: ok as long as you use early return it's ok

webapp/src/components/App/Container/MenuItem.tsx
hdinia: ok

webapp/src/components/App/Container/MenuItem.tsx
hdinia: I disagree, it's not a reason to continue.. 
they are 2 reason why I ask for this,
1. it is soon mandatory to have accessible code so on new code we have to enforce this, the sooner the better
2. (most important) for end to end tests that we will start soon it's mandatory to be able to test properly components

hdinia: in particular `role` is important to target the components in end to end tests

webapp/src/components/App/Container/TaskIcon.tsx
skamril: No is very specific for this menu item, it's the only one with a badge

webapp/package.json
skamril: In our case the GitHub repo is our homepage (`The URL to the project homepage.`), and `repository` must point to the `.git`. Maybe the homepage will be https://antares-web.readthedocs.io/ in the future.

The link in about page is for the homepage of the project.

webapp/src/components/App/Container/MenuItem.tsx
skamril: But it is not correct to ask to do something that has not yet been asked for and has never been used.
Managing only the main menu has no interest. And it's not the purpose of this PR.

webapp/src/components/common/dialogs/DigestDialog/DigestTabs.tsx
skamril: What do you mean by empty? Empty array or nil value?

If it's empty array the condition `matrices.X &&` is always true. 
If it can be a nil value this type must be updated:

```typescript
export interface DigestData {
  area: DigestMatrixData;
  districts: DigestMatrixData;
  flowLinear: DigestMatrixData;
  flowQuadratic: DigestMatrixData;
}
```

webapp/package.json
hdinia: again, homepage = prod url for any app. It can be the repo url but only for packages not apps. Remove `homepage` and define a constant `REPO_URL` or use the `repository` property

webapp/src/components/App/Container/MenuItem.tsx
hdinia: The other day when writing end to end tests, I was having trouble because of this, it's indeed required now, we have to do it on all new code and when we will add end to end tests we will have several components to fix (add accessibility)

webapp/src/components/common/dialogs/DigestDialog/DigestTabs.tsx
hdinia: The type is correct, static types give us compile-time guarantees, but they don't protect against runtime surprises

As I am not sure about this case, you can remove it but before make sure to **test it properly.** 
Or keep the runtime check (more safe)

hdinia: edit: the runtime check is not correct here, can you update it if you decide to keep it:
```ts
const tabItems = [
    matrices.area && {
      label: "Area",
      content: () => <DigestMatrix matrix={matrices.area} />,
    },
    matrices.districts && {
      label: "Districts",
      content: () => <DigestMatrix matrix={matrices.districts} />,
    },
    matrices.flowLinear && {
      label: "Flow Linear",
      content: () => <DigestMatrix matrix={matrices.flowLinear} />,
    },
    matrices.flowQuadratic && {
      label: "Flow Quadratic",
      content: () => <DigestMatrix matrix={matrices.flowQuadratic} />,
    },
  ].filter(Boolean);
```

makdeuneuv:
revue thème1
Is it possible to go back on a long name of a favorite study? Because we see that the display is truncated and that the yellow star is placed above the name of the study. The remark is purely aesthetic.
@skamril

makdeuneuv:
revue thème2
Is it possible to reduce the space marked by the red line on the image? Because this adds nothing and would allow more display space for the name of the study and its variants. The request is purely aesthetic. @skamril

webapp/src/components/common/dialogs/DigestDialog/DigestTabs.tsx
skamril: Front code:

```typescript
client.get<DigestData>(`/v1/private/studies/${studyId}/outputs/${outputId}/digest-ui`)
```

Back code:

```python
@bp.get(
  "/private/studies/{study_id}/outputs/{output_id}/digest-ui",
  tags=[APITag.study_outputs],
  summary="Display an output digest file for the front-end",
  response_model=DigestUI,
)
```

```python
class DigestMatrixUI(AntaresBaseModel):
    columns: List[str | List[str]]
    data: List[List[str]]
    grouped_columns: bool = Field(alias="groupedColumns")


class DigestUI(AntaresBaseModel):
    area: DigestMatrixUI
    districts: DigestMatrixUI
    flow_linear: DigestMatrixUI = Field(alias="flowLinear")
    flow_quadratic: DigestMatrixUI = Field(alias="flowQuadratic")
```

So the TypeScript type is valid, and we don't need to check it based on the API documentation.
If the API returns a nil value, the backend must be fixed.

Then in your code, if we follow your logic, you must also check that `matrices` is really an object and not a string or a number for example.

skamril:
> revue thème1 Is it possible to go back on a long name of a favorite study? Because we see that the display is truncated and that the yellow star is placed above the name of the study. The remark is purely aesthetic. @skamril

It's fixed, thanks

hdinia (approved):

PR #2342, opened 2025-02-12
fix(st-storage): fix ST storage groups case for v8.6

sylvlecl:
In v8.6, short term storage groups were not simply in title-case, 
this PR implements the exact correct case.


MartinBelthle (approved):
Indeed

PR #2349, opened 2025-02-18
build(lint): use ruff instead of isort and black

MartinBelthle:


TheoPascoli (commented):
Do you mind adding how to use it in the readme ?

.github/workflows/main.yml
```diff
@@ -18,16 +18,10 @@ jobs:
...
-          requirementsFiles: "requirements-dev.txt"
-      - name: Check with black
-        uses: psf/black@stable
-        with:
-          # Version of Black should match the versions set in `requirements-dev.txt`
-          version: "~=25.1.0"
-          options: --check --diff
+      - name: Check format (ruff)
+        run: |
+          ruff check antarest/ tests/
```
sylvlecl, line 23: Is it possible to define the directories in pyproject.toml ?
Would be more convenient to ensure consistency between CI and local execution.

sylvlecl: Actually it does not seem very convenient, you can only define patterns in the `include` configuration property, not simply directories.

PR #2350, opened 2025-02-18
feat(area): replace update config in table mode areas

TheoPascoli:


antarest/study/storage/variantstudy/model/command/update_area_properties.py
```diff
@@ -0,0 +1,98 @@
...
+
+        for area_id, area_folder in self.dict_area_folder.items():
+            self.update_area_optimization(area_id, study_data)
+            self.update_adequacy_patch(area_id, study_data)
+
+        output, _ = self._apply_config(study_data.config)
+
+        return output
+
+    def update_thermal_properties(self, study_data: FileStudy) -> None:
```
sylvlecl: we'd better make the methods private (starting with `_`)

antarest/study/storage/variantstudy/model/command/update_area_properties.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+
+    # Overloaded metadata
+    # ===================
+
+    command_name: CommandName = CommandName.UPDATE_AREA_PROPERTIES
+
+    # Command parameters
+    # ==================
+    dict_area_folder: Dict[str, AreaFolder]
```
MartinBelthle: It seems that you designed this command to update properties for several areas. It would be nice to says it explicitely in the command name and doc

antarest/study/storage/variantstudy/model/command/update_area_properties.py
```diff
@@ -0,0 +1,98 @@
...
+    list_thermal_area_properties: List[ThermalAreasProperties]
```
MartinBelthle: As we discussed it for other PRs of the same type, these objects can be partially initialized and so we need to change the code accordingly

MartinBelthle (commented):
I think you should also use your command inside the file `properties_management.py` to replace the method `set_field_values` which is used to update area properties for one area only

antarest/study/business/areas/properties_management.py
```diff
@@ -35,160 +37,66 @@
...
-    energy_cost_unsupplied: float
-    energy_cost_spilled: float
-    non_dispatch_power: bool
-    dispatch_hydro_power: bool
-    other_dispatch_power: bool
-    spread_unsupplied_energy_cost: float
-    spread_spilled_energy_cost: float
-    filter_synthesis: Set[str]
-    filter_by_year: Set[str]
+    average_unsupplied_energy_cost: float = 0.0
```
sylvlecl: Do we really want to rename those fields ?
This is a breaking change of the API, it will probably break some clients other than the UI (R packages, antares-craft)

antarest/study/storage/variantstudy/model/command/update_areas_properties.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+
+    # Overloaded metadata
+    # ===================
+
+    command_name: CommandName = CommandName.UPDATE_AREAS_PROPERTIES
+
+    # Command parameters
+    # ==================
+    areas_folder: Dict[str, AreaFolder]
```
sylvlecl: We should strive to not use too much format/filesystem related wording and classes in this business layer. "folder" is clearly related to the current underlying format. We should either create another one or rename it, I think.

antarest/study/storage/variantstudy/model/command/update_areas_properties.py
```diff
@@ -0,0 +1,106 @@
...
+            thermal_properties["spilledenergycost"].update(thermal_area_property.spilled_energy_cost)
+            thermal_properties["unserverdenergycost"].update(thermal_area_property.unserverd_energy_cost)
+        study_data.tree.save(thermal_properties, ["input", "thermal", "areas"])
+
+    def update_adequacy_patch(self, area_id: str, study_data: FileStudy) -> None:
+        area_folder = self.areas_folder.get(area_id)
+
+        if area_folder and area_folder.adequacy_patch:
+            new_config = area_folder.adequacy_patch.to_config()
+            adequacy_patch_properties = study_data.tree.get(["input", "areas", area_id, "adequacy_patch"])
```
sylvlecl: we could avoid to build the path list twice (here and below)

antarest/study/storage/variantstudy/model/command/update_areas_properties.py
sylvlecl: Also,very important:
in this `AreaFolder` class, most fields are not optional, which will make the behaviour different from other updates:
if we use as input `{"fr": AreaFolder()}`, this will override the existing properties with the default properties.

sylvlecl (changes requested):
First set of comments:
I think we should try not to re-use the existing `AreaFolder` class, for 2 reasons: naming/dependence to the file tree format, and more importantly behaviour, because we risk to override properties with default values.

TheoPascoli:
> First set of comments: I think we should try not to re-use the existing `AreaFolder` class, for 2 reasons: naming/dependence to the file tree format, and more importantly behaviour, because we risk to override properties with default values.

I see. I did a rollback and I'll try to find a better way

MartinBelthle (changes requested):
As we discussed I think the parameters of the command should be objects with optional fields

antarest/study/storage/variantstudy/model/command/common.py
```diff
@@ -30,6 +30,7 @@ class FilteringOptions:
 
 class CommandName(Enum):
     CREATE_AREA = "create_area"
+    UPDATE_AREAS_PROPERTIES = "update_area_properties"
```
MartinBelthle: I would write "update_areas_properties" to fit with the name you wrote

antarest/study/business/areas/properties_management.py
```diff
@@ -10,185 +10,63 @@
...
-            is_in_version = start_ver <= study_ver <= end_ver
-            if not is_in_version:
-                return None
 
-            try:
-                val = file_study.tree.get(field_info["path"].format(area=area_id).split("/"))
-            except (ChildNotFoundError, KeyError):
-                return field_info["default_value"]
+        current_thermal_props = file_study.tree.get(THERMAL_PATH)
+        current_optim_properties = file_study.tree.get([s.format(area_id=area_id) for s in OPTIMIZATION_PATH])
```
MartinBelthle: Weird writing the .format on the list. Even if I don't know how we could do otherwise

antarest/study/business/areas/properties_management.py
```diff
@@ -10,185 +10,63 @@
...
+        current_adequacy_patch = (
+            file_study.tree.get([s.format(area_id=area_id) for s in ADEQUACY_PATCH_PATH])
+            if study.version >= STUDY_VERSION_8_3
+            else {}
+        )
 
-            encode = field_info.get("encode") or (lambda x: x)
-            return encode(val)
+        area_properties = AreaPropertiesProperties(
```
MartinBelthle: The name AreaPropertiesProperties is not really clear

antarest/study/business/areas/properties_management.py
```diff
@@ -10,185 +10,63 @@
...
+            thermal_properties=ThermalAreasProperties(**current_thermal_props),
+            optimization_properties=OptimizationProperties(**current_optim_properties),
+            adequacy_properties=AdequacyPathProperties(**current_adequacy_patch),
+        )
 
-        return PropertiesFormFields.model_construct(**{name: get_value(info) for name, info in FIELDS_INFO.items()})
+        return AreaProperties.model_validate(area_properties.get_area_properties(area_id))
```
MartinBelthle: Naming everything AreaProperties makes the code really hard to understand

antarest/study/business/model/area_properties_model.py
```diff
@@ -0,0 +1,144 @@
...
+
+        return values
+
+
+@all_optional_model
+class AreaPropertiesUpdate(AntaresBaseModel, extra="forbid", populate_by_name=True):
+    energy_cost_unsupplied: float = Field(
+        validation_alias=AliasChoices("averageUnsuppliedEnergyCost", "energyCostUnsupplied")
+    )
+    energy_cost_spilled: float = Field(validation_alias=AliasChoices("averageSpilledEnergyCost", "energyCostSpilled"))
```
MartinBelthle, line 102: Not sure to understand how you use the alias choices

antarest/study/business/model/area_properties_model.py
```diff
@@ -0,0 +1,144 @@
...
+
+class AreaProperties(AntaresBaseModel, extra="forbid", populate_by_name=True, alias_generator=to_camel_case):
+    energy_cost_unsupplied: float = 0.0
+    energy_cost_spilled: float = 0.0
+    non_dispatch_power: bool = True
+    dispatch_hydro_power: bool = True
+    other_dispatch_power: bool = True
+    spread_unsupplied_energy_cost: float = 0.0
+    spread_spilled_energy_cost: float = 0.0
+    filter_synthesis: Set[str] = set(FILTER_OPTIONS)
```
MartinBelthle: Couldn't you use the `comma_separated_enum_list` you introduced for the links to avoid using the encode/decode filters ?

antarest/study/business/areas/properties_management.py
MartinBelthle: you're using a model validate on an already created object

antarest/study/storage/variantstudy/model/command/update_area_properties.py
```diff
@@ -0,0 +1,127 @@
...
+            else {}
+        )
+
+        return current_thermal_props, current_optim_properties, current_adequacy_patch
+
+    @override
+    def to_dto(self) -> CommandDTO:
+        args = {}
+        for area_id, area_properties in self.properties.items():
+            args[area_id] = area_properties.model_dump(mode="json", exclude_none=True)
```
MartinBelthle: Seems that you're not using this `args` variable

MartinBelthle (changes requested):
I find the naming `AreaPropertiesProperties` really confusing.
Also it could be nice to add UTs for the command

antarest/study/business/areas/properties_management.py
TheoPascoli: This is the best way I found to iterate on the list and formatting what needs to be formatted

antarest/study/business/model/area_properties_model.py
TheoPascoli: The variable name that I receive from the area properties and the table mode are differents for the same property. The AliasChoices let me build the object with the 2 aliases.

antarest/study/business/areas/properties_management.py
TheoPascoli: Indeed, but the first properties is a business word, while the second one is related to the properties in the ini files. Let me know if you have suggestions

antarest/study/business/model/area_properties_model.py
TheoPascoli: After spending some time on it, it does not seem easy to make this change...

antarest/study/business/model/area_properties_model.py
```diff
@@ -0,0 +1,144 @@
...
+from antarest.study.storage.rawstudy.model.filesystem.config.area import (
+    AdequacyPatchMode,
+    AdequacyPathProperties,
+    OptimizationProperties,
+    ThermalAreasProperties,
+)
+
+FILTER_OPTIONS = ["hourly", "daily", "weekly", "monthly", "annual"]
+THERMAL_PATH = ["input", "thermal", "areas"]
+OPTIMIZATION_PATH = ["input", "areas", "{area_id}", "optimization"]
```
sylvlecl: we'd better provide functions to create those paths and avoid unnecessary formatting elsewhere:
```python
def get_optimization_path(area_id: str) -> List[str]:
    return ["input", "areas", area_id, "optimization"]
```

antarest/study/storage/variantstudy/model/command/update_area_properties.py
```diff
@@ -0,0 +1,123 @@
...
+
+        return output
+
+    def save_area_properties(
+        self, study_data: FileStudy, area_id: str, current_properties: AreaPropertiesProperties
+    ) -> None:
+        study_data.tree.save(current_properties.thermal_properties.model_dump(by_alias=True), THERMAL_PATH)
+        study_data.tree.save(
+            current_properties.optimization_properties.model_dump(by_alias=True),
+            [s.format(area_id=area_id) for s in OPTIMIZATION_PATH],
```
sylvlecl: see comment in the other file:
it would be simpler to have a function to create the path, instead of the constant + formatting

(same thing for other paths)

antarest/study/business/model/area_properties_model.py
sylvlecl: Is it just because it's a set and not a list ? 
Note that also we should not use a default value as is, but a "default_factory" instead, because the set is mutable.
Some code could change the default value for all other and future instances of the class.

antarest/study/storage/variantstudy/model/command/update_area_properties.py
```diff
@@ -0,0 +1,123 @@
...
+                adequacy_properties=AdequacyPathProperties(**current_adequacy_patch),
+            )
+
+            new_properties = current_properties.get_area_properties(area_id=area_id).model_copy(
+                update=area_properties.model_dump(exclude_none=True)
+            )
+
+            current_properties.set_area_properties(area_id, new_properties)
+
+            self.save_area_properties(study_data, area_id, current_properties)
```
sylvlecl: There is some room for optimization here I think:
the thermal properties file could be udpated only once at the end, instead of read/written once for each area.

antarest/study/storage/variantstudy/model/command/update_area_properties.py
```diff
@@ -0,0 +1,123 @@
...
+        if self.study_version >= STUDY_VERSION_8_3:
+            study_data.tree.save(
+                current_properties.adequacy_properties.model_dump(),
+                [s.format(area_id=area_id) for s in ADEQUACY_PATCH_PATH],
+            )
+
+    def fetch_area_properties(
+        self, study_data: FileStudy, area_id: str
+    ) -> Tuple[Dict[str, Any], Dict[str, Any], Dict[str, Any]]:
+        current_thermal_props = study_data.tree.get(THERMAL_PATH)
```
sylvlecl: See other comment: it would be better not to read again that data for every area

antarest/study/business/model/area_properties_model.py
sylvlecl: Seems nice, it's a little hidden though: could you add a docstring on the class explaining that ?

sylvlecl (changes requested):
Looks good overall, minor comments + some optimization possible

antarest/study/storage/variantstudy/model/command/update_area_properties.py
TheoPascoli: Yes, done

antarest/study/storage/variantstudy/model/command/update_area_properties.py
TheoPascoli: Indeed

antarest/study/business/model/area_properties_model.py
TheoPascoli: From the properties tabs, I receive a List, from the table mode tab I receiv a str. Moreover, in the OptimizationProperties, the two filter fields are str. Its hard (long) to make everything work together.
If it is worth to spend the time on this issue, I can do it

antarest/study/business/model/area_properties_model.py
TheoPascoli: Done !

sylvlecl (approved):

PR #2351, opened 2025-02-18
fix(STS): upper case STS names cause errors on update

sylvlecl:
STS manager creates INI modifications which use the lower case ID of the STS
for the INI file modification.

However, INI file nodes don't handle case consistently:
- when retrieving data, we use case insensitive matching
- when updating or removing data, we use exact matching

As a consequence, the update of STS which have capitals in their INI section do not work
correctly, because the update_config command will create a new INI section in
lower case, instead of updating the existing one.

**The fix consists in doing case insensitive matching also for updates and deletes,**


sylvlecl:
Closed in favor of simpler #2353

PR #2352, opened 2025-02-18
feat(renewable): add update renewable cluster command

TheoPascoli:


antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
```diff
@@ -0,0 +1,78 @@
...
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand, OutputTuple
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+_RENEWABLE_CLUSTER_PATH = "input/renewables/clusters/{area_id}/list/{cluster_id}"
+
+
+class UpdateRenewablesCluster(ICommand):
```
MartinBelthle: Even though the creation command is called `Renewables` I believe the best wording is without the `s` as it's only concerning one cluster.

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
```diff
@@ -0,0 +1,78 @@
...
+        return output
+
+    @override
+    def to_dto(self) -> CommandDTO:
+        return CommandDTO(
+            action=self.command_name.value,
+            args={
+                "area_id": self.area_id,
+                "cluster_id": self.cluster_id,
+                "properties": self.properties.model_dump(),
```
MartinBelthle: I think we should add mode='json' to ensure the enum are stored as str inside DB

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
```diff
@@ -0,0 +1,78 @@
...
+                message=f"The renewable cluster '{self.cluster_id}' in the area '{self.area_id}' has been updated.",
+            ),
+            {},
+        )
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        path = _RENEWABLE_CLUSTER_PATH.format(area_id=self.area_id, cluster_id=self.cluster_id)
+
+        study_data.tree.save(self.properties.model_dump(by_alias=True), path.split("/"))
```
MartinBelthle: From a design perspective I would prefer to have a RenewableProperties object with only the fields to update registered and do the GET , update here. 
This way the DTO would also have the modified args only (if we add exclude unset inside the model dump)

MartinBelthle (changes requested):

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
MartinBelthle: if we want to introduce the same command for multiple updates (could be used inside table mode for instance), we'll have to make the difference clear and removing the `s` helps IMO

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
MartinBelthle: The only issue being that we open the file twice. Once in the manager and once inside the command ...
But you could use the cache (file_study.config) inside the manager as I believe it contains all the info you need

TheoPascoli: I actually like what I've done. I find it more readable, straightforward, and concise. Indeed, the DTO includes all fields (even those not set by the user) and aside from that, I don't see any downside. We can talk about it or I can ask for another opinion

MartinBelthle: IMO it's a pretty big downside but we can ask Sylvain what he thinks about this indeed

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
```diff
@@ -0,0 +1,78 @@
...
+
+from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfig
+from antarest.study.storage.rawstudy.model.filesystem.config.renewable import RenewablePropertiesType
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand, OutputTuple
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+_RENEWABLE_CLUSTER_PATH = "input/renewables/clusters/{area_id}/list/{cluster_id}"
```
sylvlecl, line 28: Note: I think we should stop to do this, the path acutally used for the tree is a list, we should juste create the list without having this path that we build just for parsing it right after.
We could also change this in another PR if we want to keep homogeneity.

TheoPascoli: I will create another PR to make this change everywhere

antarest/study/business/model/renewable_cluster_model.py
```diff
@@ -0,0 +1,67 @@
...
+
+@all_optional_model
+@camel_case_model
+class RenewableClusterUpdateDTO(RenewableProperties):
+    """
+    Model representing the external data structure required to edit an existing renewable cluster.
+    """
+
+    class Config:
+        populate_by_name = True
```
MartinBelthle, line 32: Do we need to add this as RenewableProperties inherits from ItemProperties where `populate_by_name` is True ?

antarest/study/business/model/renewable_cluster_model.py
```diff
@@ -0,0 +1,67 @@
...
+    RenewablePropertiesType,
+    create_renewable_properties,
+)
+
+
+@all_optional_model
+@camel_case_model
+class RenewableClusterUpdateDTO(RenewableProperties):
+    """
+    Model representing the external data structure required to edit an existing renewable cluster.
```
MartinBelthle: perhaps I would replace `external` by `API` as to me external is a bit unclear (does the DB is external ?)

antarest/study/business/model/renewable_cluster_model.py
```diff
@@ -0,0 +1,67 @@
...
+    class Config:
+        populate_by_name = True
+        alias_generator = to_kebab_case
+
+
+@all_optional_model
+@camel_case_model
+class RenewableClusterCreation(RenewableProperties):
+    """
+    Model representing the data structure required to create a new Renewable cluster within a study.
```
MartinBelthle, line 39: I would also add API here (or external if you prefer)

antarest/study/business/model/renewable_cluster_model.py
```diff
@@ -0,0 +1,67 @@
...
+    def to_properties(self, study_version: StudyVersion) -> RenewablePropertiesType:
+        values = self.model_dump(exclude_none=True)
+        return create_renewable_properties(study_version=study_version, data=values)
+
+
+@all_optional_model
+@camel_case_model
+class RenewableClusterOutput(RenewableConfig):
+    """
+    Model representing the output data structure to display the details of a renewable cluster.
```
MartinBelthle, line 54: Same

antarest/study/storage/variantstudy/model/command/common.py
```diff
@@ -45,6 +45,7 @@ class CommandName(Enum):
     REMOVE_THERMAL_CLUSTER = "remove_cluster"
     CREATE_RENEWABLES_CLUSTER = "create_renewables_cluster"
     REMOVE_RENEWABLES_CLUSTER = "remove_renewables_cluster"
+    UPDATE_RENEWABLE_CLUSTER = "update_renewables_cluster"
```
MartinBelthle: I would change the string to remove the extra "s"

antarest/study/business/model/renewable_cluster_model.py
```diff
@@ -0,0 +1,67 @@
...
+    RenewableConfig,
+    RenewableProperties,
+    RenewablePropertiesType,
+    create_renewable_properties,
+)
+
+
+@all_optional_model
+@camel_case_model
+class RenewableClusterUpdateDTO(RenewableProperties):
```
sylvlecl: Better to remove the `DTO` suffix, this is a business layer class, it should not have any technical suffix.

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
```diff
@@ -0,0 +1,86 @@
...
+    # ===================
+
+    command_name: CommandName = CommandName.UPDATE_RENEWABLE_CLUSTER
+
+    # Command parameters
+    # ==================
+
+    area_id: str
+    cluster_id: str
+    properties: Dict[str, Any]
```
MartinBelthle: IMO it was better when this was typed. I think you can type it as a `RenewableClusterUpdateInternal`

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> OutputTuple:
+        return (
```
MartinBelthle, line 57: Here I think we should update the study_data object (to update the cache with the new values)

antarest/study/business/model/renewable_cluster_model.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    Model representing the external data structure required to edit an existing renewable cluster.
+    """
+
+    class Config:
+        populate_by_name = True
+
+
+@all_optional_model
+class RenewableClusterUpdateInternal(RenewableProperties):
```
sylvlecl: I think this additional class is not necessary, is it ?
`RenewableClusterUpdate` should be enough

MartinBelthle (changes requested):

antarest/study/business/model/renewable_cluster_model.py
sylvlecl: I would remove both, this logic can exists by itself without the API layer (if you mean REST API), it does not need to refer to it.

antarest/study/business/model/renewable_cluster_model.py
sylvlecl: I prefer not, it's clear this way I think, it does not need to know about the web API layer

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
sylvlecl: Agreed, we can have `RenewableClusterUpdate` here. This will allow to validate the input data.

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
```diff
@@ -0,0 +1,86 @@
...
+            {},
+        )
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        path = _RENEWABLE_CLUSTER_PATH.format(area_id=self.area_id, cluster_id=self.cluster_id)
+
+        properties = study_data.tree.get(path.split("/"), depth=1)
+
+        new_properties = RenewableClusterUpdateInternal.model_validate(self.properties).model_dump(
```
sylvlecl: Here we will not need it anymore with `self.properties` typed as `RenewableClusterUpdate`

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
```diff
@@ -0,0 +1,86 @@
...
+                message=f"The renewable cluster '{self.cluster_id}' in the area '{self.area_id}' has been updated.",
+            ),
+            {},
+        )
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        path = _RENEWABLE_CLUSTER_PATH.format(area_id=self.area_id, cluster_id=self.cluster_id)
+
+        properties = study_data.tree.get(path.split("/"), depth=1)
```
sylvlecl: Here it would be best to validate what we get as `RenewablePropertiesType` with `create_renewable_properties`

antarest/study/business/model/renewable_cluster_model.py
TheoPascoli: We dont, children inherit from those properties unless they are overridden. If a parent override those, we will not have them here anymore so I think we should keep it

antarest/study/business/model/renewable_cluster_model.py
TheoPascoli: I removed the suffix

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
TheoPascoli: I typed it again with `RenewableClusterUpdate`

antarest/study/business/model/renewable_cluster_model.py
TheoPascoli: I removed it

antarest/study/business/areas/renewable_management.py
```diff
@@ -248,37 +198,24 @@ def update_cluster(
...
-        ]
-        study.add_commands(commands)
+        study.add_commands([command])
+
+        file_study = study.get_files()
+        values = {}
+        for renewable in file_study.config.areas[area_id].renewables:
+            if renewable.id == cluster_id:
+                values = renewable.model_dump(exclude={"id"})
+                for key, value in cluster_data.model_dump(exclude_unset=True, exclude_none=True).items():
```
MartinBelthle: I think you can replace this `for` loop which is a bit weird with an update:
```python
values = renewable.model_dump(exclude={"id"})
values.update(cluster_data.model_dump(exclude_unset=True, exclude_none=True))
```
Also you can add a break to be a bit faster

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
```diff
@@ -0,0 +1,93 @@
...
+    area_id: str
+    cluster_id: str
+    properties: RenewableClusterUpdate
+
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> OutputTuple:
+        for index, renewable in enumerate(study_data.areas[self.area_id].renewables):
+            if renewable.id == self.cluster_id:
+                properties = renewable.model_dump()
+                for key, value in self.properties.model_dump(exclude_unset=True, exclude_none=True).items():
```
MartinBelthle: Same remark here

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
```diff
@@ -0,0 +1,93 @@
...
+        path = _RENEWABLE_CLUSTER_PATH.format(area_id=self.area_id, cluster_id=self.cluster_id)
+
+        current_renewable_config = create_renewable_config(
+            study_version=self.study_version, **study_data.tree.get(path.split("/"))
+        )
+
+        current_properties = current_renewable_config.model_dump(exclude={"id"})
+        current_properties.update(self.properties.model_dump(exclude_unset=True))
+
+        study_data.tree.save(current_properties, path.split("/"))
```
MartinBelthle: Minor: you can do the split only once (line 64)

tests/integration/study_data_blueprint/test_renewable.py
```diff
@@ -439,7 +439,7 @@ def test_lifecycle(
                 "tsInterpretation": "production-factor",
             },
         )
-        assert res.status_code == 404, res.json()
+        assert res.status_code == 500, res.json()
```
MartinBelthle: Not sure what caused this change but we have to keep 404 errors here.
Actually I think it's because you're applying the command before knowing if the cluster exists

MartinBelthle (changes requested):

MartinBelthle (dismissed):

TheoPascoli (commented):
Do not merge. It does not work right now

MartinBelthle (approved):

PR #2353, opened 2025-02-19
fix(ini): ignore case when performing update and delete of ini files

sylvlecl:
ANT-2732

STS manager creates INI modifications which use the lower case ID of the STS
for the INI file modification.

However, INI file nodes don't handle case consistently:
 - when retrieving data, we use case insensitive matching
 - when updating or removing data, we use exact matching

As a consequence, the update of STS which have capitals in their INI section do not work
correctly, because the update_config command will create a new INI section in
lower case, instead of updating the existing one.

**The quick fix consists in doing case insensitive matching also for updates and deletes.**


antarest/study/storage/rawstudy/model/filesystem/ini_file_node.py
```diff
@@ -249,33 +391,7 @@ def delete(self, url: Optional[List[str]] = None) -> None:
...
-                    self.config,
-                    f"Cannot delete section: Section [{section_name}] not found",
-                ) from None
-
-        elif url_len == 2:
-            section_name, key_name = url
-            try:
-                section = data[section_name]
-            except KeyError:
-                raise IniFileNodeWarning(
```
smailio, line 267: So now when the user sens a path that doesn't exist instead of the IniFileNodeWarning we'll get a KeyError raised by one of the get_part function above ?

sylvlecl: Actually the warnings were only used for the `delete` part, to log instead of getting an error. Indeed I should restore them to keep the same behaviour, thanks!

sylvlecl: I restored the warning logs in last commits

TheoPascoli (approved):

PR #2356, opened 2025-02-19
feat(thermal): add update thermal cluster command

TheoPascoli:


antarest/study/business/areas/thermal_management.py
```diff
@@ -330,36 +262,38 @@ def update_cluster(
...
+
+        thermal_cluster = next((thermal for thermal in area.thermals if thermal.id == cluster_id), None)
+        if thermal_cluster is None:
+            raise ThermalClusterNotFound(path, cluster_id)
+
+        versionned_cluster = create_thermal_config(
+            study_version=study.version, **thermal_cluster.model_dump(exclude_unset=True, exclude_none=True)
+        )
+
+        updated_thermal = create_thermal_config(
```
MartinBelthle: This is a bit harsh. Splitting this in several lines would help readability I think

antarest/study/storage/variantstudy/model/command/update_thermal_cluster.py
```diff
@@ -0,0 +1,107 @@
...
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> OutputTuple:
+        for index, thermal in enumerate(study_data.areas[self.area_id].thermals):
+            if thermal.id == self.thermal_cluster_id:
+                # Set the object to the correct version
+                versioned_cluster = create_thermal_config(
+                    study_version=self.study_version, **thermal.model_dump(exclude_unset=True, exclude_none=True)
+                )
+                # Update the object with the new properties
+                updated_thermal = create_thermal_config(
```
MartinBelthle: Same remark

MartinBelthle (commented):
Except the 2 comments on readability, seems ok to merge as it's the same code as the one for renewables

antarest/study/storage/variantstudy/model/command/update_thermal_cluster.py
```diff
@@ -0,0 +1,107 @@
...
+                    study_version=self.study_version,
+                    **versioned_cluster.model_copy(
+                        update=self.properties.model_dump(exclude_unset=True, exclude_none=True)
+                    ).model_dump(exclude_unset=True, exclude_none=True),
+                )
+
+                study_data.areas[self.area_id].thermals[index] = updated_thermal
+                break
+
+        return (
```
MartinBelthle, line 62: Also I just realized that here we should probably return a status=False if we don't find the right cluster. Meaning putting this return instead of the break and putting a False here

antarest/study/storage/variantstudy/model/command/update_thermal_cluster.py
```diff
@@ -0,0 +1,107 @@
...
+    area_id: str
+    thermal_cluster_id: str
+    properties: ThermalClusterUpdate
+
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> OutputTuple:
+        for index, thermal in enumerate(study_data.areas[self.area_id].thermals):
+            if thermal.id == self.thermal_cluster_id:
+                # Set the object to the correct version
+                versioned_cluster = create_thermal_config(
```
MartinBelthle: Also, here we should probably use `create_thermal_properties` instead of config no ?

TheoPascoli: I need the ID to save the new object so I think I need a config object. However, I replaced with properties in the manager because we excluded the ID at the end anyway

antarest/study/storage/variantstudy/model/command/update_thermal_cluster.py
TheoPascoli: Indeed

antarest/study/storage/variantstudy/model/command/update_thermal_cluster.py
MartinBelthle: Mmmh right

antarest/study/business/areas/thermal_management.py
```diff
@@ -41,85 +43,16 @@
 from antarest.study.storage.variantstudy.model.command.replace_matrix import ReplaceMatrix
 from antarest.study.storage.variantstudy.model.command.update_config import UpdateConfig
 
-__all__ = (
-    "ThermalClusterInput",
-    "ThermalClusterCreation",
-    "ThermalClusterOutput",
-    "ThermalManager",
-)
+__all__ = "ThermalManager"
```
MartinBelthle: Minor: perhaps this __all__ is not needed now. This would improve readability IMO

antarest/study/business/areas/thermal_management.py
```diff
@@ -330,36 +263,41 @@ def update_cluster(
...
+            data=thermal_cluster.model_dump(exclude_unset=True, exclude_none=True, exclude={"id"}),
+        )
+
+        updated_versioned_thermal = versioned_thermal.model_copy(
+            update=cluster_data.model_dump(exclude_unset=True, exclude_none=True)
+        )
+
+        thermal_cluster_properties = create_thermal_properties(
+            study_version=study.version,
+            data=updated_versioned_thermal.model_dump(),
```
MartinBelthle: Don't you have issues with the fact that you're not using mode  = 'json' ?

antarest/study/storage/variantstudy/model/command/update_thermal_cluster.py
```diff
@@ -0,0 +1,120 @@
...
+    # Command parameters
+    # ==================
+
+    area_id: str
+    thermal_cluster_id: str
+    properties: ThermalClusterUpdate
+
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> OutputTuple:
+        for index, thermal in enumerate(study_data.areas[self.area_id].thermals):
```
MartinBelthle, line 51: Same remark as the one for st storage about the architecture

antarest/study/storage/variantstudy/model/command/update_thermal_cluster.py
```diff
@@ -0,0 +1,120 @@
...
+                "properties": self.properties.model_dump(mode="json", exclude_unset=True),
+            },
+            study_version=self.study_version,
+        )
+
+    @override
+    def get_inner_matrices(self) -> List[str]:
+        return []
+
+    def update_thermal_config(self, thermal: ThermalConfigType) -> ThermalConfigType:
```
MartinBelthle, line 104: This is pretty nice and documented. Could you do the same inside the ST storage PR ?

antarest/study/business/areas/thermal_management.py
TheoPascoli: Indeed

antarest/study/business/areas/thermal_management.py
TheoPascoli: They are totally equals

antarest/study/storage/variantstudy/model/command/update_thermal_cluster.py
TheoPascoli: sure

antarest/study/storage/variantstudy/model/command/update_thermal_cluster.py
TheoPascoli: This is better

antarest/study/business/model/thermal_cluster_model.py
```diff
@@ -0,0 +1,61 @@
...
+        if not name:
+            raise ValueError("name must not be empty")
+        return name
+
+    def to_properties(self, study_version: StudyVersion) -> ThermalPropertiesType:
+        values = self.model_dump(mode="json", by_alias=False, exclude_none=True)
+        return create_thermal_properties(study_version=study_version, data=values)
+
+
+@all_optional_model
```
sylvlecl: I know it was this way before, but I don't think we should have optionals here, the optionals should be only for user input (updates, creation), not for outputs, since we will provide default values.

tests/integration/study_data_blueprint/test_thermal.py
```diff
@@ -953,7 +953,7 @@ def test_variant_lifecycle(self, client: TestClient, user_access_token: str, var
         assert actions == [
             "create_area",
             "create_cluster",
-            "update_config",
+            "update_thermal_cluster",
```
sylvlecl, line 956: nice  :)

antarest/study/storage/variantstudy/model/command/update_thermal_cluster.py
```diff
@@ -0,0 +1,118 @@
...
+        # Set the object to the correct version
+        versioned_thermal = create_thermal_config(
+            study_version=self.study_version, **thermal.model_dump(exclude_unset=True, exclude_none=True)
+        )
+        # Update the object with the new properties
+        updated_versioned_thermal = versioned_thermal.model_copy(
```
sylvlecl, line 110: Not a MUST, but I really think we should introduce a utility method for this update of pydantic models, because it's not obvious to understand what it does, and to know what parameters should be used (exclude_unset etc).

sylvlecl (changes requested):
Very minor comments + we should add a test in test_command_factory for this new command

antarest/study/storage/variantstudy/model/command/update_thermal_cluster.py
TheoPascoli: I removed a block that was completely useless and I think it is easier to understand now.

sylvlecl (approved):

PR #2359, opened 2025-02-20
chore(patch)!: remove patch service

MartinBelthle:
This PR introduces several breaking changes:
- Removes the endpoint PUT  `/studies/{uuid}/areas/{area_id}`
- Removes the endpoint PUT `/studies/{study_id}/outputs/{output_id}/reference`
- Inside endpoint PUT `/studies/{uuid}` , removes the object properties (body and response) `scenario`, `status` and `doc`
- Same for the GET endpoint response
- Inside endpoint GET `/studies/{study_id}/outputs`, removes the args `referenceStatus` and `synchronized` from the object response
- Inside endpoint GET `/studies/{uuid}/areas` when called with `ui=False`, change response structure to use camelcase instead of ini-case


smailio (approved):
looks good to me

PR #2360, opened 2025-02-20
chore(dev): share pre-commit configuration for local dev

sylvlecl:
Share configuration file for pre-commit tool, which is useful in dev environment
for executing mypy and ruff on commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,28 @@
+repos:
+-   repo: local
+    hooks:
+    -   id: mypy
+        name: Run mypy
+        entry: mypy
+        language: system
+        files: ^antarest/
+        types: [python]
+
+-   repo: local
+    hooks:
+    -   id: ruff-format
+        name: Ruff format
+        entry: ruff
+        args: ["format", "antarest", "tests"]
+        language: system
+        types: [python]
+
+-   repo: local
+    hooks:
+    -   id: ruff-check
+        name: Ruff check
+        entry: ruff
+        args: ["check", "--fix", "antarest", "tests"]
+        language: system
+        types: [python]
+
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,6 +2,7 @@
 -r requirements-desktop.txt
 ruff~=0.9.6
 mypy~=1.15.0
+pre-commit~=4.1.0
 pyinstaller==6.10.0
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+-   repo: local
+    hooks:
+    -   id: ruff-check
+        name: Ruff check
+        entry: ruff
+        args: ["check", "--fix", "antarest", "tests"]
+        language: system
+        types: [python]
+
```
smailio, line 28: FYI on my local conf I also use isort, ruff replace isort but I still use isort to be sure I'm consistent with the check done at integration. 

```
  - repo: https://github.com/pycqa/isort
    rev: 7de182933fd50e04a7c47cc8be75a6547754b19c
    hooks:
      - id: isort
        name: isort (python)

```

We can try just ruff for now though. If the issue never arises there's no point to configure an additional hook.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,28 @@
+repos:
+-   repo: local
```
smailio, line 2: Just out of curiosity, why you choose to use local repo ?

.pre-commit-config.yaml
MartinBelthle: Not sure to understand what you mean by `the check done at integration` but since we moved to ruff we no longer use isort in the CI

.pre-commit-config.yaml
```diff
@@ -0,0 +1,28 @@
...
+    hooks:
+    -   id: mypy
+        name: Run mypy
+        entry: mypy
+        language: system
+        files: ^antarest/
```
MartinBelthle, line 8: Why do we have to specify the files here ? As running mypy in command line already understands it only concerns the antarest folder

sylvlecl: I THINK that it allows precommit to not run this check at all when files are modified outside of this folder.

TheoPascoli (approved):

.pre-commit-config.yaml
smailio: I meant the checks done by github action. You're right as we use ruff the point doesn't hold anymore.

.pre-commit-config.yaml
sylvlecl: It was easy to setup this way, I did not find another solution which looked "standard".

PR #2367, opened 2025-02-24
feat(xpansion): add candidate commands

MartinBelthle:


antarest/study/storage/variantstudy/model/command/update_xpansion_candidate.py
```diff
@@ -0,0 +1,84 @@
...
+    # Overloaded metadata
+    # ===================
+
+    command_name: CommandName = CommandName.UPDATE_XPANSION_CANDIDATE
+
+    # Command parameters
+    # ==================
+
+    candidate_name: str
+    new_properties: XpansionCandidateInternal
```
sylvlecl: for consistency sake with other commands, we'd better name this properties

antarest/study/business/model/xpansion_model.py
```diff
@@ -213,3 +220,63 @@ class XpansionSettingsUpdate(XpansionSettings):
...
+        return name
+
+    @field_validator("link", mode="before")
+    def validate_link(cls, link: str) -> str:
+        if " - " not in link:
+            raise WrongLinkFormatError(f"The link must be in the format 'area1 - area2'. Currently: {link}")
+        return link
+
+
+class XpansionCandidateInternal(XpansionCandidateBase, alias_generator=to_kebab_case):
```
sylvlecl: we'd better use plain names, not "Internal": `XpansionCandidateUpdate` and `XpansionCandidateCreation` for example.
The 2 differ at least by the fact that name is optional for the update but not the creation.
Also, link and annual_cost_per_mw should be optionals, I guess.

antarest/study/storage/variantstudy/model/command/xpansion_common.py
```diff
@@ -50,3 +56,38 @@ def apply_create_resource_commands(
...
+        "already_installed_direct_link_profile",
+        "already_installed_indirect_link_profile",
+    ]:
+        if link_file := getattr(xpansion_candidate_dto, attr, None):
+            if link_file not in existing_files:
+                raise XpansionFileNotFoundError(f"The '{attr}' file '{link_file}' does not exist")
+
+
+def assert_link_exist(file_study: FileStudy, xpansion_candidate_dto: XpansionCandidateInternal) -> None:
+    area1, area2 = xpansion_candidate_dto.link.split(" - ")
```
sylvlecl: We should not need to know in several places that the serialization format of the link is "area1 - area2":
it would be better to have a small model with area1 and area2 field, and a parser and a serializer to format it into the desired format (both in the DTO and in the business layer classes)

antarest/study/storage/variantstudy/model/command/create_xpansion_candidate.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Overloaded metadata
+    # ===================
+
+    command_name: CommandName = CommandName.CREATE_XPANSION_CANDIDATE
+
+    # Command parameters
+    # ==================
+
+    candidate: XpansionCandidateInternal
```
sylvlecl: See other comments, we should have a XpansionCandidateCreation

antarest/study/storage/variantstudy/model/command/create_xpansion_candidate.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
+        return CommandOutput(status=True, message="ok"), {}
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        candidates_obj = study_data.tree.get(["user", "expansion", "candidates"])
```
sylvlecl: I think we should use a pydantic model to validate what we receive from the file study tree

antarest/study/storage/variantstudy/model/command/update_xpansion_candidate.py
sylvlecl: See other comments, we'd better have a `XpansionCandidateUpdate` here, with optional fields

antarest/study/storage/variantstudy/model/command/update_xpansion_candidate.py
```diff
@@ -0,0 +1,84 @@
...
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
+        return CommandOutput(status=True, message="ok"), {}
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        # Checks candidate validity
+        assert_candidate_is_correct(study_data, self.new_properties)
+        candidates = study_data.tree.get(["user", "expansion", "candidates"])
+        if self.new_properties.name != self.candidate_name:
```
sylvlecl: for consistency with other commands, we should not be required to have the name in the command AND in the properties, if we don't want to change it. We will need to take into account here the possibility that name is None

tests/variantstudy/model/command/test_xpansion_candidates.py
```diff
@@ -0,0 +1,241 @@
...
+max-units = 7
+
+"""
+        )
+
+        # Updates one
+        cmd = UpdateXpansionCandidate(
+            candidate_name="cdt_1",
+            new_properties=XpansionCandidateInternal(
+                name="cdt_1", link="at - be", annual_cost_per_mw=30, max_investment=100, direct_link_profile="capa1.txt"
```
sylvlecl, line 98: we should not need to duplicate the name and the link definition here

sylvlecl (changes requested):
Changes requested on naming and on optional fields for update

antarest/study/business/model/xpansion_model.py
```diff
@@ -213,3 +220,83 @@ class XpansionSettingsUpdate(XpansionSettings):
...
+
+        illegal_name_characters = [" ", "\n", "\t", "\r", "\f", "\v", "-", "+", "=", ":", "[", "]", "(", ")"]
+        for char in name:
+            if char in illegal_name_characters:
+                raise IllegalCharacterInNameError(f"The character '{char}' is not allowed in the candidate name")
+
+        return name
+
+    @field_validator("link", mode="before")
+    def validate_link(cls, link: str) -> str:
```
sylvlecl: it's better to move it to the xpansion_link validator

antarest/study/storage/variantstudy/model/command/replace_xpansion_candidate.py
```diff
@@ -0,0 +1,90 @@
...
+    # Overloaded metadata
+    # ===================
+
+    command_name: CommandName = CommandName.REPLACE_XPANSION_CANDIDATE
+
+    # Command parameters
+    # ==================
+
+    candidate_name: str
+    new_properties: XpansionCandidate
```
sylvlecl: still think we should rename it to just properties for consistency sake

sylvlecl (changes requested):
Very minor stuff

antarest/study/business/model/xpansion_model.py
sylvlecl: As discussed, the current behaviour is to actually REPLACE the whole candidate, because sometimes we want to remove fields. With an update class as usual, this would not be possible.
Hence we decided to renamed the command as `ReplaceXpansionCandidate` because it's different from other update commands.

antarest/study/storage/variantstudy/model/command/replace_xpansion_candidate.py
MartinBelthle: Indeed

PR #2368, opened 2025-02-25
feat(xpansion): add `remove_xpansion_resource` command

MartinBelthle:


antarest/study/business/xpansion_management.py
```diff
@@ -21,23 +21,25 @@
...
+    XpansionSettingsUpdate,
+)
 from antarest.study.business.study_interface import StudyInterface
 from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
 from antarest.study.storage.variantstudy.model.command.create_xpansion_configuration import CreateXpansionConfiguration
 from antarest.study.storage.variantstudy.model.command.remove_xpansion_configuration import RemoveXpansionConfiguration
+from antarest.study.storage.variantstudy.model.command.remove_xpansion_resource import (
+    RemoveXpansionResource,
+    checks_resource_deletion_is_allowed,
+    get_resource_dir,
```
sylvlecl: a little weird to have it in the "remove" command, maybe we should have a "utils" for xpansion related files ?

sylvlecl (changes requested):
We need to add some unit tests for command_factory.
Ideally we should have some for the command itself too, like test_remove_xpansion_resource

PR #2369, opened 2025-02-25
fix(xpansion): allow matrices normalization

### LINKED ISSUE #2362: Upgrading Antares Xpansion study bug
1. You can import study SmallTestFiveCandidates which is one of the examples when you download Antares Xpansion. That study has 8.0 Antares version tag.
2. In debug mode you can clearly see TS as it is shown on following screenshot:

Image

3. After upgrading Antares study to 8.8 version, you can no longer see TS in debug mode or in the UI and files get .link extension, ass it is shown on following screenshot:

Image

4. When i download that study on my machine in corrupted/modified files I found address of matrix as it is show on following screenshot:

Image

MartinBelthle:
Fix [ANT-2881]

### DIFF
--- a/antarest/study/storage/rawstudy/model/filesystem/root/user/expansion/matrix_resources.py
+++ b/antarest/study/storage/rawstudy/model/filesystem/root/user/expansion/matrix_resources.py
@@ -9,13 +9,26 @@
 # SPDX-License-Identifier: MPL-2.0
 #
 # This file is part of the Antares project.
+from typing_extensions import override
 
 from antarest.study.storage.rawstudy.model.filesystem.bucket_node import BucketNode
 from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfig
 from antarest.study.storage.rawstudy.model.filesystem.context import ContextServer
+from antarest.study.storage.rawstudy.model.filesystem.inode import TREE
 from antarest.study.storage.rawstudy.model.filesystem.matrix.input_series_matrix import InputSeriesMatrix
 
 
 class ExpansionMatrixResources(BucketNode):
     def __init__(self, context: ContextServer, config: FileStudyTreeConfig):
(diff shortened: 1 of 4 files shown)


antarest/study/storage/rawstudy/model/filesystem/root/user/expansion/matrix_resources.py
```diff
@@ -9,13 +9,26 @@
...
     def __init__(self, context: ContextServer, config: FileStudyTreeConfig):
         super().__init__(context, config, None, default_file_node=InputSeriesMatrix)
+
+    @override
+    def build(self) -> TREE:
+        children: TREE = {}
+        for resource in self.config.path.iterdir():
+            if resource.is_file():
+                # Removes the .link suffix if the matrix is normalized
+                file_name = resource.name.removesuffix(".link")
```
sylvlecl: I don't hink we should have to do this here, what's the purpose ?
If the file as a .link extension, it should not be a valid content for the matrix, so I expect this will not work correctly ?
In any case, It seems that it should be handled by the InputSeriesMatrix class instead, the .link stuff should not be known outside of it.

sylvlecl (changes requested):

antarest/study/storage/rawstudy/model/filesystem/root/user/expansion/matrix_resources.py
MartinBelthle: The issue is that the node was not instanciated wit the right config. Its attribute `self.config.path` contained the .link if it was normalized which is not supposed to be the case
Elsewhere in the code we build the children based on the areas or we assume we know the file names. Here we don't and can't. (only other place like this is the outputs where we can't have normalized matrices).

The other way around could be to perform this removal at the input_series_matrix instantiation
Not sure it's better

PR #2370, opened 2025-02-25
fix(hydro_management): fix read and write in hydro.ini

maugde:
Fix [ANT-2734](https://gopro-tickets.rte-france.com/browse/ANT-2734)

Antares Web use area_id to retrieve and save data about areas. However, as described in the ticket, in the case of an area creating using another tool, if this area has capital characters in its name, AW won't be able to recognize it.

In this PR, we allow the program to read and save the area_id regardless its case.

Two tests for read and save operations has been added too.

### DIFF
--- a/antarest/study/business/areas/hydro_management.py
+++ b/antarest/study/business/areas/hydro_management.py
@@ -117,25 +117,25 @@ class HydroManager:
 
     # @staticmethod
     # def get_id(area_id: str, field_dict: Dict[str, FieldInfo]) -> str:
-        """
-        Try to match the current area_id with the one from the original file.
-        These two ids could mismatch based on their character cases since the id from
-        the filesystem could have been modified with capital letters.
-        We first convert it into lower case in order to compare both ids.
+    """
+    Try to match the current area_id with the one from the original file.
+    These two ids could mismatch based on their character cases since the id from
+    the filesystem could have been modified with capital letters.
+    We first convert it into lower case in order to compare both ids.
 
-        Returns the area id from the file if both values matched, the initial area id otherwise.
-        """
-     #   for file_area_id in field_dict:
-     #        if LowerCaseIdentifier.generate_id(file_area_id) == area_id:
-     #            return file_area_id
-     #    return area_id
+    Returns the area id from the file if both values matched, the initial area id otherwise.
+    """
+    #   for file_area_id in field_dict:
+    #        if LowerCaseIdentifier.generate_id(file_area_id) == area_id:
+    #            return file_area_id
+    #    return area_id
 
     # def get_hydro_config(self, study) -> Dict[str, Dict[str, FieldInfo]]:
-        """
-            Returns a dictionary of hydro configurations
-        """
-        # file_study = self.storage_service.get_storage(study).get_raw(study)
-        # return file_study.tree.get(HYDRO_PATH.split("/"))
+    #     """
+    #         Returns a dictionary of hydro configurations
+    #     """
+    #   file_study = self.storage_service.get_storage(study).get_raw(study)
+    #   return file_study.tree.get(HYDRO_PATH.split("/"))
 
     def __init__(self, command_context: CommandContext) -> None:
         self._command_context = command_context
--- a/antarest/study/business/areas/hydro_management.py
+++ b/antarest/study/business/areas/hydro_management.py
@@ -17,8 +17,11 @@ from pydantic import Field
 from antarest.study.business.all_optional_meta import all_optional_model
 from antarest.study.business.study_interface import StudyInterface
 from antarest.study.business.utils import FieldInfo, FormFieldsBaseModel
+#from antarest.study.model import Study
+# from antarest.study.storage.storage_service import StudyStorageService
 from antarest.study.storage.variantstudy.model.command.update_config import UpdateConfig
 from antarest.study.storage.variantstudy.model.command_context import CommandContext
+# from antarest.study.storage.rawstudy.model.filesystem.config.identifier import LowerCaseIdentifier
 
 INFLOW_PATH = "input/hydro/prepro/{area_id}/prepro/prepro"
 
@@ -108,6 +111,32 @@ FIELDS_INFO: Dict[str, FieldInfo] = {
 
 
 class HydroManager:
+    # def __init__(self, storage_service: StudyStorageService) -> None:
+    #    self.storage_service = storage_service
+    #    self.area_id = ""
+
+    # @staticmethod
+    # def get_id(area_id: str, field_dict: Dict[str, FieldInfo]) -> str:
+        """
+            Try to match the current area_id with the one from the original file.
+            These two ids could mismatch based on their character cases since the id from
+            the filesystem could have been modified with capital letters.
+            We first convert it into lower case in order to compare both ids.
+
+            Returns the area id from the file if both values matched, the initial area id otherwise.
+        """
+     #   for file_area_id in field_dict:
+     #        if LowerCaseIdentifier.generate_id(file_area_id) == area_id:
+     #            return file_area_id
+     #    return area_id
+
+    # def get_hydro_config(self, study) -> Dict[str, Dict]:
+        """
+            Returns a dictionary of hydro configurations
+        """
+        # file_study = self.storage_service.get_storage(study).get_raw(study)
+        # return file_study.tree.get(HYDRO_PATH.split("/"))
+
     def __init__(self, command_context: CommandContext) -> None:
         self._command_context = command_context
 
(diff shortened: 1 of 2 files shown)


tests/study/business/areas/test_hydro_management.py
```diff
@@ -0,0 +1,183 @@
...
+                },
+            }
+        }
+    }
+}
+
+
+class TestHydroManagement:
+    @staticmethod
+    def create_study_service(
```
smailio: Why this isn't a fixture ?

tests/study/business/areas/test_hydro_management.py
```diff
@@ -0,0 +1,183 @@
...
+
+        study = self.create_study_interface(tmp_path)
+
+        # add som areas
+        areas = ["AreaTest1", "AREATEST2", "area_test_3"]
+
+        # gather initial data of the area
+        for area in areas:
+            # get actual value
+            data = study_service.hydro_manager.get_field_values(study, area).dict()
```
smailio: dict is depreceated use model_dump

antarest/study/business/areas/hydro_management.py
```diff
@@ -164,6 +198,7 @@ def get_inflow_structure(self, study: StudyInterface, area_id: str) -> InflowStr
         """
         # NOTE: Focusing on the single field "intermonthly-correlation" due to current model scope.
         path = INFLOW_PATH.format(area_id=area_id)
+        # file_study = self.storage_service.get_storage(study).get_raw(study)
```
smailio: remove comment

tests/study/business/areas/test_hydro_management.py
```diff
@@ -0,0 +1,183 @@
...
+        tree.save(hydro_ini_content)
+
+        file_study = Mock(spec=FileStudy, config=config, tree=FileStudyTree(Mock(spec=ContextServer), config))
+
+        study = Mock(spec=StudyInterface, version=860)
+        study.get_files.return_value = file_study
+        return study
+
+    @pytest.mark.unit_test
+    def test_get_field_values(
```
smailio: Both tests pass when I comment the get_id, reformat_hydro_config code. 

I'm not sure the test properly cover the use case ? 
image

antarest/study/business/areas/hydro_management.py
```diff
@@ -110,18 +111,42 @@ class ManagementOptionsFormFields(FormFieldsBaseModel):
 class HydroManager:
     def __init__(self, command_context: CommandContext) -> None:
         self._command_context = command_context
+        self.area_id = ""
```
TheoPascoli: You never use this variable. I think you can delete it

antarest/study/business/areas/hydro_management.py
```diff
@@ -110,18 +111,42 @@ class ManagementOptionsFormFields(FormFieldsBaseModel):
...
+        """
+        Try to match the current area_id with the one from the original file.
+        These two ids could mismatch based on their character cases since the id from
+        the filesystem could have been modified with capital letters.
+        We first convert it into lower case in order to compare both ids.
+
+        Returns the area id from the file if both values matched, the initial area id otherwise.
+        """
+        for file_area_id in field_dict:
+            if LowerCaseIdentifier.generate_id(file_area_id) == area_id:
```
TheoPascoli: What is the purpose of using `LowerCaseIdentifier` instead of `file_area_id.lower()` ?

antarest/study/business/areas/hydro_management.py
```diff
@@ -110,18 +111,42 @@ class ManagementOptionsFormFields(FormFieldsBaseModel):
...
+                return file_area_id
+        return area_id
```
TheoPascoli: Also, for this iterator, you can do it in a more "pythonic" way with the `next()` method.
Something like: 
```
return next(
        (file_area_id for file_area_id in field_dict
         if file_area_id.lower() == area_id),
        area_id
    )
```
You can check the documentation or ask me if you need more informations

antarest/study/business/areas/hydro_management.py
```diff
@@ -110,18 +111,42 @@ class ManagementOptionsFormFields(FormFieldsBaseModel):
...
+
+    @staticmethod
+    def get_hydro_config(study: StudyInterface) -> Dict[str, Dict[str, FieldInfo]]:
```
TheoPascoli: You can add a underscore in front of the method name to make it private, as you only use it in this class

antarest/study/business/areas/hydro_management.py
```diff
@@ -138,13 +163,22 @@ def set_field_values(
         """
         commands: List[UpdateConfig] = []
 
+        hydro_config = self.get_hydro_config(study)
+
+        # create a reformatted hydro config dictionary so the fields name can match `FIELDS_INFO` fields name
+        reformat_hydro_config = {}
+        for key, value in hydro_config.items():
+            reformat_hydro_config[key.replace("-", "_").replace(" ", "_")] = value
```
TheoPascoli: Do you really need those three lines ? I removed them and the unit tests still passes

TheoPascoli (changes requested):
Hello, nice PR ! I have just small remarks. Feel free to ask me if you need more informations

antarest/study/business/areas/hydro_management.py
sylvlecl: Actually @maugde , I think the whole "setter" part is not needed anymore following this evolution, which make the INI file updates case-insensitive:
https://github.com/AntaresSimulatorTeam/AntaREST/pull/2353

So I think we can remove those changes. Sorry for not spotting that common part earlier.

tests/study/business/areas/test_hydro_management.py
```diff
@@ -0,0 +1,176 @@
...
+            study_service.hydro_manager.set_field_values(study, new_field_values, area)
+
+            # retrieve edited
+            new_data = study_service.hydro_manager.get_field_values(study, area).model_dump()
+
+            # check if the intra daily modulation was modified
+            for field, value in new_data.items():
+                if field == "intra_daily_modulation":
+                    assert initial_data[field] != new_data[field]
+                    assert initial_data[field] == 5
```
sylvlecl: I think this should test that `new_data` is 5 instead

tests/study/business/areas/test_hydro_management.py
```diff
@@ -0,0 +1,176 @@
...
+            new_field_values = ManagementOptionsFormFields(**initial_data)
+            study_service.hydro_manager.set_field_values(study, new_field_values, area)
+
+            # retrieve edited
+            new_data = study_service.hydro_manager.get_field_values(study, area).model_dump()
+
+            # check if the intra daily modulation was modified
+            for field, value in new_data.items():
+                if field == "intra_daily_modulation":
+                    assert initial_data[field] != new_data[field]
```
sylvlecl, line 148: here I think we should have the equality, since the `initial_data` has been modified with intra_daily_modulation = 5

tests/study/business/areas/test_hydro_management.py
```diff
@@ -0,0 +1,176 @@
...
+                    "AreaTest1": 1,
+                },
+            }
+        }
+    }
+}
+
+
+@pytest.fixture(name="study_service")
+def study_service_fixture(
```
sylvlecl: Actually now we don't want the `StudyService` anymore in that part of the code, only `StudyInterface` is needed.
You can just build the `HydroManager` to test it, without the study service, see how it's done in other managers tests.

tests/study/business/areas/test_hydro_management.py
```diff
@@ -0,0 +1,176 @@
...
+def study_interface_fixture(tmp_path: Path) -> StudyInterface:
+    study_id = "study_test"
+    study_path = tmp_path.joinpath(f"tmp/{study_id}")
+    config = build(study_path, study_id)
+    tree = FileStudyTree(Mock(spec=ContextServer), config)
+    tree.save(hydro_ini_content)
+
+    file_study = Mock(spec=FileStudy, config=config, tree=FileStudyTree(Mock(spec=ContextServer), config))
+
+    study = Mock(spec=StudyInterface, version=860, path=study_path, id=study_id)
```
sylvlecl: Here you can use a `FileStudyInterface` instead of a mock:
this way, the underlying file will be actually modified by the commands.

sylvlecl (changes requested):
A few adaptations are needed, in particular in the unit test which I think does not check what we want

antarest/study/business/areas/hydro_management.py
maugde: The purpose was to use a validator in case the id needed more validation steps. However it is indeed not really useful in this case so I'll change for `.lower()` instead.

sylvlecl (approved):

PR #2371, opened 2025-02-26
feat(xpansion): add `create_xpansion_resource` commands

MartinBelthle:


antarest/study/business/xpansion_management.py
```diff
@@ -381,49 +388,50 @@ def update_xpansion_constraints_settings(
...
-
-        file_study.tree.save(data)
-
     def add_resource(
         self,
         study: StudyInterface,
         resource_type: XpansionResourceFileType,
-        files: List[UploadFile],
+        file: UploadFile,
     ) -> None:
```
TheoPascoli: This method returns None whereas it is supposed to return "Any" in the API definition

antarest/study/storage/variantstudy/model/command/create_xpansion_matrix.py
```diff
@@ -0,0 +1,93 @@
...
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand
+from antarest.study.storage.variantstudy.model.command.xpansion_common import (
+    apply_config_create_resource_commands,
+    apply_create_resource_commands,
+)
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+class AbstractCreateXpansionMatrix(ICommand):
```
TheoPascoli, line 34: This does not implement all the abstracts methods

antarest/study/business/xpansion_management.py
```diff
@@ -381,49 +388,50 @@ def update_xpansion_constraints_settings(
...
+        # checks the file doesn't already exist
+        keys = get_resource_dir(resource_type)
         file_study = study.get_files()
-        self._add_raw_files(file_study, files, resource_type)
+        if filename in file_study.tree.get(keys):
+            raise XpansionFileAlreadyExistsError(f"File '{filename}' already exists")
+
+        # parses the content
+        content = file.file.read()
+        if isinstance(content, str):
```
TheoPascoli: It can really be a str ?

antarest/study/business/xpansion_management.py
```diff
@@ -381,49 +388,50 @@ def update_xpansion_constraints_settings(
...
-        for file in files:
-            content = file.file.read()
-            if isinstance(content, str):
-                content = content.encode(encoding="utf-8")
-            assert file.filename is not None
-            buffer[file.filename] = content
-
-        file_study.tree.save(data)
-
     def add_resource(
```
TheoPascoli: I think this method is a bit too long. Could you divide it in smaller parts ?

TheoPascoli (changes requested):
Small comments

antarest/study/business/xpansion_management.py
MartinBelthle: Indeed, I'll change ANy to None inside the blueprint

antarest/study/business/xpansion_management.py
MartinBelthle: read() returns `AnyStr` which is `AnyStr = TypeVar('AnyStr', bytes, str)` inside the typing doc

antarest/study/storage/variantstudy/model/command/create_xpansion_matrix.py
MartinBelthle: My IDE says the same thing but mypy is silent. Weird. We have the same issue with AbstratcBindingConstraintCommand but I won't modify it here

antarest/study/storage/variantstudy/model/command/xpansion_common.py
```diff
@@ -20,3 +25,28 @@ def get_resource_dir(resource_type: XpansionResourceFileType) -> list[str]:
     elif resource_type == XpansionResourceFileType.WEIGHTS:
         return ["user", "expansion", "weights"]
     raise NotImplementedError(f"resource_type '{resource_type}' not implemented")
+
+
+def apply_config_create_resource_commands(
```
TheoPascoli, line 30: Do you plan to use those two methods elsewhere ? Why not put them above the class that needs them

TheoPascoli (changes requested):

antarest/study/storage/variantstudy/model/command/xpansion_common.py
MartinBelthle: I use it inside 2 different files, create_constraint and create_matrix. So i put it there to avoid weird dependencies

TheoPascoli (approved):

antarest/study/storage/variantstudy/model/command/create_xpansion_matrix.py
```diff
@@ -0,0 +1,101 @@
...
+    # Command parameters
+    # ==================
+
+    filename: str
+    matrix: List[List[MatrixData]] | str = Field(validate_default=True)
```
sylvlecl, line 39: @MartinBelthle @TheoPascoli an after the fact comment:
we should strive to not have `List[List[float]]` in memory, it's really heavy in memory and slow to process, we should use numpy arrays instead anytime we can

PR #2374, opened 2025-02-28
feat(hydro): add update hydro management option command

TheoPascoli:


antarest/study/business/model/hydro_management_model.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) 2025, RTE (https://www.rte-france.com)
+#
+# See AUTHORS.txt
```
MartinBelthle: I would rename this file hydro_model to fit with the other file names inside this directory

antarest/study/business/model/hydro_management_model.py
```diff
@@ -0,0 +1,56 @@
...
+    inter_monthly_correlation: float = Field(
+        default=0.5,
+        ge=0,
+        le=1,
+        description="Average correlation between the energy of a month and that of the next month",
+        title="Inter-monthly correlation",
+    )
+
+
+class HydroManagementOptions(
```
MartinBelthle: You can use `@all_optional_model`

antarest/study/storage/variantstudy/command_factory.py
```diff
@@ -89,6 +90,7 @@
     CommandName.CREATE_ST_STORAGE.value: CreateSTStorage,
     CommandName.REMOVE_ST_STORAGE.value: RemoveSTStorage,
     CommandName.UPDATE_ST_STORAGE.value: UpdateSTStorage,
+    CommandName.UPDATE_HYDRO_MANAGEMENT.value: UpdateHydroManagement,
```
MartinBelthle: I would rename the command `update_hydro_properties` as I find this name misleading

antarest/study/web/study_data_blueprint.py
```diff
@@ -441,22 +441,22 @@ def remove_district(
...
+    ) -> HydroManagementOptions:
         logger.info(
             msg=f"Getting Hydro management config for area {area_id} of study {uuid}",
             extra={"user": current_user.id},
         )
         params = RequestParameters(user=current_user)
         study = study_service.check_study_access(uuid, StudyPermissionType.READ, params)
         study_interface = study_service.get_study_interface(study)
-        return study_service.hydro_manager.get_field_values(study_interface, area_id)
+        return study_service.hydro_manager.get_hydro_management_options(study_interface, area_id)
```
MartinBelthle: Same i would replace management by properties

antarest/study/web/study_data_blueprint.py
```diff
@@ -441,22 +441,22 @@ def remove_district(
         "/studies/{uuid}/areas/{area_id}/hydro/form",
         tags=[APITag.study_data],
         summary="Get Hydro config values for form",
-        response_model=ManagementOptionsFormFields,
+        response_model=HydroManagementOptions,
```
MartinBelthle: Same I would rename in `HydroProperties`

antarest/study/storage/variantstudy/model/command/update_hydro_management.py
```diff
@@ -0,0 +1,84 @@
+# Copyright (c) 2025, RTE (https://www.rte-france.com)
+#
+# See AUTHORS.txt
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
```
MartinBelthle, line 10: I don't like the fact that we're handling `inter_monthly_correlation` separately :/ 
IMO we should put all of this inside the same command

antarest/study/storage/variantstudy/model/command/update_hydro_management.py
```diff
@@ -0,0 +1,84 @@
...
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand, OutputTuple
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+def normalize_key(key: str) -> str:
+    return key.replace(" ", "").replace("-", "").replace("_", "").lower()
+
+
+def update_current_for_area(area: str, current: dict[str, Any], new: dict[str, Any]) -> dict[str, Any]:
```
MartinBelthle: These 2 methods are really weird :/ I prefer what we usually do with Pydantic aliases, even if that means introducing a new object

MartinBelthle (changes requested):
I don't quite like the approach we took as we still have one `update_config` :/
We'll have to ask Sylvain what he thinks about this

antarest/study/business/model/hydro_management_model.py
TheoPascoli: I can not because `@all_optional_model` erases default values. I need default values because the hydro object can not contains all values in base depending on what field are True etc

antarest/study/storage/variantstudy/model/command/update_hydro_management.py
TheoPascoli: `inter_monthly_correlation` is in another tab. I can not save it at the same time. However, I can create a command for this field

antarest/study/web/study_data_blueprint.py
smailio: Also if you didn't already, think to sync with the front end team as they like to match the name of their front end object to the backend one. Theirs is still named ManagementOptionsFormFields.

antarest/study/business/model/hydro_management_model.py
MartinBelthle: I see. But then there's an issue as the API model should be all optional as it's used for updates. Unless you use exclude_unset=True

antarest/study/storage/variantstudy/command_factory.py
TheoPascoli: In this command, I am really updating what is inside the management option tab and nothing else. I think its good to have this level of specificity

antarest/study/storage/variantstudy/model/command/update_hydro_management.py
sylvlecl: On this matter, there is indeed a choice on the granularity of commands:
- either we consider ALL hydro properties as one big object, and then it follows that we should have only one command for everything (we would need to make multiple "save" in that command but it's not an issue)
- or we consider that the different "tabs" of hydro indeed correspond to quite independent objects, and then it makes sense to have several commands

Here, I think it's OK to have several commands because indeed the 2 are quite independent, see documentation https://antares-simulator.readthedocs.io/en/stable/user-guide/solver/02-inputs/#hydro .

Bu then, it may make sense to rename again "hydro properties" to "hydro management", actually (I was confused, I thought management was a technical term but it's actually the "business" wording here).

antarest/study/storage/variantstudy/command_factory.py
```diff
@@ -89,6 +91,8 @@
     CommandName.CREATE_ST_STORAGE.value: CreateSTStorage,
     CommandName.REMOVE_ST_STORAGE.value: RemoveSTStorage,
     CommandName.UPDATE_ST_STORAGE.value: UpdateSTStorage,
+    CommandName.UPDATE_HYDRO_PROPERTIES.value: UpdateHydroProperties,
```
sylvlecl: we should add some round trip tests in test_command_factory

antarest/study/business/model/hydro_model.py
```diff
@@ -0,0 +1,69 @@
...
+
+from pydantic import Field
+
+from antarest.core.serde import AntaresBaseModel
+from antarest.core.utils.string import to_camel_case
+
+HYDRO_PATH = ["input", "hydro", "hydro"]
+
+
+class HydroProperties(AntaresBaseModel, extra="forbid", populate_by_name=True, alias_generator=to_camel_case):
```
sylvlecl: For consistency sake with for example thermal cluster update, this should be named `***Update`, and we should have None as default values, otherwise partial updates will override properties with default values.

See other comment: after all I think it makes sense to keep the name "HydroManagement" instead of HydroProperties

antarest/study/storage/variantstudy/model/command/update_hydro_management.py
```diff
@@ -0,0 +1,79 @@
...
+    # Overloaded metadata
+    # ===================
+
+    command_name: CommandName = CommandName.UPDATE_HYDRO_PROPERTIES
+
+    # Command parameters
+    # ==================
+
+    area_id: str
+    properties: HydroProperties
```
sylvlecl: For consistency sake with for example thermal clusters, this should be a `HydroManagementUpdate`

antarest/study/storage/variantstudy/model/command/update_hydro_management.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> OutputTuple:
+        return CommandOutput(status=True, message=f"Hydro properties in '{self.area_id}' updated."), {}
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
```
sylvlecl, line 51: We need to be consistent between our implementations of update. For me we should follow the example of thermal clusters:
1. the command contains a `HydroManagementUpdate` which is "partial" (it has optionals)
2. we read from the study under modification a complete mode, for example `HydroManagementProperties`, with more or less no optionals
3. we update this complete model with the "update" object
4. we write it back to file
 
I know the structure of the file here is different, but this should not change this logic.

antarest/study/business/model/inflow_model.py
```diff
@@ -0,0 +1,40 @@
...
+# This file is part of the Antares project.
+from pydantic import Field
+
+from antarest.core.serde import AntaresBaseModel
+from antarest.core.utils.string import to_camel_case
+
+INFLOW_PATH = ["input", "hydro", "prepro", "{area_id}", "prepro", "prepro"]
+
+
+class InflowProperties(AntaresBaseModel, extra="forbid", populate_by_name=True, alias_generator=to_camel_case):
```
sylvlecl: I think the wording InflowStructure was ok, it's the business view wording, it's more specific than "properties"

antarest/study/storage/variantstudy/model/command/update_inflow_properties.py
```diff
@@ -0,0 +1,73 @@
...
+    # Overloaded metadata
+    # ===================
+
+    command_name: CommandName = CommandName.UPDATE_INFLOW_PROPERTIES
+
+    # Command parameters
+    # ==================
+
+    area_id: str
+    properties: InflowProperties
```
sylvlecl: Here too we should have an `InflowStructureUpdate` with optionals and None default values, for consistency sake

antarest/study/storage/variantstudy/model/command/update_hydro_management.py
```diff
@@ -0,0 +1,79 @@
...
+        current_hydro = study_data.tree.get(HYDRO_PATH)
+        new_hydro = self.properties.model_dump(exclude_unset=True)
+
+        area_id = get_hydro_id(area_id=self.area_id, field_dict=current_hydro)
```
sylvlecl: Here I think we should be more cautious: for the simulator, it's valid to have one of the properties for area "FR" and another one for area "fr", so we'd better not assume that the case is the same for all properties unfortunately :(

antarest/study/business/areas/hydro_management.py
```diff
@@ -10,203 +10,86 @@
...
         file_study = study.get_files()
-        return file_study.tree.get(HYDRO_PATH.split("/"))
+        hydro_config = file_study.tree.get(HYDRO_PATH)
 
-    def get_field_values(self, study: StudyInterface, area_id: str) -> ManagementOptionsFormFields:
-        """
-        Get management options for a given area
-        """
-        hydro_config = self._get_hydro_config(study)
+        new_area_id = get_hydro_id(area_id, hydro_config)
```
sylvlecl: same comment as in update: here we'd better not assume that IDs have the same case for all fields.

A possibility would be to have a pydantic model just to read the actual underlying file structure with LowerCaseStr, like

```python
# there may be a better name ...
class HydroManagementProperties(AntaresBaseModel):
    inter_daily_breakdown: Dict[LowerCaseStr, float]
    intra_daily_modulation: Dict[LowerCaseStr, float]
    ...

    def get_area_management(self, area_id: str) -> HydroManagement:
    ...

    def set_area_management(self, area_id: str, hydro: HydroManagement) -> None:
    ...
```

This way we are sure to get lower case stuff.
Then we can have a method in this pydantic model to get the data for only one area, which will be our `HydroManagement` object, which we will update with a `HydroManagementUpdate` object, then we will write back the `HydroManagement` to a consolidated `HydroManagementProperties` and dump it to file

antarest/study/business/areas/hydro_management.py
```diff
@@ -10,203 +10,86 @@
...
+        args = {k: v[new_area_id] for k, v in hydro_config.items() if new_area_id in v}
 
-        def get_value(field_info: FieldInfo) -> Any:
-            path = field_info["path"]
-            target_name = path.split("/")[-1]
-            field_dict = hydro_config.get(target_name, {})
-            return field_dict.get(self._get_id(area_id, field_dict), field_info["default_value"])
+        validated_args = HydroPropertiesInternal.model_validate(args).model_dump()
```
sylvlecl: Why not building the result without this intermediate ?

sylvlecl (changes requested):
I think we need to change a little 3 main things:
- ensure fields of the update are really optional (None default values)
- handle the case where area ID has different case in the same file
- be consistent with thermal cluster in the naming and implementation:
  having a `HydroManagementUpdate` for the command data, maybe a `HydroManagementProperties` (or another name?) for the ini file correspondance, and a plain `HydroManagement` as the main business model, which is retrieved when we get data from the manager

antarest/study/web/study_data_blueprint.py
TheoPascoli: After talking with Hatim, it is fine like that

TheoPascoli:
> I think we need to change a little 3 main things:
> 
>     * ensure fields of the update are really optional (None default values)
> 
>     * handle the case where area ID has different case in the same file
> 
>     * be consistent with thermal cluster in the naming and implementation:
>       having a `HydroManagementUpdate` for the command data, maybe a `HydroManagementProperties` (or another name?) for the ini file correspondance, and a plain `HydroManagement` as the main business model, which is retrieved when we get data from the manager

I think it looks good now. Thanks for your feedback, it really helped me going through this

antarest/study/storage/variantstudy/model/command/update_hydro_management.py
```diff
@@ -0,0 +1,76 @@
...
+)
+from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfig
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand, OutputTuple
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+class UpdateHydroProperties(ICommand):
```
sylvlecl: Probably best to rename it to UpdateHydroManagement, too

antarest/study/storage/variantstudy/model/command/update_hydro_management.py
sylvlecl: New implementation looks good :)

antarest/study/business/areas/hydro_management.py
```diff
@@ -10,169 +10,48 @@
...
-            return field_dict.get(self._get_id(area_id, field_dict), field_info["default_value"])
+        hydro_properties = HydroManagementProperties(**file_study.tree.get(HYDRO_PATH))
 
-        return ManagementOptionsFormFields.model_construct(
-            **{name: get_value(info) for name, info in FIELDS_INFO.items()}
-        )
+        return hydro_properties.get_hydro_management(area_id)
 
-    def set_field_values(
+    def update_hydro_properties(
```
sylvlecl: here too we should use "management" to be consistent: `update_hydro_management`

antarest/study/business/areas/hydro_management.py
```diff
@@ -10,169 +10,48 @@
...
-    @staticmethod
-    def _get_hydro_config(study: StudyInterface) -> Dict[str, Dict[str, FieldInfo]]:
-        """
-        Returns a dictionary of hydro configurations
-        """
-        file_study = study.get_files()
-        return file_study.tree.get(HYDRO_PATH.split("/"))
-
-    def get_field_values(self, study: StudyInterface, area_id: str) -> ManagementOptionsFormFields:
+    def get_hydro_properties(self, study: StudyInterface, area_id: str) -> HydroManagement:
```
sylvlecl: here too we should use "management" to be consistent: `get_hydro_management`

sylvlecl (changes requested):
Looks good! :) 
Just some last renamings are needed for consistency then we merge it

PR #2386, opened 2025-03-06
fix(perfs, variant): avoid loading study from disk for snapshot generation

sylvlecl:
This PR solves the following performance issues when generating snapshots:

- the study "config" is always read from disk, which is very slow,
  whereas we can use the config from cache as long as we don't
  perform a generation from scratch

- the study config was loaded twice instead of just once, again
  losing more time reading from disk

**The fix includes a modification on variant tree handling, that would deserve further work:**
- on "rebase" (commands history has changed), the snapshot and caches are invalidated
- on "advance" (commands have been appended), the snapshot and caches are kept,
  but the snapshot is marked stale through the update time

**In order to achieve this, it also fixes an issue which is a pre-requisite:**
when loading config from cache, currently, we don't instantiate the objects
based on the study version. For example, you can get a `Thermal870Config`
for a study version 7.0.
This causes some bugs when the config is used after reload from the cache.
It is now fixed, using validators that depend on the provided study version.

**Last results**:
As an example, for a large variant study, the creation of a cluster drops for around 1s to 0.1-0.2s.


antarest/study/storage/variantstudy/snapshot_generator.py
```diff
@@ -118,7 +122,7 @@ def generate_snapshot(
             variant_study.snapshot = VariantStudySnapshot(
                 id=variant_study_id,
                 created_at=datetime.datetime.utcnow(),
-                last_executed_command=variant_study.commands[-1].id if variant_study.commands else None,
+                last_executed_command=(variant_study.commands[-1].id if variant_study.commands else None),
```
TheoPascoli: are those parenthesis necessary ?

TheoPascoli (commented):
Hello, not much too say, that looks good !
Have you tried to compare performances before and after ? Or is it something that can only be better so we dont need to care too much ?

sylvlecl:
> Hello, not much too say, that looks good ! Have you tried to compare performances before and after ? Or is it something that can only be better so we dont need to care too much ?

Actually I had not compared !
And result is not as good as I expected, because one of 2 generation from disk is still performed, so for example on my test study, creating a thermal cluster goes from ~ 1s to ~0,5s, but it should be shorter than this.
The first generation is performed because when a snapshot is invalidated (when we add a command), we also remove the config from cache. It should not be necessary because in any case since the snapshot is invalidated, the config will never be used as is but be updated when generating the new snapshot.
I'll try to make changes in this direction.

antarest/study/storage/variantstudy/snapshot_generator.py
```diff
@@ -208,11 +212,14 @@ def _read_additional_data(self, file_study: FileStudy) -> StudyAdditionalData:
         study_additional_data = StudyAdditionalData(horizon=horizon, author=author)
         return study_additional_data
 
+    def _invalidate_cache(self, study_id: str) -> None:
```
MartinBelthle: Minor: Do we need to introduce this method as we could directly use `remove_from_cache` which even has a clearer name IMO

antarest/study/storage/variantstudy/variant_command_generator.py
```diff
@@ -110,14 +110,19 @@ def _generate(
     def generate(
         self,
         commands: List[List[ICommand]],
-        dest_path: Path,
+        dest_path: Optional[Path] = None,
         metadata: Optional[VariantStudy] = None,
         delete_on_failure: bool = True,
-        listener: Optional[ICommandListener] = None,
```
MartinBelthle, line 116: I don't understand how the code still works. As IMO we need the listener here in the case we're launching TS generation on the variant (see line 132)

antarest/study/storage/variantstudy/variant_command_generator.py
```diff
@@ -110,14 +110,19 @@ def _generate(
...
+        study: Optional[FileStudy] = None,
     ) -> GenerationResultInfoDTO:
         # Build file study
         logger.info("Building study tree")
-        study = self.study_factory.create_from_fs(dest_path, "", use_cache=False)
+        if not study:
```
MartinBelthle: Seems that this method is only used once. And if so, we never use dest_path and study is always given so we should adapt the code

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -420,24 +423,56 @@ def _get_variant_study(
...
+    def on_variant_rebase(self, study: VariantStudy) -> None:
+        """
+        This variant has been "rebased" in the sense of git (history changed):
+        it will need a generation from scratch, and children need
+        to be rebased too.
+        """
+        self._invalidate_snapshot(study)
+        self.on_parent_change(study.id)
+
+    def on_parent_change(self, study_id: str) -> None:
```
MartinBelthle, line 453: Not sure to understand how these 2 functions call each other. Imagine we have a parent that has 3 children all beign children of another. 

Calls being made:
- on variant_rebase on 1st
- on parent_change on 1nd
- on variant_rebase on 2nd
- on parent_change on 2nd
- on variant_rebase on 3rd
- on parent change on 3rd
- on variant_rebase on 3rd
- on parent change on 3rd

So it seems to me we're doing too many operations

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -420,24 +423,56 @@ def _get_variant_study(
...
+        self.repository.save(
+            metadata=study,
+            update_modification_date=True,
+        )
+        self.on_parent_change(study.id)
+
+    def get_children(self, parent_id: str) -> List[VariantStudy]:
+        return self.repository.get_children(parent_id=parent_id)
+
+    def on_variant_rebase(self, study: VariantStudy) -> None:
```
MartinBelthle, line 444: Really nice to have separated the cases in methods.
I think the naming lacks an info about commands as when I read the code where these methods are called I don't know what they are supposed to do. But I don't have a better proposition

tests/storage/test_model.py
```diff
@@ -66,6 +74,30 @@ def test_file_study_tree_config_dto():
...
+
+    config_dict = config_dto.model_dump()
+    parsed_config = FileStudyTreeConfigDTO.model_validate(
+        config_dict, context=study_version_context(STUDY_VERSION_7_0)
+    ).to_build_config()
+    assert parsed_config == config
+
+
+def test_file_study_tree_config_round_trip(study_tree_config):
+    config = study_tree_config
```
MartinBelthle: Minor but why creating a variable before each one of your tests ?

MartinBelthle (changes requested):

antarest/study/storage/variantstudy/variant_command_generator.py
sylvlecl: I fixed the calls as discussed

antarest/study/storage/variantstudy/variant_study_service.py
sylvlecl: Actually the goal of the naming here is more to say on what event it's supposed to be called, than what it does exactly, indeed.
(But there are some explanations in the docstring about the implementation details)

antarest/study/storage/variantstudy/variant_study_service.py
sylvlecl: As discussed, I added in the docstring that it gets direct children only.

tests/storage/test_model.py
sylvlecl: I removed this, it did not make sense indeed

antarest/study/storage/variantstudy/repository.py
```diff
@@ -55,7 +55,7 @@ def session(self) -> Session:
 
     def get_children(self, parent_id: str) -> List[VariantStudy]:
         """
-        Get the children of a variant study in chronological order.
+        Get the direct children of a variant study in chronological order.
```
MartinBelthle, line 58: I think we can also rename the method

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -420,24 +423,59 @@ def _get_variant_study(
...
+        It will need a snapshot generation (NOT from scratch),
+        and children need to be notified of their parent change.
+        """
+        self.repository.save(
+            metadata=study,
+            update_modification_date=True,
+        )
+        self.on_parent_change(study.id)
+
+    def get_children(self, parent_id: str) -> List[VariantStudy]:
```
MartinBelthle, line 438: Same remark here

MartinBelthle (approved):
Minor comments. I let Theo put his own review

TheoPascoli (approved):
Really good use of TypeAdapter and some others methods.
We are probably underutilizing our tools, which could likely provide us with a lot more than we suspect

antarest/study/storage/variantstudy/variant_study_service.py
sylvlecl: Actually from my point of view adding "direct" is a little redundant, children are not descendant nor grandchildren :) 
I think in the future we'll add a get_descendant or something like that for performance reasons, so it will be even more clear.

PR #2388, opened 2025-03-06
perf(ui-results): prevent infinite rendering in results matrices

hdinia:


webapp/src/components/App/Singlestudy/explore/Results/ResultDetails/index.tsx
```diff
@@ -197,25 +200,29 @@ function ResultDetails() {
...
 
-  const handleItemTypeChange: ToggleButtonGroupProps["onChange"] = (
-    _,
-    newValue: OutputItemType,
-  ) => {
-    if (newValue && newValue !== itemType) {
-      setItemType(newValue);
-    }
-  };
+  const handleItemTypeChange: ToggleButtonGroupProps["onChange"] = useCallback(
```
skamril: Give the type to the hook useCallback instead, so the type of the function arguments will infer automatically:

```typescript
const handleItemTypeChange = useCallback<ToggleButtonGroupProps["onChange"]>((_event, newValue) => ...);
```

skamril (approved):

PR #2393, opened 2025-03-11
feat(xpansion): add `update_xpansion_settings` command

MartinBelthle:


antarest/study/business/xpansion_management.py
```diff
@@ -72,59 +76,19 @@ def delete_xpansion_configuration(self, study: StudyInterface) -> None:
...
-            excludes.add("yearly_weights")
-
-        config_obj = updated_settings.model_dump(mode="json", by_alias=True, exclude=excludes)
-        file_study.tree.save(config_obj, ["user", "expansion", "settings"])
-
-        if new_xpansion_settings.sensitivity_config:
-            sensitivity_obj = new_xpansion_settings.sensitivity_config.model_dump(mode="json", by_alias=True)
-            file_study.tree.save(sensitivity_obj, ["user", "expansion", "sensitivity", "sensitivity_in"])
-
+        checks_settings_are_correct(new_xpansion_settings, file_study)
```
TheoPascoli: This method returns a set that is not used

antarest/study/business/xpansion_management.py
```diff
@@ -72,59 +76,19 @@ def delete_xpansion_configuration(self, study: StudyInterface) -> None:
...
+        command = UpdateXpansionSettings(
+            settings=new_xpansion_settings, command_context=self._command_context, study_version=study.version
+        )
+        study.add_commands([command])
         return self.get_xpansion_settings(study)
```
TheoPascoli, line 92: Why not calling the `get_expansion_settings` directly from the common module ?

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
```diff
@@ -0,0 +1,90 @@
...
+    settings: XpansionSettingsUpdate
+
+    @override
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
+        return CommandOutput(status=True, message="Xpansion settings updated successfully"), {}
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        # Checks settings are correct
+        excludes = checks_settings_are_correct(self.settings, study_data)
```
TheoPascoli: by the method name, I would expect it to return a boolean or None. It is not clear why "checking" would return a collection

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Updates settings
+        current_settings = get_xpansion_settings(study_data)
+        new_settings = self.settings.model_dump(mode="json", exclude_none=True, exclude={"sensitivity_config"})
+        updated_settings = current_settings.model_copy(deep=True, update=new_settings)
+        config_obj = updated_settings.model_dump(mode="json", by_alias=True, exclude=excludes)
```
TheoPascoli, line 68: mode="json" does not change anything

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # Overloaded metadata
+    # ===================
+
+    command_name: CommandName = CommandName.UPDATE_XPANSION_SETTINGS
+
+    # Command parameters
+    # ==================
+
+    settings: XpansionSettingsUpdate
```
TheoPascoli, line 53: When I try to update just 1 field, I get a `settings` with multiple value set (like 10). I would expect to have only the one I want to update inside this object

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        # Checks settings are correct
+        excludes = checks_settings_are_correct(self.settings, study_data)
+
+        # Updates settings
+        current_settings = get_xpansion_settings(study_data)
+        new_settings = self.settings.model_dump(mode="json", exclude_none=True, exclude={"sensitivity_config"})
+        updated_settings = current_settings.model_copy(deep=True, update=new_settings)
```
TheoPascoli: it should work without deep copying. Maybe it is not necessary

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
```diff
@@ -0,0 +1,90 @@
...
+        return CommandOutput(status=True, message="Xpansion settings updated successfully"), {}
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        # Checks settings are correct
+        excludes = checks_settings_are_correct(self.settings, study_data)
+
+        # Updates settings
+        current_settings = get_xpansion_settings(study_data)
+        new_settings = self.settings.model_dump(mode="json", exclude_none=True, exclude={"sensitivity_config"})
```
TheoPascoli, line 66: "json" is not useful in my opinion

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
```diff
@@ -0,0 +1,90 @@
...
+        updated_settings = current_settings.model_copy(deep=True, update=new_settings)
+        config_obj = updated_settings.model_dump(mode="json", by_alias=True, exclude=excludes)
+        study_data.tree.save(config_obj, ["user", "expansion", "settings"])
+
+        # Updates sensitivity
+        if self.settings.sensitivity_config:
+            sensitivity_obj = self.settings.sensitivity_config.model_dump(mode="json", by_alias=True)
```
TheoPascoli, line 73: same here

antarest/study/storage/variantstudy/model/command/xpansion_common.py
```diff
@@ -84,3 +91,39 @@ def assert_link_exist(file_study: FileStudy, xpansion_candidate_dto: XpansionCan
...
+
+
+def get_xpansion_settings(file_study: FileStudy) -> GetXpansionSettings:
+    config_obj = file_study.tree.get(["user", "expansion", "settings"])
+    with contextlib.suppress(ChildNotFoundError):
+        config_obj["sensitivity_config"] = file_study.tree.get(["user", "expansion", "sensitivity", "sensitivity_in"])
+    return GetXpansionSettings.from_config(config_obj)
+
+
+def checks_settings_are_correct(settings: XpansionSettingsUpdate, file_study: FileStudy) -> set[str]:
```
TheoPascoli: Not necessary, but you should be able to factorise this by defining a method inside this one, ans calling it with both `additionnal_constraint` and `yearly_weights`

TheoPascoli (changes requested):
Hello Martin, I have some small remarks and improvement suggestions

antarest/study/business/xpansion_management.py
MartinBelthle: Yep it's only used inside the commands. Is it an issue ?

antarest/study/business/xpansion_management.py
MartinBelthle: Just for the log. But it can be questionned

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
MartinBelthle: Indeed. I'll try to find a better name

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
MartinBelthle: Indeed, I'll try to remove it

antarest/study/storage/variantstudy/model/command/xpansion_common.py
MartinBelthle: Why not

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
MartinBelthle: Actually it does as without it Enums are serialized as Enum and not str
You can test with this `XpansionSettingsUpdate(**{"master": "integer"})`

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
MartinBelthle: Here it doesn't but if we add an Enum one day I prefer to keep this security

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
MartinBelthle: Not sure to understand. We can see in the `test_integration_xpansion.py` that doing this
`res = xp_client.put("settings", json={"optimality_gap": 42})` creates a commandDTO with args being 
`{"settings": {"optimality_gap": 42.0}}` that contains only modified fields

antarest/study/storage/variantstudy/model/command/xpansion_common.py
```diff
@@ -84,3 +91,39 @@ def assert_link_exist(file_study: FileStudy, xpansion_candidate_dto: XpansionCan
...
+    # Specific handling for yearly_weights and additional_constraints:
+    # - If the attributes are given, it means that the user wants to select a file.
+    #   It is therefore necessary to check that the file exists.
+    # - Else, it means the user want to deselect the additional constraints file,
+    #  but he does not want to delete it from the expansion configuration folder.
```
smailio: minor : can you put this in a normal doc string ?

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
TheoPascoli: On localhost in debug mode, I noticed that this object contains a lot more properties than what I want to update. Maybe Im missing something or testing it wrong

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
TheoPascoli: Alright. 
I tested like `updated_settings.model_dump(...) == updated_settings.model_dump(mode="json", ...)` and I got True. But indeed they are not the same. 
After doing some research, the == test with underlying value of the enum, that's why they seem equal but they are not

antarest/study/business/xpansion_management.py
TheoPascoli: Not really actually. It just creates some confusion as I expect a return value to be used. Maybe you can create a `validate_settings -> None` and a `get_invalid_settings -> set`

MartinBelthle: I could do 2 functions but I think they would look a lot alike. I'll keep the code as is for the moment and you'll tell me what you think in the next review

antarest/study/storage/variantstudy/model/command/xpansion_common.py
```diff
@@ -100,30 +100,30 @@ def get_xpansion_settings(file_study: FileStudy) -> GetXpansionSettings:
...
-        try:
-            weights_url = ["user", "expansion", "weights", weights_file]
-            file_study.tree.get(weights_url)
-        except ChildNotFoundError:
-            msg = f"Additional weights file '{weights_file}' does not exist"
-            raise XpansionFileNotFoundError(msg) from None
-    else:
-        excludes.add("yearly_weights")
+    for field in ["additional_constraints", "yearly_weights"]:
+        if file := settings.__getattribute__(field):
```
TheoPascoli: I’m not in favor of calling dunder methods directly in this method. you can replace `__getattribute__` by `getattr`

antarest/study/business/model/xpansion_model.py
```diff
@@ -142,7 +142,10 @@ class XpansionSettings(AntaresBaseModel, extra="ignore", validate_assignment=Tru
     sensitivity_config: Optional[XpansionSensitivitySettings] = None
 
     @model_validator(mode="before")
-    def validate_float_values(cls, values: MutableMapping[str, Any]) -> MutableMapping[str, Any]:
+    def validate_float_values(cls, values: dict[str, Any]) -> dict[str, Any]:
+        if isinstance(values, AntaresBaseModel):
```
TheoPascoli: What is the point of this condition ? 
"before" validator get raw input which is mostly dict or primitives and cannot be classes like this one

TheoPascoli (changes requested):

antarest/study/storage/variantstudy/model/command/xpansion_common.py
MartinBelthle: First I put getattr but mypy says the class doesn't have a getattr method :/

antarest/study/business/model/xpansion_model.py
MartinBelthle: During the round-trip we can already have an object here. Anis had the same issue with update_binding_constraints

antarest/study/storage/variantstudy/model/command/xpansion_common.py
TheoPascoli: Really ? I can't reproduce it on my side with `if file := getattr(settings, field, None):`

MartinBelthle: Indeed, I don't know why it didn't work before. Really weird. I'm pushing that and then merging

antarest/study/storage/variantstudy/model/command/update_xpansion_settings.py
TheoPascoli: The code is fine, but the front send all the fields in the request. That's why you rewrite even the fields that you donc update

MartinBelthle: Indeed, we'll have to work on that when refactoring the way we handle Xpansion inside the API / front-end

antarest/study/business/model/xpansion_model.py
MartinBelthle: You were right I removed it

PR #2394, opened 2025-03-11
feat(workspaces): forbid additional workspaces on desktop

smailio:
This PR is about external workspaces. 

This PR only cover the first part of the feature described in the doc bellow. When the app start if the desktop mode is configured it would raise an error if other workspace than default are configured. 

https://gopro-collaboratif.rte-france.com/pages/viewpage.action?pageId=451514074


antarest/core/config.py
```diff
@@ -654,14 +659,18 @@ class Config:
...
     root_path: str = ""
     api_prefix: str = ""
+    desktop_mode: bool = False
 
     @classmethod
     def from_dict(cls, data: JSON) -> "Config":
         defaults = cls()
+        desktop_mode = data.get("desktop_mode", defaults.desktop_mode)
+        storage_config = StorageConfig.from_dict(data["storage"]) if "storage" in data else defaults.storage
+        StorageConfig._validate_workspaces(storage_config.workspaces, desktop_mode)
```
TheoPascoli: This method has been declared as private. Is it ok to still call it from here ?

TheoPascoli (commented):
Hello Anis, just one question

antarest/core/config.py
smailio: No you're right it doesn't make sense anymore to set this method as private

antarest/core/config.py
```diff
@@ -654,14 +659,18 @@ class Config:
     tasks: TaskConfig = TaskConfig()
     root_path: str = ""
     api_prefix: str = ""
+    desktop_mode: bool = False
```
sylvlecl, line 634: Note that in order to take this into account in the desktop version we build, you need to add this configuration property in `resources/antares-desktop-fs/config.yaml`.

We also need to update the installer so that it adds that property on existing installations :(
See here: 
https://github.com/AntaresSimulatorTeam/antares-web-installer/blob/main/src/antares_web_installer/config/__init__.py

An alternative would be to not put it in the config file, but to pass this boolean flag when we start the desktop app, in `antarest.desktop.systray_app` (call to `fastapi_app`)

antarest/core/core_blueprint.py
```diff
@@ -59,6 +63,22 @@ def version_info() -> Any:
             version=antarest_version,
             gitcommit=get_commit_id(config.resources_path),
             dependencies=get_dependencies(),
+            desktop_mode=config.desktop_mode,
         )
 
+    @bp.get(
+        "/desktop_mode",
```
sylvlecl, line 69: Nice, I guess this will be the way to add dedicated features on the front-end side :)

sylvlecl (commented):
In addition to this work, we need to actually configure the desktop app to use the new boolean flag: this can be done either in the config and in the installer, or in the startup method of the desktop app

antarest/core/core_blueprint.py
smailio: Oh about that, I forgot to write about it in the PR. We can either add an endoint, or just add it to the version info dto. Or do both. In this PR I did both, but my favorite choice is in the version info dto.

antarest/core/core_blueprint.py
```diff
@@ -59,6 +63,22 @@ def version_info() -> Any:
             version=antarest_version,
```
smailio: add desktop mode doc

antarest/study/repository.py
```diff
@@ -348,11 +349,11 @@ def _search_studies(
         if study_filter.archived is not None:
             q = q.filter(entity.archived == study_filter.archived)
         if study_filter.name:
```
smailio, line 351: The code bellow fix (minor) issues that are not directly related to this PR. It just happened that these issue were noticed when I was working on this PR. Tell me if you want me to move this code in its own PR.

sylvlecl (approved):

PR #2400, opened 2025-03-12
fix(study): allow creating old studies with new version format

sylvlecl:
Currently, when posting a study creation request with version="8.8",
it's written as a raw "8.8" in the study files.

This causes failure of the simulation, since it's not a valid format before version 9.0.

The PR ensures that versions are correctly parsed to an object,
and written to underlying storage according to the required formatting.


antarest/study/model.py
```diff
@@ -60,11 +60,12 @@
...
 STUDY_VERSION_8_7 = StudyVersion.parse("8.7")
 STUDY_VERSION_8_8 = NEW_DEFAULT_STUDY_VERSION
+STUDY_VERSION_9_0 = StudyVersion.parse("9.0")
 STUDY_VERSION_9_1 = StudyVersion.parse("9.1")
 STUDY_VERSION_9_2 = StudyVersion.parse("9.2")
 
-StudyVersionStr = Annotated[StudyVersion, BeforeValidator(StudyVersion.parse), PlainSerializer(str)]
-StudyVersionInt = Annotated[StudyVersion, BeforeValidator(StudyVersion.parse), PlainSerializer(int)]
+StudyVersionStr: TypeAlias = Annotated[StudyVersion, BeforeValidator(StudyVersion.parse), PlainSerializer(str)]
+StudyVersionInt: TypeAlias = Annotated[StudyVersion, BeforeValidator(StudyVersion.parse), PlainSerializer(int)]
```
TheoPascoli, line 68: Can we get rid of one of those two types or they must live together ?

antarest/core/serde/ini_reader.py
```diff
@@ -377,3 +377,17 @@ def read(self, path: Any, **kwargs: Any) -> JSON:
         sections = super().read(path)
         obj = cast(Mapping[str, JSON], sections)
         return obj[self._section_name]
+
+
+def read_ini(content: TextIO) -> JSON:
```
MartinBelthle: Why not putting this 2 methods inside one that accepts the union ?

MartinBelthle (dismissed):

antarest/core/serde/ini_reader.py
TheoPascoli: Do you use this aside from the tests ?

sylvlecl: No! but it makes sense to have this feature in this module from my point of view, although it's not used for now in prod code.
I prefer this as opposed to have an additional module only for testing purposes.

antarest/study/model.py
sylvlecl: For now, the study DTO in the API is defiend as `int` it seems :(
Simply replacing it to a `str` will be a breaking change unfortunately, but indeed we must do it at some point ...

I will create an issue for this, thanks !

antarest/core/serde/ini_reader.py
sylvlecl: No strong argument, just to avoid an if/else, but I can change it.

TheoPascoli (dismissed):

antarest/core/serde/ini_reader.py
MartinBelthle: Not sure to understand where you you'll need an if/else as no matter what's given as a parameter you run the same code

sylvlecl: Indeed ... !

TheoPascoli (approved):

PR #2405, opened 2025-03-14
feat(variant_study): copy variant as raw study

TheoPascoli:
This PR modifies the behavior of the /copy endpoint.
For a variant study, instead of copying a variant to another variant with the same "parents," it now copies the variant as a raw study.
The behavior for copying a raw study remains unchanged.


antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -902,45 +915,42 @@ def copy(
             The newly created study.
         """
         new_id = str(uuid4())
-        study_path = str(self.config.get_workspace_path() / new_id)
-        if src_meta.additional_data is None:
+
+        if src_study.additional_data is None:
```
sylvlecl: We have quite a lot of logic duplication now between the 2 services, we'd better try to avoid it.
I guess it would make sense to call a method of the `raw_study_service` from here, since we are creating a raw study. Maybe be it could take as an argument a `Study` for metadata and a `FileStudy` for the actual content to copy: the `FileStudy` from `get_raw` will l have the correct study path and the correct output path.

tests/integration/test_integration.py
```diff
@@ -1748,6 +1748,31 @@ def test_copy(client: TestClient, admin_access_token: str, internal_study_id: st
...
+    parent_id = raw.json()
+
+    var = client.post(f"/v1/studies/{parent_id}/variants", params={"name": "variant"})
+    assert var.status_code == 200
+    variant_id = var.json()
+
+    variant_study = client.get(f"/v1/studies/{variant_id}")
+    assert variant_study.status_code == 200
+
+    res = client.put(f"/v1/studies/{variant_id}/generate")
```
sylvlecl: We must not need to call this explicitly, this should be performed by the service automatically on copy

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -902,45 +915,42 @@ def copy(
...
-                study_version=str(command.study_version),
-                user_id=command.user_id,
-                updated_at=command.updated_at,
-            )
-            for command in src_meta.commands
-        ]
+        src_path = self.get_study_path(src_study)
+        dest_path = Path(dest_study.path)
+
+        shutil.copytree(src_path, dest_path)
```
sylvlecl, line 923: To be sure that we have a snapshot and that it's up to date, we need to call `_safe_generation` at the start of the method, like for other methods of that service

tests/integration/test_integration.py
```diff
@@ -1748,6 +1748,31 @@ def test_copy(client: TestClient, admin_access_token: str, internal_study_id: st
...
+    client.get(f"/v1/tasks/{res.json()}?wait_for_completion=True")
+
+    copy = client.post(f"/v1/studies/{variant_id}/copy?dest=copied&use_task=false")
+    client.get(f"/v1/tasks/{copy.json()}?wait_for_completion=True")
```
sylvlecl: We should check that the copy has been performed correctly:
for example we could check that we get the same areas as in the base study.
We should also add some assertions about the metadata we receive for that study (the name, the author, the permissions etc)

antarest/study/storage/variantstudy/variant_study_service.py
sylvlecl: Also:
actually, for variant studies, the outputs are not in the snapshot directory, so that method will not never copy the ouputs.

The directory layout is:
```
<study_id>
    - snapshot
        - input
        - ...
    - output
```

I guess we need to have an additional copy of this output directory if the output copy is requested.

sylvlecl (changes requested):
I think 2 behaviours need to be fixed:
- the copy should trigger the snapshot generation to get the actual data to copy
- the outputs are probably not correctly copied

Then, we can probably remove some code duplication between raw and variant services.

tests/integration/test_integration.py
TheoPascoli: I added a safe generation inside the copy method

antarest/study/storage/variantstudy/variant_study_service.py
TheoPascoli: Indeed, I modified it

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -884,63 +895,37 @@ def exists(self, metadata: VariantStudy) -> bool:
...
-                version=command.version,
-                study_version=str(command.study_version),
-                user_id=command.user_id,
-                updated_at=command.updated_at,
-            )
-            for command in src_meta.commands
-        ]
+        self._safe_generation(src_study)
+
+        output_path = str(src_study.path + "/output")
```
sylvlecl: We'd better use path objects, and the constant defined above:
```python
output_path = Path(src_study.path) / OUTPUT_RELATIVE_PATH
```
(Same for the snapshot below)

or maybe a better alternative, to not duplicate that logic, would be to call `get_raw` and just use the `study_path` and `output_dir` of the result.
Note that this also ensures that the snapshot is generated, so we could also skip the explicit call to `_safe_generation`, in this case.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -884,63 +895,37 @@ def exists(self, metadata: VariantStudy) -> bool:
...
+        src_study.path += "/snapshot"
```
sylvlecl: Actually, it's dangerous to modify the `Study` object this way: when we commit later, the object in database will be modified
--> we must avoid it

Again an option may be to provide the `FileStudy` to the raw service instead of the modified `VariantStudy`

tests/integration/test_integration.py
```diff
@@ -1748,6 +1748,43 @@ def test_copy(client: TestClient, admin_access_token: str, internal_study_id: st
...
+    assert variant_study.status_code == 200
+    assert len(all_studies.json()) == 4
+
+    copied_study = client.get("/v1/studies?name=copied")
+    assert copied_study.status_code == 200
+    copied_id = next(iter(copied_study.json()))
+
+    copied_areas = client.get(f"/v1/studies/{copied_id}/areas")
+    assert copied_areas.json() == client.get(f"/v1/studies/{parent_id}/areas").json()
+
```
sylvlecl, line 1788: We should add a test for the copy of output, since we have seen that it's not that obvious to make it work :)

sylvlecl (changes requested):
I think the implementation has an issue (modifying a study in database), + we should add some test for the outputs (I guess it's not very easy to do indeed :/ )

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -214,7 +215,7 @@ def create(self, metadata: RawStudy) -> RawStudy:
     @override
     def copy(
         self,
-        src_meta: RawStudy,
+        src_meta: RawStudy | VariantStudy,
```
MartinBelthle: You could replace with just  `Study` I believe

tests/integration/test_integration.py
```diff
@@ -1748,6 +1748,43 @@ def test_copy(client: TestClient, admin_access_token: str, internal_study_id: st
     assert res["public_mode"] == "READ"
 
 
+def test_copy_variant_as_raw(client: TestClient, admin_access_token: str, internal_study_id: str) -> None:
```
MartinBelthle: We should add comments inside this test to make it easier to read

MartinBelthle (changes requested):
Minor comments and I agree with what Sylvain wrote

sylvlecl (approved):

PR #2406, opened 2025-03-17
fix(bc)!: allow terms writing in any case and allow area/cluster names in endpoint

MartinBelthle:
Fix [ANT-2974]

Breaking changes validated by the R-scripts:
-  Inside endpoint GET /bindingconstraints, the terms do no longer have an id inside the response
- Inside endpoints POST /term and POST /terms, the user can no longer give an id


antarest/study/business/binding_constraint_management.py
```diff
@@ -86,6 +87,25 @@
...
 
+def validate_and_transform_term_id(term_id: str) -> str:
+    """
+    Used to validate the term id given by the user when updating an existing one
+    """
+    if not term_id:
+        raise InvalidConstraintTerm(term_id, "You should provide an id when updating an existing term")
+    try:
+        if "%" in term_id:
+            area_1, area_2 = sorted(term_id.split("%"))
```
smailio, line 98: Why you sort here ?

antarest/study/business/binding_constraint_management.py
```diff
@@ -135,24 +154,38 @@ class ConstraintTerm(AntaresBaseModel):
...
     data: Optional[LinkTerm | ClusterTerm] = None
 
-    @field_validator("id")
-    def id_to_lower(cls, v: Optional[str]) -> Optional[str]:
-        """Ensure the ID is lower case."""
-        if v is None:
-            return None
-        return v.lower()
+
+class ConstraintTerm(AntaresBaseModel):
```
smailio, line 163: This class should be named ConstraintTermDTO

antarest/study/business/binding_constraint_management.py
```diff
@@ -135,24 +154,38 @@ class ConstraintTerm(AntaresBaseModel):
...
+    offset: Optional[int] = None
+    data: LinkTerm | ClusterTerm
 
     def generate_id(self) -> str:
-        """Return the constraint term ID for this term based on its data."""
-        if self.data is None:
-            return self.id or ""
         return self.data.generate_id()
 
+    def from_updated_term(self, updated_term: ConstraintTermUpdate) -> "ConstraintTerm":
```
smailio: Why you double quote the return type ?  "ConstraintTerm"

smailio: Also maybe add a doc to explain why you never expect "data" in this constructor method ?

antarest/study/business/binding_constraint_management.py
```diff
@@ -86,6 +87,25 @@
...
+        elif "." in term_id:
+            area, cluster = term_id.split(".")
+            return f"{transform_name_to_id(area)}.{transform_name_to_id(cluster)}"
+        else:
+            raise InvalidConstraintTerm(term_id, "Your term id is not well-formatted")
+    except Exception:
+        raise InvalidConstraintTerm(term_id, "Your term id is not well-formatted")
+
+
 class LinkTerm(AntaresBaseModel):
```
smailio, line 109: Should'nt this be renamed LintkTermDTO ?

antarest/study/business/binding_constraint_management.py
```diff
@@ -986,75 +1018,63 @@ def _update_constraint_with_terms(
...
+        if missing_terms:
+            raise ConstraintTermNotFound(binding_constraint_id, *missing_terms)
+
+        # Updates existing terms
+        for term in constraint_terms:
+            existing_term = existing_terms.pop(term.id)
+            # We only support 2 cases:
+            # 1- Renaming a term
+            # 2- Modifying a term offset and/or weight
+            if term.data and term.data.generate_id() != existing_term.data.generate_id():  # First case
```
smailio: Tbh this part of the code was hard to understand, I'm not requesting any specefic change maybe it's inherently complex. Also I don't get why we have to do support only these two cases.

antarest/study/business/binding_constraint_management.py
smailio: Ah it's because you don't allow two links about two identical areas but different ways.

antarest/study/business/binding_constraint_management.py
```diff
@@ -986,75 +1018,63 @@ def _update_constraint_with_terms(
...
         self._update_constraint_with_terms(study, constraint, sorted_terms)
 
-    def create_constraint_terms(
-        self, study: StudyInterface, binding_constraint_id: str, constraint_terms: Sequence[ConstraintTerm]
+    def update_constraint_terms(
+        self, study: StudyInterface, binding_constraint_id: str, constraint_terms: Sequence[ConstraintTermUpdate]
     ) -> None:
         """
-        Adds new constraint terms to an existing binding constraint.
+        Update or add the specified constraint terms.
```
smailio: Now all you do is rename or update but not add, no ?

antarest/study/business/binding_constraint_management.py
smailio: I realized late that this method isn't a constructor method. Of course I could have seen it earlier because it's not a classmethod nor a staticmethod but the name is so typical of constructor methods that I thought it was. Maybe rename it simply to `update_from` ?

antarest/study/business/binding_constraint_management.py
MartinBelthle: We talked with Sylvain about the naming convention `DTO` and here's the guideline we want to follow:
- A class should have the suffix `DTO` when it's a construct for the API only. For instance if the front-end needed another field, we would introduce a class `ConstraintTermDTO` with the same fields and the one for the front. And this class would not be known by the manager, only by the service calling the manager. The service will get a `ConstraintTerm` object from the manager and would transform it into a `ConstraintTermDTO` one.

Hope this is clear :) Feel free to ask questions

antarest/study/business/binding_constraint_management.py
MartinBelthle: Based on your last comment I believe the only thing to change is to rename the method. I will follow your proposition as I don't have a strong opinion on the subject

antarest/study/business/binding_constraint_management.py
MartinBelthle: Same remark as above

antarest/study/business/binding_constraint_management.py
MartinBelthle: Indeed I'll remove the `or add` inside the docstring

antarest/study/business/binding_constraint_management.py
MartinBelthle: I agree it's a bit complex. I said we only support these 2 modification for a reason but I can't remember what. I'll try to allow all modification and if I see the problem I'll answer under this comment

MartinBelthle: It was because of a test but in fact the test doesn't make sense. So i'll simplify this code, let me know if you find it easier to read

smailio (approved):
LGTM

There're tests for the use case, they also cover failing cases. Code is clean, documented and understandable. Some breaking changes but they were reviewed with the users.

PR #2408, opened 2025-03-17
fix(sts): fix create_st_storage v1 parsing

sylvlecl:
The parsing of existing commands of type `create_st_storage` raises an error, because the presence of a `cluster_name` property is assumed.

This was a mistake, the command never had such a field (unlike command `create_cluster` which inspired it).

We need to keep the v2 version anyway, because some commands may have been created with this new version number.


MartinBelthle (commented):
Could be nice to also add a test

sylvlecl:
> Could be nice to also add a test

Actually there is one, I just fixed the existing test which was wrong (the input data did not look like actual v1 data, it had a `cluster_name` field).

TheoPascoli (approved):

PR #2414, opened 2025-03-18
feat(logs): add task_id and user.id to logs

sylvlecl:
A technical evolution to:

- add a `task_id` property to logs, to be able to filter the logs of a particular task
- add `user` to all logs
- remove the explicit addition of user, this is superseded by the use of the context variable


antarest/core/logging/utils.py
```diff
@@ -210,4 +218,12 @@ def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
...
 
 
-context_filter = ContextFilter()
+@contextlib.contextmanager
+def task_context(task_id: str, user: Optional[JWTUser]) -> Iterator[None]:
+    """
+    Sets current user and task_id to specified values.
+    """
+    with current_user_context(user):
+        reset_token = _task_id.set(task_id)
```
smailio, line 227: if the code wrapped by this context manager fail, the task id won't be reset. To so you need to wrap the yield by a try finally.

sylvlecl: Indeed, thanks ! It was a misuse of the `@contextmanager` annotation

sylvlecl: I added a small unit test on the task context to check the fixed behaviour.

smailio (approved):

PR #2416, opened 2025-03-19
perf(table-mode): speed-up mass update for thermal clusters

MartinBelthle:
Performance test performed on my local env for 50 thermals inside the same area
- on `dev` branch: ~3s
- on this branch: ~200ms


antarest/study/storage/variantstudy/model/command/common.py
```diff
@@ -46,6 +46,7 @@ class CommandName(Enum):
     CREATE_THERMAL_CLUSTER = "create_cluster"
     REMOVE_THERMAL_CLUSTER = "remove_cluster"
     UPDATE_THERMAL_CLUSTER = "update_thermal_cluster"
```
TheoPascoli: You can remove this now that you changed the name

TheoPascoli (changes requested):

antarest/study/storage/variantstudy/model/command/update_thermal_clusters.py
```diff
@@ -0,0 +1,124 @@
...
+    def _apply_config(self, study_data: FileStudyTreeConfig) -> OutputTuple:
+        for area_id, value in self.cluster_properties.items():
+            if area_id not in study_data.areas:
+                return CommandOutput(status=False, message=f"The area '{area_id}' is not found."), {}
+
+            thermal_mapping: dict[str, tuple[int, ThermalConfigType]] = {}
+            for index, thermal in enumerate(study_data.areas[area_id].thermals):
+                thermal_mapping[thermal.id] = (index, thermal)
+
+            for cluster_id, properties in value.items():
```
smailio: you never use properties, I suggest to replace it by _ so the reader knows immediately that this isn't used, also  vscode won't highlight it as unused.

antarest/study/storage/variantstudy/model/command/update_thermal_clusters.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # Overloaded metadata
+    # ===================
+
+    command_name: CommandName = CommandName.UPDATE_THERMAL_CLUSTERS
+
+    # Command parameters
+    # ==================
+
+    cluster_properties: dict[str, dict[str, ThermalClusterUpdate]]
```
smailio: Minor : 

maybe we can create a type for this ? Otherwise we have to read the code bellow to understand that the first key is the area id and the second the cluster id, something like this. 

```python
AreaID = str
ClusterID = str
ThermalClusterUpdates = Dict[AreaID, Dict[ClusterID, ThermalClusterUpdate]]
```
Or like this
 
```python

@camel_case_model
class ThermalClusterUpdates(Thermal870Properties):
    clusterId: str = None
    updates : List[ThermalClusterUpdate]

@camel_case_model
class ThermalAreaUpdates(Thermal870Properties):
    areaId: str = None
    updates : List[ThermalClusterUpdates]

```
It's just a matter of readability so this is a mere suggestion, if you find it's not that much of an improvement ignore this comment.

antarest/study/storage/variantstudy/model/command/update_thermal_clusters.py
```diff
@@ -0,0 +1,124 @@
...
+                index, thermal = thermal_mapping[cluster_id]
+                study_data.areas[area_id].thermals[index] = self.update_thermal_config(area_id, thermal)
+
+        return CommandOutput(status=True, message="The thermal clusters were successfully updated."), {}
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        for area_id, value in self.cluster_properties.items():
+            ini_path = ["input", "thermal", "clusters", area_id, "list"]
+            all_clusters_for_area = study_data.tree.get(ini_path)
```
smailio: In the apply_config you return a CommandOutput(status=False) when you don't find the area, you should protect againt wrong area here too no ?

antarest/study/storage/variantstudy/model/command/update_thermal_clusters.py
```diff
@@ -0,0 +1,124 @@
...
+        output, _ = self._apply_config(study_data.config)
+
+        return output
+
+    @override
+    def to_dto(self) -> CommandDTO:
+        args: dict[str, dict[str, Any]] = {}
+        for area_id, value in self.cluster_properties.items():
+            for cluster_id, properties in value.items():
+                args.setdefault(area_id, {})[cluster_id] = properties.model_dump(mode="json", exclude_unset=True)
```
smailio, line 111: Minor : split this into more lines for readability

antarest/study/storage/variantstudy/model/command/update_thermal_clusters.py
```diff
@@ -0,0 +1,124 @@
...
+        return CommandDTO(
+            action=self.command_name.value, args={"cluster_properties": args}, study_version=self.study_version
+        )
+
+    @override
+    def get_inner_matrices(self) -> List[str]:
+        return []
+
+    def update_thermal_config(self, area_id: str, thermal: ThermalConfigType) -> ThermalConfigType:
+        # Set the object to the correct version
```
smailio, line 122: Tbh I didnt dig in this code but I see this function is identical to the one it replaces, so I guess if it used to work it'll keep working, let me know if you want me to inspect this more thoroughly.

tests/integration/study_data_blueprint/test_thermal.py
```diff
@@ -1123,3 +1124,60 @@ def test_thermal_cluster_deletion(self, client: TestClient, user_access_token: s
...
+            res.raise_for_status()
+            body[f"{area_id} / {cluster_id}"] = {"enabled": False}
+
+        # Modify all of them with the table-mode endpoint. Due to new code this should be pretty fast.
+        start = time.time()
+        res = client.put(f"/v1/studies/{study_id}/table-mode/thermals", json=body)
+        end = time.time()
+        assert res.status_code in {200, 201}
+        duration = end - start
+        assert duration < 1
```
smailio, line 1152: less than one second, impressive ! :dash: :dash:

smailio (changes requested):

antarest/study/storage/variantstudy/model/command/update_thermal_clusters.py
MartinBelthle: Yes it's the same exact code that was previously validated so we can keep it as is

antarest/study/storage/variantstudy/model/command/update_thermal_clusters.py
MartinBelthle: I addded whitespace before and after the double loop to make it easier to read

antarest/study/storage/variantstudy/model/command/update_thermal_clusters.py
MartinBelthle: Indeed. But I'll simply use the keys and not items it will do the same thing and be more efficient

antarest/study/storage/variantstudy/model/command/update_thermal_clusters.py
MartinBelthle: I understand. I quite like the first proposition so I'll implement it

PR #2418, opened 2025-03-19
fix(adequacy-patch): fix serialization issue inside the ini file

TheoPascoli:
We must have dash instead of underscore


MartinBelthle (commented):
The code is okay but it could be nice to add a test to ensure we don't encounter this bug again in the future

MartinBelthle (approved):

PR #2423, opened 2025-03-20
perf(table-mode): speed-up mass update for renewable clusters

MartinBelthle:
Same benchmark as the one inside thermal clusters PR


antarest/study/storage/variantstudy/model/command/update_renewable_clusters.py
```diff
@@ -0,0 +1,123 @@
...
+    RenewableConfig,
+    create_renewable_properties,
+)
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand, OutputTuple
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+AreaID = str
```
TheoPascoli: We now have a dedicated type for AreaID that you can use instead of str -> AreaId: TypeAlias

antarest/study/storage/variantstudy/command_factory.py
```diff
@@ -97,7 +97,7 @@
     CommandName.UPDATE_THERMAL_CLUSTERS.value: UpdateThermalClusters,
     CommandName.CREATE_RENEWABLES_CLUSTER.value: CreateRenewablesCluster,
     CommandName.REMOVE_RENEWABLES_CLUSTER.value: RemoveRenewablesCluster,
-    CommandName.UPDATE_RENEWABLE_CLUSTER.value: UpdateRenewableCluster,
+    CommandName.UPDATE_RENEWABLE_CLUSTERS.value: UpdateRenewableClusters,
```
TheoPascoli: Do you think we should match the create and remove command name by adding a "s" on renewable ?

tests/integration/study_data_blueprint/test_renewable.py
```diff
@@ -614,9 +615,66 @@ def test_variant_lifecycle(self, client: TestClient, user_access_token: str, var
...
+        res = client.post("/v1/studies", params={"name": "study_test", "version": "8.8"})
+        study_id = res.json()
+        area_id = "area_1"
+        res = client.post(f"/v1/studies/{study_id}/areas", json={"name": area_id, "type": "AREA"})
+        res.raise_for_status()
+
+        # Creates 50 renewable clusters inside the same area
+        body = {}
+        for k in range(50):
+            cluster_id = f"th_{k}"
```
TheoPascoli: Minor : If "th" is a c/c from the thermal PR, you can rename it

TheoPascoli (changes requested):

antarest/study/storage/variantstudy/model/command/common.py
```diff
@@ -48,7 +48,7 @@ class CommandName(Enum):
     UPDATE_THERMAL_CLUSTERS = "update_thermal_clusters"
     CREATE_RENEWABLES_CLUSTER = "create_renewables_cluster"
     REMOVE_RENEWABLES_CLUSTER = "remove_renewables_cluster"
-    UPDATE_RENEWABLE_CLUSTER = "update_renewable_cluster"
```
sylvlecl, line 51: This removal will break existing studies that have the previous command.
It's acceptable because it has not been released yet, but we should still search for existing studies on test platforms that have those commands, and either remove them or fix them.

antarest/study/storage/variantstudy/command_factory.py
MartinBelthle: Mmmh indeed it would be better to harmonize

MartinBelthle: I added the s everywhere (file, class and string)

antarest/study/storage/variantstudy/model/command/update_renewable_clusters.py
MartinBelthle: Indeed

antarest/study/storage/variantstudy/model/command/update_renewables_clusters.py
```diff
@@ -0,0 +1,123 @@
...
+
+        return CommandOutput(status=True, message="The renewable clusters were successfully updated."), {}
+
+    @override
+    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        for area_id, value in self.cluster_properties.items():
+            ini_path = ["input", "renewables", "clusters", area_id, "list"]
+
+            try:
+                all_clusters_for_area = study_data.tree.get(ini_path)
```
sylvlecl, line 85: we could validate the whole list here, instead of doing it later item-wise.

Generally speaking, we should validate objects as soon as we read them, unlike the current practice where we use dicts everywhere.
(yes there could be a minor performance cost here because we will also validate items that we don't modify, but it's a better tradeoff in my opinion)

antarest/study/storage/variantstudy/model/command/update_renewables_clusters.py
```diff
@@ -0,0 +1,123 @@
...
+            except ChildNotFoundError:
+                return CommandOutput(status=False, message=f"The area '{area_id}' is not found.")
+
+            for cluster_id, properties in value.items():
+                if cluster_id not in all_clusters_for_area:
```
sylvlecl, line 90: We will probably have problems with IDs, names,  lower case etc. here:
in cluster files, the section names are cluster names, not IDs, according to the simulator.

What are we expected to use in the command object ? the ID ? the name ?
If it's the ID (lower case), we will surely have problems.

I think the safest is to match based on a transformation to ID first.

antarest/study/storage/variantstudy/model/command/update_renewable_cluster.py
```diff
@@ -1,102 +0,0 @@
...
-            ),
-            {},
-        )
-
-    @override
-    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
-        path = _RENEWABLE_CLUSTER_PATH.format(area_id=self.area_id, cluster_id=self.cluster_id).split("/")
-
-        current_renewable_properties = create_renewable_properties(
-            study_version=self.study_version, data=study_data.tree.get(path)
```
sylvlecl, line 75: This is hidden (which is ugly), but here the search was performed in a case-insensitive way.

tests/integration/study_data_blueprint/test_renewable.py
```diff
@@ -614,9 +615,66 @@ def test_variant_lifecycle(self, client: TestClient, user_access_token: str, var
...
             "create_renewables_cluster",
-            "update_renewable_cluster",
+            "update_renewables_clusters",
             "replace_matrix",
             "create_renewables_cluster",
             "replace_matrix",
             "remove_renewables_cluster",
         ]
+
+    def test_update_multiple_renewable_clusters(self, client: TestClient, user_access_token: str) -> None:
```
sylvlecl, line 625: Not great that we don't have unit tests for that command and for the renewable manager !!

antarest/study/storage/variantstudy/model/command/update_renewables_clusters.py
```diff
@@ -0,0 +1,123 @@
...
+)
+from antarest.study.storage.rawstudy.model.filesystem.config.validation import AreaId
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand, OutputTuple
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+ClusterID = str
+RenewableClusterUpdates = dict[AreaId, dict[ClusterID, RenewableClusterUpdate]]
```
sylvlecl, line 31: This type should be defined close to RenewableClusterUpdate, and used also in the manager class.

Note that dictionaries don't make great input objects, we don't really know what's expected. For example in ` Mapping[str, Mapping[str, RenewableClusterUpdate]]` what the first string is supposed to be, and the second one ? With a named field in a class it would be more clear. Not asking to change this now though, it's quite homogeneous with the rest of the application

sylvlecl (commented):
More review

antarest/study/storage/variantstudy/model/command/update_renewables_clusters.py
MartinBelthle: I agree with the first comment. I'll modify this

For the rest of the comment I don't totally agree as okay it's not a whole object but IMO putting AreaId and ClusterId inside the type is pretty clear

antarest/study/storage/variantstudy/model/command/update_renewables_clusters.py
MartinBelthle: Mmmmmh you're right. I'll add the transformation to id and also test cases

PR #2424, opened 2025-03-20
fix(commands): use ids inside newly introduced update commands

MartinBelthle:


MartinBelthle:
I put read_ini instead of IniReader inside the next PR

antarest/study/storage/variantstudy/model/command/update_thermal_clusters.py
```diff
@@ -81,19 +79,25 @@ def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = N
...
                     return CommandOutput(
                         status=False,
                         message=f"The thermal cluster '{cluster_id}' in the area '{area_id}' is not found.",
                     )
                 # Performs the update
                 new_properties_dict = properties.model_dump(mode="json", by_alias=False, exclude_unset=True)
-                current_properties_obj = create_thermal_properties(
-                    self.study_version, all_clusters_for_area[cluster_id]
-                )
+                current_properties_obj = clusters_by_id[cluster_id][1]
```
sylvlecl: To make it more clear (and also to avoid searching the dict twice) we could at least unpack the tuple like this:
```
cluster_key, cluster = clusters_by_id[cluster_id]
```

sylvlecl: To go further, since this will be a quite common scheme to index by ID, we could create a small utility class for this, that we can re-use in multiple commands.

For example an `IdMapping` that would take as input the original dictionary, and expose a getter based on the ID

sylvlecl (changes requested):
Minor request for clarification, the logic seems right

PR #2425, opened 2025-03-20
perf(table-mode): speed-up mass update for short term storages

MartinBelthle:


TheoPascoli (dismissed):

antarest/study/storage/variantstudy/model/command/update_st_storages.py
```diff
@@ -0,0 +1,140 @@
...
+                if storage_id not in storages_by_id:
+                    return CommandOutput(
+                        status=False,
+                        message=f"The short-term storage '{storage_id}' in the area '{area_id}' is not found.",
+                    )
+                # Performs the update
+                new_properties_dict = properties.model_dump(mode="json", by_alias=False, exclude_unset=True)
+                current_properties_obj = storages_by_id[storage_id][1]
+                updated_obj = current_properties_obj.model_copy(update=new_properties_dict)
+                all_clusters_for_area[storages_by_id[storage_id][0]] = updated_obj.model_dump(
```
sylvlecl: Same request for clarification as in the other PR:
at least unpack the tuple with named arguments, but I think we can do better with a utility class that will embed the logic of the indexing based on the ID

sylvlecl (changes requested):
Same minor comment as in other PR about clarification of mapping code

PR #2427, opened 2025-03-21
feat(ui-studies): display the study list in the tree even if folder list is loading

skamril:


webapp/src/components/App/Studies/StudyTree/index.tsx
```diff
@@ -59,55 +65,60 @@ function StudyTree() {
...
-      const newSubFolders = await fetchSubfolders(itemId);
-      // use union to prioritize new subfolders
-      const nextSubfolders = R.unionWith(R.eqBy(R.prop("path")), newSubFolders, subFolders);
-      setSubFolders(nextSubfolders);
-      const nextStudiesTree = insertFoldersIfNotExist(studiesTree, nextSubfolders);
-      setStudiesTree(nextStudiesTree);
-    } catch {
+      const newSubFolders = await api.getFolders(workspace, subPath.join("/"));
+
+      if (newSubFolders.length > 0) {
```
smailio, line 86: you do this for optimizations ? Because I guess the unionWith will return a new reference and a a state update is triggered for nothing ?

webapp/src/components/App/Studies/StudyTree/index.tsx
```diff
@@ -59,55 +65,60 @@ function StudyTree() {
...
+    const isValidWorkspace = workspace && workspace !== DEFAULT_WORKSPACE_NAME;
+
+    if (!isAbsolutePath || !isValidWorkspace) {
+      return null;
     }
-    // usefull flag to display "loading..." message
-    setItemsLoading([...itemsLoading, itemId]);
-    // fetch subfolders and inesrt them to the tree
+
+    setItemsLoading((prev) => [...prev, itemId]);
```
smailio, line 80: is it usefull here to use a callback instead of just giving the new state ?

webapp/src/components/App/Studies/StudyTree/StudyTreeNode.tsx
```diff
@@ -12,44 +12,45 @@
...
-    <TreeItemEnhanced itemId={id} label={studyTreeNode.name} onClick={() => onNodeClick(id)}>
+    <TreeItemEnhanced
+      itemId={path}
+      label={name}
+      onClick={() => onNodeClick(node.path)}
+      loading={isLoading}
+    >
+      {hasUnloadedChildren && (
+        <TreeItemEnhanced
+          itemId={`${path}//loading`}
```
smailio, line 45: this double slash is normal ?

webapp/src/components/App/Studies/StudyTree/index.tsx
skamril: Yes because the `setItemsLoading` is called inside a Promise, so to be sure to have the last value of `itemsLoading` we need to use a callback (cf. race condition)

webapp/src/components/App/Studies/StudyTree/StudyTreeNode.tsx
skamril: No, I add it to made to a distinction with valid path, in case with have a child named `loading`

webapp/src/components/App/Studies/StudyTree/index.tsx
skamril: That's it

smailio (approved):

PR #2429, opened 2025-03-25
fix(db): fix relationship issues inside tests

MartinBelthle:


TheoPascoli (dismissed):

antarest/dbmodel.py
```diff
@@ -10,7 +10,20 @@
...
 from antarest.core.persistence import Base as PersistenceBase
+from antarest.core.tasks.model import TaskJob
+from antarest.launcher.model import JobResult
+from antarest.login.model import Identity
 
 Base = PersistenceBase
+
+# Define a one-to-many relationship with `JobResult`.
+# If an identity is deleted, all the associated job results are detached from the identity.
+Identity.job_results = relationship(JobResult, back_populates="owner", cascade="save-update, merge")
```
sylvlecl: I think it works a little "by chance" because we import this `antarest.dbmodel` from `conftest_db.py`.

Since that's the case, it seems to me that it would be enough to just import the 3 involved models here, while still keeping the relationship definitions in the classes.

If this works, I think it would be a slightly better solution, what do you think ?

MartinBelthle: Yep it's way better

PR #2430, opened 2025-03-25
refactor(thermal): improve classes design for thermal clusters

sylvlecl:
The goal of that work is to propose a new schema for business models classes,
starting with thermal clusters.

The main points are:
- separate business models for `ThermalCluster` themselves, their update, and their creation
- creation and update become plain python functions that can be re-used elsewhere
- no more version-related classes: however business classes can be validated on demand against
  a given study version
- separate "model" for serialization to INI file. Serialization and parsing depend on business model,
  not the other way around

A change in serialization too:
- in "file tree config", we don't use anymore the INI-like serialization, but camel case serialization, like any other think exposed in our API
- in commands, we don't use anymore the INI-like serialization either, which should be specific to the INI format either


antarest/study/storage/variantstudy/model/command/update_thermal_clusters.py
```diff
@@ -14,12 +14,15 @@
 from typing_extensions import override
 
 from antarest.core.exceptions import ChildNotFoundError
-from antarest.study.business.model.thermal_cluster_model import ThermalClusterUpdate
+from antarest.study.business.model.thermal_cluster_model import (
```
MartinBelthle, line 18: Don't quite see how you handled this but I just saw an issue we have in `dev` branch inside this code. The command expects `ThermalClusterUpdates` which has all field no matter the version. So we should check that the user didn't fill a v8.7 field with a study in version 8.1 for instance. I'm implementing a quick validator inside st-storage for the 9.2 but I won't touch the thermal as it would probably conflict with you PR

sylvlecl: I did not handle this yet actually (it will raise an exception when applying the command, but not before).
Because the commands have the study_version, we could check the validity as an "after" class validator.

sylvlecl: I added those new validators in commands in last commit.

TheoPascoli (dismissed):

antarest/core/exceptions.py
```diff
@@ -525,9 +525,9 @@ def __init__(self, message: str) -> None:
         super().__init__(HTTPStatus.BAD_REQUEST, message)
 
 
-class InvalidFieldForVersionError(HTTPException):
+class InvalidFieldForVersionError(HTTPException, ValueError):
```
MartinBelthle, line 528: As we discussed, we shouldn't have HTTP notions deep in the code. To fix that we could introduce an HTTP handler. It doesn't have to be done inside this PR but it could be nice as it we do nothing the following PRs will have the same behavior

antarest/study/business/areas/thermal_management.py
```diff
@@ -370,7 +352,7 @@ def duplicate_cluster(
 
         study.add_commands(commands)
 
-        return ThermalClusterOutput(**new_config.model_dump(mode="json", by_alias=False))
+        return self.get_cluster(study, area_id, new_id)
```
MartinBelthle: Seems that you added an i/o operation :/

antarest/study/business/model/thermal_cluster_model.py
```diff
@@ -9,56 +9,373 @@
...
+                group="Gas",
+                name="Gas Cluster XY",
+                enabled=False,
+                unit_count=100,
+                nominal_capacity=1000.0,
+                gen_ts="use global",
+                co2=7.0,
+            ).model_dump(mode="json")
+
+    name: Optional[str] = None  # TODO SL: better ignore this, not handled correctly
```
MartinBelthle: Several TODO SL left in this file. Don't know what you want to do with them

antarest/study/business/model/thermal_cluster_model.py
```diff
@@ -9,56 +9,373 @@
...
+        for field in ["nh3", "so2", "nox", "pm2_5", "pm5", "pm10", "nmvoc", "op1", "op2", "op3", "op4", "op5"]:
+            _initialize_field_default(cluster, field, 0)
+
+    if version >= STUDY_VERSION_8_7:
+        _initialize_field_default(cluster, "cost_generation", ThermalCostGeneration.SET_MANUALLY)
+        _initialize_field_default(cluster, "efficiency", 100.0)
+        _initialize_field_default(cluster, "variable_o_m_cost", 0.0)
+
+
+def create_thermal_cluster(cluster_data: ThermalClusterCreation, version: StudyVersion) -> ThermalCluster:
```
MartinBelthle, line 390: I'm a bit confused about the architecture decision we're taking on these methods. IMO they should be inside the thermal.py file

antarest/study/web/study_data_blueprint.py
```diff
@@ -1898,13 +1898,12 @@ def delete_renewable_clusters(
         path="/studies/{uuid}/areas/{area_id}/clusters/thermal",
         tags=[APITag.study_data],
         summary="Get thermal clusters for a given area",
-        response_model=Sequence[ThermalClusterOutput],
```
MartinBelthle, line 1901: Why did you remove this `response_model` ?

tests/integration/studies_blueprint/assets/test_synthesis/raw_study.synthesis.json
```diff
@@ -18,219 +18,354 @@
...
+          "volatilityPlanned": 0.0,
+          "lawForced": "uniform",
+          "lawPlanned": "uniform",
+          "marginalCost": 40.0,
+          "spreadCost": 0.0,
+          "fixedCost": 0.0,
+          "startupCost": 0.0,
+          "marketBidCost": 40.0,
+          "co2": 0.0,
+          "nh3": null,
```
MartinBelthle, line 159: This is kind of a breaking change for the R scripts. We should see with them if that's an issue

tests/integration/study_data_blueprint/test_table_mode.py
```diff
@@ -397,7 +397,7 @@ def test_lifecycle__nominal(
         if study_version >= 860:
             _solar_values["so2"] = 8.25
         if study_version >= 870:
-            _solar_values.update({"costGeneration": "useCostTimeseries", "efficiency": 87, "variableOMCost": -12.5})
+            _solar_values.update({"costGeneration": "useCostTimeseries", "efficiency": 87, "variableOMCost": 12.5})
```
MartinBelthle, line 400: You added the check on the fact that `variable_om_cost` should be higher than 0 inside this PR. Is it wanted ?

tests/integration/study_data_blueprint/test_thermal.py
```diff
@@ -483,21 +519,28 @@ def test_lifecycle(self, client: TestClient, user_access_token: str, internal_st
         # The thermal cluster properties should not have been updated.
         res = client.get(f"/v1/studies/{internal_study_id}/areas/{area_id}/clusters/thermal/{fr_gas_conventional_id}")
         assert res.status_code == 200, res.json()
-        assert res.json() == fr_gas_conventional_cfg
+        assert res.json() == expected
 
         # Update with a pollutant. Should succeed even with versions prior to v8.6
+        # TODO SL: chek if it's still necessary ? Should raise instead ?
```
MartinBelthle: We need to adress that. I'm almost sure we don't want this behavior anymore

tests/integration/study_data_blueprint/test_thermal.py
```diff
@@ -483,21 +519,28 @@ def test_lifecycle(self, client: TestClient, user_access_token: str, internal_st
...
         res = client.patch(
             f"/v1/studies/{internal_study_id}/areas/{area_id}/clusters/thermal/{fr_gas_conventional_id}",
             json={"nox": 10.0},
         )
-        assert res.status_code == 200
+        if version >= STUDY_VERSION_8_6:
+            assert res.status_code == 200, res.json()
+        else:
+            assert res.status_code == 422, res.json()
```
MartinBelthle, line 528: Could be nice to check the description inside res.json

tests/variantstudy/model/command/test_update_thermal_cluster.py
```diff
@@ -53,17 +57,16 @@ def _set_up(self, study: FileStudy, command_context: CommandContext, area_name:
...
         }
-        study.tree.save(thermal, ["input", "thermal", "clusters", area_id, "list", thermal_id])
-        thermal.update({"id": thermal_id})
-        thermal_config = create_thermal_config(study_version=study.config.version, **thermal)
+        study.tree.save(thermal, ["input", "thermal", "clusters", area_id, "list", thermal_name])
+        thermal_config = parse_thermal_cluster(study_version=study.config.version, data=thermal)
         study.config.areas[area_id].thermals.append(thermal_config)
 
     def test_update_thermal(self, empty_study_810: FileStudy, command_context: CommandContext):
+        # TODO SL: have a name different from ID
```
MartinBelthle: Seems you didn't adress this TODO

MartinBelthle (changes requested):

antarest/study/business/model/thermal_cluster_model.py
sylvlecl: You mean `thermal.py` in filesystem package ? This one should only handle parsing/serialization.

Here the `create_thermal_cluster` is a simple business-level operation (like update_thermal_cluster), which only operates on business models defined here, so its place is here together with those business models.

antarest/study/web/study_data_blueprint.py
sylvlecl: It's redundant, when it's the same thing as the method return type, see https://fastapi.tiangolo.com/tutorial/response-model/#response_model-parameter

tests/integration/studies_blueprint/assets/test_synthesis/raw_study.synthesis.json
sylvlecl: Indeed, I checked in the packages code that the synthesis is not used.

Note that not long ago, the synthesis did not return all thos information, but only the `enabled` flag. Not sure why we decided to return more ... I don't think it was a really conscious decision. I guess it was to have more information in the "config" on the backend side. It's a techncial debt that the 2 are linked together: we should distinguish what we want in cache on the backend side, from what we consider as a synthesis of the study.

The front still has a model with only the name, the ID, and the enabled flag.

tests/integration/study_data_blueprint/test_table_mode.py
sylvlecl: Yes, the simulator considers it invalid to have a negative value here.
```cpp
    if (variableomcost < 0.)
    {
        logs.error() << "Thermal cluster: " << parentArea->name << '/' << pName
                     << ": The variable operation & maintenance cost must be positive or null";
        variableomcost = 0.;
        ret = false;
    }
```

tests/integration/study_data_blueprint/test_thermal.py
sylvlecl: Actually I changed the behaviour, I will remove that comment. I also checked that it's not an issue from the front-end side.

tests/variantstudy/model/command/test_update_thermal_cluster.py
sylvlecl: I actually removed the possibility to rename (checked with users), since it does not work as expected (it unsyncs the section name and property "name" in INI files which does not make sense from the simulator point of view).

I will remove that comment.

tests/integration/study_data_blueprint/test_thermal.py
sylvlecl: Could be considered redundant, the actual error message is already checked in the unit tests

antarest/study/business/model/thermal_cluster_model.py
sylvlecl: I handled those:
- we now ignore the name in inputs
- data validation is implemented for all 3 models, with annotated types where relevant to avoid repetition

antarest/study/business/areas/thermal_management.py
sylvlecl: Indeed, I restored the previous behaviour.

antarest/core/exceptions.py
sylvlecl: Agreed.
It needs more thinking though:
if I just create a new exception that does not inherit from HTTPException, it will not translate correctly to 4** status code.
Indeed we could add a specific handler, but I am not sure it's the best solution:
we should only return a 4** when the error is in the request input data, which is not guaranteed. For example the errors might be in existing study data or in stored commands.

So maybe we'll need to separate more the request input validation from other processing...

I prefer to think more about the solution before actually doing something.

antarest/study/business/model/thermal_cluster_model.py
MartinBelthle: As we discussed, it was blurry at the time but I agree now

antarest/study/business/areas/thermal_management.py
MartinBelthle: I still see it :/

antarest/study/business/model/thermal_cluster_model.py
```diff
@@ -9,56 +9,390 @@
...
-from antarest.study.storage.rawstudy.model.filesystem.config.thermal import (
-    Thermal870Config,
-    Thermal870Properties,
-    ThermalPropertiesType,
-    create_thermal_properties,
-)
+from antarest.core.serde import AntaresBaseModel
+from antarest.study.business.enum_ignore_case import EnumIgnoreCase
+from antarest.study.model import STUDY_VERSION_8_6, STUDY_VERSION_8_7
+from antarest.study.storage.rawstudy.model.filesystem.config.identifier import transform_name_to_id
```
MartinBelthle, line 44: Minor: My IDE says that we didn't add `transform_name_to_id` inside the file `identifier.py`. I think we should do it

tests/integration/study_data_blueprint/test_thermal.py
```diff
@@ -51,11 +51,12 @@
...
 
 from antarest.core.utils.string import to_camel_case
+from antarest.study.business.model.thermal_cluster_model import ThermalCluster
+from antarest.study.model import STUDY_VERSION_8_6, STUDY_VERSION_8_7
 from antarest.study.storage.rawstudy.model.filesystem.config.identifier import transform_name_to_id
-from antarest.study.storage.rawstudy.model.filesystem.config.thermal import ThermalProperties
 from tests.integration.utils import wait_task_completion
 
-DEFAULT_PROPERTIES = ThermalProperties(name="Dummy").model_dump(mode="json")
+DEFAULT_PROPERTIES = ThermalCluster(name="Dummy").model_dump(mode="json")
```
MartinBelthle: We don't use this it seems

MartinBelthle (changes requested):
Minor review. Once it's fixed we can merge without another review

PR #2431, opened 2025-03-26
chore(tests): remove useless zip and introduce new study fixtures

MartinBelthle:


tests/conftest.py
```diff
@@ -46,3 +66,135 @@ def write_clean_ini(path: Path, txt: str) -> None:
         path.write_text(clean_ini)
 
     return write_clean_ini
+
+
+@pytest.fixture(name="matrix_service")
+def matrix_service_fixture() -> MatrixService:
```
sylvlecl: Note that there is another matrix fixture in a nested `conftest.py`, I guess the nested one takes precedence, but it could be misleading. I don"t have any alternative to propose though :)

Also, cannot we just re-use the `InMemorySimpleMatrixService` instead of this mock ?
The study does not need a full MatrixService, only a `ISimpleMatrixService`

tests/storage/business/test_xpansion_manager.py
```diff
@@ -69,129 +52,98 @@ def make_link(link_manager: LinkManager, study: StudyInterface) -> None:
...
-def test_delete_xpansion_configuration(xpansion_manager: XpansionManager, tmp_path: Path) -> None:
+def test_delete_xpansion_configuration(
+    xpansion_manager: XpansionManager, tmp_path: Path, empty_study_810: FileStudy
+) -> None:
     """
     Test the deletion of a configuration.
     """
-    study = make_empty_study(tmp_path, 810)
-
+    empty_study = FileStudyInterface(empty_study_810)
```
sylvlecl: actually I think just `study` was a better name: the study does not remain empty for long (same in existing tests though, tha could be part of another PR)

sylvlecl (changes requested):
Looks good: I just propose to take this opportunity to remove the matrix service mock and re-use the in memory implementation instead

PR #2433, opened 2025-03-27
feat(desktop): manage external studies

smailio:


antarest/study/repository.py
```diff
@@ -316,6 +316,7 @@ def _search_studies(
         # efficiently (see: `AbstractStorageService.get_study_information`)
         entity = with_polymorphic(Study, "*")
 
+        escape_char = "\\"
```
sylvlecl, line 319: Does this change works for both sqlite and postgresql ?
After some reading, I expect it does, well done.

antarest/study/service.py
```diff
@@ -984,6 +984,67 @@ def create_study(
...
+        """
+
+        author = self.get_user_name(params)
+        now = datetime.now(timezone.utc)
+        path = str(study_folder.path)
+        raw = RawStudy(
+            id=str(uuid4()),
+            name=study_folder.path.name,
+            folder=path,
+            workspace="external",
```
sylvlecl: We could have a constant next to DEFAULT_WORKSPACE_NAME

antarest/study/web/explorer_blueprint.py
```diff
@@ -60,6 +62,45 @@ def list_dir(
         logger.info(f"Listing directory {path} in workspace {workspace}")
         return explorer.list_dir(workspace, path)
 
+    @bp.post(
```
sylvlecl: Maybe we could not at all add those routes when desktop_mode is not enabled:
it would be more secure.

antarest/study/storage/explorer_service.py
```diff
@@ -76,3 +93,56 @@ def list_workspaces(
...
+        if not self.config.desktop_mode:
+            logger.warning("Called open api when desktop mode was off")
+            raise ExternalWorkspaceDisabled("Desktop mode is not enabled in the configuration")
+
+        # check path is a study folder and we have read permission
+        if not is_study_folder(path):
+            raise HTTPException(HTTPStatus.UNPROCESSABLE_ENTITY, f"Path {path} is not a study folder")
+
+        # check path is not in a filtered folder
+        if should_ignore_folder_for_scan(path, [".*"], []):
```
sylvlecl: Actually I am not sure we need this: if the user wants explicitly to open a study, I don't think it's useful to add some constraint on it

antarest/study/storage/explorer_service.py
```diff
@@ -76,3 +93,56 @@ def list_workspaces(
...
+            raise HTTPException(HTTPStatus.UNPROCESSABLE_ENTITY, "Cannot open a file in a filtered folder")
+
+        # check if path is inside the default workspace folder
+        default_workspace = get_workspace_from_config(self.config, DEFAULT_WORKSPACE_NAME, default_allowed=True)
```
sylvlecl, line 109: Nice to have identified that specific case!

antarest/study/service.py
```diff
@@ -984,6 +984,67 @@ def create_study(
...
+                type=EventType.STUDY_CREATED,
+                payload=raw.to_json_summary(),
+                permissions=PermissionInfo.from_study(raw),
+            )
+        )
+
+        logger.info("External study %s created by user %s", raw.id, params.get_user_id())
+        return str(raw.id)
+
+    def delete_external_study(self, study: Study) -> None:
```
sylvlecl, line 1042: I am wondering: 
what are the consequences if the user calls the usual "delete" endpoint for such a study ?

I guess it will raise because I think by default deletion of raw studies is not allowed, but to be checked.

An alternative would be to go through the existing `delete_study` and add the specific case of external studies there. For this, we would need a convention to identify external studies:
should we rely on the "external" workspace for this ?

antarest/core/core_blueprint.py
```diff
@@ -59,6 +63,22 @@ def version_info() -> Any:
             version=antarest_version,
             gitcommit=get_commit_id(config.resources_path),
             dependencies=get_dependencies(),
+            desktop_mode=config.desktop_mode,
         )
 
+    @bp.get(
```
sylvlecl: I think we had decided on keeping only the boolean in version info (however no problem to keep it if it's finally useful).

antarest/study/service.py
```diff
@@ -984,6 +985,68 @@ def create_study(
...
+
+        author = self.get_user_name(params)
+        now = datetime.now(timezone.utc)
+        path = str(study_folder.path)
+        folder = f"{EXTERNAL_WORKSPACE_NAME}/{path}"
+        raw = RawStudy(
+            id=str(uuid4()),
+            name=study_folder.path.name,
+            folder=folder,
+            workspace=EXTERNAL_WORKSPACE_NAME,
```
sylvlecl, line 1018: I was wondering if having a workspace name which does not have any configuration could be an issue, but it does not seem so, by reading the use of `workspace` in the code (not used very much, apart from checking if a study is a "managed" one).
We should carefully test the behaviour though, to be sure.

sylvlecl (changes requested):
Looks good to me overall, I lef minor comments.
It's also a little difficult to be sure that everything works correctly with this "fictitious" workspace, we'll need to test it quite in depth.

antarest/study/web/explorer_blueprint.py
smailio: That's right. Shoud I remove the check I do inside the service ? Or do you think it makes sense to keep it ?

antarest/study/service.py
smailio: It's the first thing I checked, the call failed and I didnt dig into to it back then and I just went with my own version to avoid introducing side effects. 

Indeed we can update `delete_study`, I tried in this PR it seems to work.

antarest/core/core_blueprint.py
smailio: I just forgot to remove it :sweat_smile:

antarest/study/service.py
smailio: Turns out it wasn't that easy I broke multiple tests, working on it.

smailio: Done.

antarest/study/web/explorer_blueprint.py
sylvlecl: I think it's ok to keep them, safer for the future, although indeed it will be redundant now.

PR #2434, opened 2025-03-27
fix(area): fix bug on allocation and correlation views

maugde:
Fix #[ANT-2960](https://gopro-tickets.rte-france.com/browse/ANT-2960)

This PR must fix bugs that happen when asking for the correlation view, the allocation view and their sub views (`view all correlations` and `view all allocations`). They exist only on a few studies (i.e. https://antares-web.rte-france.com/studies/5030aa35-3a3a-4153-843c-902189e9b6f7/explore/debug?path=input%2Fhydro%2Fprepro%2Fcorrelation on production).

I updated the checks in order to make them as permissive as the external tools which could modify studies too.

The bugs fixed are :
- accepting correlation.ini files that could have an empty "annual" property
- reading allocation.ini files that could have another format ("[" instead of "[[")
- respect rules of external tool that convert `min-up-time` and `min-down-time` values that exceed bounds to 1 if the value is lesser than 1 and 168 if the value is greater than 168

### DIFF
--- a/antarest/study/business/allocation_management.py
+++ b/antarest/study/business/allocation_management.py
@@ -139,7 +139,8 @@ class AllocationManager:
         if not allocation_data:
             raise AllocationDataNotFound(area_id)
 
-        return allocation_data.get("[allocation]", {})  # type: ignore
+        # allocation format can differ from the number of '[' (i.e. [[allocation]] or [allocation])
+        return allocation_data.get("[allocation]", allocation_data.get("allocation", {}))  # type: ignore
 
     def get_allocation_form_fields(
         self, all_areas: List[AreaInfoDTO], study: StudyInterface, area_id: str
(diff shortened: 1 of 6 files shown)


antarest/study/storage/rawstudy/model/filesystem/config/thermal.py
```diff
@@ -135,15 +135,15 @@ def get_id(self) -> str:
     )
     min_up_time: int = Field(
         default=1,
-        ge=1,
```
sylvlecl: Note that to make it have the same behaviour as the simulator, I think we should also crop it to 1 or 168 if it's outside of the range [1,168].

antarest/study/business/allocation_management.py
```diff
@@ -139,7 +139,8 @@ def get_allocation_data(self, study: StudyInterface, area_id: str) -> Dict[str,
         if not allocation_data:
             raise AllocationDataNotFound(area_id)
 
-        return allocation_data.get("[allocation]", {})  # type: ignore
+        # allocation format can differ from the number of '[' (i.e. [[allocation]] or [allocation])
```
sylvlecl, line 142: well spotted, this is unbelievable ...

Actually, when the simulator reads the file, it seems to not even care about what's in the section name.
However when it writes to the file, it writes `[[allocation]]`

sylvlecl (approved):

PR #2437, opened 2025-03-31
fix(variants): don't raise Exeptions when deleting a variant

MartinBelthle:
Fix [ANT-3046]


antarest/study/service.py
```diff
@@ -1420,15 +1420,12 @@ def delete_study(self, uuid: str, children: bool, params: RequestParameters) ->
             else:
                 raise StudyDeletionNotAllowed(study.id, "Study has variant children")
 
-        self.repository.delete(study.id)
+        # If the study is a variant, and its snapshot is generating,
```
sylvlecl, line 1424: I think it would be more appropriate to move this code to `variant_service.delete`, to avoid this specific case handling

sylvlecl (commented):
Is it just possible to move the specific handling of the variant into variant_study service ?

As discussed, at some point we'll need to review broadly the management of concurrent access to studies. This fix only fiexes a specific case, for example we'll still be able to trigger a delete and launch a generation during the deletion, with unclear outcome.

sylvlecl (approved):
Ok as a quick fix. The deletion method would deserve some rework like many else: there are many specific case handling, that could probably be clearer with some abstraction.

PR #2438, opened 2025-03-31
chore(variant): remove obsolete `execute_or_add_commands` method

MartinBelthle:


antarest/study/service.py
```diff
@@ -238,11 +237,13 @@ def __init__(
         repository: StudyMetadataRepository,
         storage_service: StudyStorageService,
         event_bus: IEventBus,
+        interface_method: Callable[[], Callable[[Study], StudyInterface]],
```
maugde: Maybe i am missing something but I'm wondering if it is possible to simplify the type of `interface_method` and its call

sylvlecl: Agreed, I think there is one un-necessary level of callable ?
Also we could name it `study_supplier` or `study_provider` to be more explicit: the callable **provides** a study interface.
We could also define our own small generic type in like `antares.core.typing`:
```python
T = TypeVar("T")
Supplier: TypeAlias = Callable[[], T]
```
Then use:
```python
study_supplier: Supplier[StudyInterface]
```

MartinBelthle: If I just type it as `interface_method: Callable[[Study], StudyInterface]` mypy is unhappy as it says `interface_method=lambda: study_service.get_study_interface` is a Callable of that . Don't know what you want me to do :/
And why not for the supplier but i don't understand how is that supposed to work as there's no Study specified and the method expects a study

sylvlecl (dismissed):
Nice to remove this, as discussed, ok after renaming and simplification of callable

PR #2439, opened 2025-03-31
feat(study): copy study with specific path

TheoPascoli:


antarest/core/config.py
```diff
@@ -693,17 +693,18 @@ def from_yaml_file(cls, file: Path, res: Optional[Path] = None) -> "Config":
             data["resources_path"] = res
         return cls.from_dict(data)
 
-    def get_workspace_path(self, *, workspace: str = DEFAULT_WORKSPACE_NAME) -> Path:
+    def get_workspace_path(self, *, workspace: str = DEFAULT_WORKSPACE_NAME, destination_path: str = "") -> Path:
```
sylvlecl: The "computation" of the subpath in the workspace should not go here:
 - first and most importantly, this is very misleading for the internal workspace, where the path is only materialized in database: the path of the study on disk remains <workspace_path>/<study_id>
 - then, this module is more about plain config data, it should not have too much knowledge of what the config is used for. In particular, "destination" is not an appropriate wording because it is very linked to the copy feature, whereas that method could be used for anything else than copy

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -227,46 +228,53 @@ def copy(
...
+                author=src_study.additional_data.author,
+                patch=src_study.additional_data.patch,
             )
         dest_id = str(uuid4())
         dest_study = RawStudy(
             id=dest_id,
             name=dest_name,
             workspace=DEFAULT_WORKSPACE_NAME,
-            path=str(self.config.get_workspace_path() / dest_id),
+            path=str(self.config.get_workspace_path(destination_path=destination_path) / dest_id),
```
sylvlecl: The destination path must not go there, but in the `folder` property.
Actually, when you look at the current mechanism we have the following data in the `Study` class:
- `path` is for the actual path on disk (honestly, I don't understand why we have this for internal studies, it's redundant with the workspace and the ID of the study, and it will break everything if we want to move the installation on disk ... more tehcnical debt again).
- `folder` is for the logical folder in which this study is

So it's very important the path remains `<workspace>/<id>`, but folder should be updated to `destination_path`

tests/integration/test_integration.py
```diff
@@ -1748,6 +1748,107 @@ def test_copy(client: TestClient, admin_access_token: str, internal_study_id: st
...
+        f"/v1/studies/{parent_id}/areas",
+        json={"name": "area1", "type": "AREA"},
+    )
+    client.post(
+        f"/v1/studies/{parent_id}/areas",
+        json={"name": "area2", "type": "AREA"},
+    )
+    variant = client.post(f"/v1/studies/{raw.json()}/variants", params={"name": "variant"})
+
+    dest_path = tmp_path / "internal_workspace" / "specific_path"
```
sylvlecl: Actually the study should not go into that path, we keep the internal studies flat, only the `folder` property should change

sylvlecl (changes requested):
Not exactly the expected behaviour: the destination path should only update the logical `folder` of the new study, not the actual path on disk

TheoPascoli:
> Not exactly the expected behaviour: the destination path should only update the logical `folder` of the new study, not the actual path on disk

I changed the code accordingly. Now I update the folder

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -270,6 +274,7 @@ def build_raw_study(self, dest_name: str, groups: Sequence[str], src_study: Stud
             additional_data=additional_data,
             public_mode=PublicMode.NONE if groups else PublicMode.READ,
             groups=groups,
+            folder=str(Path(destination_folder) / dest_id),
```
sylvlecl: This is so ridiculous that we need to append the ID :( ... but I know this is the way for consistency with the current mechanisms.

It's safer to not use `Path` here but `PurePosixPath`:
- pure because it does not relate to a real filesystem path
- posix because by convention we use slash separators, even on windows

antarest/study/common/studystorage.py
```diff
@@ -86,14 +86,22 @@ def exists(self, metadata: T) -> bool:
         """
 
     @abstractmethod
-    def copy(self, src_meta: T, dest_name: str, groups: Sequence[str], with_outputs: bool = False) -> T:
+    def copy(
+        self,
+        src_meta: T,
+        dest_name: str,
+        groups: Sequence[str],
+        destination_folder: str,
```
sylvlecl: let's use a `PurePosixPath` already here

antarest/study/service.py
```diff
@@ -1168,6 +1168,7 @@ def copy_study(
         group_ids: List[str],
         use_task: bool,
         params: RequestParameters,
+        destination_folder: str,
```
sylvlecl: let's use a PurePosixPath already here

sylvlecl (changes requested):
Really minor stuff about using pure posix path, otherwise it's good to go

sylvlecl (approved):

PR #2441, opened 2025-04-02
feat(ui-matrix): add copy paste and keybinds support

hdinia:


webapp/src/components/common/Matrix/components/MatrixGrid/index.tsx
```diff
@@ -141,6 +145,105 @@ function MatrixGrid({
...
+                      kind: GridCellKind.Number,
+                      data: value,
+                      displayData: formatGridNumber({ value }),
+                      allowOverlay: true,
+                    },
+                  });
+                }
+              }
+            }
+          }
```
skamril, line 222: Removes duplications: (I added a colum check also)

```typescript
const updates: GridUpdate[] = [];

const addItemUpdate = (item: Item) => {
  const column = columns[item[0]];
  
  if (!column.editable || column.type !== GridCellKind.Number) {
    return;
  }

  const coordinates = gridToData(item);

  if (coordinates) {
    updates.push({
      coordinates,
      value: {
        kind: GridCellKind.Number,
        data: value,
        displayData: formatGridNumber({ value }),
        allowOverlay: true,
      },
    });
  }
};

// Handle range selection
if (gridSelection.current?.range) {
  const { x, y, width, height } = gridSelection.current.range;

  for (let col = x; col < x + width; col++) {
    for (let row = y; row < y + height; row++) {
      addItemUpdate([col, row]);
    }
  }
}
// Handle row selections
else if (gridSelection.rows.length > 0) {
  for (const rowIndex of gridSelection.rows) {
    for (let col = 0; col < columns.length; col++) {
      addItemUpdate([col, rowIndex]);
    }
  }
}
// Handle column selections
else if (gridSelection.columns.length > 0) {
  for (const colIndex of gridSelection.columns) {
    for (let row = 0; row < rows; row++) {
      addItemUpdate([colIndex, row]);
    }
  }
}
```

webapp/src/components/common/Matrix/components/MatrixGrid/index.tsx
```diff
@@ -141,6 +145,105 @@ function MatrixGrid({
...
+  const handleKeyDown = useCallback(
+    (event: GridKeyEventArgs) => {
+      // Fill selection with value (Ctrl+Shift+Enter)
+      if (event.shiftKey && event.ctrlKey && event.key === "Enter") {
+        if (
+          gridSelection.current?.range ||
+          gridSelection.rows.length > 0 ||
+          gridSelection.columns.length > 0
+        ) {
+          const userInput = prompt(t("matrix.fillSelection.numberPrompt"));
```
skamril, line 157: Maybe better to use a dialog from MUI to keep consistent.

This shortcut works like this in Excel, or it's a specific one?

skamril (changes requested):

webapp/src/components/common/Matrix/components/MatrixGrid/index.tsx
```diff
@@ -141,6 +145,105 @@ function MatrixGrid({
...
+
+          if (userInput === null) {
+            return;
+          }
+
+          const value = Number(userInput);
```
skamril, line 164: Checks if `userInput` is empty because `Number("")` gives `0`. Use `trim()` also.

skamril (changes requested):

webapp/src/components/common/Matrix/components/MatrixGrid/index.tsx
hdinia: Yes we can do it in a future improvement

webapp/src/components/common/Matrix/components/MatrixGrid/index.tsx
hdinia: thanks

skamril (approved):

PR #2443, opened 2025-04-02
fix(ui): theme issues

skamril:
* CSS file and assets path in index.html were only loaded in home page
* Height issue with filter toolbar in Tasks view
* Add true result link in Results view


webapp/src/components/App/Tasks/JobTableView.tsx
```diff
@@ -57,37 +57,46 @@ const FILTER_LIST: Array<TaskView["type"]> = [
...
   content: TaskView[];
   refresh: () => void;
 }
 
-function JobTableView(props: PropType) {
+function JobTableView(props: Props) {
   const { content, refresh } = props;
   const [t] = useTranslation();
-  const [sorted, setSorted] = useState<string>();
+  const [dateOrder, setDateOrder] = useState<NonNullable<TableSortLabelProps["direction"]>>("desc");
```
hdinia, line 68:
```ts
const [sortDirection, setSortDirection] = useState<'asc' | 'desc'>('desc');
```

webapp/src/components/App/Tasks/JobTableView.tsx
```diff
@@ -97,12 +106,13 @@ function JobTableView(props: PropType) {
     setFilterRunningStatus(!filterRunningStatus);
   };
 
-  useEffect(() => {
-    setCurrentContent(applyFilter(content));
-  }, [content, applyFilter]);
+  const handleRequestDateSort = () => {
+    setDateOrder(dateOrder === "asc" ? "desc" : "asc");
+  };
```
hdinia, line 111:
```ts
const handleToggleSortDirection = () => {
  setSortDirection(prevDirection => prevDirection === 'asc' ? 'desc' : 'asc');
};
```

webapp/src/components/App/Tasks/JobTableView.tsx
```diff
@@ -57,37 +57,46 @@ const FILTER_LIST: Array<TaskView["type"]> = [
...
-        filteredContent = filteredContent.filter((o) => o.type === filterType);
-      }
-      return filteredContent;
-    },
-    [filterType, filterRunningStatus],
-  );
+  // Refresh launcher metrics every minute
+  useInterval(launcherMetrics.reload, 60_000);
+
+  const displayContent = useMemo(() => {
```
hdinia, line 80: `filteredTasks`

hdinia (changes requested):

PR #2447, opened 2025-04-07
feat(study): copy only selected outputs

TheoPascoli:


antarest/study/common/studystorage.py
```diff
@@ -92,6 +92,7 @@ def copy(
         dest_name: str,
         groups: Sequence[str],
         destination_folder: PurePosixPath,
+        allowed_outputs: List[str],
```
MartinBelthle: Minor: Not a huge fan of the naming. On the other hand, I suck at finding good names. Perhaps `output_ids` would be better to emphasize the fact that you consider them as ids

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -240,11 +255,13 @@ def copy(
         src_path = self.get_study_path(src_meta)
         dest_path = self.get_study_path(dest_study)
 
-        shutil.copytree(src_path, dest_path)
+        shutil.copytree(src_path, dest_path, ignore=shutil.ignore_patterns("output"))
+
+        if with_outputs:
```
MartinBelthle: IMO we should raise an Exception if `with_outputs` is False and we have a non-empty `allowed_outputs` list

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -240,11 +255,13 @@ def copy(
...
+            src_output_path = src_path / "output"
```
MartinBelthle: Minor: Could be done ine one line (line 264 ?)

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -45,6 +46,18 @@
 logger = logging.getLogger(__name__)
 
 
+def copy_output_folders(src_output_path: Path, dest_output_path: Path, allowed_outputs: List[str]) -> None:
```
MartinBelthle: Why not simply using a shutil.copytree ?

Also change the line 58. IMO we should raise an Exception if the name doesn't exist

antarest/study/web/studies_blueprint.py
```diff
@@ -356,6 +356,7 @@ def copy_study(
         uuid: str,
         dest: str,
         with_outputs: bool = False,
+        allowed_outputs: Annotated[list[str] | None, Query()] = None,
```
MartinBelthle: Why not just putting: `list[str] | None = None` ?

tests/integration/test_integration.py
```diff
@@ -1802,7 +1802,7 @@ def test_copy_as_variant_with_outputs(client: TestClient, admin_access_token: st
     # Copy of the variant as a reference study
     copy = client.post(
         f"/v1/studies/{variant.json()}/copy",
-        params={"dest": "copied", "with_outputs": True, "use_task": True},  # type: ignore
+        params={"dest": "copied", "with_outputs": True, "use_task": True, "allowed_outputs": ["output.txt"]},  # type: ignore
```
MartinBelthle: Forcing the user to do so is a breaking change. I think we should consider the flag `with_outputs` to say we want to export all outputs. Doing so will avoid a breaking change.

Or if we want to introduce this breaking change for some reason, we should document it. And I don't think we want that as for an user exporting all its outputs will be hard as he will have to fill them all.

tests/integration/test_integration.py
```diff
@@ -1843,6 +1843,63 @@ def test_copy_variant_with_specific_path(client: TestClient, admin_access_token:
...
+
+    copy = client.post(
+        f"/v1/studies/{variant.json()}/copy",
+        params={
+            "dest": "copied",
+            "with_outputs": True,
+            "use_task": True,
+            "allowed_outputs": ["output_0.txt", "output_1.txt"],
+        },
+        # type: ignore
```
MartinBelthle: Why do you have to do that ? it's not great :/

MartinBelthle (changes requested):

antarest/study/common/studystorage.py
TheoPascoli: Thanks for the suggestion

antarest/study/web/studies_blueprint.py
```diff
@@ -356,6 +356,7 @@ def copy_study(
         uuid: str,
         dest: str,
         with_outputs: bool = False,
+        output_ids: Annotated[list[str] | None, Query()] = None,
```
sylvlecl: Actually, for query parameters, we cannot distinguish between an empty list and a `None`. So it would probably be simpler to have a non optional and an empty list as the default here ?

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -925,9 +927,9 @@ def copy(
         shutil.copytree(src_path, dest_path)
 
         src_path = cast(Path, file_study.config.output_path)
-        if with_outputs and src_path.exists():
+        if with_outputs and src_path.exists() and output_ids:
```
sylvlecl: See other comment: we need to decide, if output_ids is empty, if this means that we want to copy all outputs.
I think it makes sense, since we can already copy no output at all with the 'with_outputs=False' option.

antarest/study/web/studies_blueprint.py
```diff
@@ -382,7 +384,7 @@ def copy_study(
         group_ids = [sanitize_string(gid) for gid in group_ids]
         uuid_sanitized = sanitize_uuid(uuid)
         destination_name_sanitized = escape(dest)
-
+        outputs = output_ids or []
```
sylvlecl: We need to decide on the following questions:
 1. if the user calls `/copy?output_ids=my_output`, he will now get an error, because the default values of `with_outputs` is False. Do we want to allow it ? (implicitly making with_output true in that case)
 2. if the list is empty, do we want co copy ALL outputs ? I think we should, otherwise we need to select the outputs 1 by 1 which is not convenient. It would also be a breaking change compared to current behaviour.

Any opinion on this @MartinBelthle @TheoPascoli ?

If we answer yes to 1, we need to add some logic, in particular:
- make `with_outputs` optional with None default
- if `output_ids` is not empty, implicitly with_outputs is True. If it's False, raise a validation error

In the end I think we should test the following behaviours:

- `copy?with_outputs=True&output_ids=my-output` -> ok, copies only my-output
- `copy?output_ids=my-output` -> **to be decided**, exactly the same or raise ?
- `copy?with_outputs=False&output_ids=my-output` -> raise, we have an inconsistency
- `copy?with_outputs=True` -> copies all outputs
- `copy?with_outputs=False` -> copies no output
- `copy` -> same, copies no output

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -45,6 +45,17 @@
...
```
sylvlecl: naming: maybe just "outputs" or "selected_outputs" or "filtered_outputs" ?  ("allowed" sounds weird, allowed to do what ? also allowed does not mean required, it sounds like we can copy them but we don't have to)

sylvlecl (changes requested):
Looks good overall, one thing though:
I think we need to clarify the behaviour for the new argument `output_ids` when it's empty, and its interaction with `with_outputs`.

antarest/study/web/studies_blueprint.py
sylvlecl: Actually I tried it and it does not work indeed: fastapi tries to recognize which argument is a query parameter and which one is the body, if it's not a "singular type", it will not be recognized as a query param:
https://fastapi.tiangolo.com/tutorial/body/#request-body-path-parameters

> If the parameter is of a singular type (like int, float, str, bool, etc) it will be interpreted as a query parameter.

antarest/study/web/studies_blueprint.py
TheoPascoli: I agree with you — with_outputs should be implicitly set to True if at least one ID is provided. Also, as Martin suggested, an empty list should imply copying all outputs.
I'll proceed with changes accordingly.

antarest/study/web/studies_blueprint.py
TheoPascoli: I set it this way to not have a mutable value as default. We don't want that

sylvlecl: You may use `Query(default_factory=list)` for that purpose (but indeed it's not that different)

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -240,11 +260,9 @@ def copy(
         src_path = self.get_study_path(src_meta)
         dest_path = self.get_study_path(dest_study)
 
-        shutil.copytree(src_path, dest_path)
+        shutil.copytree(src_path, dest_path, ignore=shutil.ignore_patterns("output"))
```
sylvlecl, line 263: Nice optimization !

sylvlecl (approved):

PR #2448, opened 2025-04-07
chore(outputs): create a dedicated output service to separate concerns

MartinBelthle:


antarest/launcher/main.py
```diff
@@ -21,13 +21,13 @@
...
 from antarest.launcher.service import LauncherService
 from antarest.launcher.web import create_launcher_api
-from antarest.study.service import StudyService
+from antarest.study.storage.output_service import OutputService
 
 
 def build_launcher(
     app_ctxt: Optional[AppBuildContext],
     config: Config,
-    study_service: StudyService,
```
sylvlecl, line 30: I would prefer to keep both dependencies explicit here, the output service AND the study service:
it's kind of an implementation detail of the output service that it has the study service.
The launcher service needs:
1. the study service to get the input
2. the output service to write the outptu

it's natural that it has both in its dependencies.

antarest/service_creator.py
```diff
@@ -126,7 +127,9 @@ def create_event_bus(app_ctxt: Optional[AppBuildContext], config: Config) -> Tup
 
 def create_core_services(
     app_ctxt: Optional[AppBuildContext], config: Config
-) -> Tuple[ICache, IEventBus, ITaskService, FileTransferManager, LoginService, MatrixService, StudyService]:
+) -> Tuple[
```
sylvlecl: I think it could be more clear to have a small dataclass `CoreServices` here with all services

antarest/study/web/output_blueprint.py
```diff
@@ -0,0 +1,528 @@
...
+        current_user: JWTUser = Depends(auth.get_current_user),
+    ) -> Any:
+        logger.info(f"Fetching output list for study {study_id}")
+        study_id = sanitize_uuid(study_id)
+        params = RequestParameters(user=current_user)
+        content = output_service.get_study_sim_result(study_id, params)
+        return content
+
+    @bp.get(
+        "/studies/{uuid}/areas/aggregate/mc-ind/{output_id}",
```
sylvlecl: It makes me realize that this URL is wrong :( 
it should be structured the opposite, with the output resource identified first: 
`/studies/<id>/outputs/<id>/aggregate/mc-ind/areas`

Could we add this new one and redirect the existing one towards it, for backward compat ?
Same for other endpoints ?

The idea is that all output-related URL should start with "/studies/<i>/outputs/<id>".

antarest/study/storage/output_service.py
```diff
@@ -0,0 +1,531 @@
...
+        output_path = Path(study.path) / "output" / output_id
+        if not is_output_archived(output_path):
+            if not output_path.exists():
+                raise OutputNotFound(output_id)
+            raise OutputAlreadyUnarchived(output_id)
+
+        archive_task_names = OutputService._get_output_archive_task_names(study, output_id)
+        task_name = archive_task_names[1]
+
+        study_tasks = self.study_service.task_service.list_tasks(
```
sylvlecl: the task service should be an explicit dependency of this one, if it uses it directly, it should not be used from the study service, it's an implementation detail of that one

antarest/study/storage/output_service.py
```diff
@@ -0,0 +1,531 @@
...
+from antarest.study.storage.study_download_utils import StudyDownloader, get_output_variables_information
+from antarest.study.storage.utils import assert_permission, is_managed, is_output_archived, remove_from_cache
+from antarest.worker.archive_worker import ArchiveTaskArgs
+
+logger = logging.getLogger(__name__)
+
+
+class OutputService:
+    def __init__(self, study_service: StudyService):
+        self.study_service = study_service
```
sylvlecl: let's take the habit of making dependencies protected `_study_service`

antarest/study/storage/output_service.py
```diff
@@ -0,0 +1,531 @@
...
+
+    def get_digest_file(self, study_id: str, output_id: str, params: RequestParameters) -> DigestUI:
+        study = self.study_service.get_study(study_id)
+        assert_permission(params.user, study, StudyPermissionType.READ)
+        file_study = self.study_service.storage_service.get_storage(study).get_raw(study)
```
sylvlecl: We should add a top level method to the `study_service` to get the `FileStudy` more simply.

antarest/study/storage/output_service.py
```diff
@@ -0,0 +1,531 @@
...
+        """
+        study = self.study_service.get_study(study_id)
+        assert_permission(params.user, study, StudyPermissionType.READ)
+        logger.info(
+            "study %s output listing asked by user %s",
+            study_id,
+            params.get_user_id(),
+        )
+
+        return self.study_service.storage_service.get_storage(study).get_study_sim_result(study)
```
sylvlecl: Ideally, we should not need to go to the study service to get some information about the outputs, only this service should be involved.

As a first step, we could add the 2 storage services as explicit dependencies of this one.

In the end, I think we should also separate output-related features from those storage services, but that would be too much work for now I think.

antarest/study/storage/output_service.py
```diff
@@ -0,0 +1,531 @@
...
+            output_uuid: output id
+            params: request parameters
+        """
+        study = self.study_service.get_study(study_uuid)
+        assert_permission(params.user, study, StudyPermissionType.READ)
+        self.study_service.assert_study_unarchived(study)
+
+        logger.info(f"Exporting {output_uuid} from study {study_uuid}")
+        export_name = f"Study output {study.name}/{output_uuid} export"
+        export_file_download = self.study_service.file_transfer_manager.request_download(
```
sylvlecl: The file transfer manager must be an explicit dependency of this one, we should not know the study_service uses it

antarest/study/storage/output_service.py
```diff
@@ -0,0 +1,531 @@
...
+            params: request parameters
+
+        Returns:
+
+        """
+        study = self.study_service.get_study(uuid)
+        assert_permission(params.user, study, StudyPermissionType.WRITE)
+        self.study_service.assert_study_unarchived(study)
+        self.study_service.storage_service.get_storage(study).delete_output(study, output_name)
+        self.study_service.event_bus.push(
```
sylvlecl: Same here, the event bus must be an explicit dependency of this one, we must not use the one of the study service

sylvlecl (changes requested):
Looks good overall: however we need to make the dependencies of the output service explicit, instead of relying on implementation details of the study service

antarest/launcher/main.py
MartinBelthle: Indeed. I kept the distinction inside the launcher service but i can also explicitely put it here

antarest/study/web/output_blueprint.py
MartinBelthle: I see what you mean. Indeed it should be reorganized. I'll introduce 4 new endpoints and hide the other ones

antarest/study/storage/output_service.py
MartinBelthle: I introduced it and named it get_file_study

antarest/study/storage/output_service.py
MartinBelthle: I agree with the observation but I don't understand what you recommended as the first step :/

antarest/service_creator.py
MartinBelthle: I also introduced a class `Services`

antarest/study/common/outputstorage.py
```diff
@@ -0,0 +1,83 @@
...
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import BinaryIO, Generic, List, Optional, TypeVar
+
+from antarest.study.model import Study, StudySimResultDTO
+
+T = TypeVar("T", bound=Study)
+
+
+class IOutputStorageService(ABC, Generic[T]):
```
sylvlecl: I propose to simplify the name to `IOutputStorage` and to remove the generic, I think it causes more issues than it adds value (we can downcast in the implementation if needed).

antarest/study/storage/storage_service.py
```diff
@@ -62,3 +63,43 @@ def get_storage(self, study: Study) -> IStudyStorageService[Union[RawStudy, Vari
             The study storage service associated with the study type.
         """
         return self.raw_study_service if isinstance(study, RawStudy) else self.variant_study_service
+
+
+class OutputStorageService:
```
sylvlecl: The name is misleading (looks like the interface we have defined above), we can name it something like OutputStorageProvider.

Actually it would probably make things simpler to have an interface (or typealias)  OutputStorageProvider, and the implementation will dispatch to raw or variant service. The implementation could even be as simple as as lambda or function.

This would allow to test the output service with a dummy implementation which does not require actual raw and variant implementations.

antarest/study/storage/storage_service.py
```diff
@@ -62,3 +63,43 @@ def get_storage(self, study: Study) -> IStudyStorageService[Union[RawStudy, Vari
...
+        """
+        Get the appropriate study storage service based on the type of study.
+
+        Args:
+            study: The study object for which the storage service is required.
+
+        Returns:
+            The study storage service associated with the study type.
+        """
+        return self.raw_study_service if isinstance(study, RawStudy) else self.variant_study_service
```
sylvlecl: we'd better check both types and raise if it's another one (in case of future addition).
Note:
you may use the new `match` here  :)
```python
match study:
    case RawStudy():
        return self.raw_study_service
    case VariantStudy():
        return self.variant_study_service
    case _:
        raise ...
```

antarest/study/common/outputstorage.py
```diff
@@ -0,0 +1,83 @@
+# Copyright (c) 2025, RTE (https://www.rte-france.com)
```
sylvlecl: Maybe while we are on that kind of change, we can try to make the modules more consistent:
- rename studystorage.py and outputstorage.py to study_storage.py and output_storage.py
- move them to `storage` where they belong
remove `common`

antarest/study/storage/storage_service.py
```diff
@@ -62,3 +63,43 @@ def get_storage(self, study: Study) -> IStudyStorageService[Union[RawStudy, Vari
...
+        Initialize the ``OutputStorageService`` with raw and variant study services.
+
+        Args:
+            raw_study_service: The service for managing raw studies.
+            variant_study_service: The service for managing variant studies.
+        """
+        self.raw_study_service = raw_study_service
+        self.variant_study_service = variant_study_service
+
+    def get_storage(self, study: Study) -> IOutputStorageService[Union[RawStudy, VariantStudy]]:
```
sylvlecl: Will be simpler without the generic here

antarest/study/storage/output_service.py
```diff
@@ -0,0 +1,540 @@
...
+            ids_to_consider: list of areas or links ids to consider, if empty, all areas are selected
+            params: request parameters
+            mc_years: list of monte-carlo years, if empty, all years are selected (only for mc-ind)
+
+        Returns: the aggregated data as a DataFrame
+
+        """
+        study = self._study_service.get_study(uuid)
+        assert_permission(params.user, study, StudyPermissionType.READ)
+        study_path = self._storage_service.raw_study_service.get_study_path(study)
```
sylvlecl: Actually the design seems broken here and in the aggregate manager:
- the aggregate manager takes a `study_path` and then relies on the fact that outputs are stored in `study_path / output` in all cases, but it should not need to assume this. It would be better to provide him the path to the specific output we are working on, instead
- then IOutputStorage could have a method get_output_path() (which would indeed return study_path / output / <output_id> for both implementations but we must not rely on it)

This will avoid to reference explicitly the raw and variant storages in the whole class.
Then we could only have a dependency to an `OutputStorageProvider` in the constructor, which would make this service dependencies cleaner.

sylvlecl (commented):
Nice, a few more suggestions about naming and component dependencies

antarest/study/storage/output_service.py
sylvlecl: An afterthought:
the best, simpler design would probably be the following one actually:
remove the `OutputServiceProvider` interface, only use a `IOutputStorage`.

An `IOutputStorage` implementation, for example `OutputStorageDispatcher` will redirect to the "right" implementation depending on the study type.
This will simplify calls from:
```python
self._storage_provider.get_storage(study).export_output(study, ...)
```
to just
```python
self._storage.export_output(study, ...)
```

antarest/main.py
```diff
@@ -282,17 +279,14 @@ def handle_all_exception(request: Request, exc: Exception) -> Any:
 
     application.include_router(api_root)
 
-    if config.server.services and Module.WATCHER.value in config.server.services:
-        watcher = cast(Watcher, services["watcher"])
-        watcher.start()
+    if services.watcher:
```
sylvlecl, line 282: Nice it's far more readable

sylvlecl (approved):

PR #2449, opened 2025-04-07
refactor(commands): remove apply config

TheoPascoli:


TheoPascoli:
> Copilot reviewed 50 out of 50 changed files in this pull request and generated no comments.

Thank you Copilot great review

antarest/study/storage/variantstudy/model/command/create_cluster.py
```diff
@@ -103,8 +103,7 @@ def _validate_against_version(self) -> Self:
         validate_thermal_cluster_against_version(self.study_version, self.parameters)
         return self
 
-    @override
-    def _apply_config(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
+    def validate_data(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
```
sylvlecl: naming: it's maybe a little misleading to name this way, because actually it does not only validate the data (which looks like a read only operation), it also updates the "config". Maybe we can user the `update_in_config` wording you have used elsewhere to keep consistent ? 

Also, we can now do simpler than return a dict, since we are not constrained anymore by the parent class signature: we could for example just return the cluster ID.
The successful `CommandOutput` can be created at the end of the `_apply` method instead of here, it seems a little weird now to create it here.

antarest/study/storage/variantstudy/model/command/create_district.py
```diff
@@ -56,8 +56,7 @@ def validate_district_name(cls, val: str) -> str:
             raise ValueError("Area name must only contains [a-zA-Z0-9],&,-,_,(,) characters")
         return val
 
-    @override
-    def _apply_config(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
+    def validate_data(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
```
sylvlecl: Same here, we could also simplify the return type

antarest/study/storage/variantstudy/model/command/create_xpansion_matrix.py
```diff
@@ -90,10 +80,6 @@ class CreateXpansionCapacity(AbstractCreateXpansionMatrix):
 
     command_name: CommandName = CommandName.CREATE_XPANSION_CAPACITY
 
-    @override
-    def _apply_config(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
-        return apply_config_create_resource_commands(self.filename, XpansionResourceFileType.CAPACITIES)
```
sylvlecl, line 95: The function `apply_config_create_resource_commands` can now be removed and `apply_create_resource_commands` simplified.

antarest/study/storage/variantstudy/model/command/remove_area.py
```diff
@@ -65,19 +65,15 @@ def _remove_area_from_sets_in_config(self, study_data_config: FileStudyTreeConfi
...
         self._remove_area_from_links_in_config(study_data_config)
         self._remove_area_from_sets_in_config(study_data_config)
 
         remove_area_cluster_from_binding_constraints(study_data_config, self.id)
 
-        return (
-            CommandOutput(status=True, message=f"Area '{self.id}' deleted"),
-            {},
-        )
+        return CommandOutput(status=True, message=f"Area '{self.id}' deleted")
```
sylvlecl: In this kind of case where there is no error case, I think it would be more straight forward to not return anything here, and create the `CommandOuput` at the end of the `_apply` method

antarest/study/storage/variantstudy/model/command/remove_link.py
```diff
@@ -152,7 +151,7 @@ def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = N
 
             self._remove_link_from_scenario_builder(study_data)
 
-        return self._apply_config(study_data.config)[0]
+        return self.remove_from_config(study_data.config)[0]
```
sylvlecl: It looks like the second part of the tuple is never used, we can remove it.
Same in `_check_link_exists`, it seems

antarest/study/storage/variantstudy/model/command/remove_renewables_cluster.py
```diff
@@ -132,7 +131,7 @@ def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = N
 
         # Deleting the renewable cluster in the configuration must be done AFTER
         # deleting the files and folders.
-        return self._apply_config(study_data.config)[0]
+        return self.remove_from_config(study_data.config)[0]
```
sylvlecl: We can remove the second part of the tuple here too

antarest/study/storage/variantstudy/model/command/remove_st_storage.py
```diff
@@ -43,8 +43,7 @@ class RemoveSTStorage(ICommand):
     area_id: str = Field(description="Area ID", pattern=r"[a-z0-9_(),& -]+")
     storage_id: str = Field(description="Short term storage ID", pattern=r"[a-z0-9_(),& -]+")
 
-    @override
-    def _apply_config(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
+    def remove_from_config(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
```
sylvlecl: We can remove the tuple here too

antarest/study/storage/variantstudy/model/command/update_config.py
```diff
@@ -54,16 +54,15 @@ class UpdateConfig(ICommand):
     target: str
     data: _Data
 
-    @override
-    def _apply_config(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
+    def update_in_config(self, study_data: FileStudyTreeConfig) -> CommandOutput:
```
sylvlecl: we can remove the return here, and create the command output in `_apply`

antarest/study/storage/variantstudy/model/command/update_district.py
```diff
@@ -42,9 +42,7 @@ class UpdateDistrict(ICommand):
     output: Optional[bool] = None
     comments: Optional[str] = None
 
-    @override
-    def _apply_config(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
-        base_set = study_data.sets[self.id]
+    def update_in_config(self, study_data: FileStudyTreeConfig) -> Tuple[CommandOutput, Dict[str, Any]]:
```
sylvlecl: we could simplify the return type, replacing the dict, here too

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -784,50 +783,6 @@ def generate(
 
         return self.generate_task(variant_study, denormalize, from_scratch=from_scratch)
 
-    def generate_study_config(
```
sylvlecl, line 787: :heart:

sylvlecl (changes requested):
Nice simplifications :) 
A few commands can be simplified further to make it complete.

The fact that very few tests needed to be updated confirms that we did not tested really correctly the `_apply_config` (although tested through the `_apply`, but not independently).

sylvlecl (approved):
Nice, thanks for the simplification !

PR #2450, opened 2025-04-07
feat(matrix): support matrices with headers inside the matrixstore

MartinBelthle:


antarest/matrixstore/web.py
```diff
@@ -70,8 +71,8 @@ def create_by_importation(
             return service.create_by_importation(file, is_json=json)
         raise UserHasNotPermissionError()
 
-    @bp.get("/matrix/{id}", tags=[APITag.matrix], response_model=MatrixDTO)
-    def get(id: str, user: JWTUser = Depends(auth.get_current_user)) -> Any:
+    @bp.get("/matrix/{id}", tags=[APITag.matrix])
+    def get(id: str, user: JWTUser = Depends(auth.get_current_user)) -> MatrixDTO | None:
```
sylvlecl: I know it's the current behaviour, but we should return a 404 if the matrix does not exist, not return a `null` text

antarest/core/config.py
```diff
@@ -40,22 +38,38 @@ class InternalMatrixFormat(StrEnum):
     PARQUET = "parquet"
     FEATHER = "feather"
 
-    def load_matrix(self, path: Path) -> npt.NDArray[np.float64]:
-        if self == InternalMatrixFormat.TSV or path.stat().st_size == 0:
-            return np.loadtxt(path, delimiter="\t", dtype=np.float64, ndmin=2)
+    def load_matrix(self, path: Path) -> pd.DataFrame:
```
sylvlecl: we must move all that parsing logic outside of this `config` module, and unit-test it.
We can either create a new `antarest.core.matrix` or just move it to `antarest.matrixstore`

antarest/matrixstore/service.py
```diff
@@ -72,7 +70,7 @@ def __init__(self, matrix_content_repository: MatrixContentRepository) -> None:
         self.matrix_content_repository = matrix_content_repository
 
     @abstractmethod
-    def create(self, data: List[List[MatrixData]] | npt.NDArray[np.float64]) -> str:
+    def create(self, data: pd.DataFrame) -> str:
```
sylvlecl, line 81: Great to remove the support for python lists

tests/matrixstore/test_service.py
```diff
@@ -40,23 +42,25 @@
...
 
 
 class TestMatrixService:
     def test_create__nominal_case(self, matrix_service: MatrixService) -> None:
         """Creates a new matrix object with the specified data."""
         # when a matrix is created (inserted) in the service
-        data: MatrixType = [[1, 2, 3], [4, 5, 6]]
-        matrix_id = matrix_service.create(data)
+        data = TEST_MATRIX
+        df_to_save = pd.DataFrame(data)
```
sylvlecl, line 67: I guess we should add some tests with actual user-defined headers, not only the default ones created by pandas.

antarest/matrixstore/repository.py
```diff
@@ -170,12 +169,7 @@ def get(self, matrix_hash: str) -> MatrixContent:
                 break
         if not storage_format:
             raise FileNotFoundError(str(matrix_path.with_suffix("")))
-        matrix = storage_format.load_matrix(matrix_path)
-        matrix = matrix.reshape((1, 0)) if matrix.size == 0 else matrix
-        data = matrix.tolist()
-        index: List[int | str] = list(range(matrix.shape[0]))
```
sylvlecl, line 176: So actually the data model was prepared for column names, but we actually always returned a list of integers ?

tests/matrixstore/test_repository.py
```diff
@@ -49,17 +50,15 @@ def test_bucket_lifecycle(self, tmp_path: Path) -> None:
...
-        assert aid == repo.save(a)
-
-        bid = repo.save(b)
+        aid = repo.save(pd.DataFrame(a))
+        bid = repo.save(pd.DataFrame(b))
         assert aid != bid
 
-        assert matrix_content_a == repo.get(aid)
-        assert matrix_content_b == repo.get(bid)
+        assert matrix_content_a.eq(repo.get(aid)).all().all()
```
sylvlecl: pandas provides methods for dataframe comparison:
- either https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html
- or https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.equals.html

antarest/core/config.py
```diff
@@ -40,22 +38,38 @@ class InternalMatrixFormat(StrEnum):
...
+        if path.stat().st_size == 0:
+            return pd.DataFrame()
+        if self == InternalMatrixFormat.TSV:
+            # The legacy format is TSV, so we have to handle both cases
+            # To know if we're opening a legacy matrix or not we have to seek the first bytes of the file
+            # Legacy
+            header = None
+            index_col = None
+            with open(path, "r") as f:
+                if f.read(1) == "\t":
```
sylvlecl: So this relies on the fact that the top left cell is empty, right ?
I wonder if it's solid enough, in particular for further evolutions that might not be aware of that subtlety. Otherwise we could version matrix format in the database together with its identifier, to decide if it's an old one or a new one.

To be decided

antarest/matrixstore/repository.py
```diff
@@ -222,8 +215,7 @@ def save(self, content: List[List[MatrixData]] | npt.NDArray[np.float64]) -> str
         #    of the floating point numbers which can introduce rounding errors.
         # However, this method is still a good approach to calculate a hash value
         # for a non-mutable NumPy Array.
-        matrix = content if isinstance(content, np.ndarray) else np.array(content, dtype=np.float64)
-        matrix_hash = hashlib.sha256(matrix.data).hexdigest()
+        matrix_hash = hashlib.sha256(np.ascontiguousarray(content.to_numpy()).data).hexdigest()
```
sylvlecl: There is a quite important issue here:
2 matrices with the same numeric content but different headers and indices will have the same hash, and therefore we won't be able to save both of them.

We need to think about a solution!
The simplest approach is probably to includes the headers and indices in the hash.
That means we won't be able to "de-duplicate" matrices with same content but different headers/index. I don't think it's a real issue.
If we go for this, another consequence is that new matrices will never match existing ones. it sounds OK at first glance.

This makes me think:
the generation of the hash, which is the basic component for "de-duplication", is too hidden here. We should make it a separate method and unit test it against explicit hashes to ensure we don't create regressions on the matrix hashes.

sylvlecl (changes requested):
A few comments: an important one is about how we generated hashes. If we don't take into account the headers and indices, we won't be able to save 2 matrices with same content but different headers.

tests/matrixstore/test_repository.py
MartinBelthle: It didn't work before because i forgot a numpy post treatment but now it's okay :)

antarest/matrixstore/service.py
```diff
@@ -72,11 +70,11 @@ def __init__(self, matrix_content_repository: MatrixContentRepository) -> None:
...
```
sylvlecl, line 81: As discussed a little:
I think we should not re-use the notion of "index" that pandas has for analytics purpose, but which has no benefit for us. It will make things easier to only take into account named columns (we will not need to wonder how to handle the index).

So if we stick with pandas for now:
- we should document that index  is ignored by the matrix store in a docstring
- in the impl, we should not write the index when serializing
- in the impl, we should not take into account the index when computing the hash

antarest/matrixstore/service.py
```diff
@@ -114,20 +112,12 @@ def __init__(self, matrix_content_repository: MatrixContentRepository):
...
-        return MatrixDTO.model_construct(
-            id=matrix_id,
-            width=len(data.columns),
-            height=len(data.index),
-            index=data.index,
-            columns=data.columns,
-            data=data.data,
-        )
+    def get(self, matrix_id: str) -> pd.DataFrame:
+        return self.matrix_content_repository.get(matrix_id, matrix_version=2)
```
sylvlecl: We should have a constant for the current version, where we document the changes in a comment

antarest/matrixstore/repository.py
```diff
@@ -134,6 +135,26 @@ def delete(self, matrix_hash: str) -> None:
         logger.debug(f"Matrix {matrix_hash} deleted")
 
 
+@dataclass
+class MatrixMetaData:
+    hash: str
+    new: bool
+
+
+def calculates_hash(df: pd.DataFrame) -> str:
```
sylvlecl: detail: maybe just `compute_hash` or at least remove the "s" in calculate

antarest/matrixstore/repository.py
```diff
@@ -134,6 +135,26 @@ def delete(self, matrix_hash: str) -> None:
         logger.debug(f"Matrix {matrix_hash} deleted")
 
 
+@dataclass
+class MatrixMetaData:
```
sylvlecl: The name is too vague, matrix metadata could contain info about the matrix itself such as a description, labels, columns names or whatever:
here it's just the result of a creation request, maybe just name it `MatrixCreationResult` ?

Also we can make the class frozen (better to do it anytime we can).

antarest/matrixstore/uri_resolver_service.py
```diff
@@ -50,25 +48,11 @@ def extract_id(uri: str) -> Optional[str]:
...
-                "data": data.data,
-                "index": data.index,
-                "columns": data.columns,
-            }
-        df = pd.DataFrame(
-            data=data.data,
-            index=data.index,
-            columns=data.columns,
-        )
+            return {"data": df.to_numpy().tolist(), "index": list(df.index), "columns": list(df.columns)}
```
sylvlecl, line 54: I think this is the same as `df.to_dict(orient="split")` which could be faster

antarest/matrixstore/service.py
```diff
@@ -194,18 +171,19 @@ def create(self, data: List[List[MatrixData]] | npt.NDArray[np.float64]) -> str:
...
-            # Do not use the `timezone.utc` timezone to preserve a naive datetime.
-            created_at = datetime.utcnow()
-            matrix = Matrix(
-                id=matrix_id,
-                width=shape[1],
-                height=shape[0],
-                created_at=created_at,
-            )
-            self.repo.save(matrix)
+        matrix_metadata = self.matrix_content_repository.save(data)
```
sylvlecl, line 218: I think we should check that the dataframe is compliant with our conventions / limitations:
- have a kind of whitelist of the dtypes that are accepted:
  numeric types, strings, and probably datetime. We may add more types when needed in the future.
- check that index is a range index (it will not be saved and will be loaded as a range again when read)

antarest/matrixstore/repository.py
```diff
@@ -134,6 +135,26 @@ def delete(self, matrix_hash: str) -> None:
...
+    # Checks dataframe dtype to infer if the matrix could correspond to a legacy format
+    legacy_format = False
+    if all(np.issubdtype(dtype, np.number) for dtype in df.dtypes):
+        shape = df.shape
+        # We also need to check the headers to see if they correspond to the default ones
+        if df.index.equals(pd.RangeIndex(0, shape[0])) and df.columns.equals(pd.RangeIndex(0, shape[1])):
+            legacy_format = True
+
+    if not legacy_format:
+        df = pd.concat([util.hash_pandas_object(df), util.hash_pandas_object(df.columns)])
```
sylvlecl: I think we should use the parameter index=False for consistency, if we choose that we just ignore the index.

antarest/matrixstore/repository.py
```diff
@@ -134,6 +135,26 @@ def delete(self, matrix_hash: str) -> None:
...
+    return hashlib.sha256(np.ascontiguousarray(df.to_numpy(dtype=np.float64)).data).hexdigest()
```
sylvlecl, line 187: - it's kind of weird to convert to float 64 when we have produced integers with the hash method of pandas. Actually, we are not even consistent, because for new arrays, floats and integers with the same values will produce different hashes, for example:
```python
    integers = pd.DataFrame(data=[[1,2], [3, 4]])
    print(pd.util.hash_pandas_object(integers))
#0    9519509150087416772
#1     997946931316031040

    floats = pd.DataFrame(data=[[1.0,2.0], [3.0, 4.0]])
    print(pd.util.hash_pandas_object(floats))
#0    11803621802037679806
#1     7682785268321734489
```
see comment about adding the column types in the hash. For me it's OK to consider different types different. If we don't want this we should probably convert everything to the target type before saving it.
- we could add a comment explaining the contiguity part (hashlib requires C contiguous arrays, while this is not the general behaviour of pandas to store its data this way). Note that it's a shame because it will require a copy of the memory to reorganize it.

antarest/matrixstore/repository.py
sylvlecl: More generally, we need to be careful of what we include in the computation of the hash:
- columns names since they are part of the data, OK
- the array content, OK
- I think we should also include the type of each column (by its name for example):
  otherwise it's easy to create a hash conflict for example by creating a column of int where we but the hashes of strings instead of strings themselves. Different numeric data (floats and ints) will also have the same hashed value

sylvlecl: Also, I think it would be more clear and straightforward to feed the hash object with the different parts of the hash, instead of going through a concatenated dataframe, for example:
```
column_names_hashes =  util.hash_pandas_object(df.columns, index=False)
row_hashes = util.hash_pandas_object(df, index=False)
df_hash =  hashlib.sha256(column_names_hashes)
df_hash.update(row_hashes)
```

antarest/matrixstore/parsing.py
```diff
@@ -0,0 +1,57 @@
...
+        return pd.read_parquet(path)
+    elif matrix_format == InternalMatrixFormat.FEATHER:
+        return pd.read_feather(path)
+    else:
+        raise NotImplementedError(f"Internal matrix format '{matrix_format}' is not implemented")
+
+
+def save_matrix(matrix_format: InternalMatrixFormat, dataframe: pd.DataFrame, path: Path) -> None:
+    if matrix_format == InternalMatrixFormat.TSV:
+        dataframe.to_csv(path, sep="\t", float_format="%.6f")
```
sylvlecl: See other comments, let's not write the index here.
We could probably reduce even more the precision, but let's keep it that way for now

antarest/study/storage/variantstudy/business/matrix_constants/hydro/v7.py
```diff
@@ -10,11 +10,10 @@
...
-from typing import List
 
-from antarest.matrixstore.model import MatrixData
+import pandas as pd
 
-credit_modulations: List[List[MatrixData]] = [[1.0] * 101] * 2
-inflow_pattern: List[List[MatrixData]] = [[1.0]] * 365
-max_power: List[List[MatrixData]] = [[0.0, 24.0, 0.0, 24.0]] * 365
-reservoir: List[List[MatrixData]] = [[0.0, 0.5, 1.0]] * 365
+credit_modulations = pd.DataFrame(data=[[1] * 101] * 2)
```
sylvlecl: probably better to type it as floats, not integers (either by restoring the `.`, or using an explicit `dtype`).
Same for all matrices that represent floating point data (most of them ?)

sylvlecl (changes requested):
A few more comments:
- I think we can still clarify/make more robust/document a bit the hashing function
- I think we need to be explicit about the fact that we don't want to consider the index of dataframes, this will only cause trouble down the road, I think (instead let's use plain columns)

antarest/matrixstore/repository.py
MartinBelthle: The `dtype=float64` was to ensure floats and integers df are treated the same way with legacy format.
With the new format we don't really care about that it was just to not duplicate existing matrices

antarest/matrixstore/uri_resolver_service.py
MartinBelthle: I think I saw a difference, i don't remember which one. I'll handle this in the next PR I'll be working on to simplify and speed-up the uri resolver service

sylvlecl (approved):

PR #2451, opened 2025-04-08
feat(aggregator_manager): enable aggregation size on endpoint

maugde:
Resolve [ANT-2718](https://gopro-tickets.rte-france.com/browse/ANT-2718)

The purpose of this PR is to use a variable to manage the maximum size of the results of aggregation operations instead of a static value. The default value of this variable should be equal to 200 and can be modified with the `application.yaml` file config.

The PR is also linked to [this MR](https://devin-source.rte-france.com/antares/antares-cd/-/merge_requests/80) in order to set the default values of each environment.

### DIFF
--- a/antarest/study/business/aggregator_management.py
+++ b/antarest/study/business/aggregator_management.py
@@ -90,15 +90,6 @@ class MCAllLinksQueryFile(StrEnum):
     ID = "id"
 
 
-def _checks_estimated_size(nb_files: int, df_bytes_size: int, nb_files_checked: int, maximum_size: int) -> None:
-    """
-    A `maximum_size` equivalent to 100 Mo corresponds to a 15 seconds task.
-    """
-    estimated_df_size = nb_files * df_bytes_size // (nb_files_checked * 10**6)
-    if estimated_df_size > maximum_size:
-        raise FileTooLargeError(estimated_df_size, maximum_size)
-
-
 def _columns_ordering(df_cols: List[str], column_name: str, is_details: bool, mc_root: MCRoot) -> Sequence[str]:
     # original columns
     org_cols = df_cols.copy()
@@ -366,7 +357,7 @@ class AggregatorManager:
             MCAllAreasQueryFile.DETAILS_RES,
         ]
         final_df = pd.DataFrame()
-        nb_files = len(files)
+
         for k, file_path in enumerate(files):
             df = self._process_df(file_path, is_details)
 
@@ -380,7 +371,8 @@ class AggregatorManager:
 
             # The following formula is the more accurate one compared to the final csv file.
             estimated_binary_size = final_df.memory_usage().sum()
-            _checks_estimated_size(nb_files, estimated_binary_size, k, self.aggregation_results_max_size)
+            if estimated_binary_size > self.aggregation_results_max_size * 10**6:
+                raise FileTooLargeError(estimated_binary_size, estimated_binary_size)
 
             column_name = AREA_COL if self.output_type == "areas" else LINK_COL
             new_column_order = _columns_ordering(list_of_df_columns, column_name, is_details, self.mc_root)
--- a/antarest/study/business/aggregator_management.py
+++ b/antarest/study/business/aggregator_management.py
@@ -379,12 +379,9 @@ class AggregatorManager:
             if not list_of_df_columns or set(list_of_df_columns) == {CLUSTER_ID_COL, TIME_ID_COL}:
                 return pd.DataFrame()
 
-            # checks if the estimated dataframe size does not exceed the limit
-            # This check is performed on 10 aggregated files to have a more accurate view of the final df.
-            if k == 10:
-                # The following formula is the more accurate one compared to the final csv file.
-                estimated_binary_size = final_df.memory_usage().sum()
-                _checks_estimated_size(nb_files, estimated_binary_size, k, self.aggregation_results_max_size)
+            # The following formula is the more accurate one compared to the final csv file.
+            estimated_binary_size = final_df.memory_usage().sum()
+            _checks_estimated_size(nb_files, estimated_binary_size, k, self.aggregation_results_max_size)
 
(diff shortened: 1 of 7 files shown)


MartinBelthle (commented):
Seems okay to me. We could add a little unit test to ensure that we're able to read a different data than the default one.
I'll let Sylvain take a look

tests/storage/test_config.py
```diff
@@ -112,3 +112,12 @@ def test_storage_config_from_dict_validiation_errors(storage_config_default):
 
     with pytest.raises(ValueError):
         StorageConfig.from_dict(data)
+
+
+@pytest.mark.unit_test
+def test_storage_default_aggregation_results_max_size_value(storage_config_default):
```
sylvlecl: Would be nice to also test that the parameter is correctly taken into account.
Bad luck, it appears there is no unit test on the aggregator manager ... (only integration tests through the API).
I think we should create a unit test for this feature (with small limits obviously :) ).

sylvlecl (changes requested):
The implementation looks good, I think we shuld add a unit test though to check the behaviour. Unfortunately we need to create one from scratch, it seems there is no unit test for that component today ...

tests/storage/test_config.py
```diff
@@ -112,3 +113,12 @@ def test_storage_config_from_dict_validiation_errors(storage_config_default):
 
     with pytest.raises(ValueError):
         StorageConfig.from_dict(data)
+
+
+@pytest.mark.unit_test
```
MartinBelthle, line 118: As we discussed, I think we should add a UT to ensure we're able to read the field with a different value than 200

tests/study/business/areas/assets/aggregator_manager/viz_2.zip
MartinBelthle, line 1: We shouldn't have to introduce this binary. What you should do instead is using the STA-mini study as it already has outputs and just set the limit lower in your test

MartinBelthle (changes requested):

tests/study/business/areas/assets/aggregator_manager/viz_2.zip
maugde: I first tried to use the STA-mini study but it does not have enough aggregated files to test the maximum size, because of this check:
```
# checks if the estimated dataframe size does not exceed the limit
# This check is performed on 10 aggregated files to have a more accurate view of the final df.
if k == 10:
    # The following formula is the more accurate one compared to the final csv file.
    estimated_binary_size = final_df.memory_usage().sum()
    _checks_estimated_size(nb_files, estimated_binary_size, k, self.aggregation_results_max_size)
```
I tried to modify a duplicated version of the study in order to match the requirements but I didn't find a simple way to produce 200 Mo data for this test.

MartinBelthle: Okay I see.
I discussed this issue with Sylvain and we think that we should remove this `if` check and just do the calculation each time. This way you can test it and also we won't have to infer the size anymore

MartinBelthle:
Also nothing to do with the PR but you can remove this `MC_TEMPLATE_PARTS = "economy/{mc_root}"` inside the aggregator_management.py (line 28) as we don't use it anymore

And we should add `/aggregate` inside the newly introduced endpoint `"/studies/{uuid}/outputs/{output_id}/mc-ind/areas"` to fit with the 3 other ones. It's a typo I did in a recent PR

antarest/study/business/aggregator_management.py
```diff
@@ -91,8 +90,10 @@ class MCAllLinksQueryFile(StrEnum):
     ID = "id"
 
 
-def _checks_estimated_size(nb_files: int, df_bytes_size: int, nb_files_checked: int) -> None:
-    maximum_size = 100  # size in Mo that corresponds to a 15 seconds task.
+def _checks_estimated_size(nb_files: int, df_bytes_size: int, nb_files_checked: int, maximum_size: int) -> None:
```
MartinBelthle: We don't need this method anymore. You can just compare the estimated_binary_size with `self.aggregation_results_max_size`

tests/study/business/test_aggregator_manager.py
```diff
@@ -0,0 +1,67 @@
...
+
+import pytest
+
+from antarest.core.exceptions import FileTooLargeError
+from antarest.study.business.aggregator_management import AggregatorManager, MCIndAreasQueryFile
+from antarest.study.storage.rawstudy.model.filesystem.matrix.matrix import MatrixFrequency
+
+
+@pytest.fixture(name="get_output_path")
+def get_output_path_fixture(tmp_path: Path, project_path: Path) -> Path:
```
MartinBelthle: Really minor but do we need to introduce a fixture for something we're only going to use once ?

MartinBelthle (changes requested):
Minor changes

MartinBelthle (dismissed):

antarest/study/business/aggregator_management.py
```diff
@@ -375,12 +369,10 @@ def _build_dataframe(self, files: Sequence[Path]) -> pd.DataFrame:
...
-            # checks if the estimated dataframe size does not exceed the limit
-            # This check is performed on 10 aggregated files to have a more accurate view of the final df.
-            if k == 10:
-                # The following formula is the more accurate one compared to the final csv file.
-                estimated_binary_size = final_df.memory_usage().sum()
-                _checks_estimated_size(nb_files, estimated_binary_size, k)
+            # The following formula is the more accurate one compared to the final csv file.
+            estimated_binary_size = final_df.memory_usage().sum()
+            if estimated_binary_size > self.aggregation_results_max_size * 10**6:
+                raise FileTooLargeError(estimated_binary_size, estimated_binary_size)
```
sylvlecl: I think the second argument should be the max size

sylvlecl (approved):

PR #2455, opened 2025-04-11
fix(ui-clusters): handle area/cluster navigation by properly updating URL segments

skamril:


webapp/src/components/App/Singlestudy/explore/Modelization/Areas/AreasTab.tsx
```diff
@@ -28,31 +28,6 @@ function AreasTab({ renewablesClustering }: Props) {
...
-    if (areaIndex >= 0 && areaIndex + 1 < pathSegments.length) {
-      // replace only the segment after 'area' with the new areaId
-      pathSegments[areaIndex + 1] = areaId.toString();
-
-      const newPath = pathSegments.join("/");
-      if (newPath !== currentPath) {
-        navigate(newPath, { replace: true });
-      }
-    }
-  }, [areaId, navigate, location.pathname]);
```
hdinia, line 55: you remove automatic selection of the first area when the component mounts

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/index.tsx
```diff
@@ -34,13 +34,40 @@ function Areas() {
...
+      pathSegments[areaIdIndex] = newAreaId;
+
+      // If the we are in a cluster page, we don't keep the cluster ID
+      if (["thermal", "storages", "renewables"].includes(pathSegments[nextPageIndex])) {
+        pathSegments = pathSegments.slice(0, nextPageIndex + 1);
+      }
+
+      navigate(pathSegments.join("/"), { replace: true });
+    }
+  };
```
hdinia: This only updates the URL when an area is explicitly clicked via `handleAreaClick`
You can add an effect to handle the case where **no area is initially selected** when the component mounts

hdinia: I think you need to keep the effect in `AreasTab` (but I am not sure about the right fix here)

webapp/src/components/App/Singlestudy/explore/Modelization/index.tsx
```diff
@@ -54,14 +54,16 @@ function Modelization() {
       }
     };
 
+    const areaPath = [basePath, "area", encodeURI(areaId || areas[0]?.id || "")].join("/");
```
hdinia, line 57: Any reason to do this ? If not can you revert back to the simpler version

hdinia (changes requested):

hdinia (approved):

PR #2457, opened 2025-04-15
fix(desktop): use antares-solver_windows instead of rte-antares--installer-64bits

smailio:


scripts/package_antares_web.sh
```diff
@@ -19,7 +19,7 @@ RESOURCES_DIR="${PROJECT_DIR}/resources"
 ANTARES_SOLVER_DIR="${DIST_DIR}/AntaresWeb/antares_solver"
 
 if [[ "$OSTYPE" == "msys"* ]]; then
-  ANTARES_SOLVER_FOLDER_NAME="rte-antares-$ANTARES_SOLVER_FULL_VERSION-installer-64bits"
```
MartinBelthle, line 22: Do we still need this `ANTARES_SOLVER_FOLDER_NAME` variable ?

MartinBelthle (commented):
If you've tested that the Desktop version is correctly built, it's okay for me.

TheoPascoli (dismissed):

scripts/package_antares_web.sh
```diff
@@ -66,9 +66,7 @@ rm $ANTARES_SOLVER_ZIPFILE_NAME
 
 if [[ "$OSTYPE" == "msys"* ]]; then
   echo "INFO: Moving executables in '$ANTARES_SOLVER_DIR'..."
-  mv "$ANTARES_SOLVER_FOLDER_NAME/bin/antares-$ANTARES_SOLVER_VERSION-solver.exe" "$ANTARES_SOLVER_DIR"
-  mv "$ANTARES_SOLVER_FOLDER_NAME/bin/sirius_solver.dll" "$ANTARES_SOLVER_DIR"
-  mv "$ANTARES_SOLVER_FOLDER_NAME/bin/zlib1.dll" "$ANTARES_SOLVER_DIR"
+  mv "$ANTARES_SOLVER_DIR/solver/Release/"* "$ANTARES_SOLVER_DIR"
```
sylvlecl, line 69: Note that this leaves an empty dir in the final package `AntaresWeb/antares_solver/solver/Release`, we'd better remove it.

scripts/package_antares_web.sh
sylvlecl: Indeed it seems that it's only used for removal later but does not seem to remove anything ?

sylvlecl (changes requested):
Minor comments.

Not related to that work, but with all those if windows do this else do that, I wonder if it would not be more maintainable with 2 separate scripts, or at least 2 separate blocks of code in the script.

sylvlecl (approved):

PR #2460, opened 2025-04-15
fix(ui-digest): make first column fixed

hdinia:
ANT-2996

Enhanced DataGridViewer component with several usability improvements:
- Fixed first column to ensure users can always see area names when scrolling horizontally
- Added row selection capability with checkbox markers and "select all" functionality in top-left corner
- Allow copy of the grid content
- Maintained existing column selection (click group header to select all sub-variables or Shift+click for multiple sub-headers)
- Implemented number formatting for string-type numeric values to ensure consistent display regardless of data format received from API

### DIFF
--- a/webapp/src/components/common/DataGridViewer.tsx
+++ b/webapp/src/components/common/DataGridViewer.tsx
@@ -21,25 +21,29 @@ import {
 } from "@glideapps/glide-data-grid";
 import { useMemo } from "react";
 import { formatGridNumber } from "./Matrix/shared/utils";
-import DataGrid from "@/components/common/DataGrid";
+import DataGrid, { type RowMarkers } from "@/components/common/DataGrid";
+import { isNumericValue } from "@/utils/numberUtils";
 
 type CellValue = number | string;
 
 interface DataGridViewerProps {
   data: CellValue[][];
   columns: GridColumn[];
+  freezeColumns?: number;
+  rowMarkers?: RowMarkers;
 }
 
-function DataGridViewer({ data, columns }: DataGridViewerProps) {
+function DataGridViewer({ data, columns, freezeColumns, rowMarkers }: DataGridViewerProps) {
   const getCellContent = useMemo(
     () => (cell: Item) => {
       const [col, row] = cell;
       const value = data[row]?.[col];
 
-      if (typeof value === "number") {
+      // Handle numeric values (both as numbers and as strings)
+      if (isNumericValue(value)) {
         return {
           kind: GridCellKind.Number,
-          data: value,
+          data: Number(value),
           displayData: formatGridNumber({ value, maxDecimals: 3 }),
           decimalSeparator: ".",
           thousandSeparator: " ",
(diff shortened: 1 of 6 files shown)


webapp/src/components/common/DataGridViewer.tsx
```diff
@@ -72,8 +82,10 @@ function DataGridViewer({ data, columns }: DataGridViewerProps) {
       rows={data.length}
       columns={columns}
       getCellContent={getCellContent}
-      rowMarkers="none"
+      rowMarkers={rowMarkers}
       getCellsForSelection
+      freezeColumns={fixedFirstColumn ? 1 : 0}
```
skamril: Why not using `freezeColumns` in props instead of the limited `fixedFirstColumn` prop?

webapp/src/components/common/Matrix/shared/utils.ts
```diff
@@ -33,6 +33,31 @@ import { getCurrentLanguage } from "@/utils/i18nUtils";
 import { Aggregate, Column, TIME_FREQUENCY_CONFIG } from "./constants";
 import { groupHeaderTheme } from "../styles";
 
+/**
+ * Checks if a string value is a valid numeric representation that can be parsed.
+ *
+ * @param value - The value to check
+ * @returns true if the string can be converted to a valid number, false otherwise
+ */
+export function isNumericString(value: string | number): boolean {
```
skamril: `isNumeric` more appropriate, because the function allow number value.

`isNumericString(1)` returns `true` but it's not a numeric string.

`: boolean` can be removed.

webapp/src/components/common/DataGridViewer.tsx
```diff
@@ -20,26 +20,34 @@ import {
...
   columns: GridColumn[];
+  freezeColumns?: number;
+  rowMarkers?: RowMarkers;
 }
 
-function DataGridViewer({ data, columns }: DataGridViewerProps) {
+function DataGridViewer({
+  data,
+  columns,
+  freezeColumns = 0,
```
skamril: `freezeColumns` is already 0 by default and rowMarkers is already "none" by default.
Don't set them here allows to keep the same default values on the two components.

webapp/src/components/common/Matrix/shared/utils.ts
```diff
@@ -33,6 +33,31 @@ import { getCurrentLanguage } from "@/utils/i18nUtils";
...
+  if (typeof value === "number") {
```
skamril: ⚠️ `typeof NaN === "number"`

and 

```typescript
if (typeof value !== "string") {
  return false;
}
```

is unuseful after the first check.

skamril (changes requested):

webapp/src/components/common/DataGridViewer.tsx
hdinia: its intended to be set here

webapp/src/components/common/Matrix/shared/utils.ts
hdinia: done

webapp/src/components/common/Matrix/shared/utils.ts
hdinia: the check does serve as a runtime guard

skamril (dismissed):

skamril (approved):

PR #2463, opened 2025-04-17
refactor(matrix): remove `ContextServer` and simplify `UriResolverService`

MartinBelthle:
It also speeds-up the code:
- When saving a matrix, in some cases, we used to load the corresponding matrix inside the matrixstore just to ensure it existed. Now we only check its existence without loading it
- Inside the matrix parsing process, in some cases we did an useless roundtrip
DataFrame  -> JSON -> DataFrame. I removed it


antarest/study/storage/rawstudy/model/filesystem/matrix/matrix.py
```diff
@@ -109,7 +108,7 @@ def normalize(self) -> None:
             return
 
         matrix = self.parse_as_dataframe()
-        uuid = self.context.matrix.create(matrix)
+        uuid = self.context.resolver.matrix_service.create(matrix)
```
sylvlecl: To be honest this does not really improve the design:
the "URI resolver" is responsible for doing a mapping between URIs and some data (matrices), the fact that it contains a matrix service should be an implementation detail, not something we rely on (matrix service should be a hidden member).

If we need a matrix service here, it was indeed more clear/explicit to have it straight in the context.

antarest/study/storage/rawstudy/model/filesystem/context.py
```diff
@@ -12,11 +12,9 @@
 
 from dataclasses import dataclass
 
-from antarest.matrixstore.service import ISimpleMatrixService
 from antarest.matrixstore.uri_resolver_service import UriResolverService
 
 
-@dataclass
+@dataclass(frozen=True)
 class ContextServer:
```
sylvlecl: If we really want to simplify things, I think we should just remove that class.

In this filesystem layer, we only need an interface for matrices which:
1. knows how to load a matrix from a URI
2. knows how to store a matrix and return a URI which points to it

And that's all.
We are not going to build more complex things on this filesystem layer, so that class can be removed, and we can just keep one object similar to the current `UriResolverService` with a hidden matrix service and an additional `store_matrix` method which returns a URI. Could be named something like MatrixUriMapper.

We could also make it an abstract class to mock it more easily / have a dummy implementation, but not a must have.

sylvlecl (changes requested):
See comments, I think we should simplify further by removing the "context" class and have only one object responsible for get/load of matrices, in that filesystem layer

antarest/study/storage/rawstudy/model/filesystem/matrix/matrix.py
MartinBelthle: I agree. I just forgot to introduce a method `create_matrix` inside the UriResolver to hide this

antarest/study/storage/rawstudy/model/filesystem/context.py
MartinBelthle: Okay. I wanted to keep this class just in case we'd want to provide other things than the matrixstore inside the context but I'm not sure we'll need to do this so I'll remove it as you proposed.
And I'll rename the UriResolver class

sylvlecl: Yes I agree, it made sense to keep a kind of container class for further needs. I just hope we won't need to add more in the future, and handle things at a "higher" level than the file level instead (like DAO), so it sounds OK to just remove it :)

antarest/matrixstore/uri_resolver_service.py
```diff
@@ -9,60 +9,29 @@
...
-import re
-from typing import Optional, Tuple
+from antarest.matrixstore.service import MATRIX_PROTOCOL_PREFIX, ISimpleMatrixService
 
-from antarest.core.model import JSON
-from antarest.matrixstore.service import ISimpleMatrixService
 
-
-class UriResolverService:
+class MatrixUriMapper:
```
sylvlecl: We should rename the file accordingly.

antarest/matrixstore/uri_resolver_service.py
```diff
@@ -9,60 +9,29 @@
...
     def __init__(self, matrix_service: ISimpleMatrixService):
         self.matrix_service = matrix_service
```
sylvlecl: We should use this opportunity to make the service private

antarest/study/storage/rawstudy/model/filesystem/factory.py
```diff
@@ -55,11 +53,10 @@ class StudyFactory:
 
     def __init__(
         self,
-        matrix: ISimpleMatrixService,
-        resolver: UriResolverService,
+        resolver: MatrixUriMapper,
```
sylvlecl: we should rename resolver / context in here too

sylvlecl (changes requested):
Minor comments

sylvlecl (approved):

PR #2464, opened 2025-04-17
fix(hydro): allow negative values for hydro allocation coefficients

hdinia:
**ANT-3014**

## Description
This PR modifies both the frontend and backend validation to accept negative values for hydro allocation coefficients.

## Context
The hydro allocation heuristic in Antares distributes daily turbining credits based on weighted residual consumption across areas. For example:
- By default, a weight of 1 for FR means allocation is proportional to FR's residual consumption
- With weights of 2 for FR and 1 for IT, allocation for FR becomes proportional to (2 × FR residual consumption + 1 × IT residual consumption)

The simulator actually accepts any real number as weights, including negative values, but our validation was previously enforcing positive-only values.

## Changes
- Frontend: Removed the min:0 validation constraint in the number field editor
- Backend: Removed checks for negative coefficients in model validators
- Backend: Updated coefficient filtering to retain all values, not just positive ones

## Testing
Manual testing confirms that negative values can now be entered and saved in the hydro allocation form.


antarest/study/business/allocation_management.py
```diff
@@ -49,14 +49,11 @@ def check_allocation(self) -> "AllocationFormFields":
             raise ValueError("allocation must not contain duplicate area IDs")
 
         for a in allocation:
-            if a.coefficient < 0:
-                raise ValueError("allocation must not contain negative coefficients")
-
             if numpy.isnan(a.coefficient):
                 raise ValueError("allocation must not contain NaN coefficients")
 
-        if sum(a.coefficient for a in allocation) <= 0:
```
MartinBelthle, line 58: Why did you remove this check ? Is it not relevant anymore ?

Anyway the current code is garbage we shouldn't have to perform the validation at 2 different places, it's error prone

hdinia: If the sum of coefficients is not positive, Pydantic will throw a ValidationError, and we must allow negative coefficients. The simulator actually accepts any real number as weights, including negative values

MartinBelthle: I agree we should allow negative values but we can allow them while still checking the sum is positive.
If it doesn't make sense we can remove it i don't mind but i just wanted to point out the difference to ensure we know what we're doing

MartinBelthle: Other than that the code seems legit

hdinia: I replaced it with this check:

```python
if all(a.coefficient == 0 for a in allocation):
    raise ValueError("At least one allocation coefficient must be non-zero")
```

Previously, we were checking for at least one positive coefficient. With the requested changes, the nuance is that we now only need to ensure that at least one coefficient is non-zero. We no longer care about the sum, since negative values are now allowed.

That's how I interpreted the user's request, do you think there's any issue with this?

MartinBelthle: I discussed with Jean-Marc and what he wants is:
- Allow negative values -> :white_check_mark: 
- Checks that at least one coefficient is not null -> :white_check_mark: 
- Ensures the sum is positive -> You removed that check, we should put it back

MartinBelthle (changes requested):
See my last comment

antarest/study/business/allocation_management.py
hdinia: mybad thanks, I'll put it back

sylvlecl (approved):

PR #2465, opened 2025-04-23
fix(matrixstore): ensure we can read legacy null matrix

MartinBelthle:


antarest/matrixstore/parsing.py
```diff
@@ -42,6 +45,10 @@ def load_matrix(matrix_format: InternalMatrixFormat, path: Path, matrix_version:
 
 
 def save_matrix(matrix_format: InternalMatrixFormat, dataframe: pd.DataFrame, path: Path) -> None:
+    if dataframe.empty:
+        path.touch()
```
sylvlecl: I don't think we should do this for other formats than TSV: the read will probably fail.

Even for TSVs, empty does not mean that there is no headers, I think we will want to store headers to read them again.

sylvlecl (approved):

PR #2466, opened 2025-04-23
feat(launcher): support multiple launchers

TheoPascoli:


tests/launcher/test_web.py
```diff
@@ -144,21 +144,9 @@ def test_get_solver_versions() -> None:
 @pytest.mark.parametrize(
     "solver, status_code, expected",
     [
-        pytest.param(
```
sylvlecl, line 147: we should have a test for when the passed ID is not one of the configured launchers

tests/launcher/test_local_launcher.py
```diff
@@ -165,13 +166,14 @@ def test_select_best_binary():
...
         callbacks=Mock(),
         event_bus=Mock(),
         cache=Mock(),
     )
 
-    assert local_launcher._select_best_binary("600") == binaries["700"]
-    assert local_launcher._select_best_binary("700") == binaries["700"]
-    assert local_launcher._select_best_binary("710") == binaries["800"]
-    assert local_launcher._select_best_binary("1100") == binaries["1000"]
+    assert local_launcher._select_best_binary("600", "id") == binaries["700"]
```
sylvlecl: It's weird that we need to specify the id here, since we already have the launcher object

antarest/launcher/service.py
```diff
@@ -603,50 +603,49 @@ def download_output(self, job_id: str) -> FileDownloadTaskDTO:
...
-                raise KeyError("Default launcher is slurm but it is not registered in the config file")
-
-        # local load calculation
-        local_used_cpus = sum(
-            LauncherParametersDTO.from_launcher_params(job.launcher_params).nb_cpu or 1
-            for job in self.job_result_repository.get_running()
-        )
+        launcher = self.config.launcher.get_launcher_by_id(launcher_id)
+
+        if isinstance(launcher, SlurmConfig):
```
sylvlecl: I think we can use this work to improve a little the design: 
we should move the method to the launcher implementations, instead of having an if/else here

antarest/launcher/adapters/local_launcher/local_launcher.py
```diff
@@ -43,26 +43,28 @@ class LocalLauncher(AbstractLauncher):
...
         self.job_id_to_study_id: Dict[str, Tuple[str, Path, subprocess.Popen]] = {}  # type: ignore
         self.logs: Dict[str, str] = {}
 
-    def _select_best_binary(self, version: str) -> Path:
-        local = self.config.launcher.local
-        if local is None:
-            raise LauncherInitException("Missing parameter 'launcher.local'")
-        elif version in local.binaries:
+    def _select_best_binary(self, version: str, launcher_id: str) -> Path:
+        local = self.config.launcher.get_launcher_by_id(launcher_id)
```
sylvlecl: We should use `self` instead of retrieving another launcher here.

Also, the launcher object should **never** have acces to the whole `Config` containing other launchers configurations:
we need to change the constructors of the `AbstractLauncher` class heirarchy  so that it only receives its own config (instance of LocalConfig or SlurmConfig

antarest/core/config.py
```diff
@@ -417,38 +435,36 @@ class LauncherConfig:
...
-        local = LocalConfig.from_dict(data["local"]) if "local" in data else defaults.local
-        slurm = SlurmConfig.from_dict(data["slurm"]) if "slurm" in data else defaults.slurm
         batch_size = data.get("batch_size", defaults.batch_size)
+        for launcher in data["launchers"]:
+            if launcher["type"] == Launcher.LOCAL:
+                launchers.append(LocalConfig.from_dict(launcher))
+            elif launcher["type"] == Launcher.SLURM:
+                launchers.append(SlurmConfig.from_dict(launcher))
+            else:
+                continue
```
sylvlecl: we should raise an error here instead.
Note that you may use a `match` instead of if/else

antarest/launcher/adapters/local_launcher/local_launcher.py
```diff
@@ -43,26 +43,28 @@ class LocalLauncher(AbstractLauncher):
...
+        launcher_id: str,
         callbacks: LauncherCallbacks,
         event_bus: IEventBus,
         cache: ICache,
     ) -> None:
         super().__init__(config, callbacks, event_bus, cache)
-        if self.config.launcher.local is None:
-            raise LauncherInitException("Missing parameter 'launcher.local'")
-        self.local_workspace = self.config.launcher.local.local_workspace
+        launcher = config.launcher.get_launcher_by_id(launcher_id)
```
sylvlecl: see other comment: here we should only provide a LocalConfig to the constructor, instead of the whole config + id

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -145,17 +145,18 @@ class SlurmLauncher(AbstractLauncher):
     def __init__(
         self,
         config: Config,
+        launcher_id: str,
```
sylvlecl: Same as in local implementation: we should provide a `SlrumConfig` here, not the whole config + id

antarest/launcher/model.py
```diff
@@ -46,6 +46,7 @@ class LauncherParametersDTO(AntaresBaseModel):
     auto_unzip: bool = True
     output_suffix: Optional[str] = None
     other_options: Optional[str] = None
+    launcher_id: str = "local_id"
```
sylvlecl: We should have a `None` default here (in that case, at launch time we will use the default launcher)

antarest/launcher/service.py
```diff
@@ -662,15 +661,11 @@ def get_solver_versions(self, solver: str) -> List[str]:
...
-        local_config = self.config.launcher.local
-        slurm_config = self.config.launcher.slurm
-        default_config = self.config.launcher.default
-        versions_map = {
-            "local": sorted(local_config.binaries) if local_config else [],
-            "slurm": sorted(slurm_config.antares_versions_on_remote_server) if slurm_config else [],
-        }
-        versions_map["default"] = versions_map[default_config]
-        return versions_map[solver]
+        config = self.config.launcher.get_launcher_by_id(solver)
```
sylvlecl: Same here, we should move the if/else to the 2 implementations of the launcher instead

antarest/launcher/service.py
```diff
@@ -662,15 +661,11 @@ def get_solver_versions(self, solver: str) -> List[str]:
         Raises:
             KeyError: if the configuration is not "default", "slurm" or "local".
```
sylvlecl: We need to update docstrings mentioning explicitly that restriction: it now only depends on the configuration

antarest/core/config.py
```diff
@@ -479,12 +491,19 @@ def get_time_limit(self, launcher: Launcher) -> TimeLimitConfig:
...
-        if launcher_config is None:
-            raise InvalidConfigurationError(launcher)
-        return launcher_config.time_limit
+        config = self.get_launcher_by_id(launcher)
+        return config.time_limit
+
+    def get_slurm_configs(self) -> List[SlurmConfig]:
+        return [cfg for cfg in self.configs or [] if isinstance(cfg, SlurmConfig)]
+
+    def get_launcher_by_id(self, launcher_id: str) -> LocalConfig | SlurmConfig:
```
sylvlecl: Naming detail: the `_by_id` part is redundant

antarest/core/config.py
```diff
@@ -479,12 +491,19 @@ def get_time_limit(self, launcher: Launcher) -> TimeLimitConfig:
...
+        if launcher_id == "default":
```
sylvlecl: We should not rely on such a hidden naming convention. It could also cause naming conflicts if someone wants (I don't know why) to actually call a config "default".
Instead, we should accept a None default value in the API, an return the default config in that case.

antarest/launcher/adapters/factory_launcher.py
```diff
@@ -32,14 +32,17 @@ def build_launcher(
...
-                callbacks,
-                event_bus,
-                cache,
-                retrieve_existing_jobs=True,
-            )
+        for cfg in config.launcher.configs or []:
+            if isinstance(cfg, SlurmConfig):
+                dict_launchers[cfg.id] = SlurmLauncher(
+                    config,
+                    cfg.id,
```
sylvlecl: As mentioned in implementations: it's here that we should only provide the `cfg` as an argument, not the whole config + id

antarest/launcher/web.py
```diff
@@ -185,10 +166,10 @@ def get_engines() -> Any:
         summary="Get the SLURM cluster or local machine load",
         response_model=LauncherLoadDTO,
     )
-    def get_load() -> LauncherLoadDTO:
+    def get_load(launcher_id: str = "local") -> LauncherLoadDTO:
```
sylvlecl: Everywhere in the API, we should be consistent:
default launcher id is None, and in that case, we should interprete it as the default one (as defined in configuration, not the raw string "default")

The `Launcher` enum must be removed

resources/antares-desktop-fs/config.yaml
```diff
@@ -19,10 +19,12 @@ storage:
       path: ./studies/
 
 launcher:
-  local:
-    binaries:
-      VER: ANTARES_SOLVER_PATH
-    local_workspace: ./local_workspace
+  launchers:
+    - id: local_id
```
sylvlecl: just `local` will do

antarest/core/config.py
```diff
@@ -300,6 +303,9 @@ def from_dict(cls, data: JSON) -> "LocalConfig":
         Returns: object NbCoresConfig
         """
         defaults = cls()
+        _id = data.get("id", "id")
```
sylvlecl: All those field should be mandatory and not have a default value:
id, name, type. If one is absent, we should raise an InvalidConfigurationError

antarest/core/config.py
```diff
@@ -284,6 +284,9 @@ def __post_init__(self) -> None:
 class LocalConfig:
     """Sub config object dedicated to launcher module (local)"""
 
+    id: str = ""
```
sylvlecl: We should not provide a default value here, the field is mandatory.
Same for name.

antarest/core/config.py
```diff
@@ -284,6 +284,9 @@ def __post_init__(self) -> None:
...
+    name: str = ""
+    type: Launcher = Launcher.LOCAL
```
sylvlecl: This should be a `ClassVar`

antarest/core/config.py
```diff
@@ -333,6 +342,9 @@ class SlurmConfig:
     Sub config object dedicated to launcher module (slurm)
     """
 
+    id: str = ""
```
sylvlecl: Same comments for that class as for local config

antarest/core/config.py
```diff
@@ -417,38 +435,36 @@ class LauncherConfig:
...
-    local: Optional[LocalConfig] = None
-    slurm: Optional[SlurmConfig] = None
+    configs: Optional[List[LocalConfig | SlurmConfig]] = None
     batch_size: int = 9999
 
     @classmethod
     def from_dict(cls, data: JSON) -> "LauncherConfig":
+        launchers: List[LocalConfig | SlurmConfig] = []
         defaults = cls()
         default = data.get("default", cls.default)
```
sylvlecl, line 435: Now we need to check that default is actually defined in the list of configurations

sylvlecl (changes requested):
A few requests to polish the implementation

tests/launcher/test_web.py
TheoPascoli: It does not seem really easy to add a test here with the mocks and the new hierarchy. I can spend more time on it but I couldn't make it in a reasonable amount of time

antarest/core/config.py
```diff
@@ -475,12 +484,82 @@ def get_time_limit(self, launcher: Launcher) -> TimeLimitConfig:
...
+        except StopIteration:
+            raise InvalidConfigurationError(launcher_id)
+
+    def get_load(self, running_jobs: List[JobResult], launcher_id: Optional[str]) -> LauncherLoadDTO:
+        """
+        Get the load of the SLURM cluster or the local machine.
+        """
+        launcher = self.get_launcher(launcher_id)
+
+        if isinstance(launcher, SlurmConfig):
```
sylvlecl: It's important to isolate as much as possible the specificities of each launcher in their implementation (`LocalLauncher` and `SlurmLauncher`) --> the methods `get_load` and `get_solver_versions` methods go there instead. The config should remain a pure data class with no additional logic inside.

antarest/launcher/adapters/abstractlauncher.py
```diff
@@ -62,7 +62,7 @@ class LauncherCallbacks(NamedTuple):
 class AbstractLauncher(ABC):
     def __init__(
         self,
-        config: Config,
+        config: LocalConfig | SlurmConfig,
```
sylvlecl: we can remove it from this class instead, it's not useful to have it in this abstract class

antarest/launcher/adapters/factory_launcher.py
```diff
@@ -32,14 +32,16 @@ def build_launcher(
...
-        if config.launcher.slurm is not None:
-            dict_launchers["slurm"] = SlurmLauncher(
-                config,
-                callbacks,
-                event_bus,
-                cache,
-                retrieve_existing_jobs=True,
-            )
+        for cfg in config.launcher.configs or []:
+            if isinstance(cfg, SlurmConfig):
```
sylvlecl, line 36: In relation to the previous comment: this is indeed the only legitimate place to do the dispatch slurm/local, all other specificities should be handled by the launchers (by the implementation class) that we instantiate here.

antarest/launcher/adapters/local_launcher/local_launcher.py
```diff
@@ -42,15 +42,14 @@ class LocalLauncher(AbstractLauncher):
...
+        config: LocalConfig,
         callbacks: LauncherCallbacks,
         event_bus: IEventBus,
         cache: ICache,
     ) -> None:
         super().__init__(config, callbacks, event_bus, cache)
-        if self.config.launcher.local is None:
-            raise LauncherInitException("Missing parameter 'launcher.local'")
-        self.local_workspace = self.config.launcher.local.local_workspace
+        self.local_config: LocalConfig = cast(LocalConfig, self.config)
```
sylvlecl: we don't need the cast here

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -144,18 +144,17 @@ def apply_adequacy_patch(self, launcher_params: LauncherParametersDTO) -> None:
...
         self,
-        config: Config,
+        config: SlurmConfig,
         callbacks: LauncherCallbacks,
         event_bus: IEventBus,
         cache: ICache,
         use_private_workspace: bool = True,
         retrieve_existing_jobs: bool = False,
     ) -> None:
+        if not isinstance(config, SlurmConfig):
```
sylvlecl: We don't really need this now, this will be guaranteed by the type checker in CI

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -144,18 +144,17 @@ def apply_adequacy_patch(self, launcher_params: LauncherParametersDTO) -> None:
...
+            raise TypeError(f"Launcher {config.id} is not a SlurmConfig")
         super().__init__(config, callbacks, event_bus, cache)
-        if config.launcher.slurm is None:
-            raise LauncherInitException("Missing parameter 'launcher.slurm'")
-
-        self.slurm_config: SlurmConfig = config.launcher.slurm
+        self.slurm_config: SlurmConfig = cast(SlurmConfig, self.config)
```
sylvlecl: we don't need the cast anymore

antarest/launcher/service.py
```diff
@@ -603,74 +601,28 @@ def download_output(self, job_id: str) -> FileDownloadTaskDTO:
...
-        local_config = self.config.launcher.local
-        slurm_config = self.config.launcher.slurm
-        default_config = self.config.launcher.default
-        versions_map = {
-            "local": sorted(local_config.binaries) if local_config else [],
-            "slurm": sorted(slurm_config.antares_versions_on_remote_server) if slurm_config else [],
-        }
-        versions_map["default"] = versions_map[default_config]
-        return versions_map[solver]
+        return self.config.launcher.get_solver_versions(launcher_id)
```
sylvlecl: In relation with previous comment: here we should ask the the launcher implementation instead:
```python
return self.launchers[cluster_id].get_solver_versions()
```

antarest/launcher/service.py
```diff
@@ -603,74 +601,28 @@ def download_output(self, job_id: str) -> FileDownloadTaskDTO:
...
-
-        # local load calculation
-        local_used_cpus = sum(
-            LauncherParametersDTO.from_launcher_params(job.launcher_params).nb_cpu or 1
-            for job in self.job_result_repository.get_running()
-        )
 
-        # The cluster load is approximated by the percentage of used CPUs.
-        cluster_load_approx = min(100.0, 100 * local_used_cpus / (os.cpu_count() or 1))
+        return self.config.launcher.get_load(self.job_result_repository.get_running(), launcher_id)
```
sylvlecl: In relation with previous comment: here we should ask the the launcher implementation instead:
```python
return self.launchers[cluster_id].get_load()
```
we don't need to provide the launcher with more information such as running jobs, since he knows itself which jobs are running

tests/launcher/test_web.py
sylvlecl: Ok, the test actually does not make sense, it's anyway enough to check this at the service level, which is already done in the commented tests.

tests/launcher/test_service.py
```diff
@@ -305,112 +304,210 @@ def test_service_get_jobs_from_database(self, db_session) -> None:
     @pytest.mark.parametrize(
         "config, solver, expected",
         [
+            # pytest.param(
```
sylvlecl: In general, we should either completely delete tests or make them pass (otherwise we'll end up with too many commented stuff :) ) :
here it seems that the commented out test actually pass, so I think we can re-activate them

sylvlecl (changes requested):
Just a few more comments: the local/slurm specificities need to go in implementations of the launcher interfaces, not in the config classes

antarest/launcher/adapters/local_launcher/local_launcher.py
```diff
@@ -218,3 +214,22 @@ def kill_job(self, job_id: str) -> None:
...
                 None,
             )
+
+    @override
+    def get_solver_versions(self) -> List[str]:
+        return sorted(self.local_config.binaries)
+
+    @override
+    def get_load(self) -> LauncherLoadDTO:
+        local_used_cpus = len(self.job_id_to_study_id) or 1
```
sylvlecl: The formulae needs following fixes:
- we need to take into account the nb_core used per computation, which is defined in the configuration (see previous code in the service)
- if the list of jobs is empty we should have 0 instead of 1

sylvlecl (approved):

antarest/launcher/adapters/local_launcher/local_launcher.py
sylvlecl: Done in pair programming

PR #2477, opened 2025-05-02
fix(ui-studytree): minor fixes and improvement

smailio:


webapp/src/components/App/Studies/StudyTree/StudyTreeNode.tsx
```diff
@@ -16,18 +16,26 @@ import TreeItemEnhanced from "@/components/common/TreeItemEnhanced";
 import * as R from "ramda";
 import { useMemo } from "react";
 import { useTranslation } from "react-i18next";
-import type { StudyTreeNodeProps } from "./types";
+import type { StudyTreeNodeProps, StudyTreeNode } from "./types";
+
+function prioritizeDefault(folderA: StudyTreeNode, folderB: StudyTreeNode): number {
+  if (folderA.name === "default") {
```
smailio: use constant

webapp/src/components/App/Studies/StudyTree/StudyTreeNode.tsx
```diff
@@ -16,18 +16,26 @@ import TreeItemEnhanced from "@/components/common/TreeItemEnhanced";
...
   const { hasChildren, children, path, name } = node;
   const isLoading = itemsLoading.includes(node.path);
   const hasUnloadedChildren = hasChildren && children.length === 0;
   const { t } = useTranslation();
 
-  const sortedChildren = useMemo(
-    () => R.sortBy(R.compose(R.toLower, R.prop("name")), children),
-    [children],
-  );
+  const nameSort = R.sortWith([prioritizeDefault, R.ascend(R.compose(R.toLower, R.prop("name")))]);
```
smailio: prioritize default only if name  == root

skamril (dismissed):

webapp/src/components/App/Studies/StudyTree/index.tsx
```diff
@@ -17,37 +17,45 @@ import useEnqueueErrorSnackbar from "@/hooks/useEnqueueErrorSnackbar";
...
   const enqueueErrorSnackbar = useEnqueueErrorSnackbar();
   const dispatch = useAppDispatch();
   const [t] = useTranslation();
 
-  useUpdateEffect(() => {
-    const nextStudiesTree = insertFoldersIfNotExist(initialStudiesTree, subFolders);
-    setStudiesTree(nextStudiesTree);
-    // subFolders isn't a dependency because we don't want to trigger this code
+  useEffect(() => {
+    api.getWorkspaces().then((nextWorkspaces) => {
```
skamril: @sylvlecl Do we need a condition to call the API only in desktop mode?

sylvlecl: Indeed it's not gonna be useful for the web deployment.
Actually even in desktop mode I guess it will rarely be needed (only when the user adds a new disk and does not refresh the page?).

skamril (approved):

webapp/src/components/App/Studies/StudyTree/index.tsx
smailio: On desktop it's usefull in the case a user wants to explore a workspace in which he hasn't scanned any study yet.

PR #2479, opened 2025-05-05
chore(git): fix conflicts between master and dev due to squash merge of v2.19.2 (master part)

sylvlecl:
**Explanation:**
Release v2.19.2 has not been done correctly according to git flow:
squash merges have been performed in master then in dev, which now causes
conflicts between them (seen on release/v2.20.0 branch).

We now need to fix that by actually merging again branch hotfix/v2.19.2 in master and in dev (release/v2.20.0 actually).

This is the "merge in master" part. We need to leave a message commit in master explaining it.
Note that this merge does not bring any actual difference, but is necessary for conflict resolution.

### DIFF
--- a/antarest/__init__.py
+++ b/antarest/__init__.py
@@ -19,9 +19,9 @@ from pathlib import Path
 
 # Standard project metadata
 
-__version__ = "2.19.1"
+__version__ = "2.19.2"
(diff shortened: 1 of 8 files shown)


sylvlecl:
Closing, we'll merge master into release branch instead.

PR #2480, opened 2025-05-05
fix(scan): use workspace and study path as key of missing studies

smailio:
When I removed a workspace from the conf I had an issue where the scan deleted the studies from that workspace and just after the scan saved the studies again. 

In the UI the studies were deleted, so they were removed, because the scan emitted a DELETE event, and a few seconds later they would come back. 

The cause of this bug is that the scan used two different keys in two different dict that are both used to identify weither a study was missing, which makes no sense. 

It's strange that this bug wasn't noticed so far, so maybe it was a regression.


tests/storage/test_service.py
```diff
@@ -276,50 +276,96 @@ def test_study_listing(db_session: Session) -> None:
...
     me = RawStudy(
         id="e",
         path="e",
+        folder="e",
+        name="e",
         created_at=now,
         missing=datetime.utcnow() - timedelta(MAX_MISSING_STUDY_TIMEOUT - 1),
+        workspace="workspace1",
     )
-    fc = StudyFolder(path=Path("c"), workspace=DEFAULT_WORKSPACE_NAME, groups=[])
```
smailio, line 300: It doens't make sense to use default workspace here as the studies uder default workspace aren't scanned and "default" is kind of a reserved keyword.

sylvlecl (approved):

PR #2484, opened 2025-05-07
feat(aggregation): streamed aggregation

sylvlecl:
The aggregation of outputs, which is actually just a concatenation of several tables, is currently performed fully in-memory.
For large studies, this can quickly use a huge amount of memory (10ths of Go), whereas it's not actually needed to hold the concatenated table in memory.

This PR implements a streamed concatenation:
the concatenated files are read 1 by 1 and appended to the underlying target file, without building a concatenated table in memory.
More specifically, we now use a `Generator` of `Dataframe` instead of a concatenated dataframe, as input to the serialization method.

Test results on real outputs (1000 MC years) show a 2x speedup and no accumulation of memory.


antarest/study/business/aggregator_management.py
```diff
@@ -13,12 +13,12 @@
...
 from enum import Enum, StrEnum
 from pathlib import Path
-from typing import Any, Dict, List, MutableSequence, Optional, Sequence
+from typing import Any, Dict, Iterator, List, MutableSequence, Optional, Sequence
 
 import numpy as np
 import pandas as pd
 
-from antarest.core.exceptions import FileTooLargeError, MCRootNotHandled, OutputNotFound, OutputSubFolderNotFound
+from antarest.core.exceptions import MCRootNotHandled, OutputNotFound, OutputSubFolderNotFound
```
MartinBelthle, line 21: As discussed, we should remove the `aggregation_results_max_size` at the instantiation here as we don't use it anymore

antarest/core/serde/matrix_export.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def _csv_stream_writer(sep: str, decimal: str) -> DataframeStreamWriter:
+    def writer(path: Path, dataframes: Iterator[pd.DataFrame]) -> None:
+        _write_dataframes_stream_csv(path, sep, decimal, dataframes)
+
+    return writer
+
+
+def write_dataframes_stream_excel(path: Path, dataframes: Iterator[pd.DataFrame]) -> None:
```
MartinBelthle: Seems to me that the 2 following writers should be private

antarest/core/serde/matrix_export.py
```diff
@@ -0,0 +1,201 @@
...
+            columns = df.columns
+        else:
+            if any(columns != df.columns):
+                raise ValueError("Cannot append dataframe to file, columns are different from initial dataframe.")
+        yield df
+
+
+def _write_dataframes_stream_csv(path: Path, sep: str, decimal: str, dataframes: Iterator[pd.DataFrame]) -> None:
+    headers = True
+    mode: Literal["w", "a"] = "w"
```
MartinBelthle: Minor: we should choose between the ` mode="r+" if append else "w"` and this solution to harmonize methods

tests/core/serde/test_matrix_export.py
```diff
@@ -0,0 +1,105 @@
...
+)
+def test_csv_stream_writer(
+    tmp_path: Path, dataframes: Iterator[pd.DataFrame], format: TableExportFormat, expected_delimiter: str
+) -> None:
+    export_path = tmp_path / "export.csv"
+    writer = format.get_stream_writer()
+    writer(path=export_path, dataframes=dataframes)
+    parsed = pd.read_csv(export_path, delimiter=expected_delimiter)
+
+    expected = pd.DataFrame(data=[(10, 11), (12, 13), (12, 13), (14, 15)], columns=["A", "B"])
```
MartinBelthle: Minor: We could declare this DF inside a global var at the start of the file and use it in every test to improve readability

MartinBelthle (changes requested):

MartinBelthle (approved):
Seems good to me

PR #2486, opened 2025-05-14
fix(study)!: rename variable for study copy

TheoPascoli:


antarest/study/web/studies_blueprint.py
```diff
@@ -332,7 +332,7 @@ def copy_study(
 
         Args:
         - `uuid`: The identifier of the study you wish to duplicate.
-        - `dest`: The destination workspace where the study will be copied.
+        - `study_name`: The destination workspace where the study will be copied.
```
sylvlecl: we need to adapt the docstring

sylvlecl:
We need to adapt the front before merging:
probably better to wait for the work of @skamril on the front-end part of variant copies.

TheoPascoli:
> We need to adapt the front before merging: probably better to wait for the work of @skamril on the front-end part of variant copies.

The front-end has been adapted

sylvlecl (approved):
Ok!
Let's not include it in v2.21.0 though

PR #2490, opened 2025-05-16
perf(db): remove N+1 requests when getting users and groups list

MartinBelthle:
[ANT-2893]


antarest/login/service.py
```diff
@@ -564,33 +569,46 @@ def get_all_users(self, details: Optional[bool] = False) -> List[Union[UserInfo,
...
+            roles_per_user.setdefault(role.identity_id, []).append(
+                RoleDTO(
+                    group_id=role.group_id,
+                    group_name=role.group.name,
+                    identity_id=role.identity_id,
+                    type=role.type,
+                )
+            )
+
+        if user.is_site_admin():
```
TheoPascoli, line 607: Could you please refactor the conditions to separate the admin and non-admin cases more explicitly, and handle the details flag in a clearer way? I have trouble reading this method

antarest/login/service.py
```diff
@@ -518,31 +520,34 @@ def get_all_groups(self, details: bool = False) -> List[Union[GroupDetailDTO, Gr
```
TheoPascoli, line 521: Right above, the documentation is not clear : `details: get all user information, including users.` What does it mean ?

antarest/login/service.py
```diff
@@ -564,33 +569,46 @@ def get_all_users(self, details: Optional[bool] = False) -> List[Union[UserInfo,
```
TheoPascoli, line 571: In the doc : `Returns: list of groups`. I think it would be better to say that it returns a list of users

TheoPascoli (changes requested):
Hello, just a few remarks

smailio (dismissed):

antarest/login/service.py
MartinBelthle: I'll explicit a bit more

antarest/login/service.py
MartinBelthle: Indeed, I'll explicit a bit more

antarest/login/service.py
MartinBelthle: I agree but I don't know how to simplify it. I added comments in the code to help readability

TheoPascoli (approved):

PR #2495, opened 2025-05-21
fix(ui-playlist): some issues

skamril:
- Use decimal.js-light to calculate the weights sum, to prevent issue with floats
- Rename label
- Padding issue
- Reduce the dialog size

### DIFF
--- a/webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
+++ b/webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
@@ -25,6 +25,7 @@ import { DEFAULT_WEIGHT } from "@/services/api/studies/config/playlist/constants
 import type { Playlist, PlaylistData } from "@/services/api/studies/config/playlist/types";
 import { appendColon } from "@/utils/i18nUtils";
 import { Box, Button, ButtonGroup } from "@mui/material";
+import { Decimal } from "decimal.js-light";
 import * as R from "ramda";
 import * as RA from "ramda-adjunct";
 import { useMemo, useRef, useState } from "react";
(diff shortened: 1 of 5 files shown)


webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
```diff
@@ -98,7 +100,7 @@ function ScenarioPlaylistDialog({ study, open, onClose }: Props) {
       (acc, [_, { status, weight }]) => {
         if (status) {
           acc.selected += 1;
-          acc.sumWeights += weight;
+          acc.sumWeights = new Decimal(acc.sumWeights).plus(weight).toNumber();
```
hdinia, line 104: you can add a comment to explain the issue

hdinia, line 104: Using a lib for this only seems not ideal, do you have any reason to not do this in plain JS ? 
I dont know the issue you having, but have you tried something like this `acc.sumWeights = Number.parseFloat((acc.sumWeights + weight).toFixed(2));`

or if you dealing with floating point errors you can also consider Number.EPSILON 
```ts
function roundToPrecision(num, decimals = 2) {
  const multiplier = 10 ** decimals;
  return Math.round((num + Number.EPSILON) * multiplier) / multiplier;
}


acc.sumWeights = roundToPrecision(acc.sumWeights + weight);
```

hdinia (changes requested):

webapp/src/components/App/Singlestudy/explore/Configuration/General/dialogs/ScenarioPlaylistDialog/index.tsx
skamril: I must not round the result (I already asked). Vanilla JS gives wrong results with float values in some case.
e.g.: 
`0.1+0.1+0.1+0.1+0.1+0.1+0.1+0.1+0.1+0.1` gives `0.999999....`.
`0.1 + 0.2` gives `0.24400000000000002`

The lib allows to fix this issue.

skamril: Ok

hdinia (approved):

PR #2496, opened 2025-05-21
feat(ui-study,ui-studies): uniformize context menu

skamril:


webapp/public/locales/fr/main.json
```diff
@@ -645,8 +646,7 @@
   "study.success.studyIdCopy": "Identifiant de l'étude copié !",
   "study.success.jobIdCopy": "Identifiant de la tâche copié !",
   "study.success.modifiedStudy": "L'étude {{studyname}} a été modifié avec succès",
-  "studies.question.delete": "Êtes-vous sûr de vouloir supprimer cette étude ?",
-  "studies.deleteSubvariants": "Supprimer toutes les sous variantes",
+  "studies.question.delete": "Êtes-vous sûr de vouloir supprimer cette étude ainsi que tous ses sous-variants ?",
```
sylvlecl: We should stick with "variant**e**" for consistency sake

webapp/src/components/App/shared/studies/dialogs/ExportModal/ExportFilter/Filter/style.ts
sylvlecl, line 1: Is it on purpose to have it in `components/App/shared` and not in `components/common` ?
Maybe it's too specific to studies to the business to go into common ?

webapp/src/hooks/useConfirm.ts
```diff
@@ -106,6 +106,7 @@ function useConfirm<TData = undefined>() {
     yes: yesRef.current,
     no: noRef.current,
     cancel: cancelRef.current,
+    setData,
```
sylvlecl, line 109: Not sure it's relevant to expose this in the return type ?
The data is supposed to be set through `showConfirm`, `setData` does not seem used anywhere else.

sylvlecl (commented):
Looks good, minor comments/question

webapp/src/components/App/shared/studies/dialogs/ExportModal/ExportFilter/Filter/style.ts
skamril: Yes, I move it in `components/App/shared` because for me it's too specific to studies.
I would like after to move `SingleStudy` into `Studies`, and move the `shared` folder in it.

webapp/src/hooks/useConfirm.ts
skamril: I used it for a checkbox in the delete confirmation, but I finally removed it.
`showConfirm` for the inital data, and `setData` for the `onChange` of the checkbox.
I think it could be useful, that's why I left it.

webapp/public/locales/fr/main.json
skamril: Ok

sylvlecl (approved):

PR #2509, opened 2025-05-28
refactor(renewable): introduce the DAO layer and improve classes design

MartinBelthle:


antarest/study/business/areas/thermal_management.py
```diff
@@ -123,7 +123,7 @@ def update_thermals_props(
             old_thermals_by_ids = old_thermals_by_areas[area_id]
             for thermal_id, update_cluster in update_thermals_by_ids.items():
                 # Update the thermal cluster properties.
-                old_cluster = old_thermals_by_ids[thermal_id]
+                old_cluster = old_thermals_by_ids[thermal_id.lower()]
```
sylvlecl: Actually we should not lower here, the input is supposed to be a `LowerCaseId`

tests/variantstudy/test_command_factory.py
```diff
@@ -1060,6 +1050,6 @@ def test_parse_create_renewable_cluster_dto_v1(command_factory: CommandFactory):
     assert len(commands) == 1
```
sylvlecl, line 1059: We should add some unit test to check that we still parse correctly v2 commands

antarest/study/storage/variantstudy/command_factory.py
```diff
@@ -178,12 +178,9 @@ def to_command(self, command_dto: CommandDTO) -> List[ICommand]:
         """
         args = command_dto.args
         if isinstance(args, dict):
-            # In some cases, pydantic can modify inplace the given args.
```
sylvlecl, line 181: Are we sure it's OK to remove this ?
It's still true that validation should not modify the input dict.

antarest/study/storage/variantstudy/model/command/update_renewables_clusters.py
```diff
@@ -43,63 +41,27 @@ class UpdateRenewablesClusters(ICommand):
...
+            if area_id not in all_renewables:
+                return command_failed(f"The area '{area_id}' is not found.")
 
-            try:
-                all_clusters_for_area = study_data.tree.get(ini_path)
-            except ChildNotFoundError:
-                return CommandOutput(status=False, message=f"The area '{area_id}' is not found.")
+            new_clusters = []
+            for cluster_id, new_properties in value.items():
+                lowered_id = cluster_id.lower()
```
sylvlecl: We should not lower, we have a `LowerCaseId` as input so it must already be a valid ID

antarest/study/dao/file/file_study_renewable_dao.py
```diff
@@ -0,0 +1,171 @@
...
+            # may raise KeyError if the path is missing
+            clusters = file_study.tree.get(path.split("/"), depth=5)
+            # may raise KeyError if "list" is missing
+            clusters = {area_id: cluster_list["list"] for area_id, cluster_list in clusters.items()}
+        except KeyError:
+            raise RenewableClusterConfigNotFound(path)
+
+        renewables_by_areas: dict[str, dict[str, RenewableCluster]] = {}
+        for area_id, cluster_obj in clusters.items():
+            for cluster_id, cluster in cluster_obj.items():
```
sylvlecl, line 56: the key is not the cluster **id**, but the cluster **name**

antarest/study/dao/file/file_study_renewable_dao.py
```diff
@@ -0,0 +1,171 @@
...
+                lowered_id = cluster_id.lower()
```
sylvlecl: we'd better just use `cluster_obj.id`, the objects is responsible for knowing its own ID. The lower case character is an implementation detail that we don't want to know here

antarest/study/dao/file/file_study_renewable_dao.py
```diff
@@ -0,0 +1,171 @@
...
+_ALL_CLUSTERS_PATH = "input/renewables/clusters"
+
+
+class FileStudyRenewableDao(RenewableDao, ABC):
+    @abstractmethod
+    def get_file_study(self) -> FileStudy:
+        pass
+
+    @override
+    def get_all_renewables(self) -> dict[str, dict[str, RenewableCluster]]:
```
sylvlecl, line 40: The typing is not explicit about the keys here, we must absolutely document in docstring that they are area **id** and cluster **id**.

It would be better  if we could make it clear in the typing, not sure of the best way to do this.

tests/variantstudy/test_command_factory.py
```diff
@@ -416,40 +416,33 @@
...
             },
             study_version=STUDY_VERSION_8_8,
         ),
         {
             "area_id": "area_name",
             "parameters": {
-                "enabled": True,
-                "group": "other res 1",
                 "name": "cluster_name",
-                "nominalcapacity": 0.0,
```
sylvlecl, line 438: Why not keep a test with all parameters ?

sylvlecl (changes requested):
A few comments, the important part is to add unit tests for previous command versions parsing

antarest/study/storage/variantstudy/command_factory.py
MartinBelthle: Yes because we're already doing the copy inside the `to_single_command` method which is called just after

antarest/study/dao/file/file_study_renewable_dao.py
MartinBelthle: Yep, we discussed it and I did it inside the STS PR but I forgot to put it back here

antarest/study/dao/file/file_study_renewable_dao.py
MartinBelthle: Yes. But I still need to do .lower() because the renewable id is not in lower case

antarest/study/dao/file/file_study_renewable_dao.py
MartinBelthle: I'll add the docstring (for thermals too)

tests/variantstudy/test_command_factory.py
MartinBelthle: I put back a test with all params

sylvlecl (approved):

PR #2514, opened 2025-06-03
feat(ui-study): add copy dialog and rename action in context menu

skamril:
- allow to define a name for the new study
- extra: update ConfirmDialog

ANT-2886


webapp/public/locales/en/main.json
```diff
@@ -308,6 +308,7 @@
   "study.managedStudy": "Managed study",
   "study.properties": "Properties",
   "study.upgrade": "Upgrade study",
+  "study.copy": "Save as reference study",
```
sylvlecl: This should not be the general "copy" phrasing, only for variants.
For other studies, we should stick with "copy"

webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+      withOutputs: false,
+    });
+  };
+
+  ////////////////////////////////////////////////////////////////
+  // JSX
+  ////////////////////////////////////////////////////////////////
+
+  return (
+    <FormDialog
```
sylvlecl, line 57: I think it would be better to have a "Cancel" instead of a "Close" for the cancel button, for all the actions of the menu (the case for launch, but not for other actions).

But maybe out of the scope of this PR

webapp/src/utils/validation/string.ts
```diff
@@ -83,10 +86,14 @@ export function validateString(
     maxLength = 255,
   } = options;
 
+  if (!value) {
```
sylvlecl, line 89: Note that this will make "" allowed, but not "   ".
It seemed weird to me at first, but after all it looks OK when using the form.

webapp/public/locales/en/main.json
sylvlecl: (Done)

sylvlecl (dismissed):
Looks, good, we should rename "close" to "cancel" in actions dialogs in a future work

webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
skamril: Yes I will change it in another PR

sylvlecl (approved):

PR #2515, opened 2025-06-03
feat(matrix): add different handling of matrix for managed and unmanaged study

TheoPascoli:


antarest/matrixstore/matrix_uri_mapper.py
```diff
@@ -25,22 +29,87 @@ def build_matrix_uri(id: str) -> str:
...
+        pass
+
+
+class MatrixUriMapperManaged(MatrixUriMapper):
+    def __init__(self, matrix_service: ISimpleMatrixService) -> None:
+        self._matrix_service = matrix_service
+
+    @override
+    def get_matrix(self, uri: str) -> pd.DataFrame:
+        print("MANAGED")
```
MartinBelthle: You can remove all the `print` in this file

antarest/study/storage/variantstudy/snapshot_generator.py
```diff
@@ -105,6 +105,7 @@ def generate_snapshot(
             # The snapshot is generated, we also need to de-normalize the matrices.
             file_study = self.study_factory.create_from_fs(
                 snapshot_dir,
+                is_managed(root_study),
```
MartinBelthle: In the case of generating a variant, the studies are always managed

tests/study/storage/variantstudy/test_snapshot_generator.py
```diff
@@ -870,13 +870,14 @@ def test_generate__nominal_case(
...
-        # We expect 5 queries:
+        # We expect 6 queries:
         # - 1 query to fetch the ancestors of a variant study,
         # - 1 query to fetch the root study (with owner and groups for permission check),
         # - 1 query to fetch the list of variants with snapshot, commands, etc.,
+        # - 1 query to fetch raw study information,
         # - 1 query to update the variant study additional_data,
         # - 1 query to insert the variant study snapshot.
-        assert len(db_recorder.sql_statements) == 5, str(db_recorder)
+        assert len(db_recorder.sql_statements) == 6, str(db_recorder)
```
MartinBelthle, line 880: Why do we need to add a request ? Pehaps we can remove it as we know we're in the case of a managed study

antarest/study/storage/utils.py
```diff
@@ -399,7 +399,7 @@ def ignore_outputs(directory: str, _: Sequence[str]) -> Sequence[str]:
     duration = "{:.3f}".format(stop_time - start_time)
     with_outputs = "with outputs" if outputs else "without outputs"
     logger.info(f"Study '{study_dir}' exported ({with_outputs}, flat mode) in {duration}s")
-    study = study_factory.create_from_fs(dest, "", use_cache=False)
+    study = study_factory.create_from_fs(dest, True, "", use_cache=False)
```
MartinBelthle: I think we can enter this code with an unmanaged study (exporting an unmanaged study is dumb but possible)

antarest/launcher/extensions/adequacy_patch/extension.py
```diff
@@ -110,7 +110,7 @@ def after_export_flat_hook(
     ) -> None:
         logger.info("Applying adequacy patch postprocessing script")
         study = self.study_service.storage_service.raw_study_service.study_factory.create_from_fs(
-            study_export_path, study_id, use_cache=False
+            study_export_path, True, study_id, use_cache=False
```
MartinBelthle: Same here

MartinBelthle (changes requested):
Seems quite good. We should add tests for unmanaged studies though

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -129,9 +129,13 @@ def save(self, data: str | bytes | S, url: Optional[List[str]] = None) -> None:
         self._assert_url_end(url)
 
         if isinstance(data, str) and self.matrix_mapper.matrix_exists(data):
-            self.get_link_path().write_text(data)
-            if self.config.path.exists():
-                self.config.path.unlink()
+            if self.matrix_mapper.is_managed:
```
sylvlecl: This choice should be handled inside the "mapper" instead, and the mapper should not expose the is_managed property:
the idea is to actually hide this implementation detail.

But it's not obvious to do because there are several things entangled here in this class.
First, this `LazyNode` should not even know about matrices.
I think it should not even know about "link" files, which are also only used in conjunction with the matrix store, as far as I know.

So in order to clean up the design, we probably need to refactor this a little.
Need to think a little more about it, but we should move down into `MatrixNode` everything which is specific to them.

sylvlecl: Indeed link files are only used in `MatrixNode` which is the only one which actually implements normalize/denormalize.

antarest/matrixstore/matrix_uri_mapper.py
```diff
@@ -25,22 +36,126 @@ def build_matrix_uri(id: str) -> str:
     return f"{MATRIX_PROTOCOL_PREFIX}{id}"
 
 
-class MatrixUriMapper:
+class MatrixUriMapperType(StrEnum):
```
sylvlecl: I have a global proposition of renaming:
this layer of the code is one of the "lower level": it should not really know about managed or unmanaged studies, it's more the opposite.

Instead we can name those implementations by what they do: writing "normalized" or "plain/denormalized" matrices.

So we could have `NORMALIZED` and `DENORMALIZED` value for the enum, and `NormalizedMatrixUriMapper` for the names

antarest/study/storage/rawstudy/model/filesystem/factory.py
```diff
@@ -92,16 +98,19 @@ def create_from_fs(
         lock_file = os.path.join(self._lock_dir, self._lock_fmt.format(basename=lock_basename))
         with filelock.FileLock(lock_file):
             logger.info(f"🏗 Creating a study by reading the configuration from the directory '{path}'...")
-            return self._create_from_fs_unsafe(path, study_id, output_path, use_cache)
+            return self._create_from_fs_unsafe(path, is_managed, study_id, output_path, use_cache)
 
     def _create_from_fs_unsafe(
         self,
         path: Path,
+        is_managed: bool,
```
sylvlecl: Here instead of the the `is_managed`, I think it would be more explicit and understandable to have a parameter called `with_matrix_normalization`.

Then, it's the responsibility of the higher level components (the services), use the normalized form for managed studies.

antarest/study/storage/rawstudy/model/filesystem/lazy_node.py
```diff
@@ -114,30 +111,14 @@ def get_node(
     @override
     def delete(self, url: Optional[List[str]] = None) -> None:
         self._assert_url_end(url)
-        if self.get_link_path().exists():
-            self.get_link_path().unlink()
-        elif self.config.path.exists():
+        if self.config.path.exists():
             self.config.path.unlink()
 
-    def get_link_path(self) -> Path:
```
sylvlecl, line 122: Nice to have removed it from here !

antarest/study/storage/rawstudy/model/filesystem/matrix/matrix.py
```diff
@@ -82,6 +83,59 @@ def __init__(
         LazyNode.__init__(self, matrix_mapper, config)
         self.freq = freq
 
+    def get_link_path(self) -> Path:
```
sylvlecl: I think that the best thing would be to completely hide the link file into the "mapper" implementations, since it's the mapper which has the responsibility to read/write matrix data from/to disk.

To achieve this we would need to provide the `path` (not the link path) to the matrix mapper methods.
We may need to add more methods to the matrix mapper, so that it handles also the loading of the matrix, the delete etc.

Maybe there is a better solution, but in the current state the responsibilites between the matrix node and the mapper are not very clear :(

sylvlecl (commented):
Some design comments:
- proposition of renaming the mappers implementations
- I think we should clarify even further the responsibilities between the mapper and the matrix node.

antarest/matrixstore/matrix_uri_mapper.py
```diff
@@ -25,22 +36,126 @@ def build_matrix_uri(id: str) -> str:
...
+    @abstractmethod
+    def create_matrix(self, matrix: pd.DataFrame) -> str:
+        pass
+
+    @abstractmethod
+    def matrix_exists(self, uri: str) -> bool:
+        pass
+
+    @abstractmethod
+    def handle_matrix_save(self, data: "MatrixNode", matrix_uri: str, link_path: Path) -> None:
```
sylvlecl: Related to the responsibilities of the mapper vs. the ones of the node:

now this interface is not homogeneous anymore: it has methods that are independent from the nodes, and a new method which works directly on a node.
I think the 3 methods that existed before could be replaced by methods that work on nodes (or on the underlying path) instead, to encapsulate the "link" file implementation detail. The MatrixNode would just delegate its operations to it:
```python
def load_matrix(path: Path) -> pd.Dataframe:
   ...
def save_matrix(path: Path, uri: str) -> None:
   ...
def normalize(path: Path) -> None:
   ...
def denormalize(path: Path) -> None:
   ...
```

TheoPascoli:
@sylvlecl The responsibilities of each class are better separated. However, I can go further and completely hide the concept of link from matrix.py by making it call an abstraction implemented with the link mechanism

sylvlecl (approved):

sylvlecl:
> @sylvlecl The responsibilities of each class are better separated. However, I can go further and completely hide the concept of link from matrix.py by making it call an abstraction implemented with the link mechanism

Would probably be an even better design, but probably not worth the effort for now, thanks !

PR #2516, opened 2025-06-03
feat(ui-study): add destination folder in copy dialog

skamril:
extra: update move dialog

ANT-2887


webapp/src/components/App/shared/studies/dialogs/MoveStudyDialog.tsx
```diff
@@ -103,9 +104,20 @@ function MoveStudyDialog(props: Props) {
...
           control={control}
-          rules={{ validate: validatePath({ allowEmpty: true }) }}
+          rules={{
+            validate: validatePath({
+              allowEmpty: true,
+              allowToStartWithSlash: false,
+            }),
+          }}
           label={t("global.path")}
-          placeholder={t("studies.movefolderplaceholder")}
```
sylvlecl, line 108: We should check with users that it's OK to not have this hint.
Maybe it could go into a tooltip or help dialog ?

sylvlecl (commented):
Some duplication between copy and move but seems OK to me !

We should still check with users that it's OK to remove the hint for slashes.

webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
```diff
@@ -77,6 +83,24 @@ function CopyStudyDialog({ study, open, onClose }: Props) {
               validate: validateString({ allowEmpty: true }),
             }}
           />
+          <StringFE
+            name="destinationFolder"
+            label={t("global.path")}
```
sylvlecl: I think here we should use "Directory" or "Folder".
"Path" is less clear for users, and it could be considered wrong, the path would be folder + name.

Same in the move form.

webapp/src/components/App/shared/studies/dialogs/MoveStudyDialog.tsx
skamril: ok for help message

webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
skamril: ok for "Folder"

sylvlecl (approved):

PR #2519, opened 2025-06-04
fix(link): handle legacy commands

MartinBelthle:


antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -43,11 +44,31 @@ class AbstractLinkCommand(ICommand, metaclass=ABCMeta):
 
     area1: str
     area2: str
-    parameters: Optional[Dict[str, Any]] = None
+    parameters: LinkUpdate
```
sylvlecl: I know they will be the same for now, but we'd better create a `LinkCreation` class for consistency sake

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -105,7 +131,7 @@ def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener]
         if study_data.link_exists(self.area1, self.area2):
             return command_failed(f"Link between '{self.area1}' and '{self.area2}' already exists")
 
-        link = parse_link(self.parameters or {}, self.area1, self.area2)
+        link = parse_link(self.parameters.model_dump(mode="json", exclude_none=True) or {}, self.area1, self.area2)
```
sylvlecl: `parse_link` must not be used here, it should be used only to create a `Link` from INI files in the DAO, instead we must have a business method `create_link` in the model file which creates a `Link` from a `LinkCreation` object.

We also need to restore this in `create_cluster` (I see this was changed in the DAO PR)

antarest/study/storage/variantstudy/model/command/create_link.py
```diff
@@ -43,11 +44,31 @@ class AbstractLinkCommand(ICommand, metaclass=ABCMeta):
...
+    def _validate_parameters(cls, values: Dict[str, Any], info: ValidationInfo) -> Dict[str, Any]:
+        if "parameters" not in values:
+            values["parameters"] = LinkUpdate()
+
+        elif isinstance(values["parameters"], dict):
+            parameters = values["parameters"]
+            if info.context:
+                version = info.context.version
+                if version < 2:
+                    parameters.pop("area1", None)
```
sylvlecl: Actually I think we don't need to remove area1 and area2: those fields were only introduced temporarily with the DAO merge, they are not present in releases

sylvlecl (changes requested):
Request to change some implementation details

antarest/study/storage/variantstudy/model/command/create_link.py
MartinBelthle: They will indeed be the same class. And I don't think it would be better as this AbstractCommand is also the parent of the `UpdateLink` command. 
So either I do nothing or I remove this `inheritance` (first solution is easy :) but the 2nd could harmonize the links code with the thermal, renewable ... one )

antarest/study/storage/variantstudy/model/command/create_link.py
MartinBelthle: I changed the `parse` to `create` for thermal inside the renewable DAO so I won't touch this inside this PR.
But I'll do it for the links

sylvlecl (approved):

PR #2521, opened 2025-06-05
feat(ui-debug): update style of JSON view

skamril:
add overflow

### DIFF
--- a/webapp/src/components/common/JSONEditor/index.tsx
+++ b/webapp/src/components/common/JSONEditor/index.tsx
@@ -214,17 +214,16 @@ function JSONEditor({
   ////////////////////////////////////////////////////////////////
 
   return (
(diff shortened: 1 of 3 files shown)


webapp/src/components/common/JSONEditor/index.tsx
```diff
@@ -214,17 +214,17 @@ function JSONEditor({
...
-        sx={mergeSxProp(
-          {
-            overflow: "auto", // Fix when parent use `flex-direction: "column"`
-          },
-          sx,
-        )}
         ref={ref}
+        sx={{
+          // Allow to keep the menu visible when parent has scroll
+          overflow: "auto",
```
smailio: on my machine it works without this overflow isn't redundant ?

PR #2522, opened 2025-06-05
feat(matrix): add advanced filtering and bulk editing system

hdinia:
ANT-1300
Introduces a comprehensive filtering and editing system for matrix data, enabling users to precisely select data subsets and apply bulk operations.


webapp/package.json
```diff
@@ -126,7 +126,7 @@
     "vitest": "2.1.1"
   },
   "engines": {
-    "node": "22.13.0"
+    "node": ">=22.13.0"
```
skamril: @sylvlecl Does it follow the prod version?

webapp/public/locales/en/main.json
```diff
@@ -120,6 +120,46 @@
...
+    "april": "April",
+    "may": "May",
+    "june": "June",
+    "july": "July",
+    "august": "August",
+    "september": "September",
+    "october": "October",
+    "november": "November",
+    "december": "December",
+    "shortMonday": "Mon",
```
skamril: Maybe use cascade for "short" also:

```json
"short": {
  "monday": "Mon",
}
```

skamril:
The filter button is not displayed correctly in the Results view:

image

webapp/public/locales/en/main.json
```diff
@@ -838,5 +878,71 @@
...
+      "rowsFilter": "Rows Filter",
+      "columns": "Columns",
+      "rows": "Rows",
+      "type": "Filter type",
+      "range": "Range",
+      "modulo": "Modulo",
+      "list": "List",
+      "min": "Minimum",
+      "max": "Maximum",
+      "listValues": "Values (comma separated)",
```
skamril: Display "comma separated" in helper text instead of the label:

image

to

image

webapp/src/components/common/Matrix/components/FilterableMatrixGrid.tsx
```diff
@@ -0,0 +1,156 @@
...
+  rowHeaders,
+  onCellEdit,
+  onMultipleCellsEdit,
+  readOnly = true,
+  showPercent,
+  showStats = true,
+  timeFrequency,
+}: FilterableMatrixGridProps) {
+  // Initialize filter preview state
+  const [filterPreview, setFilterPreview] = useState<{
```
skamril: This state is not used except in the `useEffect`

skamril:
Use comment blocks in React component

webapp/src/components/common/Matrix/components/MatrixFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const handleTypeChangeEvent = (e: SelectChangeEvent) => {
+    handleTypeChange(e.target.value as FilterType);
+  };
+
+  const handleOperatorChangeEvent = (operator: FilterOperatorType) => {
+    handleOperatorChange(operator);
+  };
+
+  const handleRangeChange = (newValues: [number, number]) => {
+    setFilter({
```
skamril: Tip: use Immer with all `setFilter`:

```typescript
const handleRangeChange = ([min, max]: [number, number]) => {
  setFilter(
    produce((draftFilter) => {
      draftFilter.columnsFilter.range = { min, max };
    }),
  );
};
```

webapp/src/components/common/Matrix/components/MatrixFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,135 @@
...
+        )}
+
+        {filter.columnsFilter.type === FILTER_TYPES.LIST && (
+          <ListFilterControl
+            inputValue={inputValue}
+            selectedValues={filter.columnsFilter.list || []}
+            operator={filter.columnsFilter.operator}
+            onInputChange={handleListChange}
+            onKeyPress={handleKeyPress}
+            onAddValue={() => addValueToList()}
```
skamril: You can use point-free here:

```
onAddValue={addValueToList}
onAddValues={addValuesToList}
onRemoveValue={removeValueFromList}
onOperatorChange={handleOperatorChangeEvent}
onClearAll={clearAllValues}
```

webapp/src/components/common/Matrix/components/MatrixFilter/Operations.tsx
```diff
@@ -0,0 +1,113 @@
...
+  type SelectChangeEvent,
+  TextField,
+} from "@mui/material";
+import { useTranslation } from "react-i18next";
+import { Operation } from "../../shared/constants";
+import { useOperationControls } from "./hooks/useOperationControls";
+import { DESIGN_TOKENS, FORM_STYLES } from "./styles";
+import type { OperationsProps } from "./types";
+
+function Operations({ filter, setFilter, onApplyOperation }: OperationsProps) {
```
skamril, line 32: No "Filter" suffix?

webapp/src/components/common/Matrix/components/MatrixFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,135 @@
...
+        sx={ACCORDION_STYLES.summary}
+      >
+        <Box sx={{ display: "flex", alignItems: "center", gap: 1, flex: 1 }}>
+          <Typography sx={TYPOGRAPHY_STYLES.sectionTitle}>
+            {t("matrix.filter.columnsFilter")}
+          </Typography>
+        </Box>
+      </AccordionSummary>
+      <AccordionDetails sx={ACCORDION_STYLES.details}>
+        <FormControl fullWidth size="small" sx={{ mb: DESIGN_TOKENS.spacing.lg }}>
```
skamril, line 92: Use SelectFE

skamril:
Use our filter editors for matrix filters, they work without react-hook-form. Its better that using directly MUI, if a modification is made to a field it will be applied everywhere, and it permits to have less code (like for Select).

webapp/src/components/common/Matrix/components/MatrixFilter/Operations.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const handleValueChangeEvent = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const newValue = Number.parseFloat(e.target.value) || 0;
+    handleValueChange(newValue);
+  };
+
+  return (
+    <Box>
+      <Box sx={{ ...FORM_STYLES.sideBySideContainer, mt: 3 }}>
+        <Box sx={FORM_STYLES.responsiveContainer}>
+          <FormControl size="small" sx={FORM_STYLES.sideBySideFormControl}>
```
skamril, line 54: Use SelectFE

webapp/public/locales/en/main.json
skamril: Or maybe the tooltip at the right is enough, and you can remove this message completely.

skamril:
Issue with month days range (I have 14 days on the matrice so maybe this is the list filter to fix):

chrome-capture-2025-6-30

webapp/src/components/common/Matrix/components/MatrixFilter/Operations.tsx
```diff
@@ -0,0 +1,113 @@
...
+              <MenuItem value={Operation.Abs} sx={FORM_STYLES.menuItem}>
+                {t("matrix.operation.absolute")}
+              </MenuItem>
+            </Select>
+          </FormControl>
+
+          {filter.operation.type !== Operation.Abs && (
+            <Box sx={{ flex: 1 }}>
+              <TextField
+                label={t("matrix.filter.value")}
```
skamril, line 87: Use NumberFE

webapp/src/components/common/Matrix/components/MatrixFilter/constants.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  const options = TIME_FREQUENCY_INDEXING_MAP[timeFrequency];
+  return options?.[0] ?? TIME_INDEXING.DAY_OF_MONTH;
+};
+
+/**
+ * Temporal filtering options with localization keys and descriptions.
+ * Each option defines a specific way to filter time-based data.
+ */
+export const TEMPORAL_OPTIONS: readonly TemporalOption[] = [
```
skamril: `readonly TemporalOption[]` is for the array not the items, replace with `Array<Readonly<TemporalOption>>`

webapp/src/components/common/Matrix/components/MatrixFilter/RowFilter.tsx
```diff
@@ -0,0 +1,304 @@
...
+            aria-label={t("matrix.filter.removeFilter")}
+          >
+            <DeleteIcon fontSize="small" sx={{ width: 13, height: 13 }} />
+          </IconButton>
+        )}
+      </AccordionSummary>
+
+      <AccordionDetails sx={ACCORDION_STYLES.details}>
+        <Box sx={FORM_STYLES.sideBySideContainer}>
+          <FormControl size="small" sx={{ flex: 1, ...FORM_STYLES.sideBySideFormControl }}>
```
skamril, line 242: Use two SelectFE

webapp/src/components/common/Matrix/components/MatrixFilter/components/ListFilterControl.tsx
```diff
@@ -0,0 +1,280 @@
...
+  ////////////////////////////////////////////////////////////////
+  // JSX
+  ////////////////////////////////////////////////////////////////
+
+  return (
+    <>
+      <Box sx={{ ...FORM_STYLES.sideBySideContainer }}>
+        <Box sx={FORM_STYLES.responsiveContainer}>
+          {onOperatorChange && (
+            <FormControl size="small" sx={FORM_STYLES.sideBySideFormControl}>
```
skamril, line 170: Use SelectFE

webapp/src/components/common/Matrix/components/MatrixFilter/components/ListFilterControl.tsx
```diff
@@ -0,0 +1,280 @@
...
+                </MenuItem>
+                <MenuItem value={FILTER_OPERATORS.RANGE} sx={FORM_STYLES.menuItem}>
+                  {t("matrix.filter.operator.range")}
+                </MenuItem>
+              </Select>
+            </FormControl>
+          )}
+
+          <Box sx={{ flex: 1 }}>
+            <TextField
```
skamril, line 202: Use StringFE

webapp/src/components/common/Matrix/components/MatrixGrid/index.tsx
```diff
@@ -112,7 +217,17 @@ function MatrixGrid({
...
+        const [visibleCol, visibleRow] = edit.location;
+        const dataRow = getDataRowIndex(visibleRow);
+
+        // When in preview mode, we need to map the visible column index to the original column index
+        const originalCol = filterPreview.active
+          ? columns.findIndex((col) => col.id === visibleColumns[visibleCol].id)
+          : visibleCol;
+
+        const mappedLocation: Item = [originalCol, dataRow];
+        const dataCoordinates = gridToData(mappedLocation);
```
skamril, line 230: `filter(Boolean)` at the end.

webapp/src/components/common/Matrix/components/MatrixFilter/styles.ts
```diff
@@ -0,0 +1,361 @@
...
+  label: {
+    fontSize: DESIGN_TOKENS.fontSize.sm,
+  },
+  value: {
+    display: "block",
+    fontSize: DESIGN_TOKENS.fontSize.md,
+  },
+} as const satisfies Record<string, SxProps<Theme>>;
+
+export const OPERATION_STYLES = {
```
skamril, line 344: Some const are not used

skamril:
I have lag when adding a filter on a big matrice:

chrome-capture-2025-6-30 (1)

skamril:
Maybe can you use Form component with auto submit to reduce the amont of code.

webapp/src/utils/date/constants.ts
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+export const SUPPORTED_LOCALES = ["en", "fr"] as const;
+export type SupportedLocale = (typeof SUPPORTED_LOCALES)[number];
+
+export const WEEKDAY_NAMES = {
```
skamril, line 18: Why not using i18n keys?

webapp/src/utils/date/dateUtils.ts
```diff
@@ -0,0 +1,217 @@
...
+export function getHourOfYear(date: Date): number {
+  return (getDayOfYear(date) - 1) * 24 + getHours(date) + 1;
+}
+
+/**
+ * Get the current locale from language settings
+ *
+ * @returns SupportedLocale
+ */
+export function getCurrentLocale(): SupportedLocale {
```
skamril: There is `getCurrentLanguage` in i18nUtils

webapp/src/utils/date/dateUtils.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+/**
+ * Find month index from a string (1-based)
+ *
+ * @param dateStr - The string to search in
+ * @param locale - The locale to use
+ * @returns Month index (1-12) or null if not found
+ */
+export function findMonthIndex(dateStr: string, locale: SupportedLocale = "en"): number | null {
```
skamril, line 120: I think it's better to add a `parse` method in `TIME_FREQUENCY_CONFIG`, and works with `Date` object (`date.getMonth()`).

skamril (changes requested):

webapp/src/utils/date/dateUtils.ts
skamril: This message is not only for this function, but all the functions that works with custom string date

skamril: Or maybe create a cell renderer for Date to use in glide-data-grid, that allows to display a specific date format and keep the real Date object in the grid data.

webapp/public/locales/en/main.json
hdinia: Yes I agree that it's redundant with the Tooltip content, will remove it

webapp/src/components/common/Matrix/components/FilterableMatrixGrid.tsx
hdinia: done thanks

webapp/src/components/common/Matrix/components/MatrixFilter/ColumnFilter.tsx
hdinia: didnt think about Immer but it's a good use case for it, I will update all states updates to avoid nested spread operators

webapp/package.json
sylvlecl: Indeed I think it's better to stick with the version we use for the prod build, for now

webapp/src/components/common/Matrix/components/MatrixFilter/Operations.tsx
hdinia: Since the component is only used within the `MatrixFilter` module, its current name is already scoped and contextualized by its location.

webapp/src/components/common/Matrix/components/MatrixFilter/constants.ts
hdinia: I used `readonly TemporalOption[]` with the `as const` assertion  to prevent any modifications to this constant that's intended to be immutable.

with `Array<Readonly<TemporalOption>>`
```ts
// This would now be allowed -> could cause bugs
TEMPORAL_OPTIONS.push({ value: 'new', label: 'new' });
```

When needed, I create a copy with the spread operator to work with it and filter and operate on copies, not the original

so it would be `ReadonlyArray<Readonly<TemporalOption>>` or like I did with the `as const` assertion

hdinia: maybe we can a type util for this case 
```ts
type DeepReadonly<T> = {
  readonly [P in keyof T]: DeepReadonly<T[P]>;
};
```

webapp/src/components/common/Matrix/components/MatrixFilter/styles.ts
hdinia: ok thanks will check

webapp/src/components/common/Matrix/components/MatrixGrid/index.tsx
hdinia: `gridToData()` can return `null`. The function signature is `([col, row]: Item) => Item | null`, and calling `.filter()` on `null` would throw an error.

Filtering coordinates doesn't make sense: Even if `dataCoordinates` is a valid tuple like `[0, 5]`, calling `.filter(Boolean)` would remove any `0` values, which is wrong because:
`0` is a valid column and row index

You'd end up with incomplete coordinates like `[5]` instead of `[0, 5]`

The existing null check is correct, the current pattern handles the `Item | null` return properly

webapp/package.json
hdinia: We mainly use it for built-time tooling not runtime, so their is minimal to no risk with that change

Is the prod version fixed to strict 22.13.0 or it allows minor versions and patches ? Maybe ` "^22.x.x"` is better it would limit the risk of major version change but allow more flexibility for us, what do you think ? @skamril @sylvlecl

webapp/src/utils/date/dateUtils.ts
hdinia: forgot to replace it, thanks

webapp/src/utils/date/dateUtils.ts
hdinia: I've actually found a cleaner approach to handle this:

Instead of the current flow:
- Get dates from backend as strings → format them for display → retrieve these formatted strings → re-parse them...

The new approach would be:
- Get dates from backend → create two separate arrays:
  1. One for display (formatted strings based on temporality, matrix type, etc.)
  2. One with parsed Date objects
- Work directly with the Date objects array for filtering logic

This separates display concerns from data manipulation and avoids the circular dependency of formatting/parsing.

However, since this is a critical part of the matrix functionality, I'd prefer to implement this refactoring in the next PR along with the UTC display fix. This way we can properly test the changes without risking introducing bugs in the current PR.

I'll create a follow-up PR that addresses both:
1. The Date object refactoring 
2. The UTC timezone display issues

hdinia: And if possible use a CustomCell to handle Dates (instead of 2 arrays)

webapp/src/components/common/Matrix/components/MatrixFilter/ColumnFilter.tsx
hdinia: done

webapp/src/components/common/Matrix/components/MatrixGrid/index.tsx
skamril: I'm talking about this line:

```typescript
.filter((update): update is GridUpdate => update !== null);
```

webapp/package.json
sylvlecl: The prod version is in our hands, we can update it when we decide it actually.

I guess it's indeed better to allow patch versions even if it differs from the version used for prod build, the risk of breaking the build should be very limited and it will allow to benefit from fixes, I am not as sure for minor versions ?

sylvlecl: It looks like minor version may introduce "new features", not clear to me if it can break anything

webapp/src/utils/date/dateUtils.ts
skamril: A JIRA ticket will be created

skamril:
> I have lag when adding a filter on a big matrice:
> 
> chrome-capture-2025-6-30 (1) [ chrome-capture-2025-6-30 (1) ](https://private-user-images.githubusercontent.com/1954121/460625399-1c59d5fe-0a8a-4b0a-b023-e16969e15e7e.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NTE4ODIyMDEsIm5iZiI6MTc1MTg4MTkwMSwicGF0aCI6Ii8xOTU0MTIxLzQ2MDYyNTM5OS0xYzU5ZDVmZS0wYThhLTRiMGEtYjAyMy1lMTY5NjllMTVlN2UuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDcwNyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA3MDdUMDk1MTQxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9OGVjODE0MjNmYjdhZjE5YWQ3MDE3NTYzMjc3YjJkMGI5ODU3NjVhNDRmZDc0OWUwYzYyMzcxMzg2YWU0YmY3YiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.octQ5VEb0co5BpVbYIIUSZT3Xka_wIXAzt0sKUW2Qo4) [ ](https://private-user-images.githubusercontent.com/1954121/460625399-1c59d5fe-0a8a-4b0a-b023-e16969e15e7e.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NTE4ODIyMDEsIm5iZiI6MTc1MTg4MTkwMSwicGF0aCI6Ii8xOTU0MTIxLzQ2MDYyNTM5OS0xYzU5ZDVmZS0wYThhLTRiMGEtYjAyMy1lMTY5NjllMTVlN2UuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDcwNyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA3MDdUMDk1MTQxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9OGVjODE0MjNmYjdhZjE5YWQ3MDE3NTYzMjc3YjJkMGI5ODU3NjVhNDRmZDc0OWUwYzYyMzcxMzg2YWU0YmY3YiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.octQ5VEb0co5BpVbYIIUSZT3Xka_wIXAzt0sKUW2Qo4)

A JIRA ticket will be created

webapp/src/components/common/Matrix/components/MatrixFilter/constants.ts
skamril: True, but `readonly TemporalOption[]` don't protect the items:

`TEMPORAL_OPTIONS[0].label = "new label";`

hdinia: `ReadonlyArray<Readonly<TemporalOption>>` is better then

skamril (approved):

PR #2526, opened 2025-06-11
fix(launcher): job owner launches the unarchiving task

MartinBelthle:


antarest/launcher/service.py
```diff
@@ -542,12 +546,20 @@ def _import_output(
...
                                 concat_files_to_str(log_paths),
                             )
-                    return self.output_service.import_output(
-                        study_id,
-                        final_output_path,
-                        output_suffix,
-                        job_launch_params.auto_unzip,
-                    )
+
+                    if job_launch_params.auto_unzip and job_owner_id:
```
sylvlecl: We should not check autounzip here, it's simpler and makes more sense to restore the user context in all cases

PR #2531, opened 2025-06-17
fix(output): reduce memory consumption of output download

sylvlecl:
Reduces memory used by using numpy arrays instead of plain python lists.

Divides memory usage by around 3, and speeds up by around 4 according to dev env tests.

[ANT-3377]


MartinBelthle (dismissed):
Cool even though you broke some tests

MartinBelthle (approved):
I almost push the same code thank god github is smart and said I was late compared to origin

sylvlecl:
> I almost push the same code thank god github is smart and said I was late compared to origin

Woups, don't hesitate if you have any improvement compared to what I pushed.

TheoPascoli (approved):

PR #2532, opened 2025-06-18
fix(desktop): clean deleted studies desktop mode

smailio:
Scanning is disabled in desktop mode, but we still need to remove studies from the database if they no longer exist on disk.

We shouldn't rely on a full scan for this, as scanning all folders would be too resource-intensive.

Instead, we'll iterate over the studies that have already been scanned and check whether each one still exists on disk.


antarest/study/storage/rawstudy/watcher.py
```diff
@@ -237,3 +240,34 @@ def scan(
...
+        recursive scan of the filesystem. Instead, it checks previously scanned studies
+        and deletes those that are no longer present on disk.
+
+        The operation is protected by a file lock to prevent concurrent access, and logs
+        the time taken to complete the cleanup.
+
+        Raises:
+            CleanDisabled: If the application is not running in desktop mode.
+        """
+        if not self.config.desktop_mode:
```
maugde, line 260: is or will the `clean_desktop_studies()` method be called elsewhere than on line 116?

smailio: For now only needed on line 119, unless a future requirement involves exposing an endpoint for users to trigger the cleanup manually.

maugde: is it necessary to test twice the same condition ?

antarest/study/service.py
```diff
@@ -1060,6 +1061,28 @@ def sync_studies_on_disk(
...
+    def delete_missing_studies(self) -> None:
+        """
+        Removes from the database any non-archived desktop studies whose folders
+        no longer exist on disk. Used to clean up orphaned study entries.
+        """
+        all_studies = self.repository.get_all_raw()
+        desktop_studies = (
+            study
+            for study in all_studies
+            if study.workspace != DEFAULT_WORKSPACE_NAME and isinstance(study, RawStudy) and not study.archived
```
TheoPascoli, line 1074: What happens if those study have one variant ?

antarest/study/service.py
```diff
@@ -1060,6 +1061,28 @@ def sync_studies_on_disk(
...
+
+        def get_path(study: RawStudy) -> Path:
+            wp = self.config.get_workspace_path(workspace=study.workspace)
+            return wp / str(study.folder)
+
+        missing_studies = (study for study in desktop_studies if not is_study_folder(get_path(study)))
+
+        # delete orphan studies on database
+        for study in missing_studies:
+            self.repository.delete(study.id)
```
TheoPascoli: I think you can pass all the studies at once according to the signature of the delete method. So we perform only one commit

TheoPascoli (commented):
Hello, nice PR ! Just one question for me

antarest/study/service.py
smailio: We can only create variants for managed studies, no ? In which case the clean would ignore those

antarest/study/storage/rawstudy/watcher.py
smailio: To be honest, no. I designed it this way to be safe for future changes, like if we're later asked to add a button to trigger this cleanup manually.

TheoPascoli (approved):

PR #2534, opened 2025-06-18
feat(ui-studies): update name and folder fields validation

skamril:
extra: add a snackbar to open a study after its creation

ANT-3375


webapp/src/utils/validation/string.ts
```diff
@@ -21,7 +21,12 @@ interface StringValidationOptions {
   isCaseSensitive?: boolean;
   allowEmpty?: boolean;
   allowSpecialChars?: boolean;
-  specialChars?: string;
+  specialChars?:
+    | string
+    | {
+        chars: string;
+        mode: "allow" | "deny";
+      };
```
hdinia, line 29: optional: could be cleaner with a dedicated type

hdinia (approved):

PR #2535, opened 2025-06-19
fix(study-version): bug parsing study version

TheoPascoli:
ValueError: invalid literal for int() with base 10: '9.2'

### DIFF
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -2267,7 +2267,7 @@ class StudyService:
             path,
             with_index=with_index,
             with_header=with_header,
-            study_version=StudyVersion.parse(study.version),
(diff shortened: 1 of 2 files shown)


antarest/study/service.py
```diff
@@ -2267,7 +2267,7 @@ def get_matrix_with_index_and_header(
             path,
             with_index=with_index,
             with_header=with_header,
-            study_version=int(study.version),
+            study_version=StudyVersion.parse(study.version),
```
MartinBelthle: You can put `study_interface.version` to avoid doing this parsing

MartinBelthle (changes requested):

PR #2538, opened 2025-06-23
fix(ui-study,model): update delete message

skamril:
- message is different for study with or without variants
- fix with GroupDTO

ANT-2648


webapp/src/components/App/Singlestudy/NavHeader/index.tsx
```diff
@@ -27,7 +27,7 @@ interface Props {
 
 function NavHeader({ study, parentStudy, variantTree, isExplorer }: Props) {
   const tree = variantTree && findNodeInTree(study.id, variantTree);
-  const variantNb = tree ? countDescendants(tree) : 0;
+  const variantNb = tree && countDescendants(tree);
```
hdinia: The change from `tree ? countDescendants(tree) : 0` to `tree && countDescendants(tree)` makes `variantNb` potentially `undefined`, requiring defensive checks like `typeof variantNb === "number"`.

The previous ternary operator guaranteed `variantNb` was always a number, eliminating type guards and simplifying the code.

Any specific reason for this change? The ternary approach seems cleaner.

skamril: Because if `tree` is undefined we cannot have the number of variants, so 0 may be a lie.

hdinia: Ok I see why you changed it, but it seems weird to me that tree can be undefined

antarest/login/model.py
```diff
@@ -67,7 +67,7 @@ class UserCreateDTO(AntaresBaseModel):
 
 
 class GroupDTO(AntaresBaseModel):
-    id: Optional[str] = None
```
sylvlecl: Actually this changes the contract of the group creation endpoint:
it will now fail if we don't provide an ID.

We need either to change this endpoint to use another class where id is optional, and ensure we create an ID (I think this is automatic when we save to database, it creates a UUID), or ensure that we call the endpoint with an ID everywhere.

(note: this is why the python test fail, but actually the calls from the group creation dialog will also fail now)

webapp/src/components/App/Singlestudy/NavHeader/index.tsx
sylvlecl: Shouldn't we throw an error instead in `findNodeInTree` when we request a study which is not in the tree ?
This would remove the `undefined` case

sylvlecl (changes requested):
Good in principle, but we need to fix the group creation endpoint

webapp/src/components/App/Singlestudy/NavHeader/index.tsx
skamril: I made a mistake I restored the previous ternary @hdinia

antarest/login/model.py
skamril: I have removed this from this PR to merge it rapidly. Must be fixed in another one.

webapp/src/components/App/Singlestudy/NavHeader/index.tsx
skamril: I can keep the previous code, @hdinia was true.

@sylvlecl Normally it's not possible to have `undefined` here, we using the ternary instead of forcing the type with TypeScript.

webapp/src/components/App/Singlestudy/index.tsx
```diff
@@ -49,10 +50,13 @@ function SingleStudy({ isExplorer }: Props) {
     const parents = await getVariantParents(studyId);
     const parentStudy = parents.length > 0 ? parents[0] : undefined;
 
-    const root = parents.length > 0 ? parents.at(-1) : study;
-    const variantTree = root ? await getVariantTree(root.id) : undefined;
+    const root = parents.length > 0 ? parents[parents.length - 1] : study;
```
hdinia, line 53: 👍🏼

hdinia (dismissed):

hdinia (approved):

PR #2539, opened 2025-06-23
fix(study): copied study's outputs not downloadable

TheoPascoli:
This PR improves the study copying functionality by also duplicating all job records associated with the source study.
Previously, when copying a study, only the study data and optionally its outputs were duplicated. Now, all job result records from the source study are also copied to the new study in the database. This ensures that the job history and simulation results are preserved in the copied study.
Each duplicated job receives a new UUID while maintaining all other properties from the original job (except for the study ID which is updated to reference the new study).


antarest/study/repository.py
```diff
@@ -430,7 +430,7 @@ def list_duplicates(self) -> List[Tuple[str, str]]:
         Get list of duplicates as tuples (id, path).
         """
         session = self.session
-        subquery = session.query(Study.path).group_by(Study.path).having(func.count() > 1).subquery()
+        subquery = session.query(Study.path).group_by(Study.path).having(func.count() > 1)
```
TheoPascoli, line 433: This has nothing to do with the PR. It is just a fix of a SQLAlchemy warning that happens at the start of the app

antarest/study/repository.py
```diff
@@ -430,7 +430,7 @@ def list_duplicates(self) -> List[Tuple[str, str]]:
         Get list of duplicates as tuples (id, path).
         """
         session = self.session
-        subquery = session.query(Study.path).group_by(Study.path).having(func.count() > 1).subquery()
```
MartinBelthle, line 433: What does this change does ?

antarest/launcher/model.py
```diff
@@ -233,6 +233,13 @@ def __repr__(self) -> str:
             f" owner_id={self.owner_id!r})>"
         )
 
+    def __iter__(self) -> Iterator[Tuple[str, Any]]:
```
MartinBelthle: Not a huge fan of this and the fact that the service does the job. I think it would be better if the JobResult object did all by himself and just returned another JobResult object. the method could expect a study_id to do all by itself

antarest/study/service.py
```diff
@@ -1140,7 +1145,19 @@ def copy_task(notifier: ITaskNotifier) -> TaskResult:
...
+
+            jobs = self.job_result_repository.find_by_study(origin_study.id)
+
+            for job in jobs:
+                job_data = {k: copy.deepcopy(v) for k, v in job if k not in ["id", "study_id"]}
+                new_job = JobResult(**job_data)
+                new_job.study_id = study.id
+                new_job.id = str(uuid4())
+
+                self.job_result_repository.save(new_job)
```
MartinBelthle: I think we should introduce a method to save several jobs to avoid doing too much db calls

MartinBelthle (changes requested):
Even though it seems to work we should add tests

antarest/study/repository.py
TheoPascoli: I answered in a comment below

antarest/launcher/repository.py
```diff
@@ -37,6 +37,12 @@ def save(self, job: JobResult) -> JobResult:
         db.session.commit()
         return job
 
+    def save_all(self, jobs: List[JobResult]) -> List[JobResult]:
```
MartinBelthle: Minor.
Don't really see the point of return the list here. We do it inside the save method above but it doesn't make sense to me as we don't modify the object

antarest/study/service.py
```diff
@@ -1140,7 +1143,18 @@ def copy_task(notifier: ITaskNotifier) -> TaskResult:
             study = self.storage_service.get_storage(origin_study).copy(
                 origin_study, dest_study_name, group_ids, destination_folder, output_ids, with_outputs
             )
+
             self._save_study(study, group_ids)
+
+            jobs = self.job_result_repository.find_by_study(origin_study.id)
+
+            new_jobs = []
```
MartinBelthle: Minor: This could be done in comprehension:
`new_jobs = [job.copy_jobs_for_study(study.id) for job in jobs]`

antarest/study/service.py
```diff
@@ -1140,7 +1143,18 @@ def copy_task(notifier: ITaskNotifier) -> TaskResult:
             study = self.storage_service.get_storage(origin_study).copy(
                 origin_study, dest_study_name, group_ids, destination_folder, output_ids, with_outputs
             )
+
             self._save_study(study, group_ids)
+
+            jobs = self.job_result_repository.find_by_study(origin_study.id)
```
MartinBelthle: We should add a comment specifying why we're doing all this

antarest/study/service.py
```diff
@@ -1140,7 +1143,18 @@ def copy_task(notifier: ITaskNotifier) -> TaskResult:
             study = self.storage_service.get_storage(origin_study).copy(
                 origin_study, dest_study_name, group_ids, destination_folder, output_ids, with_outputs
             )
+
             self._save_study(study, group_ids)
+
```
MartinBelthle, line 1148: We discussed with Sylvain and we said we should only copy jobs related to outputs we want to copy.
This way we won't have issues with running jobs or failed jobs (cause we can't have their logs in such case) and it will be a bit clearer for the user

MartinBelthle (changes requested):

MartinBelthle (dismissed):

MartinBelthle:
One test is failing though

PR #2540, opened 2025-06-23
fix(workers): use gunicorn to autorepair load balanced workers

sylvlecl:
When using `--no-gunicorn` / `--multiple-ports` option to launch the server, we launch
multiple workers from a bash script.
The assumption is that a load balancer distributes requests between those workers.

When we introduced that behaviour, we lost the "autor repair" capabilities of 
gunicorn.
It does happen in production that workers are killed by OOM, therefore it gets 
important to restore that capability also in this launch mode.

In order to achieve this, we simply used gunicorn to launch each individual worker.

**Tested in integration environment:**
Killing one of the workers causes it to be restarted correctly.


scripts/start.sh
```diff
@@ -48,7 +48,15 @@ elif [ "$use_uvicorn" = true ]; then
...
   do
-    uvicorn antarest.wsgi:app --host 0.0.0.0 --port $((5000 + $i)) --log-level info --timeout-keep-alive 600 &
+    # we still use gunicorn in that case to restart workers in case they are killed,
+    # although each gunicorn instance has only one worker
+    gunicorn --worker-class=uvicorn.workers.UvicornWorker \
+             --bind=0.0.0.0:$((5000 + $i)) \
+             --workers=1 \
+             --log-level info \
+             --timeout 600 \
+             --keep-alive $((24 * 60 * 60)) \
```
MartinBelthle: Not familiar with all this but it seems you put a keep alive time higher than before.
From my understanding, it's used to know if the server we're trying to connect is able to answer. In this context, putting such a high value is pretty weird no ?
The gunicorn doc says we can put it higher than 1-5 sec as it's behind a load balancer, seems you took it literally

sylvlecl: To be honest I just used the value from the previously used `gunicorn.py` config.
I'll look into it.

sylvlecl: Actually we did not even enable keepalive between our nginx and the server, so it really does not make sense to have that high value.
I'll remove it

MartinBelthle (approved):

PR #2541, opened 2025-06-23
feat(matrix): add get matrices

mehdiwahada:


antarest/matrixstore/service.py
```diff
@@ -172,6 +181,14 @@ def check_dataframe_compliance(df: pd.DataFrame) -> None:
                 )
 
 
+def convert_matrix_into_model(matrix: Matrix) -> MatrixModel:
```
sylvlecl: we can make this method private (starting with _) since it's used only in this module

conf/requirements.txt
```diff
@@ -0,0 +1,51 @@
+Antares-Launcher==1.4.2
```
sylvlecl: I think this  file should be removed

antarest/matrixstore/matrix_model.py
```diff
@@ -0,0 +1,23 @@
...
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+
+from datetime import datetime
+
+from antarest.core.serde import AntaresBaseModel
+
+
+class MatrixModel(AntaresBaseModel, extra="forbid", populate_by_name=True):
```
sylvlecl: After more thinking about the naming:
we can rename it `MatrixMetadataDTO` for consistency with other objects in that API:
- returned objects end with DTO, let's keep that convention for now
- "model" is a little too vague, metadata is more specific to what this class contains (metadata about the matrix)

Also, we can move it with other classes in `model.py`.

antarest/matrixstore/service.py
```diff
@@ -389,6 +406,23 @@ def get(self, matrix_id: str) -> pd.DataFrame:
...
+    @override
+    def get_matrices(self) -> List[MatrixModel]:
+        """
+        Get a list of matrix objects from the database
+        Returns:#
+            A list of matrices of the repository
+        """
+
+        matrices = self.repo.get_matrices()
+        pydantic_friendly_matrices = []
```
sylvlecl: You may shorten this with list comprehension:
```python
return [convert_matrix_into_model(m) for m in matrices]
```
After renaming the class, we may rename the conversion method to something like "_matrix_to_dto" for consistency with the class name

tests/matrixstore/test_service.py
```diff
@@ -124,6 +125,37 @@ def test_get(self, matrix_service: MatrixService) -> None:
             with pytest.raises(MatrixNotFound, match=f"Matrix {missing_hash} doesn't exist"):
                 matrix_service.get(missing_hash)
 
+    def test_get_matrices(self, matrix_service: MatrixService):
+        parent = resource_path.parent
+        matrices = os.listdir(parent)
```
sylvlecl: Actually, I think it would be better to list explicitly the names of the matrices we want to import, otherwise the test will be too dependent on changes in this folder

tests/matrixstore/test_service.py
```diff
@@ -124,6 +125,37 @@ def test_get(self, matrix_service: MatrixService) -> None:
...
+        expected_matrices_id.sort()
+        key_word = "all"
+        with db():
+            for matrix in matrices:
+                if key_word in matrix:
+                    matrix_path = parent / matrix
+                    mat = pd.read_csv(matrix_path)
+                    matrix_service.create(mat)
+
+            for matrix in matrix_service.get_matrices():
```
sylvlecl: you may shorten this with list comprehension:
```python
actual_matrices_id = [m.id for m in matrix_service.get_matrices()].sort()
```

tests/matrixstore/test_service.py
```diff
@@ -124,6 +125,37 @@ def test_get(self, matrix_service: MatrixService) -> None:
...
+                actual_matrices_id.append(matrix.id)
+
+            actual_matrices_id.sort()
+
+            assert len(actual_matrices_id) == 7
+
+            assert actual_matrices_id == expected_matrices_id
```
sylvlecl: We should also have some assertions about the other data in the returned objects (heigh, width, etc):
actually you could have a dict of all expected values insetad of just the list of ids

antarest/matrixstore/service.py
MartinBelthle: Why not putting this as a method of the Matrix class ?

antarest/matrixstore/service.py
MartinBelthle: Could de done in comprehension and it would be easier to read with the method inside the Matrix class

Also not a huge fan of the variable name `pydantic_friendly_matrices`, IMO we should just put a doc inside the MatrixModel class or here, to explain why we're doing this

antarest/matrixstore/web.py
```diff
@@ -68,6 +69,12 @@ def create_by_importation(
         logger.info("Importing new matrix dataset")
         return service.create_by_importation(file, is_json=json)
 
+    @bp.get("/matrix", tags=[APITag.matrix], description="Return a list of matrices metadatas")
```
MartinBelthle: Not sure metadata takes an s

conf/requirements.txt
```diff
@@ -0,0 +1,51 @@
...
+antares-study-version==1.0.9
+antares-timeseries-generation~=0.1.5
```
MartinBelthle: This file should be deleted

tests/integration/test_matrix_blueprint.py
```diff
@@ -0,0 +1,41 @@
...
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+from starlette.testclient import TestClient
+
+
+def test_matrix(client: TestClient, admin_access_token: str) -> None:
+    client.headers = {"Authorization": f"Bearer {admin_access_token}"}
+
+    matrix = [[1, 2], [1, 7]]
```
MartinBelthle, line 20: We should probably create at least 2 matrices to ensure we're able to fetch more than one matrix

tests/matrixstore/test_service.py
MartinBelthle: This is a weird thing to do. If we're doing this we should at least document why we're doing this

tests/integration/test_matrix_blueprint.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+
+    matrix_id = res.json()
+
+    res = client.get("/v1/matrix")
+    assert res.status_code == 200
+    matrices = res.json()
+
+    existing_matrix = any(d.get("id") == matrix_id for d in matrices)
+    assert existing_matrix
```
sylvlecl: we need to check that the result also contains the other fields: height, width, creation date, and their values (except for the date since it will not be stable)

tests/matrixstore/test_service.py
```diff
@@ -124,6 +125,37 @@ def test_get(self, matrix_service: MatrixService) -> None:
...
+            "10d7ca7b8f935d013dd6304466668b87e25f51eb179b92bc995b6f2cc70641a7",
+            "d6b4ecdc42135e48eaa1d868ef535d396bc3953b7c3bf4f802d83964db58af77",
+            "e042c928e846ed944751d1d82edec4be25ee606816e7aefc4b64c9cb6808f2a5",
+            "ff7589e7632c1f8304949a27fa3ac86107d8ecf198980e67ae7ca062f9036789",
+            "2d62ca219846080b38e2ddf8d0f8f46a9fdeb7854234bfc751519b7502708b93",
+            "87e70e8b6af459b33be2dbd65d5c814a9225ae77112c9ea791d3abe92379334c",
+        ]
+        actual_matrices_id = []
+        expected_matrices_id.sort()
+        key_word = "all"
```
MartinBelthle: Same here, it's not clear why we're using this keyword, we should document it

MartinBelthle (changes requested):
Several remarks but overall the code seems okay

tests/integration/test_matrix_blueprint.py
```diff
@@ -0,0 +1,41 @@
...
+
+    res = client.get(f"/v1/matrix/{matrix_id}")
+    actual_matrix = res.json()
+    actual_width = actual_matrix["data"][0]
```
sylvlecl: Here you could check the whole content of `data` is as expected since it's small, not only width and height

tests/integration/test_matrix_blueprint.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+from starlette.testclient import TestClient
+
```
sylvlecl, line 15: We should move other matrix-related tests from `test_integration.py` to that new file, for consistency

sylvlecl (changes requested):
Nice, a few requests on implementation details

tests/integration/test_matrix_blueprint.py
```diff
@@ -0,0 +1,100 @@
...
+    res = client.get("/v1/matrix")
+    assert res.status_code == 200
+    matrices = res.json()
+
+    second_last_matrix = matrices[-2]
+    assert second_last_matrix["width"] == 2
+    assert second_last_matrix["height"] == 2
+    assert second_last_matrix["version"] == 2
+    assert second_last_matrix["id"] == second_last_matrix_id
```
sylvlecl, line 48: We should also check the following thing about the `created_at` field:
- it is present
- it respects the following format `2025-04-01 08:56:01.021543` for consistency with other places in the API (for example /api/v1/studies endpoint)

tests/integration/test_matrix_blueprint.py
```diff
@@ -0,0 +1,100 @@
...
+
+    last_matrix_id = res.json()
+
+    res = client.get("/v1/matrix")
+    assert res.status_code == 200
+    matrices = res.json()
+
+    second_last_matrix = matrices[-2]
+    assert second_last_matrix["width"] == 2
+    assert second_last_matrix["height"] == 2
```
sylvlecl, line 46: would be slightly better to have a not-square matrix so that width and height are different, to ensure that we don't confuse the 2

tests/matrixstore/test_service.py
```diff
@@ -124,6 +125,100 @@ def test_get(self, matrix_service: MatrixService) -> None:
...
+        with db():
+            for matrix in matrices:
+                mat = pd.read_csv(matrix)
+                matrix_service.create(mat)
+
+            actual_matrices = matrix_service.get_matrices()
+
+            assert len(actual_matrices) == len(expected_matrices)
+
+            assert actual_matrices[0].id == expected_matrices[0].id
```
sylvlecl: If we stick with this method for checking all fields except "created_at", we should just have a for loop on the list of matrices, so that we check all of them while not repeating the assertion code for all matrices.

Another possibility is to use `unittest.mock.ANY` for the created_at field:
```python
            MatrixMetadataDTO.model_construct(
                id="ff7589e7632c1f8304949a27fa3ac86107d8ecf198980e67ae7ca062f9036789",
                width=1,
                height=14,
                created_at=ANY,
                version=2,
            ),
```
And then just have an assertion on the list:
```python
assert actual_matrices == expected_matrices 
```

conf/requirements.txt
sylvlecl: Still to be deleted!

sylvlecl (changes requested):
A few more improvements on tests requested + file to be removed

sylvlecl (approved):

PR #2543, opened 2025-06-24
fix(groups): distinguish between group and creation DTO

sylvlecl:
Just clarifies the API contract regarding groups objects:
- ID is optional for creation
- ID is always defined in returned objects


antarest/login/model.py
```diff
@@ -67,10 +67,15 @@ class UserCreateDTO(AntaresBaseModel):
 
 
 class GroupDTO(AntaresBaseModel):
-    id: Optional[str] = None
+    id: str
     name: str
 
 
+class GroupCreation(AntaresBaseModel):
```
skamril: No 'DTO' suffix?

sylvlecl: I'll add it for consistency sake with other objects in that API, but it does not really have any added value to have that suffix here.

antarest/login/web.py
```diff
@@ -152,10 +152,10 @@ def groups_get_id(id: str, details: bool = False) -> Any:
             return HTTPException(status_code=404, detail=f"Group {id} not found")
 
     @bp.post("/groups", tags=[APITag.users], response_model=GroupDTO)
-    def groups_create(group_dto: GroupDTO) -> Any:
+    def groups_create(group_dto: GroupCreationDTO) -> Any:
```
MartinBelthle: We should put here that the response is GroupDTO and remove it from the line above

antarest/login/web.py
```diff
@@ -152,10 +152,10 @@ def groups_get_id(id: str, details: bool = False) -> Any:
...
         logger.info(f"Creating new group '{group_dto.name}'")
         group = Group(
-            id=escape(group_dto.id) if group_dto.id else None,
+            id=group_dto.id if group_dto.id else None,
```
MartinBelthle: Isn't it the same thing as `group_dto.id` ?  

I think you meant to put something else instead of the None as I don't see how it will change our code. If we have a group with None as an id in BD or if someone creates a group without giving an id we'll return None (even if you typed it as a string); I don't understand why mypy is quiet about that

sylvlecl: When we save to database, sqlalchemy creates a UUID.
Actually we cannot have nulls in database, since it's the primary key. See nullable=False in alembic.

MartinBelthle: We don't use `nullable=False` for the id that's why I said that but you're right it's a primary key so we don't need it

MartinBelthle: And I reiterate but shouldn't we put `id=group_dto.id` instead of `id=group_dto.id if group_dto.id else None,`, seems to me it the same thing but shorter

sylvlecl: Indeed I will change it.
For the ID in DB, we have this in alembic file, so it's indeed not nullable:
```python
    op.create_table('groups',
    sa.Column('id', sa.String(length=36), nullable=False),
```

MartinBelthle (approved):

PR #2546, opened 2025-06-24
feat(study): add editor field in study

TheoPascoli:


alembic/versions/a8c6e9d47b12_add_editor_field_to_study.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'a8c6e9d47b12'
+down_revision = '9ba7fc46d4a0'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
MartinBelthle, line 19: I'm not sure to understand why we put the field `editor` in the 2 classes. We did it for `author` but it doesn't make sense IMO.

I think it should only be present in additional_data

MartinBelthle (changes requested):

tests/storage/rawstudies/samples/v810/sample1/study.antares
```diff
@@ -4,4 +4,5 @@ caption = renewable-2-clusters-ts-prod-factor
 created = 1618413128
 lastsave = 1625583204
 author = Unknown
+editor = Unknown
```
MartinBelthle: Do we need to change this ? I think we shouldn't have to

Also we should add tests to ensure the code works correctly

Otherwise the src code seems good

MartinBelthle (changes requested):

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -103,13 +103,16 @@ def update_from_raw_meta(
...
+            if metadata.additional_data and metadata.additional_data.editor:
+                raw_meta["editor"] = metadata.additional_data.editor
+                study.tree.save(raw_meta, ["study", "antares"])
+
             metadata.name = raw_meta["caption"]
             metadata.version = str(raw_meta["version"])
             metadata.created_at = datetime.utcfromtimestamp(raw_meta["created"])
             metadata.updated_at = datetime.utcfromtimestamp(raw_meta["lastsave"])
 
             metadata.additional_data = self._read_additional_data_from_files(study)
```
MartinBelthle, line 115: We should also parse the editor field inside this method

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -103,13 +103,16 @@ def update_from_raw_meta(
         study = self.study_factory.create_from_fs(path, is_managed(metadata), study_id="")
         try:
             raw_meta = study.tree.get(["study", "antares"])
+            if metadata.additional_data and metadata.additional_data.editor:
+                raw_meta["editor"] = metadata.additional_data.editor
```
MartinBelthle, line 107: When can we enter this, I'm not  sure to understand as the point of this method is to add additional_data inside a study

tests/integration/test_integration.py
```diff
@@ -1280,6 +1282,58 @@ def zip_study(src_path: Path, dest_path: Path) -> None:
         == "Study 'A' could not be imported: AntaresWeb doesn't support the value 'hourly' for the flag 'hydro-pmax'"
     )
 
+    # 1. Create two users: 'creator' and 'importer'
```
MartinBelthle, line 1292: Really nice test. Perhaps we could separate it from the rest as the current test is already pretty long

MartinBelthle (changes requested):

antarest/study/storage/rawstudy/raw_study_service.py
TheoPascoli: During an import, the study.antares file is copy/paste without any modification. I must save it a this point otherwise the "editor" field is not modified with the importer name.
The metadata should contains the editor because I added it in StudyService.import_study

tests/integration/test_integration.py
TheoPascoli: You're right

antarest/study/storage/rawstudy/raw_study_service.py
TheoPascoli: yes

antarest/study/storage/rawstudy/raw_study_service.py
MartinBelthle: Okay I understand better now. But I think we're missing the case of a copy :/
And it would be better to put this little code inside a dedicated method with a doc to use when copying, importing or creating a study.
Also the method could do `author = self.get_user_name()` itself at the start

We'll also need to add a test case for the copy

MartinBelthle (changes requested):
Last change requested

antarest/study/storage/rawstudy/raw_study_service.py
TheoPascoli: I added a test case for the copy. I did not touch that code but I added the editor field during the copy as well.

TheoPascoli (commented):
I managed to set the editor field during creation, import, and copy — both in StudyAdditionalData and study.antares.
I can't factor it more given the current state of the code.

sylvlecl (commented):
Could be done in another PR, but we will also need to update the editor field every time the study is edited (basically, when commands are added, at least).

The logic should be similar to the update of `study.updated_at` field.
However, that one does not work 100% : it seems to work for variant studies, but not raw studies. It should be fixed too.
The update of those fields should be consistent with the emission of the DATA_EDITED event too, in principle.

MartinBelthle (approved):

PR #2555, opened 2025-06-26
build(dep): bump pydantic and fastapi

TheoPascoli:
Update to Pydantic v2.11.0 for enhanced performance


requirements.txt
```diff
@@ -15,9 +15,9 @@ antares-timeseries-generation==0.1.7
 
 # We prefer to add only the specific libraries we need for our project
 # and **manage their versions** for better control and to avoid unnecessary dependencies.
-fastapi~=0.110.3
+fastapi~=0.115.0
```
MartinBelthle: We could put 0.115.13 here to ensure we don't pick an older version

MartinBelthle (dismissed):
Nice :) Perhaps you wanted to name your PR bump instead of pump.
Also, it's not really a feature it should be tagged build I think

PR #2557, opened 2025-06-26
feat(sts): support additional constraints

MartinBelthle:
[ANT-3424]

- GET, PUT, POST and DELETE: `/studies/{uuid}/areas/{area_id}/storages/{storage_id}/additional-constraints`
- GET `/studies/{uuid}/areas/{area_id}/storages/{storage_id}/additional-constraints/{constraint_id}`
- New Table-mode view for additional constraints


antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,112 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+    NETTING = "netting"
+
+
+class AdditionalConstraintOperator(EnumIgnoreCase):
+    LESS = "less"
+    GREATER = "greater"
+    EQUAL = "equal"
+
+
+HOURS_TYPE: TypeAlias = list[list[int]] | list[int]
```
TheoPascoli: You should document this type, why it can be both, why you `type: ignore` the return value in `_hours_parser`, and why you serialize this field the way you do, especially with the early return `if isinstance(value[0], int) `

antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,112 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+    operator: AdditionalConstraintOperator = AdditionalConstraintOperator.LESS
+    hours: Hours = []
+    enabled: Optional[bool] = None
+
+
+class STStorageAdditionalConstraintUpdate(AntaresBaseModel):
+    """
+    Represents an update of a short-term storage additional constraint.
+
+    Only not-None fields will be used to update the constraint.
```
TheoPascoli, line 326: We didn't set default values for other `"creation"` model, is there a reason we have some here ?

antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,112 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+    GREATER = "greater"
+    EQUAL = "equal"
+
+
+HOURS_TYPE: TypeAlias = list[list[int]] | list[int]
+
+
+def hours_parser(value: str | HOURS_TYPE) -> HOURS_TYPE:
+    def _checks_compliance(x: Any) -> None:
+        if not isinstance(x, int) or (x < 0 or x > 168):
```
TheoPascoli: Minor: you can write `not (0 <= x <= 168) `

antarest/study/business/areas/st_storage_management.py
```diff
@@ -292,3 +318,130 @@ def duplicate_cluster(
...
+            area_id: The area ID of the short-term storage.
+            storage_id: The short-term storages ID.
+        """
+        return study.get_study_dao().get_st_storage_additional_constraints(area_id, storage_id)
+
+    def create_additional_constraints(
+        self,
+        study: StudyInterface,
+        area_id: str,
+        storage_id: str,
```
TheoPascoli, line 496: This field is not in the docstring

antarest/study/business/areas/st_storage_management.py
```diff
@@ -292,3 +318,130 @@ def duplicate_cluster(
...
+
+        Args:
+            study: The study object.
+            area_id: The area ID.
+            constraints: List of constraints to create.
+        """
+        created_constraints = [create_st_storage_constraint(c) for c in constraints]
+
+        # Checks we're not duplicating existing constraints
+        existing_constraints = study.get_study_dao().get_st_storage_additional_constraints(area_id, storage_id)
```
TheoPascoli, line 511: You should also check that we don't try to create multiple additional constraints with the same id

antarest/study/business/areas/st_storage_management.py
```diff
@@ -292,3 +318,130 @@ def duplicate_cluster(
...
+        existing_ids = {c.id for c in existing_constraints}
+        for constraint in created_constraints:
+            if constraint.id in existing_ids:
+                raise DuplicateSTStorageConstraintName(area_id, constraint.id)
+
+        # Apply the command
+        command = CreateSTStorageAdditionalConstraints(
+            area_id=area_id,
+            storage_id=storage_id,
+            constraints=constraints,
```
TheoPascoli, line 522: Can you pass the created_constraints instead ? And not create them again in the command ?

antarest/study/business/table_mode_management.py
```diff
@@ -250,6 +264,20 @@ def update_table_data(
...
             return data
+        elif table_type == TableModeType.ST_STORAGE_ADDITIONAL_CONSTRAINTS:
+            update_constraints_by_areas: STStorageAdditionalConstraintUpdates = collections.defaultdict(dict)
+            for key, values in data.items():
+                area_id, storage_id, constraint_id = key.split(" / ")
+                update_constraint = STStorageAdditionalConstraintUpdate.model_validate({"id": constraint_id, **values})
+                update_constraints_by_areas.setdefault(area_id, {}).setdefault(storage_id, []).append(update_constraint)
+            constraints_map = self._st_storage_manager.update_additional_constraints(study, update_constraints_by_areas)
+            data = {
+                f"{area_id} / {storage_id} {constraint.id}": constraint.model_dump(by_alias=True, exclude={"id"})
```
TheoPascoli: I think you're missin a "/" in the key

TheoPascoli (changes requested):
First pass in the code, very good job overall !

antarest/study/business/model/sts_model.py
TheoPascoli: I see that you removed the serializer while I was reviewing !

antarest/study/business/areas/st_storage_management.py
MartinBelthle: That's not the architecture design we chose for other commands, see `create_cluster`, `create_renewables_cluster` etc. so I prefer to stick with it

antarest/study/business/areas/st_storage_management.py
MartinBelthle: Mmmh indeed, smart

antarest/study/business/areas/st_storage_management.py
```diff
@@ -211,6 +230,13 @@ def delete_storages(
             area_id: The area ID of the short-term storage.
             storage_ids: IDs list of short-term storages to remove.
         """
+        # Checks storages are not referenced in any constraint
+        existing_constraints = study.get_study_dao().get_st_storage_additional_constraints_for_area(area_id)
+        for storage_id in storage_ids:
```
smailio: You might run into concurrency issues with this code, especially if a constraint is added while you're inside this loop. Perhaps it's assumed that this situation should never occur?

antarest/study/business/model/sts_model.py
MartinBelthle: You're right, I should put them inside the STStorageAdditionalConstraint object instead

antarest/study/business/model/sts_model.py
MartinBelthle: I found a way to remove the `type: ignore`, let me know if you need me to do more

antarest/study/business/areas/st_storage_management.py
```diff
@@ -292,3 +318,130 @@ def duplicate_cluster(
...
+        # Checks the constraint exist and builds the response.
+        new_constraints: dict[str, dict[str, list[STStorageAdditionalConstraint]]] = {}
+        existing_constraints = study.get_study_dao().get_all_st_storage_additional_constraints()
+        for area_id, value in update_constraints_by_areas.items():
+            if area_id not in existing_constraints:
+                raise AreaNotFound(area_id)
+
+            for storage_id, updated_constraints in value.items():
+                if storage_id not in existing_constraints[area_id]:
+                    raise STStorageNotFound(area_id, storage_id)
```
smailio: It might be worth performing all the checks before applying any updates. Otherwise, you could end up with a partially updated state.

antarest/study/dao/api/study_dao.py
```diff
@@ -217,3 +217,19 @@ def get_st_storage_cost_variation_injection(self, area_id: str, storage_id: str)
     @override
     def get_st_storage_cost_variation_withdrawal(self, area_id: str, storage_id: str) -> pd.DataFrame:
         return self._adaptee.get_st_storage_cost_variation_withdrawal(area_id, storage_id)
+
+    @override
+    def get_all_st_storage_additional_constraints(self) -> dict[str, dict[str, list[STStorageAdditionalConstraint]]]:
```
smailio: The return type is  complex. Imo  it’s worth introducing a type alias..

```

STSAdditionalConstraintsMap = Dict[str, Dict[str, List[STStorageAdditionalConstraint]]]
# area_id -> storage_id -> list of constraints
```

antarest/study/dao/file/file_study_st_storage_dao.py
```diff
@@ -261,6 +266,92 @@ def delete_storage(self, area_id: str, storage: STStorage) -> None:
...
+                storage_id, constraint = parse_st_storage_additional_constraint(key, value)
+                constraints.setdefault(storage_id, []).append(constraint)
+            return constraints
+        except ChildNotFoundError:
+            return {}
+
+    @override
+    def save_st_storage_constraint_matrix(self, area_id: str, constraint_id: str, series_id: str) -> None:
+        study_data = self.get_file_study()
+        study_data.tree.save(series_id, ["input", "st-storage", "constraints", area_id, f"rhs_{constraint_id}"])
```
smailio: rhs ?

MartinBelthle: Right hand side. It's the name of the matrix

antarest/study/business/areas/st_storage_management.py
MartinBelthle: Not sure to understand what's the issue, could you develop ?

antarest/study/business/areas/st_storage_management.py
MartinBelthle: I'm not doing any update, just modifying objects in memory, the comment will perform the update once we're out of the loop.
And modifying objects in memory is not really time consuming

antarest/study/dao/api/study_dao.py
MartinBelthle: Indeed I use it 8 times in the code so I will introduce it

antarest/study/business/areas/st_storage_management.py
smailio: I'm talking about the case where a constraint is added after line 235

sylvlecl: Indeed the application is clearly not robust to concurrent modifications in LOTS of places, we'll not fix it in that PR for sure :)

With the underlying storage and architecture, it would be difficult to achieve, other than holding a big lock on the study for every modification.

antarest/study/storage/variantstudy/model/command/remove_area.py
```diff
@@ -176,6 +180,13 @@ def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = N
             binding_ids = [bc.id for bc in referencing_binding_constraints]
             raise ReferencedObjectDeletionNotAllowed(self.id, binding_ids, object_type="Area")
 
+        # Checks that the area is not referenced in any st-storage constraint
```
sylvlecl: For removal I don't think we should consider those new constraints as independent objects, they should be seen as a part of the ST storage, I think, and deleted together with the ST storage itself (and the area).

antarest/study/storage/variantstudy/model/command/remove_st_storage.py
```diff
@@ -53,7 +53,14 @@ def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener]
             return command_failed(f"Short-term storage '{self.storage_id}' in area '{self.area_id}' does not exist")
 
         storage = study_data.get_st_storage(self.area_id, self.storage_id)
-        study_data.delete_storage(self.area_id, storage)
+
+        # Checks the storage is not referenced in any constraint
+        existing_constraints = study_data.get_st_storage_additional_constraints(self.area_id, self.storage_id)
```
sylvlecl: See other comment, here we should juste delete associated constraints, they are part of the ST storage behaviour, it's not an external object.

antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,114 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+    NETTING = "netting"
+
+
+class AdditionalConstraintOperator(EnumIgnoreCase):
+    LESS = "less"
+    GREATER = "greater"
+    EQUAL = "equal"
+
+
+HOURS_TYPE: TypeAlias = list[list[int]] | list[int]
```
sylvlecl: All types should be camel case, here it looks like a constant

antarest/study/business/model/sts_model.py
sylvlecl: As for me, I would prefer to have a more self explaining typing with something like
```python
class ConstraintOccurence:
    hours: list[int]  # or list[WeekHour] where week our would be constrainted to 0, 168

# then in the constraint class:
class STStorageAdditionalConstraint:
...
    occurences: list[ConstraintOccurence]
```

**MB** Why not but this will change the response model :/

**SL** To be checked with front devs, but I don't think it's an issue ? it will only make things clearer. Or we could serialize it as a list of list anyway, but I think it would be better to have a clearer JSON representation too, with the term "occurences" in there.

antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,114 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+
+
+def hours_parser(value: str | HOURS_TYPE) -> HOURS_TYPE:
+    def _checks_compliance(x: Any) -> None:
+        if not isinstance(x, int) or not (0 <= x <= 168):
+            raise ValueError(f"Hours must be integers between 0 and 168, got {x}")
+
+    if isinstance(value, str):
```
sylvlecl: Not sure why we would have a string here ?

antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,114 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+        value = ast.literal_eval(value)
```
sylvlecl: We should avoid this, there could be code injection

antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,114 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+
+
+class STStorageAdditionalConstraint(AntaresBaseModel):
+    """
+    Short-term storage additional constraints model.
+    """
+
+    model_config = ConfigDict(extra="forbid", populate_by_name=True)
+
+    id: LowerCaseId
```
sylvlecl: Do we really want to enforce lower case here ?

sylvlecl: We should be consistent with the creation and have name here

antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,114 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+    variable: Optional[AdditionalConstraintVariable] = None
+    operator: Optional[AdditionalConstraintOperator] = None
+    hours: Optional[Hours] = None
+    enabled: Optional[bool] = None
+
+
+STStorageAdditionalConstraintUpdates = dict[LowerCaseId, dict[LowerCaseId, list[STStorageAdditionalConstraintUpdate]]]
```
sylvlecl: We could at least add a comment explaining what should be the first ID and what should be the second one

antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,114 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+STStorageAdditionalConstraintsMap = dict[str, dict[str, list[STStorageAdditionalConstraint]]]
+
+
+def create_st_storage_constraint(cluster_data: STStorageAdditionalConstraintCreation) -> STStorageAdditionalConstraint:
+    """
+    Creates a short-term storage constraint from a creation request
+    """
+    args = cluster_data.model_dump(exclude_none=True, exclude={"name"})
+    args["id"] = transform_name_to_id(cluster_data.name)
```
sylvlecl: As opposed to what we've discussed:
we should either not have an ID at all, or at least keep the name in the model, otherwise we lose some information that we want to display to the user (the un-transformed name).

antarest/study/business/areas/st_storage_management.py
```diff
@@ -292,3 +319,130 @@ def duplicate_cluster(
...
+        return study.get_study_dao().get_st_storage_additional_constraints(area_id, storage_id)
+
+    def create_additional_constraints(
+        self,
+        study: StudyInterface,
+        area_id: str,
+        storage_id: str,
+        constraints: list[STStorageAdditionalConstraintCreation],
+    ) -> list[STStorageAdditionalConstraint]:
+        """
```
sylvlecl, line 499: We should throw at some point if study version is < 9.2
**edit** Seems to be done in the command

antarest/study/business/areas/st_storage_management.py
sylvlecl: Agreed to not change it here, but yes it's ugly.

It serves in particular to return the created constraints even though the command is not actually executed (variants).
But in practice, the commands are 99.99% of the time executed just after (to display it to the front-end, or to append additional commands).
It's a more general design issue that I think we should implement some time:
making the command execution "synchronous" even for variants.

antarest/study/storage/rawstudy/model/filesystem/config/model.py
```diff
@@ -85,6 +85,8 @@ class Area(AntaresBaseModel, extra="forbid"):
     filters_year: List[str]
     # since v8.6
     st_storages: List[STStorage] = []
+    # Since v9.2
+    st_storages_additional_constraints: List[STStorageAdditionalConstraint] = []
```
sylvlecl: Do we really want to have them in the config ?

Note that here the information is not enough: for each constraint, we don't know to which cluster it relates, not sure that it's useful this way.

antarest/study/storage/rawstudy/model/filesystem/config/st_storage.py
```diff
@@ -67,3 +73,55 @@ def serialize_st_storage(study_version: StudyVersion, storage: STStorage) -> dic
...
+        return self.cluster, STStorageAdditionalConstraint.model_validate(args)
+
+    @classmethod
+    def from_model(
+        cls, storage_id: str, additional_constraint: STStorageAdditionalConstraint
+    ) -> "STStorageAdditionalConstraintFileData":
+        return cls.model_validate({"cluster": storage_id, **additional_constraint.model_dump(exclude={"id"})})
+
+
+def parse_st_storage_additional_constraint(constraint_id: str, data: Any) -> tuple[str, STStorageAdditionalConstraint]:
```
sylvlecl: We should document what's the returned tuple, or create a wrapper dataclass with cluster name and the constraint

hdinia (changes requested):
When creating a storage constraint, the `name` field in the request is being transformed into an `id` field in the response, with automatic lowercasing applied. The original `name` field is not preserved.

Request:
```json
{
  "name": "Charge limit",
  "variable": "injection",
  "operator": "less",
  "enabled": true,
  "hours": []
}
```

Response:
```json
{
  "id": "charge limit",  
  "variable": "injection",
  "operator": "less",
  "hours": [],
  "enabled": true
}
```
Any particular reason or constraint that leads to this design ?

If not could we have both an `id` (generated or derived), ideally a `UUID` and preserve the original `name` field ?


```json
{
  "id": "f110046d-a423-47f6-9891-40cfb094f850",  
  "name": "Charge limit", 
  "variable": "injection",
  "operator": "less",
  "hours": [],
  "enabled": true
}

File: antarest/study/storage/variantstudy/model/command/create_st_storage_constraints.py [hunk 23]
```diff
@@ -0,0 +1,96 @@
...
+)
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+# noinspection SpellCheckingInspection
+class CreateSTStorageAdditionalConstraints(ICommand):
+    """
+    Command used to create several short-term storage constraints in an area.
```
sylvlecl: actually, in one storage of an area

antarest/study/storage/variantstudy/model/command/create_st_storage_constraints.py
```diff
@@ -0,0 +1,96 @@
...
+    # ===================
+
+    command_name: CommandName = CommandName.CREATE_ST_STORAGE_ADDITIONAL_CONSTRAINTS
+
+    # Command parameters
+    # ==================
+
+    area_id: AreaId
+    storage_id: LowerCaseId
+    constraints: list[STStorageAdditionalConstraintCreation]
```
sylvlecl, line 51: we should probably check that names are unique

antarest/study/storage/variantstudy/model/command/remove_multiple_storage_constraints.py
```diff
@@ -0,0 +1,67 @@
...
+    # Overloaded metadata
+    # ===================
+
+    command_name: CommandName = CommandName.REMOVE_MULTIPLE_ST_STORAGE_ADDITIONAL_CONSTRAINTS
+
+    # Command parameters
+    # ==================
+
+    area_id: str = Field(description="Area ID", pattern=r"[a-z0-9_(),& -]+")
+    ids: list[str]
```
sylvlecl, line 46: The constraint names will not be enough to uniquely identify one constraint:
we need the tuple (area, storage, constraint name) to identify one.

This is an issue larger than just here, because we cannot handle the INI file data as a dict this time.
We can have the same name repeated multiple times, one for each storage.

antarest/study/dao/file/file_study_st_storage_dao.py
```diff
@@ -261,6 +270,92 @@ def delete_storage(self, area_id: str, storage: STStorage) -> None:
...
+
+    @override
+    def get_st_storage_additional_constraints_for_area(
+        self, area_id: str
+    ) -> dict[str, list[STStorageAdditionalConstraint]]:
+        file_study = self.get_file_study()
+        path = ["input", "st-storage", "constraints", area_id, "additional_constraints"]
+        try:
+            constraints: dict[str, list[STStorageAdditionalConstraint]] = {}
+            ini_content = file_study.tree.get(path)
```
sylvlecl, line 296: Here we will have an issue when multiple constraints have the same name (for different storages).

We'll probably need to have an evolution of the ini reader to handle this.

Slightly different topic:
I think it would be better to have a mapping 1 to 1 between the file and the `**FileData` class, so the file data class would hold all the constraints for one area, and here we would just filter the ones that we want.

antarest/study/storage/variantstudy/model/command/update_st_storage_additional_constraints.py
```diff
@@ -0,0 +1,103 @@
...
+
+            study_data.save_st_storage_additional_constraints(area_id, new_constraints)
+
+        return command_succeeded("The short-term storage additional constraints were successfully updated.")
+
+    @override
+    def to_dto(self) -> CommandDTO:
+        args: dict[str, dict[str, list[dict[str, Any]]]] = {}
+
+        for area_id, value in self.additional_constraint_properties.items():
```
sylvlecl: There may be a reason but why don't we just dump the obect instead of looping over it ?

MartinBelthle:
> When creating a storage constraint, the `name` field in the request is being transformed into an `id` field in the response, with automatic lowercasing applied. The original `name` field is not preserved.
> 
> Request:
> 
> ```json
> {
>   "name": "Charge limit",
>   "variable": "injection",
>   "operator": "less",
>   "enabled": true,
>   "hours": []
> }
> ```
> 
> Response:
> 
> ```json
> {
>   "id": "charge limit",  
>   "variable": "injection",
>   "operator": "less",
>   "hours": [],
>   "enabled": true
> }
> ```
> 
> Any particular reason or constraint that leads to this design ?
> 
> If not could we have both an `id` (generated or derived), ideally a `UUID` and preserve the original `name` field ?
> 
> ```json
> {
>   "id": "f110046d-a423-47f6-9891-40cfb094f850",  
>   "name": "Charge limit", 
>   "variable": "injection",
>   "operator": "less",
>   "hours": [],
>   "enabled": true
> }
> ```

This is because the `name` is a complete construction it doesn't exist inside the files contrary to clusters or binding constraints for instance. So we discussed it with Sylvain and decided it didn't make sense to return it in the response

tests/variantstudy/model/command/test_create_st_storage_additional_constraints.py
```diff
@@ -0,0 +1,173 @@
...
+        assert output.status
+
+        # Create another one referencing the other storage
+        cmd = CreateSTStorageAdditionalConstraints(
+            command_context=command_context,
+            area_id="fr",
+            storage_id="sts_2",
+            constraints=[
+                STStorageAdditionalConstraintCreation(
+                    name="c3",
```
sylvlecl, line 67: We will need a test with the same name as in the other storage

sylvlecl (changes requested):
A first round, with a few "structural" comments:
 1. I think we need to handle the case where multiple constraints in the same area have the same name
 2. we should consider constraints as part of the storage, and handle them this way when deleting storages or areas
 3. we need to preserve and communicate the original name to the front

antarest/study/storage/variantstudy/model/command/remove_area.py
MartinBelthle: We could, I don't have a strong opinion. I asked Alexander and that"s not what he wanted though

antarest/study/storage/variantstudy/model/command/remove_st_storage.py
MartinBelthle: Again, that's a total different behavior from deleting a link today that is forbidden if the link is referenced in a binding constraint term

antarest/study/business/model/sts_model.py
MartinBelthle: We have a string when reading the data from the ini file

antarest/study/business/model/sts_model.py
MartinBelthle: Didn't know. I use it in antares-craft too

antarest/study/storage/rawstudy/model/filesystem/config/model.py
MartinBelthle: 1- We need the info to build the tree. Also if we want to use the tree more often, we'll need it even more.

2- Indeed, i will use a dict as I cannot change the existing `st_storages` structure to make it a dict I think. Can I ?

antarest/study/storage/variantstudy/model/command/remove_multiple_storage_constraints.py
MartinBelthle: No we can't because of the unicity of the matrix so this is irrelevant

antarest/study/dao/file/file_study_st_storage_dao.py
MartinBelthle: Same. Either i'm missing something or this is completely irrelevant

antarest/study/storage/variantstudy/model/command/update_st_storage_additional_constraints.py
MartinBelthle: Not sure to understand as `self.additional_constraint_properties` is a dict not a pydantic object

antarest/study/business/model/sts_model.py
MartinBelthle: I discussed with Hatim and Samir and they don't really agree, they would prefer to have an ID here. And I think I agree with them

antarest/study/business/model/sts_model.py
MartinBelthle: Not sure to understand, I thought we agreed on the fact that we didn't want to return a name as it didn't make sense as it wasn't written inside the INI file

antarest/study/storage/variantstudy/model/command/update_st_storage_additional_constraints.py
sylvlecl: Then we should use a `TypeAdapter` for this (making it a global var so that it's built only once).

antarest/study/business/model/sts_model.py
sylvlecl: We don't have the choice to return an un-transformed name (which could be used as the ID), otherwise the user will not get what he wrote at the time of creation.
That's why I thought at first to have only that "name" either in an "id" field or a "name" field.

But eventually it seems simpler to stick with current practices elsewhere and have both name and id, so let's stick with it for now.

sylvlecl: PS: the name is indeed written inside the INI file, it's the section name

antarest/study/storage/variantstudy/model/command/remove_st_storage.py
sylvlecl: It's completely different from a BC:
a BC is an object which links several objects together, from the outside, it's not part of one of the objects.
On the other side, a ST storage constraint is an internal object of that storage, like other properties of that storage.

I am pretty sure we don't want to go into the constraints menu to delete them before we can delete the storage which contains them :)

antarest/study/business/model/sts_model.py
sylvlecl: Ok let's keep the usual behaviour with both name and id, id being generated from the name.

antarest/study/web/study_data_blueprint.py
```diff
@@ -1815,6 +1822,80 @@ def delete_st_storages(uuid: str, area_id: str, storage_ids: Sequence[str]) -> N
...
+        )
+        study = study_service.check_study_access(uuid, StudyPermissionType.WRITE)
+        study_interface = study_service.get_study_interface(study)
+        all_constraints = study_service.st_storage_manager.update_additional_constraints(
+            study_interface, {area_id: {storage_id: constraints}}
+        )
+        return all_constraints[area_id][storage_id]
+
+    @bp.delete(
+        path="/studies/{uuid}/areas/{area_id}/storages/additional-constraints",
```
hdinia: this should not include the `storage_id` like other endpoints ?

antarest/study/business/model/sts_model.py
sylvlecl: Yes but it should be handled in the parsing, not in the model

antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,129 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+    @classmethod
+    def set_id(cls, data: dict[str, Any]) -> dict[str, Any]:
+        data["id"] = transform_name_to_id(data["name"])
+        return data
+
+    id: str
+    name: ItemName
+    variable: AdditionalConstraintVariable = AdditionalConstraintVariable.NETTING
+    operator: AdditionalConstraintOperator = AdditionalConstraintOperator.LESS
+    hours: Hours = [[]]
```
sylvlecl: Seems to have been deleted, but I really think we should have a better representation than list of list, with a list of "occurences" which themselves contain a list of "hours". This will make the code more understandable here AND in the front-end.

antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,129 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+        to_return = []
+        numbers_as_list = re.findall(r"\[(.*?)\]", s)
+        if numbers_as_list == [""]:
+            # Happens if the given string is `[]`
+            return [[]]
+        for numbers in numbers_as_list:
+            to_return.append([int(v) for v in numbers.split(",")])
+        return to_return
+
+    def _checks_compliance(x: Any) -> None:
```
sylvlecl: We could have a `WeekHour` annotated type for this

antarest/study/business/model/sts_model.py
```diff
@@ -250,3 +251,129 @@ def update_st_storage(storage: STStorage, data: STStorageUpdate, version: StudyV
...
+    EQUAL = "equal"
+
+
+HoursType: TypeAlias = list[list[int]]
+
+
+def hours_parser(value: str | HoursType) -> HoursType:
+    def _string_to_list(s: str) -> HoursType:
+        to_return = []
+        numbers_as_list = re.findall(r"\[(.*?)\]", s)
```
sylvlecl: I think we could parse it as a JSON list of list, no ?

antarest/study/storage/variantstudy/model/command/remove_area.py
sylvlecl: Still to be removed I think ?

sylvlecl (commented):
A few more comments

antarest/study/business/model/sts_model.py
sylvlecl: Actually not really, we don't have the enclosing braces, let's keep it this way.

antarest/study/storage/variantstudy/model/command/update_st_storage_additional_constraints.py
sylvlecl: We should still implement this, will be shorter and more efficient

sylvlecl (approved):

PR #2561, opened 2025-07-02
feat(ui-studies): add select/deselect all in outputs fields in copy dialog

skamril:
create a new SelectFE

ANT-3321

### DIFF
--- a/webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
+++ b/webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
@@ -119,7 +119,7 @@ function CopyStudyDialog({ study, open, onClose }: Props) {
                 startCaseLabel={false}
                 multiple
                 renderValueAs="chip"
-                onSelectAllOptions={(values) => setValue("outputIds", values as string[])}
+                onSelectAllOptions={(values) => setValue("outputIds", values)}
                 onDeselectAllOptions={() => setValue("outputIds", [])}
               />
             )}
--- a/webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
+++ b/webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
@@ -14,7 +14,7 @@
 
 import FormDialog from "@/components/common/dialogs/FormDialog";
 import FieldSkeleton from "@/components/common/fieldEditors/FieldSkeleton";
-import SelectFE from "@/components/common/fieldEditors/SelectFE";
+import NewSelectFE from "@/components/common/fieldEditors/NewSelectFE";
 import StringFE from "@/components/common/fieldEditors/StringFE";
 import Fieldset from "@/components/common/Fieldset";
 import type { SubmitHandlerPlus } from "@/components/common/Form/types";
@@ -26,7 +26,6 @@ import type { StudyMetadata, StudyOutput } from "@/types/types";
 import { validateStudyName } from "@/utils/studiesUtils";
 import FileCopyOutlinedIcon from "@mui/icons-material/FileCopyOutlined";
 import SaveAsIcon from "@mui/icons-material/SaveAs";
-import { Box, Chip } from "@mui/material";
 import { useTranslation } from "react-i18next";
 import StudyPathFE from "../StudyPathFE";
 
@@ -90,7 +89,7 @@ function CopyStudyDialog({ study, open, onClose }: Props) {
       config={{ defaultValues }}
       allowSubmitOnPristine
     >
-      {({ control }) => (
+      {({ control, setValue }) => (
         <Fieldset fullFieldWidth>
           <StringFE
             name="studyName"
@@ -104,32 +103,24 @@ function CopyStudyDialog({ study, open, onClose }: Props) {
             response={outputsRes}
             ifPending={() => (
               <FieldSkeleton>
-                <SelectFE options={[]} />
+                <NewSelectFE value="" />
               </FieldSkeleton>
             )}
             ifRejected={() => (
-              <SelectFE options={[]} helperText={t("study.error.listOutputs")} error disabled />
+              <NewSelectFE helperText={t("study.error.listOutputs")} error disabled />
             )}
             ifFulfilled={(outputs) => (
-              <SelectFE
+              <NewSelectFE
                 name="outputIds"
-                label={t("global.outputs")}
                 control={control}
+                label={t("global.outputs")}
                 defaultValue={outputs}
                 options={outputs}
                 startCaseLabel={false}
                 multiple
-                renderValue={(selected) => {
-                  if (Array.isArray(selected)) {
-                    return (
-                      <Box sx={{ display: "flex", flexWrap: "wrap", gap: 0.5 }}>
-                        {selected.map((value) => (
-                          <Chip key={value} label={value} />
-                        ))}
-                      </Box>
-                    );
-                  }
-                }}
+                renderValueAs="chip"
+                onSelectAllOptions={(values) => setValue("outputIds", values as string[])}
+                onDeselectAllOptions={() => setValue("outputIds", [])}
               />
             )}
           />
--- a/webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
+++ b/webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
@@ -118,7 +118,6 @@ function CopyStudyDialog({ study, open, onClose }: Props) {
                 options={outputs}
                 startCaseLabel={false}
                 multiple
-                renderValueAs="chip"
                 onSelectAllOptions={(values) => setValue("outputIds", values)}
                 onDeselectAllOptions={() => setValue("outputIds", [])}
               />
--- a/webapp/src/components/common/Form/useFormApiPlus.ts
+++ b/webapp/src/components/common/Form/useFormApiPlus.ts
@@ -12,6 +12,8 @@
  * This file is part of the Antares project.
  */
 
+import * as RA from "ramda-adjunct";
+import { useEffect, useMemo, useRef } from "react";
 import type {
   BatchFieldArrayUpdate,
   FieldPath,
@@ -22,14 +24,12 @@ import type {
   UseFormSetValue,
   UseFormUnregister,
 } from "react-hook-form";
-import * as RA from "ramda-adjunct";
-import { useEffect, useMemo, useRef } from "react";
 import useUpdatedRef from "../../../hooks/useUpdatedRef";
 import type {
+  AutoSubmitHandler,
   UseFormRegisterPlus,
   UseFormReturnPlus,
   UseFormSetValues,
-  AutoSubmitHandler,
 } from "./types";
 
 interface Params<TFieldValues extends FieldValues, TContext> {
@@ -136,7 +136,7 @@ function useFormApiPlus<TFieldValues extends FieldValues, TContext>(
       };
 
       const updateFieldArrayWrapper: BatchFieldArrayUpdate = (...args) => {
-        control._updateFieldArray(...args);
+        control._setFieldArray(...args);
         if (dataRef.current.isAutoSubmitEnabled) {
           dataRef.current.submit();
         }
--- a/webapp/src/components/common/fieldEditors/NewSelectFE.tsx
+++ b/webapp/src/components/common/fieldEditors/NewSelectFE.tsx
@@ -1,0 +1,189 @@
+/**
+ * Copyright (c) 2025, RTE (https://www.rte-france.com)
+ *
+ * See AUTHORS.txt
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+import reactHookFormSupport, { type ReactHookFormSupportProps } from "@/hoc/reactHookFormSupport";
+import type { SvgIconComponent } from "@mui/icons-material";
+import {
+  Box,
+  Button,
+  ButtonGroup,
+  Chip,
+  MenuItem,
+  TextField,
+  Tooltip,
+  type TextFieldProps,
+} from "@mui/material";
+import startCase from "lodash/startCase";
+import * as RA from "ramda-adjunct";
+import { useMemo } from "react";
+import type { FieldPath, FieldValues } from "react-hook-form";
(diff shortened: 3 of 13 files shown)


webapp/src/components/common/fieldEditors/NewSelectFE.tsx
```diff
@@ -0,0 +1,161 @@
...
+              flexWrap: "wrap",
+              gap: 0.5,
+            },
+          },
+        }}
+        helperText={
+          <>
+            <Box>{helperText}</Box>
+            {hasOptionActions && (
+              <ButtonGroup variant="text" size="extra-small">
```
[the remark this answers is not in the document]
skamril: Custom theme

webapp/src/components/common/Form/useFormApiPlus.ts
```diff
@@ -136,7 +136,7 @@ function useFormApiPlus<TFieldValues extends FieldValues, TContext>(
       };
 
       const updateFieldArrayWrapper: BatchFieldArrayUpdate = (...args) => {
-        control._updateFieldArray(...args);
+        control._setFieldArray(...args);
```
hdinia, line 139: I see this was likely changed due to the React Hook Form version bump, but this is exactly why using private methods like `_setFieldArray` isn't recommended they can break/change between versions without warning (as happened here). Could we refactor to use the public `useFieldArray` API instead to avoid future breaking changes?

webapp/src/components/common/fieldEditors/ColorPickerFE/index.tsx
```diff
@@ -12,20 +12,26 @@
...
 import { composeRefs } from "../../../../utils/reactUtils";
-import reactHookFormSupport from "../../../../hoc/reactHookFormSupport";
+import { rgbToString, stringToRGB } from "./utils";
 
-export type ColorPickerFEProps = Omit<TextFieldProps, "type" | "defaultChecked"> & {
+export type ColorPickerFEProps<Variant extends TextFieldVariants = TextFieldVariants> = {
   value?: string; // Format: R,G,B - ex: "255,255,255"
   defaultValue?: string;
-};
+} & Omit<TextFieldProps<Variant>, "type" | "defaultChecked">;
```
hdinia: opinion: valid but I would used a `Pick` here to be more explicit and not inherit everything (some props are invalid for a color picker)

webapp/src/components/common/fieldEditors/NewSelectFE.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+interface Option {
+  label: string;
+  value: string | number;
+  icon?: SvgIconComponent;
+  tooltip?: string;
+}
+
+type MultipleValue = Array<Option["value"]>;
+type Value = Option["value"] | MultipleValue;
```
hdinia: `Value` type allows both single values AND arrays, regardless of `multiple` prop. for example:
multiple=false` but `value={["a", "b"]}` is valid TypeScript but logically wrong

webapp/src/components/common/fieldEditors/NewSelectFE.tsx
```diff
@@ -0,0 +1,158 @@
...
+type Options = Array<string | Option>;
+
+type NewSelectFEProps<Variant extends TextFieldVariants = TextFieldVariants> = {
+  value?: Value;
+  defaultValue?: Value;
+  options?: Options;
+  emptyValue?: boolean;
+  startCaseLabel?: boolean;
+  multiple?: boolean;
+  chip?: boolean;
```
hdinia: This prop only makes sense when multiple=true

The naming could be more explicit `chip` -> `displayAsChips`

webapp/src/components/common/fieldEditors/NewSelectFE.tsx
```diff
@@ -0,0 +1,158 @@
...
+  onSelectAllOptions?: (optionValues: MultipleValue) => void;
+  onDeselectAllOptions?: VoidFunction;
```
hdinia: Same remark as for `chip` those props seems designed to be used only if multiple=true

webapp/src/components/common/fieldEditors/NewSelectFE.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+type MultipleValue = Array<Option["value"]>;
+type Value = Option["value"] | MultipleValue;
+
+type Options = Array<string | Option>;
+
+type NewSelectFEProps<Variant extends TextFieldVariants = TextFieldVariants> = {
+  value?: Value;
+  defaultValue?: Value;
+  options?: Options;
```
hdinia: options should be mandatory, no ? A select without options doesn't make sense

webapp/src/components/common/fieldEditors/NewSelectFE.tsx
```diff
@@ -0,0 +1,158 @@
...
+  emptyValue?: boolean;
+  startCaseLabel?: boolean;
+  multiple?: boolean;
+  chip?: boolean;
+  onSelectAllOptions?: (optionValues: MultipleValue) => void;
+  onDeselectAllOptions?: VoidFunction;
+} & Omit<TextFieldProps<Variant>, "select" | "type" | "value" | "defaultValue">;
+
+function NewSelectFE({
```
hdinia: General comment about this added component:

They are some types and logical prop inconsistencies 

For types you can use discriminated Union Types

```jsx
type SingleSelectProps = {
  multiple?: false;
  value?: Option["value"];
  defaultValue?: Option["value"];
  displayAsChips?: never;
  onSelectAllOptions?: never;
  onDeselectAllOptions?: never;
};

type MultipleSelectProps = {
  multiple: true;
  value?: MultipleValue;
  defaultValue?: MultipleValue;
  displayAsChips?: boolean;
  onSelectAllOptions?: (optionValues: MultipleValue) => void;
  onDeselectAllOptions?: VoidFunction;
};

type NewSelectFEProps<Variant extends TextFieldVariants = TextFieldVariants> = 
  (SingleSelectProps | MultipleSelectProps) & {
    options: Options; // Required
    emptyValue?: boolean;
    startCaseLabel?: boolean;
  } & Omit<TextFieldProps<Variant>, "select" | "type" | "value" | "defaultValue">;

```


## Component Architecture Concern

The main issue I see here is that you're mixing two distinct use cases:

1. **Single Select** - Traditional dropdown with a list of options and a single element selected at a time
2. **Multiple Select** - Chip-based multi-selection with different UX patterns

## Suggested Changes

Instead of coupling these use cases, I recommend separating them:

1. Keep existing `SelectFE` → rename to `SingleSelectFE` for single selection (traditional dropdown)
2. Create new `MultipleSelectFE` for chip-based multi-selection

## Why Separation is Better


-  **Clearer types** - No need for discriminated unions or complex conditional props
-  **Simpler logic** - Each component handles one use case well
-  **Easier testing** - Mixed uses cases complicate testing a lot
-  **Better maintainability** - Future enhancements won't risk breaking the other use case
-  **Explicit API** - Developers know exactly what they're getting

**Example API:**
```tsx
// Single selection
<SingleSelectFE value="single" options={options} />

// Multiple selection  
<MultipleSelectFE value={["a", "b"]} options={options} displayAsChips />

File: webapp/src/components/common/fieldEditors/NewSelectFE.tsx [hunk 9]
```diff
@@ -0,0 +1,158 @@
...
+          select: {
+            renderValue: chip ? renderSelectedValueAsChip : undefined,
+            multiple,
+          },
+        }}
+        helperText={
+          <>
+            <Box>{helperText}</Box>
+            {hasOptionActions && (
+              <ButtonGroup variant="text" size="extra-small">
```
hdinia: you can handle disabled state, if the list is empty for example

webapp/src/components/common/fieldEditors/NewSelectFE.tsx
```diff
@@ -0,0 +1,158 @@
...
+    );
+  };
+
+  ////////////////////////////////////////////////////////////////
+  // JSX
+  ////////////////////////////////////////////////////////////////
+
+  return (
+    <>
+      <TextField
```
hdinia: does a disabled state would make sense here ? For example if the list coming from the API is empty ?

webapp/src/components/common/fieldEditors/NewSelectFE.tsx
```diff
@@ -0,0 +1,158 @@
...
+          },
+        }}
+        helperText={
+          <>
+            <Box>{helperText}</Box>
+            {hasOptionActions && (
+              <ButtonGroup variant="text" size="extra-small">
+                {onSelectAllOptions && (
+                  <Button
+                    onClick={() => onSelectAllOptions(optionsFormatted.map(({ value }) => value))}
```
hdinia: you can refactor the options array outside (or even create a handler and move out all the logic)

```ts
const allOptionValues = useMemo(
  () => optionsFormatted.map(({ value }) => value),
  [optionsFormatted]
);
```

hdinia (changes requested):

hdinia:
You can add accessibility attribute to facilitate E2E testing we will introduce soon
 
```tsx
<TextField
  {...rest}
  select
// Missing ARIA attributes
  aria-label={rest.label ? undefined : "Select options"} 
  aria-describedby={helperText ? `${rest.id}-helper` : undefined} 
/>

File: webapp/src/components/common/Form/useFormApiPlus.ts (hunk 2 above)
Comment (skamril) [reply webapp/src/components/common/Form/useFormApiPlus.ts:139]: It's not possible to use `useFieldArray`, it's a wrapper of `control` to support auto submit. This feature will be removed when the last auto submit forms will be refactor to manual submit.

File: webapp/src/components/common/fieldEditors/ColorPickerFE/index.tsx (hunk 3 above)
Comment (skamril) [reply webapp/src/components/common/fieldEditors/ColorPickerFE/index.tsx]: No, the list for pick it's too big, all the inherit  props are giving to the TextField inside.

File: webapp/src/components/common/fieldEditors/NewSelectFE.tsx (hunk 10 above)
Comment (skamril) [reply webapp/src/components/common/fieldEditors/NewSelectFE.tsx]: In this case, it is good for the user to know that the API returns an empty list. 
A disabled field can be understood as a field that is forbidden from editing.

File: webapp/src/components/common/fieldEditors/NewSelectFE.tsx (hunk 11 above)
Comment (skamril) [reply webapp/src/components/common/fieldEditors/NewSelectFE.tsx]: Create a memo for this is overkill. All option are only needed when is `multiple` and `onSelectAllOptions` is defined, so create it every time is a no sense.

I can create an event handler, but only for this is not necessary for me.

File: webapp/src/components/common/fieldEditors/NewSelectFE.tsx (hunk 7 above)
Comment (skamril) [reply webapp/src/components/common/fieldEditors/NewSelectFE.tsx]: We can have async options, and force the `options` prop suppose to do this: `options={asyncOptions || []}` (`asyncOptions` is first undefined).

File: webapp/src/components/common/fieldEditors/NewSelectFE.tsx [hunk 12]
```diff
@@ -0,0 +1,189 @@
...
+  defaultValue?: Value[];
+  onSelectAllOptions?: (optionValues: OptionValue[]) => void;
+  onDeselectAllOptions?: VoidFunction;
+}
+
+type NewSelectFEProps<OptionValue extends Value = Value> = {
+  options?: Options<OptionValue>;
+  emptyValue?: boolean;
+  startCaseLabel?: boolean;
+  renderValueAs?: "text" | "chip";
```
hdinia: you could remove the `renderValueAs` prop entirely and always use chips for multiple selects (since that's the most intuitive UI pattern)

```ts
    slotProps={{
        ...slotProps,
        select: {
          renderValue: multiple ? renderSelectedValueAsChip : undefined,
          multiple,
        },
      }}
```

webapp/src/components/common/fieldEditors/NewSelectFE.tsx
```diff
@@ -0,0 +1,189 @@
...
+} & (SingleSelect | MultipleSelect<OptionValue>) &
+  Omit<TextFieldProps, "select" | "type" | "value" | "defaultValue">;
+
+function NewSelectFE<OptionValue extends Value = Value>({
```
hdinia, line 69: - you can rename it `SelectFEv2` to clearly indicate it's version 2 of the SelectFE component and make the migration intent obvious 

- if you do not plan to make the migration and remove the `SelectFE` soon (like just after merging this one), you can create a Jira technical story to track it


- you can also add documentation on the usage of the API, since you decided to mix the 2 use-cases (single vs multiple) with examples of usage, and add few basic unit tests on important logic

webapp/src/components/common/Form/useFormApiPlus.ts
hdinia: I think you can create a wrapper around `useFieldArray` to do the same, but it's not important

webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
```diff
@@ -108,32 +107,24 @@ function CopyStudyDialog({ study, open, onClose }: Props) {
...
-                      <Box sx={{ display: "flex", flexWrap: "wrap", gap: 0.5 }}>
-                        {selected.map((value) => (
-                          <Chip key={value} label={value} />
-                        ))}
-                      </Box>
-                    );
-                  }
-                }}
+                renderValueAs="chip"
+                onSelectAllOptions={(values) => setValue("outputIds", values as string[])}
```
hdinia: type assertion on `values` not needed i think

webapp/src/components/common/fieldEditors/NewSelectFE.tsx
```diff
@@ -0,0 +1,189 @@
...
+      </Box>
+    );
+  };
+
+  ////////////////////////////////////////////////////////////////
+  // JSX
+  ////////////////////////////////////////////////////////////////
+
+  return (
+    <>
```
hdinia: unnecessary fragment

hdinia (changes requested):

webapp/src/components/common/fieldEditors/NewSelectFE.tsx
hdinia: Ok makes sense

webapp/src/components/App/shared/studies/dialogs/CopyStudyDialog.tsx
skamril: Yes, I forgot to update after updating the types.

webapp/src/components/common/fieldEditors/NewSelectFE.tsx
skamril: I will make a new PR for the migration soon. I added a deprecated tag to the old Select.

hdinia (dismissed):

hdinia (approved):

PR #2564, opened 2025-07-03
fix(ui-results): resolve variant data display, infinite loops, and state inconsistencies

hdinia:
Issues Fixed

#### 1. **Variant Output Display Issue**
- **Problem**: After variant generation jobs complete, results remained outdated/unavailable until manual refresh
- **Root Cause**: WebSocket listener not triggering for variant job completion notifications
- **Fix**: Implemented direct output fetching as workaround while API WebSocket issue is addressed

#### 2. **Infinite Loop in Results View**
- **Problem**: Infinite re-renders when variants are not found and shouldn't be displayed
- **Root Cause**: Unstable `colHeaders` prop switching between different data sources during loading states
- **Fix**: Preserved column headers during loading transitions to prevent cascading state updates

#### 3. **Incorrect "No Data" State Display**
- **Problem**: "No data" screen shown incorrectly due to unhandled loading condition
- **Root Cause**: `if (resultColHeaders.length === 0)` condition not properly handled during data loading
- **Fix**: Added proper loading state management to prevent premature "no data" display

#### 4. **Infinite Retry Loop in useStudySynthesis**
- **Problem**: Hook would retry indefinitely on errors, even when 404 errors are expected valid states
- **Root Cause**: No mechanism to prevent retry loops for legitimate error states
- **Fix**: Implemented `hasAttempted` state logic to prevent infinite retries

### DIFF
--- a/webapp/src/components/App/Singlestudy/explore/Results/ResultDetails/index.tsx
+++ b/webapp/src/components/App/Singlestudy/explore/Results/ResultDetails/index.tsx
@@ -23,17 +23,12 @@ import { useCallback, useEffect, useMemo, useState, useRef } from "react";
 import { useTranslation } from "react-i18next";
 import { useNavigate, useOutletContext, useParams } from "react-router";
 import usePromise from "../../../../../../hooks/usePromise";
+import useStudyOutput from "../hooks/useStudyOutput";
 import useAppSelector from "../../../../../../redux/hooks/useAppSelector";
-import useStudySynthesis from "../../../../../../redux/hooks/useStudySynthesis";
-import { getAreas, getLinks, getStudyOutput } from "../../../../../../redux/selectors";
+import { getAreas, getLinks } from "../../../../../../redux/selectors";
 import { getStudyMatrixIndex } from "../../../../../../services/api/matrix";
 import { getStudyData } from "../../../../../../services/api/study";
-import {
-  StudyType,
-  type Area,
-  type LinkElement,
-  type StudyMetadata,
-} from "../../../../../../types/types";
+import { type Area, type LinkElement, type StudyMetadata } from "../../../../../../types/types";
 import { toError } from "../../../../../../utils/fnUtils";
 import { isSearchMatching } from "../../../../../../utils/stringUtils";
 import ButtonBack from "../../../../../common/ButtonBack";
@@ -49,7 +44,7 @@ import {
 import EmptyView from "../../../../../common/page/EmptyView";
 import PropertiesView from "../../../../../common/PropertiesView";
 import SplitView from "../../../../../common/SplitView/index";
-import UsePromiseCond, { mergeResponses } from "../../../../../common/utils/UsePromiseCond";
+import UsePromiseCond from "../../../../../common/utils/UsePromiseCond";
 import ListElement from "../../common/ListElement";
 import ResultFilters from "./ResultFilters";
 import { createPath, DataType, MAX_YEAR, OutputItemType, SYNTHESIS_ITEMS, Timestep } from "./utils";
@@ -63,48 +58,12 @@ function ResultDetails() {
   const { t } = useTranslation();
   const navigate = useNavigate();
 
-  const outputRes = useStudySynthesis({
-    studyId: study.id,
-    selector: (state, id) => getStudyOutput(state, id, outputId as string),
-  });
-
-  console.log("Output Response:", outputRes);
-
-  const { data: outputFromSynthesis } = outputRes;
-
-  // For variants, if output is not found in synthesis, create a minimal output object
-  const output = useMemo(() => {
-    if (outputFromSynthesis) {
-      return outputFromSynthesis;
-    }
-
-    if (outputId && study.type === StudyType.VARIANT) {
-      return {
-        id: outputId,
-        name: outputId,
-        mode: "economy" as const,
-        nbyears: MAX_YEAR,
-        synthesis: true,
-        date: new Date().toISOString(),
-        by_year: true,
-        error: false,
-      };
-    }
-
-    return undefined;
-  }, [outputFromSynthesis, outputId, study.type]);
-
-  console.log("Output Fixed:", output);
-
   const [dataType, setDataType] = useState(DataType.General);
   const [timestep, setTimestep] = useState(Timestep.Hourly);
   const [year, setYear] = useState(-1);
   const [itemType, setItemType] = useState(OutputItemType.Areas);
   const [selectedItemId, setSelectedItemId] = useState("");
   const [searchValue, setSearchValue] = useState("");
-  // Store filtered headers and their original indices separately
-  // This allows us to correctly map the data rows to their corresponding headers
-  // when some columns are filtered out
   const [resultColHeaders, setResultColHeaders] = useState<string[][]>([]);
   const [headerIndices, setHeaderIndices] = useState<number[]>([]);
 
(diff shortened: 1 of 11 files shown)


webapp/src/components/App/Singlestudy/explore/Results/ResultDetails/index.tsx
```diff
@@ -44,69 +44,144 @@ import {
...
+  const { t } = useTranslation();
+
+  return (
+    <Box>
+      <PropertiesView
+        topContent={
+          <Box sx={{ width: 1, px: 1 }}>
+            <ButtonBack onClick={onNavigateBack} />
+          </Box>
+        }
```
hdinia: this was for testing purposes, it's removed in the next commit I will push

webapp/src/redux/hooks/useStudySynthesis.ts
```diff
@@ -37,21 +37,24 @@ export default function useStudySynthesis<T>(props: UseStudySynthesisProps<T>):
   const dispatch = useAppDispatch();
   const [status, setStatus] = useState<TPromiseStatus>(PromiseStatus.Idle);
   const [error, setError] = useState<Response["error"]>();
+  const [hasAttempted, setHasAttempted] = useState(false);
```
hdinia: dont know for the other use cases if this could be problematic @skamril If you think so I can add a "retry" logic, with a `MAX_RETRY` limit

webapp/src/components/App/Singlestudy/explore/Results/ResultDetails/index.tsx
```diff
@@ -78,35 +59,59 @@ function ResultDetails() {
...
   const path = useMemo(() => {
     if (output && selectedItem && !isSynthesis) {
       return createPath({
-        output,
+        output: {
+          id: output.id,
+          name: output.name,
+          mode: output.mode,
+          nbyears: output.nbyears,
+        },
```
hdinia: revert

webapp/src/components/App/Singlestudy/explore/Results/ResultDetails/index.tsx
```diff
@@ -228,118 +235,60 @@ function ResultDetails() {
...
+      if (value !== searchValue) {
+        setSearchValue(value);
+      }
+    },
+    [searchValue],
+  );
+
+  const handleNavigateBack = useCallback(() => {
+    navigate("..");
+  }, [navigate]);
```
hdinia: replace with an inline function without `useCallback`

webapp/src/components/App/Singlestudy/explore/Results/hooks/useStudyOutput.ts
```diff
@@ -0,0 +1,169 @@
...
+  studyId,
+  outputId,
+}: UseStudyOutputOptions): UseStudyOutputReturn {
+  // Try to get output from Redux store
+  const { data: outputFromStore } = useStudySynthesis({
+    studyId,
+    selector: (state, id) => (outputId ? getStudyOutput(state, id, outputId) : undefined),
+  });
+
+  // Check if we have the output in store to avoid unnecessary API calls
```
hdinia: remove redundant comments

webapp/src/components/App/Singlestudy/explore/Results/hooks/useStudyOutput.ts
```diff
@@ -0,0 +1,169 @@
...
+      // BUG: For variants, after a generation job is initiated, we must await the job
+      // completion to display results. However, the WebSocket listener that should notify
+      // us of job completion never gets triggered for this particular case. This causes
+      // variant outputs to remain outdated/unavailable until the app is manually refreshed.
+      // As a workaround, we attempt to fetch the output directly.
+      //
+      // TODO: The API should ensure that the WebSocket listener is properly set up to notify
+      // us of job completion.
+
+      console.log("[useStudyOutput] Fetching output from API:", outputId);
```
hdinia: remove logs after review

webapp/src/components/App/Singlestudy/explore/Results/hooks/useStudyOutput.ts
```diff
@@ -0,0 +1,169 @@
...
+      console.log("[useStudyOutput] Output not in list, trying to get settings:", outputId);
+
+      const outputSettings = await getStudyData(
+        studyId,
+        `output/${outputId}/about-the-study/parameters`,
+      );
+
+      // This case should never happen, but just in case
+      // we use a hard-coded default value to be able to display the output
+      if (!outputSettings) {
```
hdinia: here we dont necessarily need to call `getStudyData`, or use a hard-coded object, I don't see the case where this should be needed, as if the api call fails to get study outputs before I think we should stop there. I may remove this @skamril

webapp/src/services/api/study.ts
```diff
@@ -107,6 +107,27 @@ export const getStudyOutputs = async (sid: string): Promise<StudyOutput[]> => {
...
+ *
+ * @param studyId - The ID of the study.
+ * @param outputId - The ID of the output to retrieve.
+ * @returns The study output if found, or null if not found.
+ */
+export const getStudyOutputById = async (
+  studyId: string,
+  outputId: string,
+): Promise<StudyOutput | undefined> => {
+  try {
```
skamril: Don't use try/catch in API function, use a try/catch when you call it instead.
We lose error tracking with this code.

webapp/src/redux/hooks/useStudySynthesis.ts
skamril: What the purpose of this?

webapp/src/components/App/Singlestudy/explore/Results/hooks/useStudyOutput.ts
```diff
@@ -0,0 +1,123 @@
...
+      // As a workaround, we attempt to fetch the output directly.
+      //
+      // TODO: The API should ensure that the WebSocket listener is properly set up to notify
+      // us of job completion.
+
+      const foundOutput = await getStudyOutputById(studyId, outputId);
+
+      if (foundOutput) {
+        return {
+          id: outputId,
```
skamril: Why do you include the `outputId`? It can be overwriting by `...foundOutput`.

webapp/src/components/App/Singlestudy/explore/Results/hooks/useStudyOutput.ts
```diff
@@ -0,0 +1,123 @@
...
+export default function useStudyOutput({
+  studyId,
+  outputId,
+}: UseStudyOutputOptions): UseStudyOutputReturn {
+  const { data: outputFromStore } = useStudySynthesis({
+    studyId,
+    selector: (state, id) => (outputId ? getStudyOutput(state, id, outputId) : undefined),
+  });
+
+  const hasOutputInStore = useMemo(() => !!outputFromStore, [outputFromStore]);
```
skamril: You don't need a useMemo for this.

webapp/src/components/App/Singlestudy/explore/Results/hooks/useStudyOutput.ts
```diff
@@ -0,0 +1,123 @@
...
+      return undefined;
+    },
+    {
+      // Only depend on studyId, outputId, and whether we have the data in store
+      // This prevents infinite loops caused by outputFromStore reference changes
+      deps: [studyId, outputId, hasOutputInStore],
+    },
+  );
+
+  const data = useMemo(() => {
```
skamril: The `outputFromStore` and  `fetchedOutput` are already memoized:

```typescript
const data = outputFromStore || fetchedOutput;
```

webapp/src/components/App/Singlestudy/explore/Results/hooks/useStudyOutput.ts
hdinia: because if coming from the API outputs does not have an id, it cannot be overwitten here but just in case I will put the spread on top

webapp/src/components/App/Singlestudy/explore/Results/hooks/useStudyOutput.ts
hdinia: yes forgot to remove `hasOutputInStore` its not even necessary after converting `outputFromStore` to a Primitive for reference stability `!!outputFromStore`

skamril (approved):

PR #2570, opened 2025-07-07
fix(clusters): perform whole data validation before modifying files

MartinBelthle:


antarest/study/storage/variantstudy/model/command/update_renewables_clusters.py
```diff
@@ -43,6 +43,12 @@ class UpdateRenewablesClusters(ICommand):
...
     def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        """
+        We validate ALL objects before saving them.
+        This way, if some data is invalid, we're not modifying the study partially only.
+        """
+        memory_mapping = {}
+
         all_renewables = study_data.get_all_renewables()
 
         for area_id, value in self.cluster_properties.items():
```
TheoPascoli, line 54: Not in this PR but Line 60, we have `lowered_id = cluster_id` that seems useless to me

TheoPascoli (changes requested):
just one fix and its good

antarest/study/storage/variantstudy/model/command/update_renewables_clusters.py
MartinBelthle: Indeed, it's a remainder of an old PR, i'll change this

TheoPascoli (approved):

PR #2575, opened 2025-07-08
feat(desktop): ui improvement

smailio:
[`ANT-3469`](https://gopro-tickets.rte-france.com/browse/ANT-3469)


antarest/study/model.py
```diff
@@ -430,12 +437,13 @@ def to_posix(cls, path: Path) -> PurePosixPath:
         return PurePosixPath(path)
 
 
-class WorkspaceMetadata(AntaresBaseModel):
+class WorkspaceDTO(AntaresBaseModel):
```
smailio, line 444: This is refactoring. it didn't make sense to suffix this object with Metadata, the convention for this kind of object is to end with DTO

webapp/src/components/App/Studies/StudyTree/index.tsx
```diff
@@ -88,27 +99,24 @@ function StudyTree() {
     // Fetch subfolders and insert them to the tree
     try {
       const newSubFolders = await api.getFolders(workspace, subPath.join("/"));
-      if (newSubFolders.length > 0) {
```
smailio, line 91: It no longer makes sense to ignore empty responses from the API. This condition was relevant a few months ago, it's not anymore. In fact, when the API returns an empty list, it means folders were removed on the server side. We need to reflect that change by syncing the local storage and removing any folders that no longer exist, to avoid keeping "ghost" folders locally.

webapp/src/services/webSocket/ws.ts
```diff
@@ -209,6 +210,7 @@ function makeStudyListener(dispatch: AppDispatch): WsEventListener {
         break;
       case WsEventType.StudyDeleted:
         dispatch(deleteStudy(e.payload));
+        deleteStudyFromLocalStorage(e.payload);
```
skamril: Use localStorageMiddleware instead for consistency:

```typescript
localStorageMiddleware.startListening({
  actionCreator: deleteStudy.fulfilled,
  effect: ({ meta }) => {
    // WebSocket
    if ("name" in meta.arg) {
      const { workspace, folder } = meta.arg;

      const folders = storage.getItem(StorageKey.StudyTreeFolders) || [];
      const filteredFolders = folders.filter(
        (f) => !(f.workspace === workspace && f.path === folder),
      );
      storage.setItem(StorageKey.StudyTreeFolders, filteredFolders);
    }
  },
});
```

antarest/study/service.py
```diff
@@ -1878,8 +1879,8 @@ def unarchive_task(notifier: ITaskNotifier) -> TaskResult:
             self.event_bus.push(
                 Event(
                     type=EventType.STUDY_EDITED,
-                    payload=study.to_json_summary(),
```
smailio, line 1883: This is a bug fix. This function load study_to_archive from database but uses study from the closure. Most of the time it works except if we try to access an attribute that's lazy laoded by sql alchemy. In fac, the bug never occur because we never do that, but I discovered when working on this feature that it could happen.

webapp/src/components/App/Studies/StudyTree/utils.ts
```diff
@@ -160,25 +172,40 @@ export function insertFoldersIfNotExist(
...
     hasChildren: true,
+    alias: workspace.disk_name,
   };
-  if (stydyTree.children.some((child) => child.name === workspace)) {
-    return stydyTree;
+  if (studyTree.children.some((child) => isNodeMatchWorkspace(child, workspace))) {
+    return {
+      ...studyTree,
+      children: studyTree.children.map((child) => {
+        if (isNodeMatchWorkspace(child, workspace)) {
```
skamril, line 194: Move the logic inside StudyTreeNone by giving a `workspace` prop instead of `alias` prop.

webapp/src/components/App/Studies/StudyTree/StudyTreeNode.tsx
```diff
@@ -40,27 +40,29 @@ export default function StudyTreeNode({
...
-  const { hasChildren, children, path, name, isStudyFolder } = node;
+  const { hasChildren, children, path, name, isStudyFolder, alias } = node;
   const isLoading = itemsLoading.includes(node.path);
   const hasUnloadedChildren =
     hasChildren && children.length === 0 && !exploredFolders.includes(node.path);
   const { t } = useTranslation();
 
   const sortedChildren = useMemo(() => {
-    const sortedByName = nameSort(children);
+    const nonStudyChildren = children.filter((s) => !s.isScannedStudy);
```
skamril: non study == not scanned study ?

webapp/src/components/App/Studies/StudyTree/index.tsx
```diff
@@ -88,27 +99,24 @@ function StudyTree() {
...
-        setSubFolders(nextSubfolders);
-
-        const nextStudyTree = insertIfNotExist(initialStudiesTree, workspaces, nextSubfolders);
-        setStudiesTree(nextStudyTree);
-        storage.setItem(StorageKey.StudyTreeFolders, nextSubfolders);
-
-        setExploredFolders((prev) => [...prev.filter((e) => e !== itemId), itemId]);
-      }
+      // use union to prioritize new subfolders
+      const thisParent = ["", workspace, ...subPath].join("/");
```
skamril: Maybe `currentParent` is more clear

webapp/src/components/App/Studies/StudyTree/index.tsx
```diff
@@ -88,27 +99,24 @@ function StudyTree() {
...
+      const otherSubfolders = subFolders.filter((f) => f.parentPath !== thisParent);
+      // Keep non-study folders and study folders that haven't been scanned yet
+      const filteredStudyFolders = newSubFolders.filter(
+        (folder) =>
+          !studies.some(
+            (study) => folder.path === study.folder && study.workspace === folder.workspace,
```
skamril: To remove duplicate code (cf. in the `useEffect`) you can create a function:

```typescript
const myFunction = (folder) =>
  !studies.some(
    (study) => folder.path === study.folder && study.workspace === folder.workspace,
  )

const filteredStudyFolders = newSubFolders.filter(myFunction)
```

webapp/src/components/App/Studies/StudyTree/utils.ts
```diff
@@ -49,8 +51,18 @@ export function buildStudyTree(studies: StudyMetadata[]) {
...
+      // this flag will be used to not display the study folder in the study tree
+      // even if the folder isn't displayed we still need it in the study tree
+      // as it makes it easier to distinguish scanned studies from unscanend studies
       if (i === path.length - 1) {
+        const studyFolderName = path[i];
+        current.children.push({
+          name: studyFolderName,
+          children: [],
+          path: current.path ? `${current.path}/${studyFolderName}` : `/${studyFolderName}`,
+          isScannedStudy: true,
```
skamril, line 62: No issue if a new folder is added inside?

skamril (dismissed):

webapp/src/components/App/Studies/StudyTree/StudyTreeNode.tsx
smailio: Yeah I don't want to display folders of scanned studies

webapp/src/components/App/Studies/StudyTree/utils.ts
smailio: it never is as this is is always the last iteration of the loop (break just after), also this isn't displayed because of the isScannedStudy flag, so the user will never click on it, so we never add a folder to it.

antarest/study/storage/explorer_service.py
```diff
@@ -78,14 +81,49 @@ def list_dir(
...
         return folders
 
-    def list_workspaces(
-        self,
-    ) -> List[WorkspaceMetadata]:
+    def list_workspaces(self) -> List[WorkspaceDTO]:
         """
-        Return the list of all configured workspace name, except the default one.
+        Return the list of all configured workspace names, except the default one.
+        On Windows, includes the disk name (volume label) in WorkspaceMetadata.
```
MartinBelthle: Minor but you changed the class name so you should change the doc

antarest/study/storage/explorer_service.py
```diff
@@ -78,14 +81,49 @@ def list_dir(
...
-            WorkspaceMetadata(name=workspace_name)
-            for workspace_name in self.config.storage.workspaces.keys()
-            if workspace_name != DEFAULT_WORKSPACE_NAME
-        ]
+        result = []
+        for workspace_name in self.config.storage.workspaces.keys():
+            if workspace_name == DEFAULT_WORKSPACE_NAME:
+                continue
+
+            if sys.platform == "win32":
```
MartinBelthle: Usually we do: `if os.name == "nt"` but perhaps it's the same thing I'm not quite sure

Also you should explain why you're doing this on WIndows only (I have no clue)

antarest/study/storage/explorer_service.py
```diff
@@ -78,14 +81,49 @@ def list_dir(
...
+                drive_letter = os.path.splitdrive(workspace_name)[0] + "\\"
+                disk_name = get_volume_label(drive_letter)
+                result.append(WorkspaceDTO(name=workspace_name, disk_name=disk_name))
+            else:
+                result.append(WorkspaceDTO(name=workspace_name))
+
+        return result
+
+
+def get_volume_label(drive_letter: str) -> str:
```
MartinBelthle: WTF is this code :shaking_face: :shaking_face: 
Is there no easier way to do so ?

MartinBelthle (changes requested):
Few remarks, you can answer next to the comment

webapp/src/components/App/Studies/StudyTree/utils.ts
hdinia: you can rename `nonScannedStudyNodes` then

webapp/src/components/App/Studies/StudyTree/StudyTreeNode.tsx
```diff
@@ -40,27 +40,29 @@ export default function StudyTreeNode({
...
+    const sortedByName = nameSort(nonStudyChildren);
```
hdinia: `nonStudyChildren` looks a bit confusing to me, consider this:

```ts
 // Filter out scanned study folders that shouldn't be displayed
const nodesToDisplay = children.filter((s) => !s.isScannedStudy);
const sortedByName = nameSort(nodesToDisplay);
```

webapp/src/components/App/Container/index.tsx
```diff
@@ -79,6 +79,7 @@ function Container({ children }: Props) {
   const currentStudyId = useAppSelector(getCurrentStudyId);
   const isWsConnected = useAppSelector(isWebSocketConnected);
   const dispatch = useAppDispatch();
+  const isDesktopMode = import.meta.env.MODE === "desktop";
```
hdinia: to avoid this feature flag check at multiple places you can create a hook `(src/hooks/useAppMode.ts)`

```ts
export function useAppMode() {
  const isDesktopMode = import.meta.env.MODE === "desktop";
  const isWebMode = !isDesktopMode;
  
  return {
    isDesktopMode,
    isWebMode
  };
}
```

then use it where needed, you can also replace `!isDesktopMode` conditionals by `isWebMode`


```jsx
  const { isWebMode } = useAppMode();
  
  {isWebMode && (
    <MenuItem
      title={t("global.signOut")}
      icon={<LogoutIcon />}
      isMenuOpen={isDrawerOpen}
      onClick={() => setOpenLogoutDialog(true)}
    />
  )}
```

hdinia (changes requested):

webapp/src/components/App/Studies/StudyTree/StudyTreeNode.tsx
hdinia: Additional remark: in this file (`StudyTreeNode.tsx`) you have named a type the same name as the component, you shouldn't redeclare 'StudyTreeNode' 2 times, you can rename either the type or the component name. If you do rename the component, please put the `export default` declaration at the bottom of the file to stay consistent with the other files

webapp/src/components/App/Studies/StudyTree/types.ts
```diff
@@ -29,6 +31,11 @@ export interface FolderDTO {
   isStudyFolder?: boolean;
 }
 
+export interface WorkspaceDTO {
+  name: string;
+  disk_name: string;
+}
+
```
hdinia: this is optional but best (we aim to eliminate `snake_case` from the frontend) 

you can use camel case

```ts
export interface WorkspaceDTO {
  name: string;
  diskName: string; 
}
```

and transform at the api call layer

```ts
export const getWorkspaces = async () => {
  const res = await client.get<any[]>(`/v1/private/explorer/_list_workspaces`);

  return res.data.map(workspace => ({
    name: workspace.name,
    diskName: workspace.disk_name // Transform to camelCase
  }));
};
```


or even better make the API return `CamelCase` response

hdinia (changes requested):

antarest/study/storage/explorer_service.py
smailio: Not that I'm aware of. 

That's how the windows API works https://learn.microsoft.com/fr-fr/windows/win32/api/fileapi/nf-fileapi-getvolumeinformationw 

I found an external library that does the same thing, but it just calls this API under the hood.

antarest/study/storage/explorer_service.py
smailio: We don't really have a concept of disk name on ubuntu as the discs are mounted as folder, so we'd display the folder name.

smailio: I added a comment

antarest/study/model.py
```diff
@@ -107,8 +117,18 @@ class StudyGroup(Base):  # type:ignore
     __tablename__ = "group_metadata"
     __table_args__ = (PrimaryKeyConstraint("study_id", "group_id"),)
 
-    group_id: str = Column(String(36), ForeignKey("groups.id", ondelete="CASCADE"), index=True, nullable=False)
-    study_id: str = Column(String(36), ForeignKey("study.id", ondelete="CASCADE"), index=True, nullable=False)
+    group_id: str = Column(
```
MartinBelthle: Minor but this don't have to change. remove the trailing comma and it will be as it was before

antarest/study/model.py
```diff
@@ -135,8 +155,18 @@ class StudyTag(Base):  # type:ignore
     __tablename__ = "study_tag"
     __table_args__ = (PrimaryKeyConstraint("study_id", "tag_label"),)
 
-    study_id: str = Column(String(36), ForeignKey("study.id", ondelete="CASCADE"), index=True, nullable=False)
-    tag_label: str = Column(String(40), ForeignKey("tag.label", ondelete="CASCADE"), index=True, nullable=False)
+    study_id: str = Column(
```
MartinBelthle: Same

antarest/study/storage/explorer_service.py
```diff
@@ -78,14 +80,53 @@ def list_dir(
...
-        ]
+        result = []
+        for workspace_name in self.config.storage.workspaces.keys():
+            if workspace_name == DEFAULT_WORKSPACE_NAME:
+                continue
+
+            if os.name == "nt":
+                # On Windows, we use Windows-specific APIs (via `get_volume_label`) to retrieve the disk name
+                # for each workspace.
+                drive_letter = os.path.splitdrive(workspace_name)[0] + "\\"
```
MartinBelthle: We should put this line inside the `get_volume_label` method to put all the crap at the same place

antarest/study/storage/explorer_service.py
MartinBelthle: Okay. if the external library hides this and is popular I think we should use it. Otherwise, we can keep it as is

antarest/study/storage/explorer_service.py
```diff
@@ -78,14 +80,53 @@ def list_dir(
...
+        ctypes.byref(serial_number),
+        ctypes.byref(max_component_length),
+        ctypes.byref(file_system_flags),
+        file_system_name_buffer,
+        ctypes.sizeof(file_system_name_buffer),
+    )
+
+    if result:
+        return volume_name_buffer.value
+    return drive_letter  # fallback
```
MartinBelthle: Perhaps add a comment saying why could we have a fallback

MartinBelthle (commented):
Minor review

hdinia (dismissed):

antarest/study/storage/explorer_service.py
smailio: Pywin32 has 5k stars and is maintained by a CPython contributor, so it's solid. Still, I’d prefer not to add a dependency just to hide a single function.

MartinBelthle (approved):

PR #2576, opened 2025-07-08
build(sqlalchemy): bump to 2.0

TheoPascoli:
- Changes the runtime to sqlalchemy 2.0
- Add some typing to ORM classes fields

A future work should include using `DeclarativeBase` in order to have full typing on ORM classes.


antarest/core/filetransfer/service.py
```diff
@@ -103,7 +103,7 @@ def set_ready(self, download_id: str, use_notification: bool = True) -> None:
         if not download:
             raise FileDownloadNotFound()
 
-        download.ready = True
+        download.ready = True  # type: ignore[assignment]
```
MartinBelthle: You could remove these type ignores by doing this:
`ready: Mapped[bool] = mapped_column(Boolean, default=False)`
Same for other type ignores in the code

antarest/core/tasks/model.py
```diff
@@ -15,9 +15,10 @@
 from enum import Enum, StrEnum
 from typing import TYPE_CHECKING, Any, List, Mapping, Optional
 
-from sqlalchemy import Boolean, Column, DateTime, ForeignKey, Integer, Sequence, String  # type: ignore
-from sqlalchemy.engine.base import Engine  # type: ignore
-from sqlalchemy.orm import relationship, sessionmaker  # type: ignore
+from sqlalchemy import Boolean, Column, DateTime, ForeignKey, Integer, Sequence, String
+from sqlalchemy.engine.base import Engine
+from sqlalchemy.orm import Mapped, relationship, sessionmaker
+from sqlalchemy.orm._orm_constructors import mapped_column
```
MartinBelthle: We shouldn't use this private package, you can just import `mapped_column` from `sqlalchemy.orm`

antarest/login/service.py
```diff
@@ -88,7 +88,7 @@ def save_group(self, group: Group) -> Group:
         Returns: group
 
         """
-        if self.groups.get_by_name(group.name):
+        if self.groups.get_by_name(cast(str, group.name)):
```
MartinBelthle: Same here, to avoid the cast you can just do `name: Mapped[str] = mapped_column(String(255))`

MartinBelthle (changes requested):
Looks really nice :), we can just go a bit further and remove other `type ignores` and `cast`

antarest/core/filetransfer/service.py
```diff
@@ -191,7 +191,7 @@ def _clean_up_expired_downloads(self, file_downloads: List[FileDownload]) -> Non
                     f"Failed to remove file download {file_download.path}",
                     exc_info=e,
                 )
-            self.repository.delete(file_download.id)
+            self.repository.delete(str(file_download.id))
```
MartinBelthle: Seems this str(...) is useless now

antarest/launcher/service.py
```diff
@@ -584,10 +584,10 @@ def export_task(_: ITaskNotifier) -> TaskResult:
                 try:
                     #
                     archive_dir(output_path, export_path, archive_format=ArchiveFormat.ZIP)
-                    self.file_transfer_manager.set_ready(export_id)
+                    self.file_transfer_manager.set_ready(str(export_id))
```
MartinBelthle: Same remark here and 3 lines after

antarest/login/service.py
MartinBelthle: Same

antarest/matrixstore/model.py
```diff
@@ -156,24 +156,24 @@ class MatrixDataSet(Base):  # type: ignore
...
-    created_at: datetime.datetime = Column(DateTime)
-    updated_at: datetime.datetime = Column(DateTime)
+    public: Mapped[bool] = mapped_column(Boolean, default=False)
+    created_at: Mapped[datetime.datetime] = mapped_column(DateTime)
+    updated_at: Mapped[datetime.datetime] = mapped_column(DateTime)
 
-    owner: Identity = relationship(Identity)
-    groups = relationship(
+    owner: Mapped[Identity] = relationship(Identity)
+    groups: Mapped[List[Group]] = relationship(  # todo check if ok
```
MartinBelthle: What about this todo ?

antarest/matrixstore/service.py
```diff
@@ -513,13 +513,13 @@ def download_matrix_list(self, matrix_list: Sequence[str], dataset_name: str) ->
         def export_task(notifier: ITaskNotifier) -> TaskResult:
             try:
                 self.create_matrix_files(matrix_ids=matrix_list, export_path=export_path)
-                self.file_transfer_manager.set_ready(export_id)
+                self.file_transfer_manager.set_ready(str(export_id))
```
MartinBelthle: Same

antarest/service_creator.py
```diff
@@ -95,15 +95,15 @@ def init_db_engine(
         extra["poolclass"] = NullPool
     elif not config.db.db_url.startswith("sqlite"):
         if config.db.pool_pre_ping:
-            extra["pool_pre_ping"] = True
+            extra["pool_pre_ping"] = True  # type: ignore[assignment]
         if config.db.pool_recycle:
-            extra["pool_recycle"] = config.db.pool_recycle
+            extra["pool_recycle"] = config.db.pool_recycle  # type: ignore[assignment]
```
MartinBelthle: To avoid all these `type: ignore` you could just put `extra: dict[str, Any] = {}` at the start

antarest/study/storage/abstract_storage_service.py
```diff
@@ -70,7 +70,8 @@ def get_study_information(
         additional_data = study.additional_data or StudyAdditionalData()
 
         try:
-            patch = Patch.model_validate(from_json(additional_data.patch or "{}"))
+            patch_str = str(additional_data.patch or "{}")
```
MartinBelthle: Don't need the str()

antarest/study/storage/rawstudy/model/filesystem/matrix/date_serializer.py
```diff
@@ -153,7 +153,7 @@ def extract_date(self, df: pd.DataFrame) -> Tuple[pd.Index[str], pd.DataFrame]:
         to_remove = cast(Sequence[Hashable], df.columns[0:4])
         body = df.drop(to_remove, axis=1)
 
-        return pd.Index(date), body
+        return pd.Index(date.astype(str)), body
```
MartinBelthle: Don't think we need to change this, simply bump your pandas stubs version and the warning should disappear

antarest/study/storage/df_download.py
```diff
@@ -91,15 +91,15 @@ def _export_file(
 
     try:
         writer(export_path)
-        file_transfer_manager.set_ready(export_id, use_notification=False)
+        file_transfer_manager.set_ready(str(export_id), use_notification=False)
```
MartinBelthle: Same. I'm not putting this every time, I'll let you check

MartinBelthle (changes requested):
Still several `type: ignore` and `str() or cast()` we can avoid

Also, minor but we could renmae the PR `build()` instead of `chore()` and `bump` instead of `upgrade`

antarest/login/service.py
```diff
@@ -625,7 +625,12 @@ def get_all_users(self, details: bool = False) -> List[Union[UserInfo, IdentityD
         user_mapping_id_to_name = {user.id: user.name for user in all_users}
         if details:
             return [
-                IdentityDTO(id=id, name=user_mapping_id_to_name[id], roles=roles_per_user[id]) for id in roles_per_user
+                IdentityDTO(
```
MartinBelthle: Minor: could be in one line as it was before

antarest/study/storage/abstract_storage_service.py
```diff
@@ -70,7 +70,8 @@ def get_study_information(
         additional_data = study.additional_data or StudyAdditionalData()
 
         try:
-            patch = Patch.model_validate(from_json(additional_data.patch or "{}"))
+            patch_str = additional_data.patch or "{}"
+            patch = Patch.model_validate(from_json(patch_str))
```
MartinBelthle: Minor: could be in one line as it was before

antarest/study/model.py
```diff
@@ -350,14 +361,14 @@ class RawStudy(Study):
...
         String(36),
         ForeignKey("study.id", ondelete="CASCADE"),
         primary_key=True,
     )
-    content_status: StudyContentStatus = Column(Enum(StudyContentStatus))
-    workspace: str = Column(String(255), default=DEFAULT_WORKSPACE_NAME, nullable=False, index=True)
-    missing = Column(DateTime, nullable=True, index=True)
+    content_status: Mapped[StudyContentStatus] = mapped_column(Enum(StudyContentStatus), nullable=True)
+    workspace: Mapped[str] = mapped_column(String(255), default=DEFAULT_WORKSPACE_NAME, nullable=False, index=True)
+    missing: Mapped[DateTime] = mapped_column(DateTime, nullable=True, index=True)
```
MartinBelthle: This should be typed as an optional as nullable=True. Same for content_status just above

tests/variantstudy/model/test_variant_model.py
```diff
@@ -307,7 +307,7 @@ def check_orm_operations(orm_execute_state):
         with current_user_context(jwt_user):
             variant_study_service.append_commands(variant_study.id, commands)
 
-        nb_queries_before = nb_queries  # store initial state
+        nb_queries = 0  # Reset query counter for the assertion
         with current_user_context(jwt_user):
             variant_study_service.get_commands(variant_study.id)  # execute database query
-        assert nb_queries_before + 1 == nb_queries  # compare with initial state to make sure database was queried once
+        assert nb_queries == 2  # Ensure only two queries were made (one for study, one for user)
```
MartinBelthle, line 313: Not sure to understand how the code worked before as it stated we want only 1 request and now wa assert we make 2

antarest/study/model.py
MartinBelthle: Also you should put `Mapped[datetime]` instead of `Mapped[DateTime]` to make mypy understand

antarest/study/service.py
```diff
@@ -1539,7 +1540,7 @@ def change_owner(self, study_id: str, owner_id: int) -> None:
             )
         )
 
-        owner_name = None if new_owner is None else new_owner.name
+        owner_name = None if new_owner is None else str(new_owner.name)
```
MartinBelthle: Is it to do "None" ? I'm not sure. If no we should simply put `new_owner.name`

MartinBelthle (changes requested):
Last round of comments, after that we're good to go :)

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -118,7 +118,7 @@ def update_from_raw_meta(
             )
             if fallback_on_default is not None:
                 metadata.name = metadata.name or "unnamed"
-                metadata.version = metadata.version or 0
+                metadata.version = metadata.version or 0  # type: ignore[assignment]
```
sylvlecl: For what it's worth we can replace 0 with "0.0" here and remove the type ignore

antarest/study/storage/utils.py
```diff
@@ -72,20 +72,14 @@
...
-        study_tree: The study tree object.
-        update_author: Specifies whether the author should be modified or not.
-            The author's name should be updated when the study is created,
-            but it is not changed if the study is copied.
-    """
     study_data_info = study_tree.get(["study"])
     study_data_info["antares"]["caption"] = metadata.name
-    study_data_info["antares"]["created"] = metadata.created_at.timestamp()
-    study_data_info["antares"]["lastsave"] = metadata.updated_at.timestamp()
+    study_data_info["antares"]["created"] = (
```
sylvlecl: I don't think we want to set the creation time to "now" if it is not defined. Same for update.
I guess before it would just have raised (calling a method on a null object), we could keep that behaviour. Or just write an empty string.

antarest/matrixstore/model.py
```diff
@@ -15,13 +15,13 @@
...
+from sqlalchemy.orm import Mapped, mapped_column, relationship
 from typing_extensions import override
 
 from antarest.core.persistence import Base
 from antarest.core.serde import AntaresBaseModel
-from antarest.login.model import GroupDTO, Identity, UserInfo
+from antarest.login.model import Group, GroupDTO, Identity, UserInfo
 
 
 class Matrix(Base):  # type: ignore
```
sylvlecl, line 27: Note that to benefit from the whole typing allowed by sqlalchemy 2.0, we need to use [DeclarativeBase](https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#step-one-orm-declarative-base-is-superseded-by-orm-declarativebase) and remove all type ignores on persistence classes.

We should probably do it in another PR, because it can be a little challenging, in particular for the `Study` class and its children :)

antarest/core/configdata/repository.py
```diff
@@ -28,22 +30,19 @@ def save(self, configdata: ConfigData) -> ConfigData:
...
-            .first()
-        )
-        return configdata
+        ).scalar_one_or_none()
 
     def get_json(self, key: str, owner: Optional[int] = None) -> Optional[JSON]:
         configdata = self.get(key, owner)
         if configdata:
-            data: JSON = from_json(configdata.value)
+            data: JSON = from_json(cast(str, configdata.value))
```
sylvlecl: I don't think we should cast here, nothing guarantees we have a string:
I guess if the value is None, we should return None

antarest/study/model.py
```diff
@@ -246,27 +253,29 @@ class Study(Base):  # type: ignore
...
-    folder = Column(String, nullable=True, index=True)
-    parent_id = Column(String(36), ForeignKey("study.id", name="fk_study_study_id"), index=True)
-    public_mode = Column(Enum(PublicMode), default=PublicMode.NONE)
-    owner_id = Column(Integer, ForeignKey(Identity.id), nullable=True, index=True)
-    archived = Column(Boolean(), default=False, index=True)
-
-    tags: List[Tag] = relationship(Tag, secondary=StudyTag.__table__, back_populates="studies")
+    name: Mapped[Optional[str]] = mapped_column(String(255), nullable=True, index=True)
+    type: Mapped[str] = mapped_column(String(50), index=True)
+    version: Mapped[str] = mapped_column(String(255), index=True, default="9.2")
```
sylvlecl: Not a big fan of adding this default value which should not be used in production.
I think you have mainly added it because of tests that don't define a version, but it would probably be better to fix those places instead.
If it's easier, we could have a utility method where version is optional instead of the plain constructor call

antarest/study/model.py
```diff
@@ -246,27 +253,29 @@ class Study(Base):  # type: ignore
...
+    author: Mapped[Optional[str]] = mapped_column(String(255), nullable=True)
+    created_at: Mapped[Optional[datetime]] = mapped_column(DateTime, nullable=True, index=True)
+    updated_at: Mapped[Optional[datetime]] = mapped_column(DateTime, nullable=True, index=True)
+    last_access: Mapped[Optional[datetime]] = mapped_column(DateTime, nullable=True)
+    path: Mapped[str] = mapped_column(String(), default=DEFAULT_WORKSPACE_NAME)
```
sylvlecl: Here too; not fan of adding a default value, it should not be used in the application itself

antarest/study/service.py
```diff
@@ -971,7 +971,7 @@ def sync_studies_on_disk(
                         study.path,
                         MAX_MISSING_STUDY_TIMEOUT,
                     )
-                    study.missing = now
+                    study.missing = now  # type: ignore[assignment]
```
sylvlecl: I think this should be removed and missing be typed as `Mapped[Optional[datetime]]` instead of `Mapped[Optional[DateTime]]`

antarest/study/service.py
```diff
@@ -980,7 +980,8 @@ def sync_studies_on_disk(
                             permissions=PermissionInfo.from_study(study),
                         )
                     )
-                if study.missing < clean_up_missing_studies_threshold:
+
+                if study.missing and cast(datetime, study.missing) < clean_up_missing_studies_threshold:
```
sylvlecl: we should not need to cast after the change above

antarest/study/storage/variantstudy/model/dbmodel.py
```diff
@@ -40,13 +40,13 @@ class VariantStudySnapshot(Base):  # type: ignore
...
 
-    id: str = Column(
+    id: Mapped[str] = mapped_column(
         String(36),
         ForeignKey("variantstudy.id", ondelete="CASCADE"),
         primary_key=True,
     )
-    created_at: datetime.date = Column(DateTime)
-    last_executed_command: Optional[str] = Column(String(), nullable=True)
+    created_at: Mapped[datetime.date] = mapped_column(DateTime)
```
sylvlecl: it's very weird that it's typed as date and not datetime, I think it was a mistake ?
In database we have a datetime

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -124,8 +124,10 @@ def _get_user_name_from_id(user_id: int) -> str:
             user_id: user id (user must exist)
         Returns: String representing the user's name
         """
-        user_obj: Identity = db.session.query(Identity).get(user_id)
-        return user_obj.name  # type: ignore  # `name` attribute is always a string
+        user_obj: Identity | None = db.session.get(Identity, user_id)
+        if user_obj is None:
+            raise ValueError(f"User with id {user_id} not found")
+        return str(user_obj.name)
```
sylvlecl, line 130: We  should probably handle the case where it's None explicitly, either by returning an empty string or "Unnamed" or something like that.

antarest/study/web/variant_blueprint.py
```diff
@@ -82,8 +82,12 @@ def create_variant(uuid: str, name: str) -> str:
                             "antares": {
                                 "version": variant_study.version,
                                 "caption": variant_study.name,
-                                "created": variant_study.created_at.timestamp(),
-                                "lastsave": variant_study.created_at.timestamp(),
+                                "created": variant_study.created_at.timestamp()
+                                if variant_study.created_at
```
sylvlecl: I think we should raise if those fields are not defined here. We have create the variant_study just above, they must be defined. The database schema is too lax, but we should be more rigorous if we know the application should ensure non-nulls.
We can add getters to the study class that return a non optional or raise

sylvlecl (changes requested):
Great!
A few requests though, mainly about "optionals" handling and default values.

Also, there will be some further work required to benefit completely of the typing, by removing type ignores on ORM classes.
We should implement this in another PR because it will require quite a lot of changes in the code, in particular our use of the `Study` hierarchy is wrong in many places.

antarest/core/configdata/repository.py
TheoPascoli: I added a check and kept the cast

tests/variantstudy/model/test_variant_model.py
TheoPascoli: No idea :D
@sylvlecl Any idea ?

sylvlecl: The second select is for study.commands, which are lazily loaded.

Maybe in 1.4 the event was slightly different for lazily executed selects ?

sylvlecl: Actually no, it was the first select to get the study from the repo which was not performed. Don't know why yet, maybe it was already present in the session "state".

sylvlecl: Indeed from the doc:
```
        :meth:`_query.Query.get` is special in that it provides direct
        access to the identity map of the owning :class:`.Session`.
        If the given primary key identifier is present
        in the local identity map, the object is returned
        directly from this collection and no SQL is emitted,
        unless the object has been marked fully expired.
        If not present,
        a SELECT is performed in order to locate the object.
```

sylvlecl (approved):

PR #2577, opened 2025-07-08
feat(storage): add coherence checks for v9.2

MartinBelthle:


antarest/study/business/model/sts_model.py
```diff
@@ -213,18 +213,36 @@ def initialize_st_storage(storage: STStorage, version: StudyVersion) -> None:
         _initialize_field_default(storage, "penalize_variation_withdrawal", False)
 
 
+def check_attributes_coherence(storage: STStorage, version: StudyVersion) -> None:
```
TheoPascoli, line 216: Maybe you can add a little docstring to clarify the business rule behind this

antarest/study/business/model/sts_model.py
```diff
@@ -213,18 +213,36 @@ def initialize_st_storage(storage: STStorage, version: StudyVersion) -> None:
...
+    if version < STUDY_VERSION_9_2:
+        if storage.efficiency > 1:
+            raise ShortTermStorageValuesCoherenceError(
+                storage.id, f"Prior to v9.2, efficiency must be lower than 1 and was {storage.efficiency}"
+            )
+    else:
+        efficiency_withdrawal = storage.efficiency_withdrawal or 1
```
TheoPascoli: if efficiency_withdrawal = 0 then 1 will be affected. Is it wanted ? Or do you want to check if its none instead

TheoPascoli (changes requested):

antarest/study/business/model/sts_model.py
MartinBelthle: Indeed, i want to check if it's None I'll change it

PR #2579, opened 2025-07-09
feat(generalconfig): add dao for config general

TheoPascoli:


antarest/study/business/model/config/general_model.py
```diff
@@ -0,0 +1,142 @@
...
+    AUTOMATIC = "Automatic"
+    CUSTOM = "Custom"
+    DERATED = "Derated"
+
+
+DayNumberType: TypeAlias = Annotated[int, Field(ge=1, le=366)]
+GENERAL = "general"
+OUTPUT = "output"
+GENERAL_PATH = f"{GENERAL_DATA_PATH}/{GENERAL}"
+OUTPUT_PATH = f"{GENERAL_DATA_PATH}/{OUTPUT}"
```
MartinBelthle: You dont' use `OUTPUT_PATH`, nor `BUILDING_MODE`

tests/integration/study_data_blueprint/test_config_general.py
```diff
@@ -45,6 +45,7 @@ def test_get_general_form_values(
             "firstJanuary": "Monday",
             "firstMonth": "january",
             "firstWeekDay": "Monday",
+            "geographicTrimming": False,
```
MartinBelthle: This is because you removed the versioning part that existed inside the model. We should put it back even if it's for v7.1

tests/variantstudy/model/command/test_update_general_config.py
```diff
@@ -0,0 +1,193 @@
...
+    WeekDay,
+)
+from antarest.study.model import STUDY_VERSION_8_8
+from antarest.study.storage.variantstudy.model.command.update_general_config import UpdateGeneralConfig
+
+
+@pytest.mark.unit_test
+def test_update_general_config_apply_dao(command_context):
+    # Mock StudyDao
+    mock_study_dao = Mock()
```
MartinBelthle: I think these tests are the kind of test we won't understand in several years.
IMO you should use the empty_study fixture and really check what's written inside the ini file

antarest/study/storage/rawstudy/model/filesystem/config/general.py
```diff
@@ -0,0 +1,63 @@
...
+    leap_year: Optional[StrictBool] = Field(default=None, alias="leapyear")
+    nb_years: Optional[PositiveInt] = Field(default=None, alias="nbyears")
+    building_mode: Optional[BuildingMode] = Field(default=None)
+    selection_mode: Optional[StrictBool] = Field(default=None, alias="user-playlist")
+    year_by_year: Optional[StrictBool] = Field(default=None, alias="year-by-year")
+    simulation_synthesis: Optional[StrictBool] = Field(default=None, alias="synthesis")
+    mc_scenario: Optional[StrictBool] = Field(default=None, alias="storenewset")
+    filtering: Optional[StrictBool] = Field(default=None)
+    geographic_trimming: Optional[StrictBool] = Field(default=None, alias="geographic-trimming")
+    thematic_trimming: Optional[StrictBool] = Field(default=None, alias="thematic-trimming")
```
MartinBelthle: Why using StrictBool instead of just bool ?

antarest/study/storage/rawstudy/model/filesystem/config/general.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def to_model(self) -> GeneralConfig:
+        return GeneralConfig.model_validate(self.model_dump(exclude_none=True))
+
+    @classmethod
+    def from_model(cls, storage: GeneralConfig) -> "GeneralFileData":
```
MartinBelthle: you should rename storage into config

antarest/study/business/model/config/general_model.py
```diff
@@ -0,0 +1,142 @@
...
+    year_by_year: StrictBool = False
+    simulation_synthesis: StrictBool = True
+    mc_scenario: StrictBool = False
+    filtering: Optional[bool] = None
+    geographic_trimming: StrictBool = False
+    thematic_trimming: StrictBool = False
+
+    @model_validator(mode="before")
+    def day_fields_validation(cls, values: Dict[str, Any] | ValidationInfo) -> Dict[str, Any]:
+        new_values = values if isinstance(values, dict) else values.data
```
MartinBelthle: I launched all tests and we never have the case where values isn't a dict, we should remove this

antarest/study/business/model/config/general_model.py
```diff
@@ -0,0 +1,142 @@
...
+    simulation_synthesis: Optional[StrictBool] = None
+    mc_scenario: Optional[StrictBool] = None
+    filtering: Optional[StrictBool] = None
+    geographic_trimming: Optional[StrictBool] = None
+    thematic_trimming: Optional[StrictBool] = None
+
+
+def update_general_config(config: GeneralConfig, new_config: GeneralConfigUpdate) -> GeneralConfig:
+    """
+    Updates a link according to the provided update data.
```
MartinBelthle: Change the doc

antarest/study/storage/rawstudy/model/filesystem/config/general.py
```diff
@@ -0,0 +1,63 @@
...
+        return cls.model_validate(storage.model_dump(exclude={"id"}))
+
+
+def get_general_config(config: GeneralConfig) -> Dict[str, Any]:
```
MartinBelthle: The 2 following methods don't seem to have the right name, there are serialization methods

antarest/study/dao/file/file_study_general_config_dao.py
```diff
@@ -0,0 +1,101 @@
...
+                return BuildingMode.CUSTOM
+            return BuildingMode.AUTOMATIC
+
+        file_study = self.get_file_study()
+        general_data = file_study.tree.get(GENERAL_DATA_PATH.split("/"))
+        general = general_data.get(GENERAL, {})
+        output = general_data.get(OUTPUT, {})
+        study_ver = file_study.config.version
+
+        config_values: Dict[str, Any] = {
```
MartinBelthle: You should use a parser and a deserializer here instead of doing this

MartinBelthle (changes requested):
Several remarks

antarest/study/storage/rawstudy/model/filesystem/config/general.py
TheoPascoli: It was StrictBool before, I changed to bool but it doesnt matter actually

TheoPascoli:
> Several remarks

Thanks for the feedback, I fixed everything

antarest/study/dao/file/file_study_general_config_dao.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @override
+    def get_general_config(self) -> GeneralConfig:
+        file_study = self.get_file_study()
+
+        tree_data = file_study.tree.get(["settings", "generaldata"])
+        config_data = tree_data.get("general", {})
+        config_data.update(tree_data.get("output", {}))
+
+        file_data = GeneralFileData.model_validate(config_data)
```
MartinBelthle: You should use the parsing  method you introduced instead of doing this

antarest/study/dao/file/file_study_general_config_dao.py
```diff
@@ -0,0 +1,56 @@
...
+        return file_data.to_model()
+
+    @override
+    def save_general_config(self, config: GeneralConfig) -> None:
+        study_data = self.get_file_study()
+
+        current_general_config = study_data.tree.get(["settings", "generaldata", "general"])
+        general_config = serialize_simulation_config(config, study_data.config.version)
+        general_config.update({k: v for k, v in current_general_config.items() if k not in general_config})
+        study_data.tree.save(general_config, ["settings", "generaldata", "general"])
```
MartinBelthle: Doing 2 `tree.save` is not optimal as we're talking about the same file.
To me we should read the entire content, modfiy it with the serializing methods and save the whole content once

antarest/study/business/model/config/general_model.py
```diff
@@ -0,0 +1,138 @@
...
+
+class BuildingMode(EnumIgnoreCase):
+    AUTOMATIC = "Automatic"
+    CUSTOM = "Custom"
+    DERATED = "Derated"
+
+
+DayNumberType: TypeAlias = Annotated[int, Field(ge=1, le=366)]
+GENERAL = "general"
+OUTPUT = "output"
```
MartinBelthle: You're not using the `OUTPUT` variable

tests/variantstudy/model/command/test_update_general_config.py
```diff
@@ -0,0 +1,73 @@
...
+# This file is part of the Antares project.
+from antarest.study.business.model.config.general_model import GeneralConfigUpdate
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.variantstudy.model.command.update_general_config import UpdateGeneralConfig
+from antarest.study.storage.variantstudy.model.command_context import CommandContext
+
+
+class TestUpdateGeneralConfig:
+    def _set_up(self, study: FileStudy, command_context: CommandContext) -> None:
+        self.study = study
```
MartinBelthle: You don't use this method

tests/variantstudy/model/command/test_update_general_config.py
```diff
@@ -0,0 +1,73 @@
...
+        properties = GeneralConfigUpdate.model_validate(args)
+
+        command = UpdateGeneralConfig(
+            parameters=properties, command_context=command_context, study_version=study.config.version
+        )
+        output = command.apply(study_data=study)
+        assert output.status
+        default_values.update({"horizon": 2030, "building_mode": "Automatic"})
+
+        general_config = study.tree.get(["settings", "generaldata", "general"])
```
MartinBelthle, line 37: We should also check the ini content before the update to do the comparison and understand the test better

antarest/study/business/model/config/general_model.py
```diff
@@ -0,0 +1,138 @@
...
+    thematic_trimming: bool | None = None
+
+    @model_validator(mode="before")
+    def day_fields_validation(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        first_day = values.get("first_day")
+        last_day = values.get("last_day")
+        leap_year = values.get("leap_year")
+        day_fields = [first_day, last_day, leap_year]
+
+        if all(v is None for v in day_fields):
```
MartinBelthle, line 90: Not sure to understand these checks on `None`. I see they existed before but they seem weird here, even the comment doesn't really make sense with the class. I'll let Sylvain do a review to give his opinion on this

MartinBelthle (changes requested):
I'll let Sylvain take a look

antarest/study/dao/file/file_study_general_config_dao.py
TheoPascoli: What parsing method are you refereing to ? With this, I'm sur to validate all the fields I need and the "to_model" converts it to the wanted object

antarest/study/dao/file/file_study_general_config_dao.py
TheoPascoli: yes thanks

antarest/study/business/model/config/general_model.py
TheoPascoli: I think if only one is modified, it becomes inconsistent. That's why it has a "all or nothing" logic.

antarest/study/dao/file/file_study_general_config_dao.py
MartinBelthle: I though you introduced one but you didn't. We should introduce one in the `general.py` file : `parse_output_and_simulation_config` or something like this.

Because as I said inside Maurane PR this class shouldn't be imported in other files, just the parsing / serializing method should (see thermal cluster example)

antarest/study/business/model/config/general_model.py
```diff
@@ -0,0 +1,137 @@
...
+GENERAL_PATH = f"{GENERAL_DATA_PATH}/{GENERAL}"
+
+
+class GeneralConfig(AntaresBaseModel):
+    model_config = ConfigDict(alias_generator=to_camel, extra="forbid", populate_by_name=True)
+
+    mode: Mode = Mode.ECONOMY
+    first_day: DayNumberType = 1
+    last_day: DayNumberType = 365
+    horizon: str | int = ""
```
sylvlecl: I think we should only have a string here, we should avoid to have weakly typed fields.
We may need to enforce it in the parsing code.

antarest/study/dao/file/file_study_general_config_dao.py
```diff
@@ -0,0 +1,58 @@
...
+        tree_data = file_study.tree.get(["settings", "generaldata"])
+
+        return parse_general_config(tree_data)
+
+    @override
+    def save_general_config(self, config: GeneralConfig) -> None:
+        study_data = self.get_file_study()
+
+        current_config = study_data.tree.get(["settings", "generaldata"])
+        current_general_config = current_config["general"]
```
sylvlecl, line 46: We don't need to get the current values: it's not an update, the config should completely override the existing one

antarest/study/dao/file/file_study_general_config_dao.py
```diff
@@ -0,0 +1,58 @@
...
+        current_output_config = current_config["output"]
+
+        general_config = serialize_simulation_config(config, study_data.config.version)
+        general_config.update({k: v for k, v in current_general_config.items() if k not in general_config})
```
sylvlecl, line 50: see above, the current config should be completely replaced by the new one

antarest/study/dao/file/file_study_general_config_dao.py
```diff
@@ -0,0 +1,58 @@
...
+
+        return parse_general_config(tree_data)
+
+    @override
+    def save_general_config(self, config: GeneralConfig) -> None:
+        study_data = self.get_file_study()
+
+        current_config = study_data.tree.get(["settings", "generaldata"])
+        current_general_config = current_config["general"]
+        current_output_config = current_config["output"]
```
sylvlecl, line 47: Same here, it's not an update, we don't wan it

antarest/study/business/model/config/general_model.py
sylvlecl: Ok to keep this for me

antarest/study/business/model/config/general_model.py
```diff
@@ -0,0 +1,137 @@
...
+    first_week_day: WeekDay = WeekDay.MONDAY
+    first_january: WeekDay = WeekDay.MONDAY
+    leap_year: bool = False
+    nb_years: PositiveInt = 1
+    building_mode: BuildingMode = BuildingMode.AUTOMATIC
+    selection_mode: bool = False
+    year_by_year: bool = False
+    simulation_synthesis: bool = True
+    mc_scenario: bool = False
+    filtering: bool | None = None
```
sylvlecl, line 79: Looks like this is forbidden for studies >= 7.10, we should check and initialize them to None

antarest/study/business/model/config/general_model.py
```diff
@@ -0,0 +1,137 @@
...
+    geographic_trimming: bool | None = None
```
sylvlecl, line 80: Both trimming parameters are required for studies >= 7.10: we should initialize those values like we do for other objects (thermal clusters etc), when we know the study version.

sylvlecl (changes requested):
Thanks!
A few small change requests:
- we need to initialize the "trimming" fields
- the DAO should not update but replace the config
- add some serialziation/parsing tests for the command

tests/variantstudy/model/command/test_update_general_config.py
```diff
@@ -0,0 +1,39 @@
...
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+from antarest.study.business.model.config.general_model import GeneralConfigUpdate
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.variantstudy.model.command.update_general_config import UpdateGeneralConfig
+from antarest.study.storage.variantstudy.model.command_context import CommandContext
+
+
+class TestUpdateGeneralConfig:
```
sylvlecl, line 18: We should also add some test in test_command_factory to check the serialization and parsing to DB.

antarest/study/business/model/config/general_model.py
```diff
@@ -0,0 +1,167 @@
...
+
+    if version <= STUDY_VERSION_7_1:
+        for field in ["geographic_trimming", "thematic_trimming"]:
+            value = getattr(config, field)
+            if value is not None:
+                raise InvalidFieldForVersionError(f"Field '{field}' is not a valid field for study version {version}")
+
+
+def initialize_default_values(config: GeneralConfig, version: StudyVersion) -> None:
+    if version <= STUDY_VERSION_7_1:
```
sylvlecl: Here I thin it's filtering that we should initialize for version < 7.1

sylvlecl (approved):

PR #2581, opened 2025-07-10
feat(allocation): replace list matrix type by numpy array

TheoPascoli:


antarest/study/business/allocation_management.py
```diff
@@ -106,6 +108,17 @@ def validate_hydro_allocation_matrix(
...
 
+    @field_serializer("data")
+    def serialize_data(self, data: npt.NDArray[np.float64]) -> List[List[float]]:
+        result: List[List[float]] = data.tolist()
+        return result
+
+    @override
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, AllocationMatrix):
+            return NotImplemented
```
smailio, line 114: you mean raise ?

antarest/study/business/allocation_management.py
```diff
@@ -106,6 +108,17 @@ def validate_hydro_allocation_matrix(
 
         return data
 
+    @field_serializer("data")
+    def serialize_data(self, data: npt.NDArray[np.float64]) -> List[List[float]]:
+        result: List[List[float]] = data.tolist()
```
smailio: why not `return data.tolist()` ?

antarest/study/business/allocation_management.py
TheoPascoli: It it indeed return. Source : "If a rich comparison method returns NotImplemented, the interpreter will try the reflected operation on the other type..."

antarest/study/business/allocation_management.py
TheoPascoli: just for mypy

sylvlecl: Note that we can use `cast` for this:
```python
return cast(List[List[float], data.tolist())
```
maybe the intent is more explicit

antarest/study/business/allocation_management.py
```diff
@@ -69,15 +69,17 @@ class AllocationMatrix(FormFieldsBaseModel):
     data: 2D-array matrix of consumption coefficients
     """
 
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
     index: Annotated[List[str], Len(min_length=1)]
     columns: Annotated[List[str], Len(min_length=1)]
-    data: List[List[float]]  # NonNegativeFloat not necessary
+    data: npt.NDArray[np.float64]
```
sylvlecl: Actually I think we could re-use the `NpArray` annotated type which was introduced in `antarest.study.model`.

We should probably move it to a common file, for example in `core.serde`

sylvlecl (commented):
I think we could re-use and share an existing annotated type `NpArray` (which could be useful elsewhere probably)

antarest/study/business/allocation_management.py
TheoPascoli: I removed it completely because with the custom NpArray type there is already a serializer

sylvlecl (approved):

PR #2582, opened 2025-07-10
feat(app-version): refactor version endpoint

TheoPascoli:
/version endpoint is slow. 300ms to 1000ms before, 5ms now.
users dont care about dependencies

### DIFF
--- a/antarest/core/core_blueprint.py
+++ b/antarest/core/core_blueprint.py
@@ -45,9 +45,8 @@ def create_utils_routes(config: Config) -> APIRouter:
         "/version",
         tags=[APITag.misc],
         summary="Get application version",
-        response_model=VersionInfoDTO,
     )
-    def version_info(with_deps: bool = False) -> Any:
+    def version_info(with_deps: bool = False) -> VersionInfoDTO:
         """
         Returns the current version of the application, along with relevant dependency information.
 
--- a/antarest/core/core_blueprint.py
+++ b/antarest/core/core_blueprint.py
@@ -17,7 +17,7 @@ from fastapi import APIRouter
 from antarest.core.config import Config
 from antarest.core.serde import AntaresBaseModel
 from antarest.core.utils.web import APITag
-from antarest.core.version_info import VersionInfoDTO, get_commit_id, get_dependencies
+from antarest.core.version_info import VersionInfoDTO, get_commit_id
 
 
 class StatusDTO(AntaresBaseModel):
@@ -62,7 +62,6 @@ def create_utils_routes(config: Config) -> APIRouter:
         return VersionInfoDTO(
             version=antarest_version,
             gitcommit=get_commit_id(config.resources_path),
-            dependencies=get_dependencies(),
         )
 
     @bp.get(
--- a/antarest/core/core_blueprint.py
+++ b/antarest/core/core_blueprint.py
@@ -17,7 +17,7 @@ from fastapi import APIRouter
 from antarest.core.config import Config
 from antarest.core.serde import AntaresBaseModel
 from antarest.core.utils.web import APITag
-from antarest.core.version_info import VersionInfoDTO, get_commit_id
+from antarest.core.version_info import VersionInfoDTO, get_commit_id, get_dependencies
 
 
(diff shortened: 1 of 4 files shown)


MartinBelthle (commented):
I agree that the users don't care about the dependencies. But at first this endpoint was also designed for developers, to check dependencies that were embarked in production for example as we don't freeze our dependencies.

So IMO what we should do is add a flag to specify if we want to have the dependencies. By default it would be set to False and the front end would use it as is but it would still be possible to access them if we want to via the APIdoc.

What do you think ?

TheoPascoli:
> I agree that the users don't care about the dependencies. But at first this endpoint was also designed for developers, to check dependencies that were embarked in production for example as we don't freeze our dependencies.
> 
> So IMO what we should do is add a flag to specify if we want to have the dependencies. By default it would be set to False and the front end would use it as is but it would still be possible to access them if we want to via the APIdoc.
> 
> What do you think ?

I added an optional parameters if we want to get the dependencies as well

antarest/core/version_info.py
```diff
@@ -81,31 +82,24 @@ def get_last_commit_from_git() -> str:
...
-        # Due to PyInstaller renaming the executable to "AntaresWebServer",
-        # accessing the "python" executable becomes impossible, resulting in complications
-        # when trying to obtain the list of installed packages using `pip freeze`.
         return {}
 
     args = [sys.executable, "-m", "pip", "freeze"]
     output = subprocess.check_output(args, encoding="utf-8")
-    lines = (line for line in output.splitlines(keepends=False) if "==" in line)
-    # noinspection PyTypeChecker
+    lines = (line for line in output.splitlines() if "==" in line)
```
MartinBelthle, line 102: We could keep the `keepends=False` to avoi having different responses on WIndows and Unix

antarest/core/version_info.py
```diff
@@ -26,7 +27,7 @@ class VersionInfoDTO(AntaresBaseModel):
     name: str = "AntaREST"
     version: str
     gitcommit: str
-    dependencies: Dict[str, str]
+    dependencies: Dict[str, str] | None = None
```
MartinBelthle: I think in any case we have a dict, I don't see where you could have None

antarest/core/core_blueprint.py
```diff
@@ -47,7 +47,7 @@ def health() -> Any:
         summary="Get application version",
         response_model=VersionInfoDTO,
     )
-    def version_info() -> Any:
+    def version_info(with_deps: bool = False) -> Any:
```
MartinBelthle: Here we could type that the response is `VersionInfoDTO` and remove it in line 48

antarest/core/version_info.py
```diff
@@ -81,31 +82,24 @@ def get_last_commit_from_git() -> str:
 def get_dependencies() -> Dict[str, str]:
     """
     Retrieve the list of installed dependencies and their versions.
+    """
+    try:
+        packages = {dist.metadata["name"]: dist.version for dist in distributions()}
```
MartinBelthle: You could just do `dist.name` and add the check here to avoid iterating twice.
`return {dist.name: dist.version for dist in distributions() if dist.name.lower() != "antarest"}`

tests/core/test_version_info.py
```diff
@@ -15,24 +15,10 @@
 
 import pytest
 
-from antarest.core.version_info import get_commit_id, get_dependencies
+from antarest.core.version_info import get_commit_id
 
 
 class TestVersionInfo:
-    @pytest.mark.unit_test
-    def test_get_dependencies(self) -> None:
```
MartinBelthle: Why did you remove this test ? I think it was usefull

MartinBelthle (changes requested):

antarest/core/version_info.py
TheoPascoli: This is the default value so I removed it

MartinBelthle (approved):

MartinBelthle:
Perhaps we should clarify in the PR title that it's the **app** version endpoint and has nothing to do with the simulator version

PR #2583, opened 2025-07-10
feat(matrix_usage_provider): add usage provider

mehdiwahada:


antarest/matrixstore/matrix_usage_provider.py
```diff
@@ -0,0 +1,20 @@
...
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+from abc import ABC, abstractmethod
+
+from antarest.matrixstore.model import MatrixReference
+
+
+class IMatrixUsageProvider(ABC):
```
sylvlecl, line 17: We can add a docstring to explain a little what's the role of that interface:
it provides some information about which matrices are used by a client of the matrix service

antarest/matrixstore/model.py
```diff
@@ -135,6 +135,11 @@ def __eq__(self, other: Any) -> bool:
         return res
 
 
+class MatrixReference(AntaresBaseModel):
```
sylvlecl, line 138: Here too we can add a small docstring to explain what it represents:
it gives some information about one reference to a matrix from a client.

antarest/matrixstore/service.py
```diff
@@ -538,3 +557,36 @@ def download_matrix(self, matrix_id: str, filepath: Path) -> None:
...
+    def get_studies_matrices(self) -> Set[str]:
+        logger.info("Getting all matrices used in raw studies")
+
+        return {
+            matrix_reference.matrix_id
+            for provider in self.usage_providers
+            for matrix_reference in provider.get_matrix_usage()
+        }
+
+    def get_used_matrices(self) -> Set[str]:
```
sylvlecl: It would be simpler that this method relies **only** on the usage providers.
In order to achieve this, we should add a usage provider for dataset matrices too (see other comment).

antarest/matrixstore/service.py
```diff
@@ -538,3 +557,36 @@ def download_matrix(self, matrix_id: str, filepath: Path) -> None:
...
+    ) -> "MatrixGarbageCollector":
+        return MatrixGarbageCollector(
+            config=config,
+            study_service=study_service,
+            matrix_service=self,
+            matrices_usage_providers=self.usage_providers,
+        )
+
+    def get_studies_matrices(self) -> Set[str]:
+        logger.info("Getting all matrices used in raw studies")
```
sylvlecl: Actually here, you will not only get matrices from raw studies, but matrices from all the registered clients ("usage providers").
I think we could remove that method and have only the generic method `get_used_matrices` instead

antarest/matrixstore/service.py
```diff
@@ -538,3 +557,36 @@ def download_matrix(self, matrix_id: str, filepath: Path) -> None:
         """
         matrix = self.get(matrix_id)
         save_matrix(InternalMatrixFormat.TSV, matrix, filepath)
+
+    def get_datasets_matrices(self) -> Set[str]:
```
sylvlecl: In order to make the source of usages homogeneous, we'd better create a `UsageProvider` for that kind of usage too.

antarest/study/storage/rawstudy/raw_study_matrix_usage_provider.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class RawStudyMatrixUsageProvider(IMatrixUsageProvider):
+    def __init__(self, config: Config, matrix_service: MatrixService):
+        self.managed_studies_path: Path = config.storage.workspaces[DEFAULT_WORKSPACE_NAME].path
+        matrix_service.register_usage_provider(self)
+
+    @override
+    def get_matrix_usage(self) -> list[MatrixReference]:
+        # renvoyer la même chose que le get_matrices associé
```
sylvlecl: Let's remove that comment

antarest/study/storage/rawstudy/raw_study_matrix_usage_provider.py
```diff
@@ -0,0 +1,39 @@
...
+        logger.info("Getting all matrices used in raw studies")
+        matrices_references = []
+        for f in self.managed_studies_path.rglob("*.link"):
+            matrix_id = extract_matrix_id(f.read_text())
+            mat_ref = MatrixReference(matrix_id=matrix_id, use_description="")
```
sylvlecl: We should add a description here, like "Used by raw study {study_id}".
For this you will need to have the study ID. In order to achieve this, I would propose to change the implementation:
instead of recursing on the whole workspace path, we will be inspecting the studies folder one by one. We can get the list of all managed studies from the `StudyMetadataRepository`, and iterate on them (the `study` class knows the `path` where it's stored).

antarest/study/storage/variantstudy/business/matrix_constants/matrix_constants_usage_provider.py
```diff
@@ -0,0 +1,32 @@
...
+class ConstantsMatrixUsageProvider(IMatrixUsageProvider):
+    def __init__(self, variant_study_service: VariantStudyService, matrix_service: MatrixService):
+        self.matrix_constants = variant_study_service.command_factory.command_context.generator_matrix_constants
+        matrix_service.register_usage_provider(self)
+
+    @override
+    def get_matrix_usage(self) -> list[MatrixReference]:
+        matrices_references = []
+        for value in self.matrix_constants.hashes.values():
+            matrix_reference = MatrixReference(matrix_id=value, use_description="")
```
sylvlecl: Here we can add a description like "Constant matrix"

antarest/study/storage/variantstudy/command_matrix_usage_provider.py
```diff
@@ -0,0 +1,58 @@
...
+            return []
+
+        variant_study_commands = [cmd for c in command_blocks for cmd in transform_to_command(c.to_dto(), c.study_id)]
+        for block in command_blocks:
+            study_id = block.study_id
+            for command in variant_study_commands:
+                for matrix in command.get_inner_matrices():
+                    command_id = str(matrix)
+                    mat_reference = MatrixReference(
+                        matrix_id=command_id, use_description=f"Used by {command_id} from study {study_id}"
```
sylvlecl: we can add "Used by **command** {command_id} from **variant** study {study_id}"

antarest/study/storage/variantstudy/command_matrix_usage_provider.py
```diff
@@ -0,0 +1,58 @@
...
+from antarest.matrixstore.service import MatrixService
+from antarest.study.service import StudyService, logger
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand
+from antarest.study.storage.variantstudy.model.dbmodel import CommandBlock
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+class CommandMatrixUsageProvider(IMatrixUsageProvider):
+    def __init__(self, study_service: StudyService, matrix_service: MatrixService):
+        self.study_service = study_service
```
sylvlecl: the study_service itself is not used, we should not keep a reference to it

antarest/study/storage/variantstudy/command_matrix_usage_provider.py
```diff
@@ -0,0 +1,58 @@
...
+        self.variant_study_service = self.study_service.storage_service.variant_study_service
```
sylvlecl: Since here we only need the variant_study_service, we should have only that service in the constructor, not the whole `StudyService`. The constructor must tell what are the real dependencies of the object, and not more

antarest/study/business/link_management.py
```diff
@@ -53,8 +53,8 @@ def _create_update_link_command(
             area1=area_from,
             area2=area_to,
             parameters=link_update_dto,
-            command_context=self._command_context,
             study_version=study.version,
+            command_context=self._command_context,
```
sylvlecl, line 57: better to revert that change in order to keep a clean history

antarest/service_creator.py
```diff
@@ -222,23 +222,10 @@ def create_explorer(config: Config, app_ctxt: Optional[AppBuildContext]) -> Expl
 
 def create_matrix_gc(
```
sylvlecl: I think we can completely remove this method since it's a one-liner

antarest/study/storage/variantstudy/business/matrix_constants/matrix_constants_usage_provider.py
```diff
@@ -0,0 +1,32 @@
...
+from typing_extensions import override
+
+from antarest.matrixstore.matrix_usage_provider import IMatrixUsageProvider
+from antarest.matrixstore.model import MatrixReference
+from antarest.matrixstore.service import MatrixService
+from antarest.study.storage.variantstudy.variant_study_service import VariantStudyService
+
+
+class ConstantsMatrixUsageProvider(IMatrixUsageProvider):
+    def __init__(self, variant_study_service: VariantStudyService, matrix_service: MatrixService):
```
sylvlecl: We should only have the `matrix_constants` in the constructor, since the behaviour only depends on that

antarest/study/storage/variantstudy/business/matrix_constants/matrix_constants_usage_provider.py
```diff
@@ -0,0 +1,32 @@
...
+        self.matrix_constants = variant_study_service.command_factory.command_context.generator_matrix_constants
+        matrix_service.register_usage_provider(self)
```
sylvlecl, line 25: So now in the 3 usage providers we have created, we register them in their constructor, which is fine. BUT, those constructors are actually never called, so they don't exist when we start the application !

We need to create them at the start of the application, when the corresponding client services are created:
- construct and register this one when creating the `GeneratorMatrixConstants` service
- construct the "raw" one when creating the raw study service
- construct the "command" one when creating the variant study service
- and then construct the "dataset" one when the matrix service is created

antarest/matrixstore/matrix_garbage_collector.py
```diff
@@ -39,15 +37,15 @@ class MatrixGarbageCollector(IService):
...
+        matrices_usage_providers: list[IMatrixUsageProvider],
     ):
         super(MatrixGarbageCollector, self).__init__()
         self.saved_matrices_path: Path = config.storage.matrixstore
-        self.managed_studies_path: Path = config.storage.workspaces[DEFAULT_WORKSPACE_NAME].path
-        self.study_service: StudyService = study_service
         self.variant_study_service: VariantStudyService = study_service.storage_service.variant_study_service
         self.matrix_service = matrix_service
+        self.matrices_usage_providers = matrices_usage_providers
         self.dataset_repository: MatrixDataSetRepository = matrix_service.repo_dataset
```
sylvlecl: Note used anymore, it should be removed, as well as the `variant_study_service`, the `dataset_repository` and `matrix_constants`

antarest/matrixstore/matrix_garbage_collector.py
```diff
@@ -39,15 +37,15 @@ class MatrixGarbageCollector(IService):
     def __init__(
         self,
         config: Config,
-        study_service: StudyService,
-        matrix_service: MatrixService,
+        study_service: "StudyService",
```
sylvlecl: It does not seem used anymore, it should be removed

antarest/matrixstore/in_memory.py
```diff
@@ -25,7 +26,8 @@ class InMemorySimpleMatrixService(ISimpleMatrixService):
     In memory implementation of matrix service, for unit testing purposes.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, matrix_content_repository: MatrixContentRepository) -> None:
```
sylvlecl: In this implementation, we don't use the matrix content repository, so we should not add it here.
Actually, we should remove it from the abstract class `ISimpleMatrixService`, it was a mistake of previous evolutions to put it there, it should only go into the `SImpleMatrixService` and `MatrixService` implementations: could you implement that small cleanup please ?

tests/matrixstore/test_matrix_usage_providers.py
```diff
@@ -0,0 +1,124 @@
...
+    (raw_study_path / f"{matrix_name2}.link").write_text(f"matrix://{matrix_name2}")
+    (raw_study_path / f"{matrix_name3}.link").write_text(f"matrix://{matrix_name3}")
+    (raw_study_path / f"{matrix_name4}.txt").write_text(f"matrix://{matrix_name4}")
+
+    matrices_references = raw_studies_matrix_usage_provider.get_matrix_usage()
+    matrices_references_id = [matrix_reference.matrix_id for matrix_reference in matrices_references]
+    matrices_references_id.sort()
+    assert len(matrices_references) == 3
+
+    for matrix_name, matrix_id in zip(matrices_name, matrices_references_id):
```
sylvlecl: you can just assert that the 2 lists are equal:
```python
assert matrices_references_id == matrices_name
```
This removes also the need for the assertion above, about the length.

tests/matrixstore/test_matrix_usage_providers.py
```diff
@@ -0,0 +1,124 @@
...
+            study_version=study_version,
+        )
+
+        db.session.add(command_block1)
+        db.session.add(command_block2)
+        db.session.commit()
+
+        matrices_references = command_matrix_usage_provider.get_matrix_usage()
+
+        assert len(matrices_references) == 4
```
sylvlecl: You could have a one line for all the assertions:
```python
assert matrices_references == [MatrixReference(matrices_id, use_description )] * 4
```

tests/matrixstore/test_matrix_usage_providers.py
```diff
@@ -0,0 +1,124 @@
...
+    constant4 = {"c4": "constant_name4"}
+    constants = [constant1, constant2, constant3, constant4]
+
+    for constant in constants:
+        constants_matrix_usage_provider.matrix_constants.hashes.update(constant)
+
+    constants_reference = constants_matrix_usage_provider.get_matrix_usage()
+    matrix_ref_ids = [ref.matrix_id for ref in constants_reference]
+    matrix_ref_ids.sort()
+    for constant_id, matrix_ref_id in zip(constants, constants_reference):
```
sylvlecl: Here too we can just assert equality of the 2 lists

tests/matrixstore/test_service.py
```diff
@@ -627,3 +627,109 @@ def _create_upload_file(filename: str, file: t.IO = None, content_type: str = ""
...
+#         db.session.add(command_block2)
+#         db.session.commit()
+#         matrices = matrix_service.get_studies_matrices()
+#         assert len(matrices) == 2
+#         assert "[[1,2,3]]" in matrices
+#         assert "[[1,2,4]]" in matrices
+
+
+@pytest.mark.unit_test
+def test_get_used_matrices(matrix_service: MatrixService):
```
sylvlecl: Ce wan remove this test, it's not useful: we just assert that the mock returns what we told it to return

tests/matrixstore/test_service.py
```diff
@@ -627,3 +627,109 @@ def _create_upload_file(filename: str, file: t.IO = None, content_type: str = ""
...
+            return_value={"matrix1", "matrix2", "matrix3", "matrix4", "matrix6", "constant_matrix"}
+        )
+        assert matrix_service.get_used_matrices() == {
+            "matrix1",
+            "matrix2",
+            "matrix3",
+            "matrix4",
+            "matrix6",
+            "constant_matrix",
+        }
```
sylvlecl: We should add a test that checks that `get_used_matrices` returns what is provided by the usage providers that have been registered:
for example we can create 2 mocks, register them to the matrix service, and check that the "matrix usage" they return are also returned by `get_used_matrices`

sylvlecl (changes requested):
A little more work is required in particular to actually register the "usage providers" with the matrix service when we create the different services (at application start).
Also, for homogeneity we should add again one usage provider for the "dataset" matrices.

+ a few suggestions about implementation details

tests/matrixstore/test_matrix_usage_providers.py
sylvlecl: This comment still holds

antarest/study/storage/rawstudy/raw_study_matrix_usage_provider.py
```diff
@@ -0,0 +1,49 @@
...
+from antarest.matrixstore.service import ISimpleMatrixService
+from antarest.study.repository import AccessPermissions, StudyFilter, StudyMetadataRepository
+
+logger = logging.getLogger(__name__)
+
+
+class RawStudyMatrixUsageProvider(IMatrixUsageProvider):
+    def __init__(self, study_metadata_repo: StudyMetadataRepository, matrix_service: ISimpleMatrixService):
+        self.study_metadata_repo = study_metadata_repo
+        # self.managed_studies_path: Path = config.storage.workspaces[DEFAULT_WORKSPACE_NAME].path
```
sylvlecl: To be deleted

antarest/matrixstore/service.py
```diff
@@ -538,3 +555,44 @@ def download_matrix(self, matrix_id: str, filepath: Path) -> None:
...
+        # datasets_matrices = self.get_datasets_matrices()
+        # studies_matrices = self.get_studies_matrices()
+        # return studies_matrices | datasets_matrices
+        return {
+            matrix_reference.matrix_id
+            for provider in self.usage_providers
+            for matrix_reference in provider.get_matrix_usage()
+        }
+
+    def create_dataset_usage_provider(self) -> "IMatrixUsageProvider":
```
sylvlecl: We can make it protected (start with `_`) since it's only meant to be used inside the class

antarest/study/storage/variantstudy/business/matrix_constants/matrix_constants_usage_provider.py
```diff
@@ -0,0 +1,32 @@
...
+# This file is part of the Antares project.
+from typing_extensions import override
+
+from antarest.matrixstore.matrix_usage_provider import IMatrixUsageProvider
+from antarest.matrixstore.model import MatrixReference
+from antarest.matrixstore.service import MatrixService
+from antarest.study.storage.variantstudy.business.matrix_constants_generator import GeneratorMatrixConstants
+
+
+class ConstantsMatrixUsageProvider(IMatrixUsageProvider):
```
sylvlecl, line 22: This one still does not seem to be created on application start:
it needs to be created inside the `GeneratorMatrixConstants` constructor, like for other "clients" of the matrix service.

tests/matrixstore/test_service.py
```diff
@@ -627,3 +627,93 @@ def _create_upload_file(filename: str, file: t.IO = None, content_type: str = ""
     headers = Headers(headers={"content-type": content_type})
     # noinspection PyTypeChecker,PyArgumentList
     return UploadFile(filename=filename, file=file, headers=headers)
+
+
```
sylvlecl: Now we can delete those commented lines, since the tests have been rewritten elsewhere.

tests/matrixstore/test_matrix_usage_providers.py
```diff
@@ -0,0 +1,144 @@
...
+        assert constant_id, matrix_ref_id
```
sylvlecl: Here I think you want to assert equality

tests/matrixstore/test_matrix_usage_providers.py
```diff
@@ -0,0 +1,144 @@
...
+
+    return command_matrix_usage_provider
+
+
+@pytest.fixture
+def constants_matrix_usage_provider(matrix_service: MatrixService):
+    matrix_constants = GeneratorMatrixConstants(matrix_service)
+    return ConstantsMatrixUsageProvider(matrix_constants, matrix_service)
+
+
```
sylvlecl, line 62: We should have a test to cover the dataset provider too:
for this you can create a dataset in the matrix service, and check that the matrices of the dataset are correctly returned.

sylvlecl (changes requested):
A few more comments, important ones are:
- we need to create the constant matrix usage provider on application start
- we need some test for the dataset usage provider

tests/matrixstore/test_matrix_usage_providers.py
```diff
@@ -0,0 +1,215 @@
...
+            matrix_service.usage_providers.append(matrix_service._create_dataset_usage_provider())
+
+            matrices_reference = [
+                reference
+                for usage_prov in matrix_service.usage_providers
+                for reference in usage_prov.get_matrix_usage()
+            ]
+
+            assert matrices_reference == [MatrixReference(matrix_id=matrix_m, use_description=use_description)] * 4
```
sylvlecl: It's weird that we have 4 of them:
it's because the provider is appended 4 times:
- once every time the creation method is called (2 times)
- once when the matrix service is created
- once in your test

I will remove those duplicates so that we have only one of them

sylvlecl (approved):

PR #2584, opened 2025-07-11
fix(ui-matrix): use UTC for date/time display to avoid DST issues

hdinia:


.github/workflows/main.yml
```diff
@@ -114,7 +114,7 @@ jobs:
           key: ${{ runner.os }}-webapp-node-modules-${{ hashFiles('webapp/package-lock.json') }}
           save-always: true
       - name: Test
-        run: npm run test
+        run: TZ=UTC npm run test
```
skamril: You can add `"pretest": "TZ=UTC",` in the package.json to use UTC in all environments.

webapp/src/components/common/Matrix/shared/__tests__/utils.integration.test.ts
```diff
@@ -0,0 +1,189 @@
...
+        start_date: "2023-03-26T00:00:00.000Z", // Last Sunday of March - DST starts in Europe
+        steps: 5,
+        first_week_size: 7,
+        level: TimeFrequency.Hourly,
+      };
+
+      const result = generateDateTime(config);
+
+      // Extract hours
+      const hours = result.map((dateStr) => {
```
skamril, line 38:
```typescript
const hours = result.map((dateStr) => getHours(new Date(dateStr)));
```

It will better that `generateDateTime` returns Date objects.

hdinia: Yes i agree, I will fix all those dates issues in the refactor about the date parsing because their is some display constraints

.github/workflows/main.yml
```diff
@@ -153,4 +153,4 @@ jobs:
         uses: sonarsource/sonarcloud-github-action@v3.1.0
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
\ No newline at end of file
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
```
smailio, line 156: so weird this is a change these two lines looks identical to me

webapp/src/tests/setup.ts
```diff
@@ -18,6 +18,9 @@ import { cleanup } from "@testing-library/react";
 import * as matchers from "@testing-library/jest-dom/matchers";
 import "./mocks/mockResizeObserver";
 
+// Set timezone to UTC for consistent test behavior across different environments
+process.env.TZ = "UTC";
```
smailio: you already pass this env variable to vitest no ?

hdinia: yes setting `TZ=UTC` in the npm script affects the entire test process, so there's no need to set it again in the setup file

.github/workflows/main.yml
hdinia: I probably messed up the indent or added an extra space, thanks didn't notice it

skamril: He removes the last empty line, but it's recommend to keep it

hdinia: on dev their is no extra line, this diff is weird

skamril (approved):

.github/workflows/main.yml
hdinia: it's due to the added line at the end of the file, but the diff does not show it for some reason, so it's good

PR #2585, opened 2025-07-11
fix(xpansion): change validation rule for xpansion field

TheoPascoli:


MartinBelthle (approved):

MartinBelthle:
Perhaps the PR shouldn't be named feat but fix

PR #2594, opened 2025-07-17
feat(study): forbid illegal characters in study name and folder path

TheoPascoli:
Ticket # 3431


antarest/core/utils/utils.py
```diff
@@ -61,6 +61,24 @@ def sanitize_uuid(uuid: str) -> str:
...
+        )
+    return name.strip()
+
+
+def validate_folder_name(name: str) -> str:
+    if "=" in name:
+        raise HTTPException(
+            status_code=http.HTTPStatus.BAD_REQUEST, detail=f"folder name {name} contains illegal character '='"
+        )
+    if name.endswith("/"):
```
sylvlecl: I guess we could just remove it in that case ?

antarest/core/utils/utils.py
```diff
@@ -61,6 +61,24 @@ def sanitize_uuid(uuid: str) -> str:
...
 
+def validate_study_name(name: str) -> str:
+    if any(char in name for char in ["=", "/"]):
+        raise HTTPException(
+            status_code=http.HTTPStatus.BAD_REQUEST, detail=f"study name {name} contains illegal characters (=, /)"
+        )
+    return name.strip()
+
+
+def validate_folder_name(name: str) -> str:
```
sylvlecl: Actually here we don't validate the name but the full path, we should probably rename it to make it clear.

antarest/core/utils/utils.py
```diff
@@ -61,6 +61,24 @@ def sanitize_uuid(uuid: str) -> str:
     return uuid
 
 
+def validate_study_name(name: str) -> str:
```
sylvlecl, line 64: We could add a small docstring to explain why:
- / to not conflict with folders, I guess
- = to be written safely in INI files, I think ...

antarest/core/utils/utils.py
TheoPascoli: Now I remove all the trailing "/"

sylvlecl (approved):

PR #2598, opened 2025-07-21
feat(optimization): implements DAO for optimization config

TheoPascoli:


antarest/study/storage/variantstudy/model/command/update_optimization_preferences.py
```diff
@@ -0,0 +1,59 @@
...
+from antarest.study.dao.api.study_dao import StudyDao
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput, command_succeeded
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+class UpdateOptimizationPreferences(ICommand):
+    """
+    Command used to update multiple areas properties
```
sylvlecl: Docstring to be fixed

antarest/study/dao/file/file_study_optimization_preferences.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @override
+    def save_optimization_preferences(self, config: OptimizationPreferences) -> None:
+        file_study = self.get_file_study()
+
+        optimization_preferences = serialize_optimization_preferences(config)
+
+        # Include fields that are in the optimization part of the generaldata.ini file but not in the optimization class
+        current_optimization_preferences = file_study.tree.get(OPTIMIZATION_PATH)
+        optimization_preferences.update(
```
sylvlecl, line 46: Not an update: we should just replace that section

tests/variantstudy/model/command/test_update_optimization_preferences.py
```diff
@@ -0,0 +1,53 @@
...
+    UnfeasibleProblemBehavior,
+)
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.variantstudy.model.command.update_optimization_preferences import (
+    UpdateOptimizationPreferences,
+)
+from antarest.study.storage.variantstudy.model.command_context import CommandContext
+
+
+class TestUpdateOptimizationPreferences:
```
sylvlecl, line 23: We need to add parsing/serialization tests in test_command_factory

antarest/study/storage/rawstudy/model/filesystem/config/optimization_preferences.py
```diff
@@ -0,0 +1,59 @@
...
+        default=None, alias="include-unfeasible-problem-behavior"
+    )
+    simplex_optimization_range: SimplexOptimizationRange | None = Field(default=None, alias="simplex-range")
+
+    def to_model(self) -> OptimizationPreferences:
+        return OptimizationPreferences.model_validate(self.model_dump(exclude_none=True))
+
+    @classmethod
+    def from_model(cls, config: OptimizationPreferences) -> "OptimizationPreferencesFileData":
+        return cls.model_validate(config.model_dump(exclude={"id"}))
```
sylvlecl: There is not ID

sylvlecl (changes requested):
More or less same comments as for the general config:
almost ok but some details to fix:
- the DAO should replace, not update
- we need some parsing:serialization test for the command

antarest/study/dao/file/file_study_optimization_preferences.py
sylvlecl: About that topic:
as you explained to me, the original issue is that other "business" methods can modify the same "optimization" section of the underlying file. Obvisouly, here we don't want to remove those other properties.

Maybe the best option would be to have a class to represent the whole "generaldata" file, that we would just update with the relevant subpart of the business method.

This implies that this class really contains the whole data that we can have in that file.

sylvlecl: So I think I will merge the PRs as is, and if we can implement that one-to-one correspondance between file and class in a further PR, that will be nice.

sylvlecl (approved):

PR #2599, opened 2025-07-21
fix(matrix): fix concurrent matrix write

sylvlecl:
It can happen that writing the same matrix in 2 concurrent threads ends up in the matrix not being written.

We now:
- enlarge the lock protection
- ensure the matrix still does not exist after lock acquisition
- avoid to check for the target format in "other formats", to make sure we don't delete what we just wrote


MartinBelthle (approved):
Nice

PR #2600, opened 2025-07-21
fix(studies): make all managed studies folder end with id

smailio:
Fix [ANT-3640](https://gopro-tickets.rte-france.com/browse/ANT-3640)

Sometimes, a study's folder does not end with its own ID (as seen in the top entry of the image below), even though this should always be the case. 

This inconsistency was observed in both staging and production environments.

It is likely the result of a bug that was live for a few days several weeks ago. The query below fixes the issue by appending the study's ID to its folder path if it’s not already present.

One consequence of this inconsistency is that sometime, when the studies are processed by the front end to build the study tree, depending on the order on which they are processed some folders are marked a study folders, they are then ignored by the front and they're not shown in the study tree, for the user it seems that some folders are randomly missing from the study tree. As you can see in the screenshot bellow the uae folder isn't in the tree although it should.

image

### DIFF
--- a/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
+++ b/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
@@ -30,7 +30,7 @@ def upgrade():
 
     rawstudy_table = table("rawstudy", column("id"), column("workspace"))
 
-    # add id to folder, handle case whith and without a trailing slash
+    # add id to folder, handle case with and without a trailing slash
     add_id_exp = case(
         (
             study_table.c.folder.like("%/"),
--- a/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
+++ b/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
@@ -31,7 +31,7 @@ def upgrade():
             )
         WHERE
             s.folder IS NOT NULL
-            AND s.folder NOT LIKE concat('%', id)
+            AND s.folder NOT LIKE '%' || id
             AND (
                 EXISTS (
                     SELECT 1 FROM rawstudy r
--- a/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
+++ b/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
@@ -23,8 +23,10 @@ def upgrade():
         UPDATE study AS s SET
             folder = (
                 CASE
-                    WHEN folder LIKE '%/' THEN concat(folder, id)
-                    ELSE concat(concat(folder, '/'), id)
+                    WHEN folder LIKE '%/' THEN 
+                        folder || id
+                    ELSE 
+                        folder || '/' || id
                 END
             )
         WHERE
--- a/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
+++ b/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
@@ -18,6 +18,12 @@ depends_on = None
 
 
 def upgrade():
+    """
+    This update fixes an incident with the folder path of some managed studies.
+    Sometimes, a study's folder does not end with its own ID
+    (as seen in the top entry of the image below), even though this should always be
+    the case). So this function add the ID to the end of the folder path if it is not already there.
+    """
     op.execute(
         sa.text("""
         UPDATE study AS s SET
@@ -44,4 +50,8 @@ def upgrade():
 
 
 def downgrade():
+    """
+    We cannot downgrade this migration because we cannot make the difference between a folder path that was
+    correctly set and one that was incorrectly set.
+    """
     pass
--- a/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
+++ b/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
@@ -1,0 +1,45 @@
+"""Make all managed studies folder path end with id
+
+Revision ID: fa59340767b0
+Revises: 9ba7fc46d4a0
+Create Date: 2025-07-21 12:16:43.213036
+
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "fa59340767b0"
+down_revision = "9ba7fc46d4a0"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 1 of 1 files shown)


alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
```diff
@@ -0,0 +1,47 @@
...
+                CASE
+                    WHEN folder LIKE '%/' THEN 
+                        folder || id
+                    ELSE 
+                        folder || '/' || id
+                END
+            )
+        WHERE
+            s.folder IS NOT NULL
+            AND s.folder NOT LIKE '%' || id
```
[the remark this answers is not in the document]
smailio: Team, this is wrong no ?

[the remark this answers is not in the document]
TheoPascoli: I thing he's right, however, study id would not appear twice in the path. Why not implement its solution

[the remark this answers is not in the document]
smailio: because it's not true that the like will match the ID elsewhere in the string, my like expression only matches when the ID appears at the end. 

image

alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def upgrade():
+    """
+    This update fixes an incident with the folder path of some managed studies.
+    Sometimes, a study's folder does not end with its own ID
+    (as seen in the top entry of the image below), even though this should always be
+    the case). So this function add the ID to the end of the folder path if it is not already there.
+    """
+    op.execute(
```
MartinBelthle: Nice to add a doc to explain.
Personal opinion but I prefer to use sqlalchemy to perform the requests as it was done in the file `00a9ceb38842_add_study_version_to_commands.py` as I find it more readable than pure SQL request

alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
```diff
@@ -0,0 +1,78 @@
...
+def upgrade():
+    """
+    This update fixes an incident with the folder path of some managed studies.
+    Sometimes, a study's folder does not end with its own ID.
+    So this function add the ID to the end of the folder path if it is not already there.
+    """
+    # Create db connection
+    bind = op.get_bind()
+
+    study_table = table(
```
MartinBelthle: Minor: could be in one line

alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
```diff
@@ -0,0 +1,78 @@
...
+        "study",
+        column("id"),
+        column("folder"),
+        column("type"),
+    )
+
+    rawstudy_table = table(
```
MartinBelthle: Same

alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
```diff
@@ -0,0 +1,78 @@
...
+    # add id to folder, handle case whith and without a trailing slash
+    add_id_exp = case(
+        (
+            study_table.c.folder.like("%/"),
+            study_table.c.folder + study_table.c.id,
+        ),
+        else_=study_table.c.folder + literal("/") + study_table.c.id,
+    )
+
+    is_in_default_workspace_subq = (
```
MartinBelthle: According to chatGPT we could write this instead:
`exists().where(and_(rawstudy_table.c.id == study_table.c.id, rawstudy_table.c.workspace == "default"))`
which I find less confusing as the `select(sa.literal(1))` is weird IMO

alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
```diff
@@ -0,0 +1,78 @@
...
+        select(sa.literal(1))
+        .where(and_(rawstudy_table.c.id == study_table.c.id, rawstudy_table.c.workspace == "default"))
+        .exists()
+    )
+
+    # managed study = in default workspace or is variant
+    is_managed = or_(is_in_default_workspace_subq, study_table.c.type == "variantstudy")
```
MartinBelthle, line 47: Why do we need to add the variant check ? To me, every variant study is in the default workspace so we don't need this

alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
```diff
@@ -0,0 +1,78 @@
...
+        column("type"),
+    )
+
+    rawstudy_table = table(
+        "rawstudy",
+        column("id"),
+        column("workspace"),
+    )
+
+    # add id to folder, handle case whith and without a trailing slash
```
MartinBelthle: typo: `with`

alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+
+    # managed study = in default workspace or is variant
+    is_managed = or_(is_in_default_workspace_subq, study_table.c.type == "variantstudy")
+
+    # We're only interested in managed studies
+    # whose folder name doesn't end with their own ID.
+    where_clause = and_(
+        study_table.c.folder.isnot(None),
+        not_(study_table.c.folder.like("%" + study_table.c.id.cast(sa.String))),
```
MartinBelthle: Why do we need to use cast(sa.String) ?

MartinBelthle (commented):
Nice :). Just a few comments / questions to understand

alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
smailio: it's really a classic SQL way to write things

alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
smailio: variant don't have an entry in the rawstudy table, we can't check their workspace

MartinBelthle: Oh yes you're right

MartinBelthle (approved):

PR #2601, opened 2025-07-21
fix(xpansion): forbid candidate deletion when referenced in sensitivity

MartinBelthle:


antarest/study/business/xpansion_management.py
```diff
@@ -141,7 +142,7 @@ def update_candidate(
 
     def delete_candidate(self, study: StudyInterface, candidate_name: str) -> None:
```
TheoPascoli, line 143: In study/sercice.py L. 2067, you return this method that returns None. You should just call it without return
And same in xpansion_studies_blueprint, L. 150, you return Any and call this that returns None

TheoPascoli (commented):
Good work, nothing much to say

PR #2604, opened 2025-07-22
feat(generaldata): remove properties in general data

TheoPascoli:


antarest/study/model.py
```diff
@@ -91,6 +92,7 @@
     STUDY_VERSION_8_7,
     STUDY_VERSION_8_8,
     STUDY_VERSION_9_2,
+    STUDY_VERSION_9_3,
```
MartinBelthle: Putting this here allows users to create 9.3 studies. Not sure we want to allow this at the moment

tests/test_resources.py
```diff
@@ -78,10 +78,11 @@
 ]
 
 FILES_SINCE_V86 = sorted(["input/st-storage", "input/st-storage/clusters", "input/st-storage/series"] + STUDY_FILES)
+FILES_SINCE_V92 = [f for f in FILES_SINCE_V86 if f not in ["logs", "output", "user"]]
```
MartinBelthle, line 81: Why did you have to introduce this ?

tests/study/storage/rawstudy/test_general_data.py
```diff
@@ -0,0 +1,199 @@
...
+
+import pytest
+from antares.study.version import StudyVersion
+from antares.study.version.create_app import CreateApp
+
+from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfig
+from antarest.study.storage.rawstudy.model.filesystem.root.settings.generaldata import GeneralData
+
+
+@pytest.mark.parametrize(
```
MartinBelthle: I think you should rather add the fixture `empty_study_930` and use all the fixtures here.
At least you could introduce the fixture

tests/study/storage/rawstudy/test_general_data.py
```diff
@@ -0,0 +1,199 @@
+# Copyright (c) 2025, RTE (https://www.rte-france.com)
```
MartinBelthle, line 1: Nice to add this test :)

MartinBelthle (commented):
Few comments

antarest/study/model.py
TheoPascoli: yes it was for testing

tests/test_resources.py
TheoPascoli: I thought it was missing a test here. nothing to do with this particular PR

tests/study/storage/rawstudy/test_general_data.py
TheoPascoli: I changed it

tests/test_resources.py
MartinBelthle: Yep I meant why did this change `f not in ["logs", "output", "user"]]`. But it's because we removed the empty folders in the `antares-study-version` package inside this version. So doesn't really matter

MartinBelthle (approved):

PR #2605, opened 2025-07-22
fix(matrix): eliminate double date parsing in temporal filters

hdinia:
### Problem

The Matrix component's temporal filtering system was experiencing severe performance degradation when processing time series data. Performance profiling revealed that the `useTemporalData` hook was spending **~1,200ms** parsing dates for a typical yearly dataset (8,760 rows), accounting for 66.8% of the total rendering time.

**Root cause**: Each date string was being parsed multiple times:
- Once in `extractValueFromDate()` 
- Again in `extractTemporalValue()`
- This resulted in 2N parse operations for N dates

### Performance Profile (Before)
```
RowFilter: 1,189.3 ms (66.8%)
├── parseFlexibleDate: 626.7 ms (35.2%)
└── parseFlexibleDate: 558.2 ms (31.4%)
Total parsing: ~1,185 ms
```

### Performance Profile (After)
```
RowFilter: ~80 ms (5-8%)
└── parseFlexibleDate: ~40 ms (2-3%)
Total parsing: ~40 ms
```

### Solution

Implemented a single-parse architecture in `useTemporalData`:
1. Parse each date string only once into a Date object
2. Extract all temporal values (day, month, hour, etc.) from the parsed Date
3. Cache the extracted values for reuse across different filter operations

### Key Changes

- Added `ParsedDateInfo` interface to store all temporal values
- Created a new `useMemo` block that parses dates once and extracts all values
- Modified value extraction logic to use pre-computed values instead of re-parsing

### Results

- **Performance improvement**: ~15x faster (from ~1,200ms to ~80ms)
- **User experience**: Filters should now open instantly with no/minimal UI freezing
- **Memory trade-off**: Minimal (~20% increase for significantly better performance)
- **Code impact**: Single file change, fully backward compatible

## Local tests:

Before:

Screenshot 2025-07-22 at 10 52 56

After:

image

### DIFF
--- a/webapp/src/components/common/Matrix/components/MatrixFilter/hooks/useTemporalData.ts
+++ b/webapp/src/components/common/Matrix/components/MatrixFilter/hooks/useTemporalData.ts
@@ -12,10 +12,12 @@
  * This file is part of the Antares project.
  */
 
+import { getDayOfYear } from "date-fns";
 import { useMemo } from "react";
+import { getHourOfYear, getWeekFromDayOfYear, parseFlexibleDate } from "@/utils/date/dateUtils";
 import type { TimeFrequencyType } from "../../../shared/types";
 import { TIME_INDEXING } from "../constants";
-import { extractValueFromDate, getDefaultRangeForIndexType } from "../utils/dateUtils";
+import { getDefaultRangeForIndexType } from "../utils/dateUtils";
 
 interface UseTemporalDataProps {
   dateTime?: string[];
@@ -23,9 +25,19 @@ interface UseTemporalDataProps {
   timeFrequency?: TimeFrequencyType;
 }
 
+interface ParsedDateInfo {
+  date: Date | null;
+  dayOfYear: number | null;
+  hourOfYear: number | null;
+  dayOfMonth: number | null;
+  week: number | null;
+  month: number | null;
+  dayHour: number | null;
+  weekday: number | null;
+}
+
 export function useTemporalData({ dateTime, isTimeSeries }: UseTemporalDataProps) {
   const indexTypeRanges = useMemo(() => {
-    // Create a map of all default ranges
     return Object.values(TIME_INDEXING).reduce(
       (acc, type) => {
         acc[type] = getDefaultRangeForIndexType(type);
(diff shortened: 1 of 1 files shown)


webapp/src/components/common/Matrix/components/MatrixFilter/hooks/useTemporalData.ts
```diff
@@ -35,14 +46,65 @@ export function useTemporalData({ dateTime, isTimeSeries }: UseTemporalDataProps
...
+            hourOfYear: null,
+            dayOfMonth: null,
+            week: null,
+            month: null,
+            dayHour: null,
+            weekday: null,
+          };
+        }
+
+        const dayOfYear = Math.floor(
```
sylvlecl: Why not `getDayOfYear` of `date-fns` ?

webapp/src/components/common/Matrix/components/MatrixFilter/hooks/useTemporalData.ts
```diff
@@ -35,14 +46,65 @@ export function useTemporalData({ dateTime, isTimeSeries }: UseTemporalDataProps
...
+        return {
+          date,
+          dayOfYear,
+          hourOfYear: getHourOfYear(date),
+          dayOfMonth: date.getDate(),
+          week: getWeekFromDayOfYear(dayOfYear),
+          month: date.getMonth() + 1,
+          dayHour: date.getHours(),
+          weekday: date.getDay() + 1,
+        };
```
sylvlecl, line 83: Nice we, can remove method extractValueFromDate when we merge

webapp/src/components/common/Matrix/components/MatrixFilter/hooks/useTemporalData.ts
```diff
@@ -35,14 +46,65 @@ export function useTemporalData({ dateTime, isTimeSeries }: UseTemporalDataProps
...
+      } catch {
```
sylvlecl, line 84: I don't think we need a catch here, what could go wrong ? We just get some basic date out of a well formed date.
Then we can remove nulls from the interface.

webapp/src/components/common/Matrix/components/MatrixFilter/hooks/useTemporalData.ts
```diff
@@ -13,19 +13,30 @@
...
+import { getDefaultRangeForIndexType } from "../utils/dateUtils";
 
 interface UseTemporalDataProps {
   dateTime?: string[];
   isTimeSeries: boolean;
   timeFrequency?: TimeFrequencyType;
 }
 
+interface ParsedDateInfo {
+  date: Date | null;
```
sylvlecl, line 29: See other comments, I think we should not have nulls in that interface: we are sure we have an actual date and that we have been able to extract the information.

sylvlecl (commented):
Nice !
It's a complement of https://github.com/AntaresSimulatorTeam/AntaREST/pull/2606 , with both improvements we should have reasonable performances.

Added some implementation details comments.

webapp/src/components/common/Matrix/components/MatrixFilter/hooks/useTemporalData.ts
sylvlecl: Actually you can ignore that comment, it will only be true once it works straight on dates instead of strings, I got confused with my merge of the 2 branches.

webapp/src/components/common/Matrix/components/MatrixFilter/hooks/useTemporalData.ts
hdinia: yes `getDayOfYear` is better

hdinia:
Moved to : https://github.com/AntaresSimulatorTeam/AntaREST/pull/2606

PR #2606, opened 2025-07-22
perf(ui-matrix-filter): remove date parsing

sylvlecl:
This PR includes 2 important performance improvements for the initialization of filterable matrices:

1. **Removal of date parsing:**
  Previously, dates were formatted to string as soon as we received the data from the backend.
  We then parsed them for filtering purposes.
  Instead in this PR, we now keep them as date objects, and only format them to string for the 
  rendering. Filtering is performed on date objects.
  Some metadata is kept for formatting in `DateTimes` object.

2. **Perform date info extraction (hour of year etc) only once:**
  Previously, the extraction was done in multiple places. We now extract all
  useful information to `DateInfo` objects which are used for filtering.
  See  https://github.com/AntaresSimulatorTeam/AntaREST/pull/2605


webapp/src/components/common/Matrix/hooks/useGridCellContent/index.ts
```diff
@@ -34,7 +47,7 @@ const cellContentGenerators: Record<ColumnType, CellContentGenerator> = {
   [Column.DateTime]: (row, col, column, data, dateTime) => ({
     kind: GridCellKind.Text,
     data: "", // Date/time columns are not editable
-    displayData: dateTime?.[row] ?? "",
+    displayData: formatDate(dateTime, row),
```
hdinia, line 50: I noticed this change introduces date formatting inside `getCellContent`:

```typescript
displayData: formatDate(dateTime, row),  // Formats on every call
```

Since Glide calls `getCellContent` for every visible cell during scrolling, resizing, etc., this means we're formatting dates repeatedly instead of using pre-computed values like other columns do.

Before: `displayData: dateTime?.[row]` → Direct array access  
After: `displayData: formatDate(dateTime, row)` → Format operation

While the performance impact is likely minimal in practice, I'm wondering if we should consider maintaining consistency with how other columns work (numbers, aggregates, headers all use pre-computed values).

As discussed in the Jira ticket, another approach could be to maintain two parallel arrays:

```typescript
// In useMatrixData
const { dates, dateTimeFormatted } = useMemo(() => {
  if (!index) {
    return { dates: [], dateTimeFormatted: [] };
  }

  const { dates, formatted } = generateDateTime(index);
  return { dates, dateTimeFormatted: formatted };
}, [index]);

// Return both
return {
  dates,                       // Date[] for filtering operations
  dateTime: dateTimeFormatted, // string[] for display
  // ...
};
```

Both approaches work, I just lean towards keeping a clear separation between data (for logic) and formatted strings (for display). What do you think?

hdinia (changes requested):

webapp/src/components/common/Matrix/hooks/useGridCellContent/index.ts
sylvlecl: I think it's better to keep it this way:
 - getCellContent is called only once for each cell for rendering, so it seems appropriate to build here what we want to display, there will be no negative impact on performance. Actually, it will use more RAM for something that is not used, if we keep it in the app state.
 - the fact that we keep the formatted date elsewhere in the code will not only use RAM, but will also confuse the future reader who may be tempted to used it for something else, which he should not

hdinia: Just one clarification though ,`getCellContent` isn't called just once per cell
It's called:
- When the cell enters the viewport (scrolling)
- When data changes
- When the grid re-renders (theme changes, selections, etc)

So during a scroll session, the same cell can have its date formatted multiple times as it leaves and re-enters the viewport

That said, I see your point about the trade-offs

hdinia: Other than that it looks good to me, I will do a final check and we can merge it

webapp/src/components/common/Matrix/components/MatrixFilter/hooks/useFilteredData.ts
```diff
@@ -128,7 +128,7 @@ const createDefaultCriteria = R.memoizeWith(
  */
 export function useFilteredData({
```
hdinia, line 129: minor: update the doc props and description to reflect `datesInfo` change

webapp/src/components/common/Matrix/components/MatrixFilter/utils/index.ts
```diff
@@ -284,7 +278,7 @@ export function processRowFilters(
     return getTemporalIndices({
       filter,
       rowFilter: rowsFilters[0],
-      dateTime,
+      datesInfo: datesInfo,
```
hdinia: no assignation needed in this case you can use shorthand notation

(same for the other occurrences)

webapp/src/components/common/Matrix/components/MatrixFilter/utils/index.ts
```diff
@@ -259,15 +253,15 @@ export function getFilteredTemporalOptions(
  * Applies AND logic between different indexingTypes
  *
  * @param filter - The filter state containing row filters configuration
- * @param dateTime - Array of date/time strings for temporal indexing
+ * @param datesInfo - Array of date/time strings for temporal indexing
```
hdinia: you can update the description too

(same for other methods that have this change)

webapp/src/components/common/Matrix/hooks/useGridCellContent/index.ts
sylvlecl: Indeed you're right, I had misunderstood the mechanism.
It's not obvious what's the best choice performance-wise:
the advantage of calling the formatting in `getCellContent` is that only what is displayed is computed, so for the initial display there will be only a few calls. But indeed if you scroll or even just hover over the cells a lot, you may call the formatting more times than if we had computed everything initially :/ 

Overall, it still seems like a better tradeoff to me for now, taking into account the code simplicity.

sylvlecl: (the ideal solution would be to not compute all formatted values up-front, but to cache them when we first compute them)

hdinia: I think the performance impact will be negligible in practice, but I did not test it just so you know

hdinia (changes requested):

webapp/src/components/common/Matrix/hooks/useGridCellContent/index.ts
hdinia: I think the performance impact will be negligible in practice, but I did not test it, just so you know

webapp/src/components/common/Matrix/components/MatrixFilter/hooks/useTemporalData.ts
```diff
@@ -15,17 +15,17 @@
...
-
+import { getDefaultRangeForIndexType } from "../utils/dateUtils";
+import type { DateInfo } from "@/components/common/Matrix/components/MatrixFilter/types";
+import { INDEX_TYPE_TO_DATEINFO_PROPERTY } from "@/components/common/Matrix/components/MatrixFilter/utils";
 interface UseTemporalDataProps {
-  dateTime?: string[];
+  datesInfo?: DateInfo[];
   isTimeSeries: boolean;
   timeFrequency?: TimeFrequencyType;
 }
```
hdinia: since we no longer use `isTimeSeries` and `timeFrequency` we can remove them from the interface, or even remove it and use directly `export function useTemporalData(datesInfo?: DateInfo[]) {`

webapp/src/utils/date/matrixDateUtils.ts
hdinia, line 1: in the same idea of removing dead code their is several methods no longer used in `src/utils/date/dateUtils`

you can remove everything except `getHourOfYear` and  `getWeekFromDayOfYear` that are still in use

hdinia (changes requested):

hdinia (approved):

PR #2609, opened 2025-07-23
perf(commands): re-use the cache more often

MartinBelthle:
[ANT-2971]


antarest/study/storage/variantstudy/model/command/common.py
```diff
@@ -23,15 +23,16 @@
 @dataclass(frozen=True)
 class CommandOutput:
     status: bool
+    should_invalidate_cache: bool
```
smailio: it would be possible to put the default value here instead of in the function command_success, up to you !

PR #2612, opened 2025-07-28
fix(alembic): alembic merge heads

smailio:
This PR resolves a divergence in the Alembic migration history by merging multiple head revisions into a single lineage. 

image


MartinBelthle (changes requested):
You should document why you're doing this inside the file name or add a docstring

MartinBelthle (approved):

PR #2613, opened 2025-07-28
feat(cluster): use free groups for thermal and renewable in v9.3

MartinBelthle:
[ANT-3593]

### DIFF
--- a/antarest/study/business/model/thermal_cluster_model.py
+++ b/antarest/study/business/model/thermal_cluster_model.py
@@ -369,9 +369,9 @@ def validate_thermal_cluster_against_version(
             _check_min_version(cluster_data, field, version)
 
     if cluster_data.group is not None and version < STUDY_VERSION_9_3:
-        # We need to be sure we're able to cast the group to a ThermalClusterGroup enum value
-        if cluster_data.group not in [e.value for e in ThermalClusterGroup]:
-            raise InvalidFieldForVersionError(f"Free groups are available since v9.3 and your study is in {version}")
+        # Performs this transformation to fit with old behavior
+        # Before, when giving a fake group, we used to write `other 1` instead and not crash.
+        cluster_data.group = ThermalClusterGroup(cluster_data.group)
 
 
 def _initialize_field_default(cluster: ThermalCluster, field: str, default_value: Any) -> None:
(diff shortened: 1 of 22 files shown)


antarest/study/business/model/thermal_cluster_model.py
```diff
@@ -140,6 +140,7 @@ def _validate_week_hours(v: int) -> int:
 Cost: TypeAlias = Annotated[float, Field(ge=0)]
 Volatility: TypeAlias = Annotated[float, Field(ge=0, le=1)]
 HoursInWeek: TypeAlias = Annotated[int, PlainValidator(_validate_week_hours)]
+Group: TypeAlias = Optional[LowerCaseStr]
```
sylvlecl, line 143: For consistency sake with ST storage, and to also catch raw modifications, we should also implement lower casing at read/write time in INI readers configuration (see list.py for ST storages)

sylvlecl (changes requested):
Thanks, see comment, we also want to handle lower casing in INI readers :(

antarest/study/business/model/thermal_cluster_model.py
sylvlecl: As discussed: this is actually already handled, it was implemented before that evolution.

tests/integration/study_data_blueprint/test_renewable.py
```diff
@@ -494,6 +494,30 @@ def test_lifecycle(
...
+        assert task.status == TaskStatus.COMPLETED, task
+
+        # Checks creation with free `group`
+        res = client.post(
+            f"/v1/studies/{internal_study_id}/areas/{area_id}/clusters/renewable",
+            headers={"Authorization": f"Bearer {user_access_token}"},
+            json={"name": "free cluster", "group": "GroupFoo"},
+        )
+        assert res.status_code == 200, res.json()
+        assert res.json()["group"] == "groupfoo"
```
sylvlecl, line 435: Actually, this works, but when you GET the cluster from the backend, it's "other 1" (seen by testing with the front):

I think you forgot to adapt the serialization classes (`*FileData`).

sylvlecl (changes requested):
See comment: the serialization to file does not seem to work

tests/integration/study_data_blueprint/test_renewable.py
MartinBelthle: Indeed, I added the test you mentioned and fixed the src code

sylvlecl (approved):

PR #2614, opened 2025-07-28
refactor(xpansion): implement dao and new class design for candidates

MartinBelthle:


antarest/study/dao/api/xpansion_dao.py
```diff
@@ -0,0 +1,44 @@
...
+    @abstractmethod
+    def get_xpansion_candidate(self, candidate_id: str) -> XpansionCandidate:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def checks_xpansion_candidate_coherence(self, candidate: XpansionCandidate) -> None:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def checks_xpansion_candidate_can_be_deleted(self, candidate_name: str) -> None:
```
sylvlecl, line 33: It seems to me that we could just have the delete method which would take a name input, and do the check there ?
It would avoid to to a check/get/delete in the remove command

antarest/study/storage/variantstudy/model/command/remove_xpansion_candidate.py
```diff
@@ -43,24 +40,10 @@ class RemoveXpansionCandidate(ICommand):
...
-            raise CandidateNotFoundError(f"The candidate '{self.candidate_name}' does not exist")
-
-        del candidates[candidate_number]
-        # Reorder keys of the dict
-        new_dict = {str(i): v for i, (k, v) in enumerate(candidates.items(), 1)}
-
-        study_data.tree.save(data=new_dict, url=["user", "expansion", "candidates"])
-
+    def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        study_data.checks_xpansion_candidate_can_be_deleted(self.candidate_name)
```
sylvlecl, line 44: See other command: it seems to me that we could just have a call to the delete method here ?

antarest/study/dao/api/xpansion_dao.py
```diff
@@ -0,0 +1,44 @@
...
+        raise NotImplementedError()
+
+
+class XpansionDao(ReadOnlyXpansionDao):
+    @abstractmethod
+    def save_xpansion_candidate(self, candidate: XpansionCandidate, old_id: Optional[str] = None) -> None:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def delete_xpansion_candidate(self, candidate: XpansionCandidate) -> None:
```
sylvlecl: Here I think we should just have the name as an input

sylvlecl (changes requested):
Minor request about the DAO interface

antarest/study/dao/api/xpansion_dao.py
sylvlecl: Actually it's necessary for the check in the manager, as you reminded me.

PR #2615, opened 2025-07-28
feat(thematic-trimming): support new 9.3 fields

smailio:


antarest/study/business/model/thematic_trimming_model.py
```diff
@@ -340,3 +393,9 @@ def initialize_with_version(thematic_trimming: ThematicTrimming, version: StudyV
     if version >= STUDY_VERSION_9_1:
         for field in _get_v_9_1_fields():
             _initialize_field_default(thematic_trimming, field, default_bool)
+
+    if version >= STUDY_VERSION_9_3:
+        for field in _get_v_9_3_fields():
+            _initialize_field_default(thematic_trimming, field, default_bool)
+        for field in exclude_from_9_3:
```
smailio, line 416: I'm open to implement this differently. I don't know if it's obvious but some fields were exluded from the v 9.3 and they need to be set to None because that's what the front end expects, let me know if I should add a comment explaining this.

PR #2616, opened 2025-07-28
fix(sts-matrix): restore deprecated enpoint as a security net

sylvlecl:
We restore temporarily the STS matrix modification endpoint, as deprecated.
This is a security net to give more time to client to adapt.

### DIFF
--- a/antarest/study/business/areas/st_storage_management.py
+++ b/antarest/study/business/areas/st_storage_management.py
@@ -10,14 +10,18 @@
 #
 # This file is part of the Antares project.
 
-from typing import Mapping, Sequence
+from typing import Any, List, Literal, Mapping, MutableMapping, Sequence, TypeAlias
 
+import numpy as np
 from antares.study.version import StudyVersion
+from pydantic import field_validator
 
 from antarest.core.exceptions import (
     DuplicateSTStorage,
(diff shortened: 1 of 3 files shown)


antarest/study/business/areas/st_storage_management.py
```diff
@@ -35,10 +39,67 @@
 from antarest.study.storage.variantstudy.model.command.update_st_storages import UpdateSTStorages
 from antarest.study.storage.variantstudy.model.command_context import CommandContext
 
+# =============
+#  Time series
+# =============
+
+
+class STStorageMatrix(AntaresBaseModel):
```
MartinBelthle, line 47: To me we shouldn't re-introduce this class even if it's temporary as it would mean we'll have to maintain it with the 9.3 and there will be some change to do (some can have several columns after v9.3)

MartinBelthle (dismissed):

antarest/study/business/areas/st_storage_management.py
sylvlecl: I removed the constraint on matrix size.
I don't think we'll need more adaptations ?

MartinBelthle: Yes we can let it as it is

PR #2617, opened 2025-07-29
fix(alembic): fix migration syntax for postgresql

sylvlecl:
It seems that sqlalchemy does not understand that the addition is between 2 strings,
and does not translate it correctly to concatenation.

### DIFF
--- a/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
+++ b/alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
@@ -7,6 +7,7 @@ Create Date: 2025-07-21 12:16:43.213036
 """
 
 from sqlalchemy.sql import and_, case, column, exists, literal, not_, or_, table, update
+from sqlalchemy.sql.functions import concat
 
 from alembic import op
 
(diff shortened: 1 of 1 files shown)


alembic/versions/fa59340767b0_make_all_managed_studies_folder_path_.py
```diff
@@ -34,7 +35,7 @@ def upgrade():
     add_id_exp = case(
         (
             study_table.c.folder.like("%/"),
-            study_table.c.folder + study_table.c.id,
+            concat(study_table.c.folder, study_table.c.id),
```
smailio, line 38: Shouldn't it be `concat(study_table.c.folder,  study_table.c.id) ` ?

smailio: Also, everywhere else in the query the + operator works ?

smailio: I noticed that with every other + there's a literal involved, so maybe in theses he managed to infer that he needed a || but not in this case.

sylvlecl: Yes, that's also what I concluded!
Well spotted for the syntax, thanks

smailio (approved):

PR #2624, opened 2025-07-29
feat(matrix): add arrow format for input matrices inside GET `/raw` endpoint

MartinBelthle:
Benchmark performed on my local env. The tuple corresponds to (time, content_size)

    '8760*1 json': (0.02, 95302),
    '8760*1 binary': (0.01, 43821),
    '8760*1 arrow compressed': (0.01, 1722),
    '8760*1 arrow uncompressed': (0.01, 71450),

    '8760*60 json': (0.08, 2688412),
    '8760*60 binary': (0.1, 3153602),
    '8760*60 arrow compressed': (0.02, 41930),
    '8760*60 arrow uncompressed': (0.02, 4226554),

    '8760*200 json': (0.32, 13306546),
    '8760*200 binary': (0.79, 14997616),
    '8760*200 arrow compressed': (0.05, 7139250),
    '8760*200 arrow uncompressed': (0.06, 14086586)

NB: Because in local we don't have the delay, the time gained would be even higher in production with arrow compression as the data is smaller

Benchmark for new JSON serialization:

    '8760*60 json': (0.1, 3739612),
    '8760*60 binary': (0.22, 4204802),
    '8760*60 arrow compressed': (0.03 41458),
    '8760*60 arrow uncompressed': (0.03, 4226554),
    '8760*60 json pandas': (0.04, 3739612)

    '8760*200 json': (0.31, 13306546),
    '8760*200 binary': (0.8, 14997616),
    '8760*200 arrow compressed': (0.06, 7139250),
    '8760*200 arrow uncompressed': (0.06, 14086586),
    '8760*200 json pandas': (0.13, 13306546),
    

So now we'll use the pandas `to_json` method to have performance gains


antarest/study/web/raw_studies_blueprint.py
```diff
@@ -100,15 +102,24 @@ def get_study_data(uuid: str, path: PATH_TYPE = "/", depth: int = 3, formatted:
...
 
         Returns the fetched data: a JSON object (in most cases), a plain text file
         or a file attachment (Microsoft Office document, TSV/TSV file...).
         """
         logger.info(f"📘 Fetching data at {path} (depth={depth}) from study {uuid}")
-        output = study_service.get(uuid, path, depth=depth, formatted=formatted)
+
+        raw_format = format
+        if not raw_format:
+            raw_format = RawDataFormat.JSON if formatted else RawDataFormat.BINARY
```
sylvlecl: "Binary" is misleading: what we return is actually text since the format is pure text.
We'd better name it something like `PLAIN`.

antarest/study/service.py
```diff
@@ -170,6 +172,13 @@
 MAX_MISSING_STUDY_TIMEOUT = 2  # days
 
 
+class RawDataFormat(EnumIgnoreCase):
```
sylvlecl: We could actually name it more accurately something like `MatrixFormat`, since it's only targetted to matrices data.

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -70,6 +73,35 @@
 PATH_TYPE = Annotated[str, Query(openapi_examples=get_path_examples())]
 
 
+class MatrixFormat(EnumIgnoreCase):
+    JSON = "json"
+    ARROW_COMPRESSED = "arrow compressed"
+    ARROW_UNCOMPRESSED = "arrow uncompressed"
+    PLAIN = "plain"
+
+    def parse_dataframe(self, dataframe: pd.DataFrame) -> Response:
```
sylvlecl: it's more serialize, not parse

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -70,6 +73,35 @@
...
+        if self == MatrixFormat.PLAIN:
+            if dataframe.empty:
+                return Response(content=b"", media_type="application/octet-stream")
+            string_buffer = io.StringIO()
+            dataframe.to_csv(string_buffer, sep="\t", header=False, index=False)
+            json_response = to_json(string_buffer.getvalue())
+            return Response(content=json_response, media_type="application/json")
```
sylvlecl: Mime type should be "text/csv" instead

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -70,6 +73,35 @@
...
+    ARROW_UNCOMPRESSED = "arrow uncompressed"
+    PLAIN = "plain"
+
+    def parse_dataframe(self, dataframe: pd.DataFrame) -> Response:
+        if self == MatrixFormat.PLAIN:
+            if dataframe.empty:
+                return Response(content=b"", media_type="application/octet-stream")
+            string_buffer = io.StringIO()
+            dataframe.to_csv(string_buffer, sep="\t", header=False, index=False)
+            json_response = to_json(string_buffer.getvalue())
```
sylvlecl: We don't need this, we should return the plain string in the response

antarest/study/service.py
```diff
@@ -2426,3 +2426,16 @@ def normalize_study(self, study: Study) -> None:
...
         """
         self.storage_service.get_storage(study).get_raw(study).tree.normalize()
+
+    def get_dataframe_if_the_url_corresponds_to_a_matrix(self, uuid: str, path: str) -> pd.DataFrame | None:
+        """
+        Returns a DataFrame if the provided url corresponds to a matrix.
+        """
+        study = self.get_study(uuid)
+        assert_permission(study, StudyPermissionType.READ)
+        file_study = self.get_file_study(study)
```
sylvlecl, line 2437: I wonder if this cannot worsen a little performances of other files retrieval, because we will build the file study twice, once here to get a None, and once later to retrieve the actual file.
We can probably find a solution to avoid this.

sylvlecl (changes requested):
A few comments

PR #2629, opened 2025-07-31
refactor(adequacy-patch): introduce DAO and new class design

MartinBelthle:


antarest/study/web/study_data_blueprint.py
```diff
@@ -651,25 +654,26 @@ def set_optimization_form_values(uuid: str, field_values: OptimizationPreference
...
-        return study_service.adequacy_patch_manager.get_field_values(study_interface)
+        return study_service.adequacy_patch_manager.get_adequacy_patch_parameters(study_interface)
 
     @bp.put(
         path="/studies/{uuid}/config/adequacypatch/form",
         tags=[APITag.study_data],
         summary="Set adequacy patch config with values from form",
     )
-    def set_adequacy_patch_form_values(uuid: str, field_values: AdequacyPatchFormFields) -> None:
+    def set_adequacy_patch_form_values(
```
sylvlecl, line 670: Something we should do in addition (probably on previous work too):
- deprecate those "/form" endpoints
- replace them with a plain "/studies/{uuid}/config/adequacypatch" endpoint which does exactly the same thing

sylvlecl (dismissed):
Comments which probably hold true for previous work too:
- about evolving the API altogether to deprecate "/form" endpoints
- we don't see in tests that changing adequacy path params now translates into a specific command. I think we shuld improve our tests on managers so that they test that indeed the expected command has been added to the list

We could handle those 2 points later

PR #2630, opened 2025-07-31
fix(matrix): don't raise when optional matrices don't exist

MartinBelthle:


sylvlecl (approved):
Nice.

At some point we may need to address the fact that we make exported studies grow with all those identical empty matrices though.

PR #2638, opened 2025-08-05
feat(study): introduce new `ReplaceStudyAuthor` command

MartinBelthle:


antarest/study/storage/variantstudy/model/command/replace_study_author.py
```diff
@@ -0,0 +1,51 @@
...
+from typing_extensions import override
+
+from antarest.study.dao.api.study_dao import StudyDao
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput, command_succeeded
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+class ReplaceStudyAuthor(ICommand):
```
sylvlecl, line 23: I think we'd better have a coarser-grained command, for example named `UpdateStudyMetadata`, where we could add other stuff if needed later (like study "caption", horizon etc).

Otherwise it could lead us to create a lot of very small commands, which will be a burden on the long run.

In particular, the study service also has a `update_study_information` methdo with other metadata. We should probably handle all that together.

antarest/study/service.py
```diff
@@ -1559,15 +1561,12 @@ def change_owner(self, study_id: str, owner_id: int) -> None:
             )
         )
 
-        owner_name = None if new_owner is None else new_owner.name
-        self._edit_study_using_command(study=study, url="study/antares/author", data=owner_name)
+        command_context = self.storage_service.variant_study_service.command_factory.command_context
+        args = {"author": new_owner.name, "study_version": study.version, "command_context": command_context}
+        command = ReplaceStudyAuthor.model_validate(args)
+        self.get_study_interface(study).add_commands([command])
```
sylvlecl, line 1567: Note that it's very weird (actually buggy I think) that we edit the actual study metadata in database, and the files through a command:
if we later remove the command, there will be a difference between the database and the file.

So after all I am not sure we should enable that.
If we have duplicate data between the database and the files, we should probably only choose one as the source of truth, which here would be the database.
Then, only at export time, we could udpate the files with the actual data.

This is not perfect but it looks more consistent.

sylvlecl (commented):
Not sure we should have such a fine-grained command, probably more a coarse-grained "update metadata" command.

Also, I think it's the opportunity to re-think a little how we handle those metadata:
the "author" is kind of a duplicate of the owner of the study.
The current behaviour where we edit the author through a command is a little buggy and can lead to inconsistencies (when we delete a command for example).

I wonder if we should not just "override" those data at export-time instead.
Actually, when we generate a snapshot, we already replace the author with the study owner, which kind of duplicates the effect of the command.

Needs more thought but needs to be streamlined.

antarest/study/dao/api/study_dao.py
```diff
@@ -95,6 +95,10 @@ def get_file_study(self) -> FileStudy:
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def change_study_owner(self, new_owner: str) -> None:
```
sylvlecl, line 99: Clearly the "owner" should not be managed at the DAO level.
Only the "author" might, if we confirm that we want to separate the 2 notions.

MartinBelthle:
Too old and a bit irrelevant now

PR #2641, opened 2025-08-05
perf(matrix): write matrices as integers when possible

MartinBelthle:
Exporting a real BP study (performed on my local env):

- **Existing code**: Around 8' and input folder is 5.5 Go unzipped
- **With the int32 check**: Around 6'15'' and input folder is 4.1 Go unzipped
- **With the int32 check + the float32 check**: Around 9'15'' and input folder is 3.7 Go unzipped

*NB*: Not sure we want to do it like this but this benchmark is interesting


sylvlecl:
Additional interesting stuff:
on a load matrix 8760x200, just specifying the formatter to not have decimals like this is **4 times slower than your method** (1.2s instead of around 300ms)
```python
df.to_csv(path, float_format="%.0f", sep="\t", header=False, index=False)
```

So in the end it seems more interesting indeed to convert to ints.
For floats matrix though we'll lose some time of course, but it seems almost negligible. For the same size:
- around 730 ms with int checking
- aound 670 ms without it

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
```diff
@@ -105,7 +105,14 @@ def write_dataframe(self, df: pd.DataFrame) -> None:
         if self.default_empty is not None and np.array_equal(df.to_numpy(dtype=np.float64), self.default_empty):
             self.config.path.touch(exist_ok=True)
         else:
-            df.to_csv(self.config.path, sep="\t", header=False, index=False)
+            # We're checking if the dataFrame could be represented with integer values.
+            # If so, we'll write it this way as it would be quicker and the file takes less place on the fs.
+            df_as_int = df.astype(np.int32)
```
sylvlecl: I think we'd better target int64 to catch larger values, the serialized form will stay the same and from my tests it does not hurt performance.

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
```diff
@@ -105,7 +105,14 @@ def write_dataframe(self, df: pd.DataFrame) -> None:
...
+            try:
+                pd.testing.assert_frame_equal(df, df_as_int, check_dtype=False)
```
sylvlecl: Maybe we should add `check_exact=True` to make explicit we don't compare with some tolerance (I think it's the case by default, but not that clear).

From what I understand, under the hood we end up converting the columns to numpy arrays ad just performe a `==` on them, so we could implement it this way, but for me it's ok to keep that pandas method.

antarest/matrixstore/matrix_uri_mapper.py
```diff
@@ -166,7 +166,7 @@ def denormalize(self, node: MatrixNode) -> None:
 
         uuid = link_path.read_text()
         matrix = self.get_matrix(uuid)
-        node.dump(matrix)
+        node.write_dataframe(matrix)
```
sylvlecl, line 169: We should probably also use that serialization method for the matrix content repo when CSV is configured.
We can have a common method somewhere in `core.serde`.

sylvlecl (approved):

PR #2647, opened 2025-08-11
chore(metadata): remove unused author field

sylvlecl:
The author field in database is never actually used, this PR cleans it up.


MartinBelthle (commented):
Indeed, it's a duplicate of the StudyAdditionalData field.

But shouldn't we do an alembic migration to remove the field ? Otherwise we'll have a difference between the alembic model and the sqlalchemy one

sylvlecl:
Actually I think we should instead move the study additional data to study class to simplify things, this separation of a few fields in a separate class is quite arbitrary, the purpose is not clear.

sylvlecl:
We should instead move all those data to Study table/class

PR #2651, opened 2025-08-12
build(sqlalchemy): complete bump to sqlalchemy 2.0

TheoPascoli:


antarest/launcher/model.py
```diff
@@ -243,7 +243,7 @@ def copy_jobs_for_study(self, study_id: str) -> typing.Self:
             for column in self.__table__.columns
             if column.key not in ["id", "study_id"]
         }
-        return JobResult(
+        return self.__class__(
```
MartinBelthle: Why not just keeping `JobResult` ?

antarest/login/service.py
```diff
@@ -288,7 +288,7 @@ def get_group_info(self, id: str) -> Optional[GroupDetailDTO]:
             logger.error("group %s not found by user %s", id, get_user_id())
             raise GroupNotFoundError()
 
-    def get_user(self, id: int) -> Optional[User]:
+    def get_user(self, id: int) -> Optional[Union[User, UserLdap]]:
```
MartinBelthle: We should use the pipe instead of Union

antarest/study/service.py
```diff
@@ -870,7 +870,7 @@ def create_study(self, study_name: str, version: Optional[StudyVersion], group_i
             additional_data=StudyAdditionalData(author=author, editor=author),
         )
 
-        raw = self.storage_service.raw_study_service.create(raw)
+        raw = cast(RawStudy, self.storage_service.raw_study_service.create(raw))
```
MartinBelthle: I think we shouldn't have all these cast.
For instance, here I think I solved the issue by removing the create method inside the IStudyStorgae and inside the variant_study_service. Then changing the signature to RawStudy fixes the mypy issue we don't need the cast anymore

antarest/study/service.py
```diff
@@ -1851,7 +1851,7 @@ def archive(self, uuid: str) -> str:
 
         def archive_task(notifier: ITaskNotifier) -> TaskResult:
             study_to_archive = self.get_study(uuid)
-            self.storage_service.raw_study_service.archive(study_to_archive)
+            self.storage_service.raw_study_service.archive(cast(RawStudy, study_to_archive))
```
MartinBelthle: Here we could just change the type inside the method to Study instead of RawStudy and remove the cast

MartinBelthle (changes requested):
We should remove the `cast` everywhere we can as it's the equivalent to having type ignore inside the DB classes

antarest/launcher/model.py
TheoPascoli: Mypy not happy

antarest/study/service.py
TheoPascoli: I made `create` return a `RawStudy` and added a cast at the end

antarest/study/service.py
TheoPascoli: I like to have the more specific type possible

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -249,12 +249,12 @@ def create(self, metadata: RawStudy) -> RawStudy:
 
         metadata.path = str(path_study)
 
-        return metadata
+        return cast(RawStudy, metadata)
```
sylvlecl: In general we'd better assert the type instead of just casting it, otherwise we may have surprises at runtime. Casting should be reserved to cases where we are 100% sure we have the expected type, here it's not the case since we could have any `Study` in argument.

That said, for this particular case: `create()` should actually not be defined in the abstract class, only the one of this implementation is used --> we can remove it from the abstract class, and type this one with `RawStudy` as input

antarest/study/service.py
```diff
@@ -1287,7 +1287,7 @@ def delete_study(self, uuid: str, children: bool) -> None:
             _ = study.workspace
             study_info = study.to_enhanced_json_summary()
 
-        if self.storage_service.variant_study_service.has_children(study):
```
sylvlecl, line 1290: hrer we don't need to cast: actually the `has_children` method should expect a `Study` type, because any type of study can have children, not only variants;

antarest/study/service.py
```diff
@@ -1352,7 +1352,7 @@ def import_study(
             public_mode=PublicMode.NONE if group_ids else PublicMode.READ,
             groups=group_ids,
         )
-        study = self.storage_service.raw_study_service.import_study(study, stream)
+        study = cast(RawStudy, self.storage_service.raw_study_service.import_study(study, stream))
```
sylvlecl: Here we should not need to cast, the method `import_study` should return a `RawStudy` instead

antarest/study/service.py
```diff
@@ -1661,7 +1661,7 @@ def set_public_mode(self, study_id: str, mode: PublicMode) -> None:
     def check_errors(self, uuid: str) -> List[str]:
         study = self.get_study(uuid)
         self.assert_study_unarchived(study)
-        return self.storage_service.raw_study_service.check_errors(study)
+        return self.storage_service.raw_study_service.check_errors(cast(RawStudy, study))
```
sylvlecl: This code is actually currently buggy:
it can be called for a variant study through the API, whereas it should not.

I think we should just remove it, since it's actually not used and not maintained. (in that case we must document it as a breaking change in release notes)

If we keep it, here we should raise an exception if it's not a RawStudy

antarest/study/service.py
```diff
@@ -1894,7 +1894,7 @@ def unarchive(self, uuid: str) -> str:
 
         def unarchive_task(notifier: ITaskNotifier) -> TaskResult:
             study_to_archive = self.get_study(uuid)
-            self.storage_service.raw_study_service.unarchive(study_to_archive)
+            self.storage_service.raw_study_service.unarchive(cast(RawStudy, study_to_archive))
```
sylvlecl: Ok here we are sure it's a RawStudy (checked above).
However, I would still prefer to have a small method like `assert_raw` which actually check the type, otherwise someone could remove the check above and we won't be covered here anymore.
```python
def assert_raw(study: Study) -> RawStudy:
    if not isinstance(study, RawStudy):
        raise <something>
    return study
```

antarest/study/service.py
```diff
@@ -2133,7 +2133,7 @@ def check_and_update_all_study_versions_in_database(self) -> None:
 
         for study in studies:
             storage = self.storage_service.raw_study_service
-            storage.check_and_update_study_version_in_database(study)
+            storage.check_and_update_study_version_in_database(cast(RawStudy, study))
```
sylvlecl: This code seems completely buggy, nothing ensures we have only raw studies.

This feature does not make sense anyway, we should just remove it :)

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -394,7 +394,7 @@ def _get_variant_study(
             raise UnsupportedOperationOnThisStudyType(study.id, "get", "variant")
 
         assert_permission(study, StudyPermissionType.READ)
-        return study
+        return cast(VariantStudy, study)
```
sylvlecl: Here it's actually not always a variant study, see the boolean parameter.

Instead of this parameter, we should separate the function in 2 functions, one which returns a variant study, the other one which returns any kind of study.

antarest/study/service.py
sylvlecl: Like for the unarchive method:
indeed we are sure it's a raw study because checked above, but I would still prefer to have an assertion in case someone removes the other one for any reason

sylvlecl (changes requested):
Great: a few more comments about each case of study casting, which should mostly be removed

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -448,15 +469,17 @@ def _invalidate_snapshot(
 
     def clear_snapshot(self, variant_study: Study) -> None:
```
sylvlecl: Actually here we should accept only variant studies: raw studies don't have snapshots.
Then we'll not need the check anymore inside the method

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -390,8 +388,31 @@ def _get_variant_study(
         if study is None:
             raise StudyNotFoundError(study_id)
 
-        if not isinstance(study, VariantStudy) and not raw_study_accepted:
-            raise UnsupportedOperationOnThisStudyType(study.id, "get", "variant")
+        assert_permission(study, StudyPermissionType.READ)
+        return cast(VariantStudy, study)
```
sylvlecl: we'd better raise at runtime if it's not a variant study, to avoid falling in unexpected behaviours if an ID of raw study is given. Note that we can re-use the `_get_study_by_id` method to avoid some code duplication!

sylvlecl (dismissed):
Left 2 more minor comments, otherwise OK. Thanks for this :)

sylvlecl (dismissed):

PR #2652, opened 2025-08-12
chore(patch): remove "patch" related dead code

sylvlecl:
The data stored in patch column in database is actually not used anymore,
as well as related classes.

This PR cleans it up.


antarest/study/storage/abstract_storage_service.py
```diff
@@ -103,9 +90,6 @@ def get_study_information(
             groups=[GroupDTO(id=group.id, name=group.name) for group in study.groups],
             public_mode=study.public_mode or PublicMode.NONE,
             horizon=additional_data.horizon,
-            scenario=patch_metadata.scenario,
```
sylvlecl, line 106: Note that those fields were actually not defined in the pydantic model, thus never returned by the API.

MartinBelthle (dismissed):

PR #2655, opened 2025-08-12
perf(matrix): return matrices as integers when possible inside endpoint

MartinBelthle:
**We've reached a point where we cannot have performance gains using better parsing / serializing methods. 
But we can still increase our perfs by transmitting lighter data.**

Benchmark performed on my local env with integers matrices (first value is time and second is size of the response):
 
**Current code**

    '8760*1 json': (0.02, 95284),
    '8760*1 plain': (0.02, 35040),
    '8760*1 arrow compressed': (0.01, 1714),
    '8760*1 arrow uncompressed': (0.01, 71450),

    '8760*200 json': (0.14, 13306546),
    '8760*200 plain': (0.8, 13245614),
    '8760*200 arrow compressed': (0.06, 7139250),
    '8760*200 arrow uncompressed': (0.06, 14086586),
          
    '8760*60 json': (0.04, 3739612),
    '8760*60 plain': (0.2, 3679200),
    '8760*60 arrow compressed': (0.03, 41458),
    '8760*60 arrow uncompressed': (0.04, 4226554)
 
 
**With int64**
 

    '8760*1 json': (0.02, 77764),
    '8760*1 plain': (0.01, 17520),
    '8760*1 arrow compressed': (0.01, 1730),
    '8760*1 arrow uncompressed': (0.01, 71466),
       
    '8760*200 json': (0.16, 9802546),
    '8760*200 plain': (0.4, 9741614),
    '8760*200 arrow compressed': (0.1, 6709210),
    '8760*200 arrow uncompressed': (0.1, 14086602),
           
    '8760*60 json': (0.05, 2688412),
    '8760*60 plain': (0.1, 2628000),
    '8760*60 arrow compressed': (0.03, 41922),
    '8760*60 arrow uncompressed': (0.04, 4226554)
 
**With int32**
 
    '8760*1 json': (0.02, 77764),
    '8760*1 plain': (0.01, 17520),
    '8760*1 arrow compressed': (0.01, 1578),
    '8760*1 arrow uncompressed': (0.01, 36426),
       
    '8760*200 json': (0.1, 9802546),
    '8760*200 plain': (0.4, 9741614),
    '8760*200 arrow compressed': (0.08, 5754930),
    '8760*200 arrow uncompressed': (0.08, 7078602),
           
    '8760*60 json': (0.05, 2688412),
    '8760*60 plain': (0.1, 2628000),
    '8760*60 arrow compressed': (0.04, 32330),
    '8760*60 arrow uncompressed': (0.03, 2124154)
 

# Conclusion
For `arrow compressed` and `arrow uncompressed`: perform the INT32 check
For `PLAIN` and `JSON`: perform the INT64 check (or INT32 to simplify the code)

### DIFF
--- a/antarest/study/web/raw_studies_blueprint.py
+++ b/antarest/study/web/raw_studies_blueprint.py
@@ -82,6 +82,8 @@ class MatrixFormat(EnumIgnoreCase):
 
     def serialize_dataframe(self, dataframe: pd.DataFrame) -> Response:
         np_type: type[np.int32] | type[np.int64] = np.int64
+        # For textual formats, int64 and int32 are represented in the same way, so we use int64 to catch bigger numbers.
+        # For the arrow format, on the other hand, the size of an int32 is half the size of an int64 so we prefer int32.
         if self in {MatrixFormat.ARROW_COMPRESSED, MatrixFormat.ARROW_UNCOMPRESSED}:
             np_type = np.int32
         dataframe = simplify_dataframe(dataframe, np_type)
--- a/antarest/study/web/raw_studies_blueprint.py
+++ b/antarest/study/web/raw_studies_blueprint.py
@@ -16,6 +16,7 @@ import logging
 from pathlib import Path, PurePosixPath
 from typing import Annotated, Any, List
 
+import numpy as np
 import pandas as pd
 from fastapi import APIRouter, Body, File, HTTPException
 from fastapi.params import Query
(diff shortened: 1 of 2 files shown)


antarest/study/web/raw_studies_blueprint.py
```diff
@@ -80,6 +81,11 @@ class MatrixFormat(EnumIgnoreCase):
     PLAIN = "plain"
 
     def serialize_dataframe(self, dataframe: pd.DataFrame) -> Response:
+        np_type: type[np.int32] | type[np.int64] = np.int64
+        if self in {MatrixFormat.ARROW_COMPRESSED, MatrixFormat.ARROW_UNCOMPRESSED}:
```
sylvlecl, line 87: We can leave a comment to explain why this is only relevant for arrow.
For textual formats, it does not make a difference since the text representation of an int is always the same. On the other side, for arrow the size of an int32 is half of the size of an int64.

Note that with that implementation, if you have an integer larger than max of int32, you will fall back to float instead of int64. It should not happen too often for our use cases since its around 2 billions.

sylvlecl (dismissed):

PR #2660, opened 2025-08-13
fix(raw): fix getting link output data

sylvlecl:
This is a quick-fix to revert a regression on getting some raw data.
Not great because it will cause some searches to be performed twice,
but at least it works.

The issue is that the previous code was relying on filesystem paths to get
the relative "URL" of a node. This does not work, because paths and URLs
don't map 1 to 1. In particular, a link folder such as "de - fr" is mapped to
a 2-levels URL `["de", "fr" ]`.

If we want to have a more robust solution we'll need to add a URL member to
each node and update it correctly. Not sure it's worth it.

### DIFF
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -126,6 +126,7 @@ from antarest.study.repository import (
 from antarest.study.storage.matrix_profile import adjust_matrix_columns_index
 from antarest.study.storage.rawstudy.model.filesystem.config.model import FileStudyTreeConfigDTO
 from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.rawstudy.model.filesystem.folder_node import FolderNode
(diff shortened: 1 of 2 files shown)


TheoPascoli (dismissed):

antarest/study/service.py
```diff
@@ -2424,9 +2424,4 @@ def get_raw_content(self, uuid: str, path: str, depth: int, formatted: bool) ->
         if isinstance(node, MatrixNode):
             return node.parse_as_dataframe()
 
-        relative_url: Sequence[str]
```
MartinBelthle, line 2427: As discussed, I think just adding 
`if isinstance(node, FolderNode):
    return node.get(url=[], depth=depth)`

before the existing code solves the issue and doesn't alter performance

sylvlecl: Actually no, it will solve the issue for folders, but not for files inside those folders.

PR #2662, opened 2025-08-14
fix(matrix): fix missing import in matrix service

sylvlecl:
Fixes Matrix GC creation on application startup.

Matrix GC creation is modified and an integration test is added,
which de facto solves the missing import issue.


MartinBelthle (commented):
Indeed we should add a UT. Also minor but you can remove the quotes around MatrixGarbageCollector line 562

PR #2664, opened 2025-08-14
feat(matrixstore): add a retention time inside the garbage collector

MartinBelthle:
A prerequisite for using the DAO inside the GenerateThermalClusterTimeSeries command

### DIFF
--- a/antarest/matrixstore/matrix_garbage_collector.py
+++ b/antarest/matrixstore/matrix_garbage_collector.py
@@ -11,6 +11,7 @@
 # This file is part of the Antares project.
 import logging
 import time
+from datetime import datetime
 from os import listdir
 from pathlib import Path
 from typing import Set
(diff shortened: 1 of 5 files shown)


antarest/matrixstore/matrix_garbage_collector.py
```diff
@@ -59,7 +61,21 @@ def clean_matrices(self) -> None:
         saved_matrices = self._get_saved_matrices()
```
sylvlecl: You can use `matrix_service.get_matrices()` here instead, this will avoid to depend on `matrix_dir` which is an internal implementation detail of the service, and to gather all matrix information twice (here and later to get the creation time).

docs/developer-guide/install/1-CONFIG.md
```diff
@@ -280,6 +280,12 @@ default:
 - **Default value:** false
 - **Description:** If `true`, matrices will never be removed. Else, the ones that are unused will.
 
+## **matrix_gc_retention_time**
```
sylvlecl, line 283: Nice ! I would have forgotten this

sylvlecl (changes requested):
I think the implementation can be slightly improved to not depend on the matrix dir anymore

antarest/matrixstore/matrix_garbage_collector.py
MartinBelthle: Indeed

sylvlecl (approved):

PR #2666, opened 2025-08-15
feat(ui-sts): add occurrences edition in additional constraints form

skamril:
ANT-3538


webapp/public/locales/en/main.json
```diff
@@ -591,7 +595,11 @@
   "study.modelization.storages.additionalConstraints.charge": "Charge",
   "study.modelization.storages.additionalConstraints.discharge": "Discharge",
   "study.modelization.storages.additionalConstraints.level": "Level",
-  "study.modelization.storages.additionalConstraints.createError": "Failed to create additional constraint",
+  "study.modelization.storages.additionalConstraints.occurrences": "Occurrences (168h/occurrence)",
```
hdinia, line 598: "(168h/occurrence)" is that important to notice on the table title ?

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/ColumnsResize.tsx
hdinia, line 1: Note: undo/redo does not work for the occurrences resize

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Checks if the occurrences array contains at least one occurrence with non-empty hours.
+ *
+ * @param occurrences - The value to check.
+ * @returns `true` if at least one occurrence has non-empty hours, `false` otherwise.
+ */
+export const isOccurrencesValid = (
+  occurrences: DeepPartial<AdditionalConstraint["occurrences"]> | undefined,
+) => {
+  return occurrences?.find((occ) => occ?.hours && occ.hours.length > 0);
+};
```
hdinia: It should not return a boolean ?

`return Boolean(occurrences?.find((occ) => occ?.hours && occ.hours.length > 0));`

Or change the name because that makes the name isOccurrencesValid a bit misleading, callers will expect a boolean.

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/index.tsx
```diff
@@ -74,17 +68,15 @@ function ConstraintForm({
...
+  const handleSubmit = async ({ dirtyValues }: SubmitHandlerPlus<ConstraintValues>) => {
+    const { name, occurrences, ...newValues } = dirtyValues;
 
     const updatedConstraints = await updateAdditionalConstraints({
       studyId,
       areaId,
       storageId,
       constraints: {
-        [constraintId]: newValues,
+        [constraintId]: { ...newValues, occurrences: occurrences?.filter(Boolean) },
```
hdinia: Here you want to remove falsy values or empty occurrences ?

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/utils.ts
```diff
@@ -91,14 +91,14 @@ const getStorageUrl = (
...
- * Formalizes a storage object by ensuring all properties are defined.
+ * Formats a storage object by ensuring all properties are defined.
  * Using condition with the study version doesn't allow TypeScript
  * to infer properties types.
  *
  * @param storage - Storage object to formalize.
  * @returns Formalized storage object with all properties defined.
  */
-function formalizeStorage(storage: Storage): FormalizedStorage {
+function formatStorage(storage: Storage): FormattedStorage {
```
hdinia: Prefer using the adapter patter naming conventions, examples: 
a common one is `adapt<object>Response` or Request

```ts
// response adapter (API → UI)
export function adaptStorageResponse(storage: Storage): AdaptedStorage {
  return {
    id: storage.id,
    enabled: storage.enabled ?? false,
    adaptedProp: storage.adapted_prop,
  };
}

// request adapter (UI → API)
export function adaptStorageRequest(storage: AdaptedStorage): Storage {
  return {
    id: storage.id,
    enabled: storage.enabled,
    adapted_prop: storage.adaptedProp,
  };
}

```

webapp/src/components/common/dialogs/DigestDialog/index.tsx
```diff
@@ -43,13 +43,7 @@ function DigestDialog({ studyId, outputId, ...dialogProps }: DigestDialogProps)
...
 
   return (
-    <OkDialog
-      {...dialogProps}
-      title="Digest"
-      okButtonText={t("global.close")}
-      fullScreen
-      sx={{ m: 5 }}
-    >
+    <OkDialog {...dialogProps} title="Digest" okButtonText={t("global.close")} fullScreen>
```
hdinia, line 46: is this change needed ? you are making it full screen if I am not wrong

webapp/src/components/common/FieldEditorButtonGroup.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { Box, Tooltip, type ButtonProps, type TextFieldProps } from "@mui/material";
+import { cloneElement } from "react";
+
+export interface FieldEditorButtonGroupProps {
+  children: [
+    React.ReactElement<Pick<TextFieldProps, "slotProps" | "size">>,
+    React.ReactElement<Pick<ButtonProps, "variant" | "size" | "sx">>,
+  ];
+  size?: TextFieldProps["size"];
+  tooltip?: string;
```
hdinia: minor: `tooltipText` since you dont allow JSX elements

webapp/src/components/common/FieldEditorButtonGroup.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+function FieldEditorButtonGroup({ children, size, tooltip }: FieldEditorButtonGroupProps) {
+  const [fieldEditor, button] = children;
```
hdinia, line 28: minor: You may need to add a runtime check, to ensure extra safety on top of the compile TS check

```ts
  if (!Array.isArray(children) || children.length !== 2) {
    console.error(
      `FieldEditorButtonGroup: Expected exactly 2 children but received ${children.length}. `,
    );

    return null;
  }
```

webapp/src/theme/components.ts
```diff
@@ -125,9 +120,8 @@ export default {
     },
   },
   MuiInputBase: {
-    defaultProps: {
-      size: "small",
-    },
+    // Setting the default size to 'small' prevents setting the size 'medium'
+    // defaultProps: { size: "small" },
```
hdinia, line 124: should be uncommented or removed if not needed

webapp/src/theme/components.ts
```diff
@@ -256,4 +250,20 @@ export default {
...
+          bottom: 20,
+          left: 20,
+          right: 20,
+          ".MuiDialog-paperFullScreen": {
+            borderRadius: theme.spacing(1 / 2),
+          },
+        }),
+      },
+    ],
+  },
```
hdinia: not sure about this change, you basically mutate/remove the dialogs "fullscreen" feature of the lib

webapp/src/utils/numberSelectionsUtils.ts
```diff
@@ -0,0 +1,151 @@
...
+export function compactSelections(selections: NumberOrRange[]) {
+  const compacted: NumberOrRange[] = R.groupWith(
+    (a, b) => a + 1 === b,
+    selectionsToNumbers(selections),
+  ).map((group) => (group.length > 1 ? [group[0], group[group.length - 1]] : group[0]));
+
+  return compacted;
+}
+
+export function selectionToString(selection: NumberOrRange) {
```
skamril, line 77: Add JSDoc

webapp/src/utils/numberSelectionsUtils.ts
```diff
@@ -0,0 +1,151 @@
...
+  return Array.isArray(selection) ? selection.join(RANGE_SEPARATOR) : selection;
+}
+
+export function selectionsToString(selections: NumberOrRange[]) {
```
skamril, line 90: Add JSDoc

webapp/src/utils/numberSelectionsUtils.ts
```diff
@@ -0,0 +1,151 @@
...
+      const [start, end] = v.split(RANGE_SEPARATOR);
+      return end ? [Number(start), Number(end)] : Number(start);
+    });
+
+  return compactSelections(selections);
+}
+
+/**
+ * Checks if a selection (number or range, or a string representation of either) is valid.
+ *
```
skamril, line 119: Add @example

webapp/src/utils/numberSelectionsUtils.ts
```diff
@@ -0,0 +1,151 @@
...
+      Number(end) <= maxValue
+    );
+  }
+
+  return isNumericValue(value) && RA.inRange(1, maxValue + 1, Number(value));
+}
+
+/**
+ * Checks if selections (numbers and/or ranges, or a string representation of them) is valid.
+ *
```
skamril, line 160: Add @example

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+export function getSelectedHoursByOccurrence({
+  hours,
+  occurrences,
+  offset,
+}: {
+  hours: number[];
+  occurrences: number;
+  offset: number;
+}) {
+  // Hour elections are only applied to the first occurrence
```
skamril: Replace by: "Only applied to the first occurrence"

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+import NumberSelectionsFE from "@/components/common/fieldEditors/NumberSelectionsFE";
+import type { ControlPlus, SubmitHandlerPlus } from "@/components/common/Form/types";
+import useConfirm from "@/hooks/useConfirm";
+import { HOURS_IN } from "@/utils/date/constants";
+import { Box, Button, ButtonGroup, Divider } from "@mui/material";
+import * as R from "ramda";
+import { useMemo, useRef, useState } from "react";
+import { useWatch } from "react-hook-form";
+import { useTranslation } from "react-i18next";
+import { type ConstraintValues } from "../../utils";
```
hdinia: minor: `import type { ConstraintValues } from ...`

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+  occurrences: ConstraintValues["occurrences"];
+  onEdit: (occurrences: ConstraintValues["occurrences"]) => void;
+}
+
+function OccurrencesTableDialog({ open, onClose, onEdit, control, occurrences }: Props) {
+  const { t } = useTranslation();
+  const [columnCount, setColumnCount] = useState(occurrences.length);
+  const [offset, setOffset] = useState(0);
+  const dataGridApiRef = useRef<DataGridFormApi<OccurrencesData>>(null);
+  const [selectedHours, setSelectedHours] = useState<number[]>(R.range(1, HOURS_IN.WEEK + 1));
```
hdinia: This should be an empty array, otherwise you could affect 1-168 without having a range of hours selected before.

This can lead to overwriting existing configuration unexpectedly 

Image


Also `offset` should be disabled if their is no hours selected in the field, to avoid such accidental manipulation of the existing config (if you change offset only you can "activate", "invert" etc)


Image

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+          const occurrenceIndex = Number(key);
+          return selectedHoursByOccurrence[occurrenceIndex]?.includes(hour)
+            ? updateFn(active)
+            : active;
+        }, rowData);
+      }, data);
+
+      setData(newData);
+    }
+  };
```
hdinia: Minor: Entered values won’t take effect until the field loses focus (after inputing values we have to click outside the field first, then click on "activate")


Image

webapp/src/components/common/dialogs/DigestDialog/index.tsx
hdinia: (Nevermind I think it's because the change you made in the theme no ?)

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+  const defaultData = useMemo<OccurrencesData>(() => occurrencesToGridData(occurrences), []);
+
+  ////////////////////////////////////////////////////////////////
+  // Event Handlers
+  ////////////////////////////////////////////////////////////////
+
+  const handleOffsetChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const value = Number(event.target.value);
+    setOffset(R.clamp(0, HOURS_IN.WEEK - 1, value));
```
hdinia, line 88: offset max is 167 ?

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+                <ButtonGroup
+                  disabled={selectedHours !== null && selectedHours.length === 0}
+                  color="secondary"
+                >
+                  <Button onClick={handleUpdateHours(R.T)}>{t("global.enable")}</Button>
+                  <Button onClick={handleUpdateHours(R.F)}>{t("global.disable")}</Button>
+                  <Button onClick={handleUpdateHours(R.not)}>{t("global.reverse")}</Button>
+                </ButtonGroup>
+                <Divider flexItem orientation="vertical" />
+                <ColumnsResize columnCount={columnCount} maxColumn={1000} onResize={handleResize} />
```
hdinia: typo: `maxColumns`

hdinia (changes requested):

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+  });
+
+  const columns = useMemo<DataGridFormProps["columns"]>(() => {
+    return Array.from({ length: columnCount }, (_, index) => ({
+      id: index.toString(),
+      title: `${t("global.occurrence")} ${index + 1}`,
+    }));
+  }, [columnCount, t]);
+
+  const defaultData = useMemo<OccurrencesData>(() => occurrencesToGridData(occurrences), []);
```
hdinia, line 76: useMemo not necessary on both

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+            </Typography>
+            <Button startIcon={<EditIcon />} onClick={toggleEditDialog}>
+              {t("global.edit")}
+            </Button>
+          </Box>
+          <TableContainer>
+            <Table>
+              <TableBody sx={{ "tr:last-child > *": { border: "none" } }}>
+                {occurrences.map(({ hours }, index) => (
+                  <TableRow key={hours.join(",") + index}>
```
hdinia: use stable key

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+        if (active) {
+          occurrencesAcc[i].hours.push(Number(hour));
+        }
+      });
+
+      return occurrencesAcc;
+    },
+    [],
+  );
+};
```
hdinia: Not 100% sure but I think `Object.values(rowData)` doesn't guarantee order. Should check it and maybe use `Object.entries()`

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export type RowData = Record<string, boolean>;
+
+export type OccurrencesData = Record<string, RowData>;
+
+export const resizeRow = R.curry((resizeValue: number, rowData: RowData) => {
+  const indexes = R.range(0, resizeValue);
+  const values = R.map((i) => rowData[i] || false, indexes);
+  return R.zipObj(indexes, values);
+});
```
hdinia, line 27: Suggestion: replace with a simple curried plain JS function

```ts
export const resizeRow = (resizeValue: number) => (rowData: RowData) => {
  return Object.fromEntries(
    Array.from({ length: resizeValue }, (_, i) => i).map((i) => [i, rowData[i] || false]),
  );
};
```

hdinia (changes requested):
General feedback: add docs and unit tests to high priority / core business logic functions

Data Transformation Functions
```typescript
// OccurrencesTableDialog/utils.ts
1. occurrencesToGridData(occurrences)
2. gridDataToOccurrences(data) 
3. getSelectedHoursByOccurrence({ hours, occurrences, offset })
4. resizeRow(resizeValue)
```

Validation Functions
```typescript
// ConstraintForm/utils.ts
5. isOccurrencesValid(occurrences)

// OccurrencesFE/utils.ts
6. validateOccurrences(v)
```

webapp/src/components/common/dialogs/DigestDialog/index.tsx
skamril: Yes I made a global change for in the theme

webapp/src/components/common/FieldEditorButtonGroup.tsx
skamril: I changed the type

webapp/public/locales/en/main.json
skamril: For me yes, but they don't say anything about it in the ticket, I'm waiting for feedback from users, I will probably have several returns of this kind after the tests

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/index.tsx
skamril: What can I use to avoid using only the index? Occurence didn't have an id.

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
skamril: Yes

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/ColumnsResize.tsx
skamril: Fixed

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/utils.ts
skamril: Fixed

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/index.tsx
skamril: Fixed

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/utils.ts
skamril: I will do like optimization form in anther PR

webapp/src/components/common/FieldEditorButtonGroup.tsx
skamril: I’d keep it consistent: we don’t runtime-check size (TS enforces number), so I don’t see the need to runtime-check children either. If misuse is a real concern, we could add a dev-only guard, but otherwise TS typing as a 2-element tuple should be enough.

webapp/src/theme/components.ts
skamril: I kept it as a comment to understand what the message is about and avoid adding it

webapp/src/theme/components.ts
skamril: We never use full screen without the margin on the whole application

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
skamril: Fixed

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
skamril: I'm waiting user feedback about this, I followed playlist behavior

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
skamril: I cannot reproduce

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
skamril: Fixed

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
skamril: `occurrencesToGridData` depending on the size of the occurrences can be resource intensive.

For the columns it's a Glide Data Grid recommendation:

> The GridColumn[] passed to the DataEditor in the columns property should be memoized to avoid excessive re-rendering.

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/utils.ts
skamril: Good remark.

For `Object.values()` (same for `Object.entries()`):

> The order of the array returned by Object.values() is the same as that provided by a [for...in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) loop.

Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values

For `for...in`:

> Within each component of the prototype chain, all non-negative integer keys (those that can be array indices) will be traversed first in ascending order by value, then other string keys in ascending chronological order of property creation.

Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

`OccurrencesData` and `RowData` use integer keys (I added a comment).

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/utils.ts
skamril: Thanks for the suggestion! Since Ramda is already heavily used in the app, I find the current Ramda version simpler and more readable than the plain JS alternative.

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/utils.ts
hdinia: Ok

webapp/src/components/common/dialogs/DigestDialog/index.tsx
hdinia: Ok

webapp/src/components/common/FieldEditorButtonGroup.tsx
hdinia: That's debatable, but okay it's an optional feedback

webapp/src/theme/components.ts
hdinia: Even if that's the case, I don't think it's a good idea to change it globally. It would be better to configure the basic dialog component like this instead and not touch the global style.

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
hdinia: ok we can wait QA/user feedback

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
hdinia: the columns are already memoized if I am not wrong (in the DataGridForm) ?

webapp/src/theme/components.ts
hdinia: Ok

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Storages/StorageConfig/AdditionalConstraints/ConstraintForm/OccurrencesFE/OccurrencesTableDialog/index.tsx
skamril: Yes but DataGridForm uses the `column` prop as dependency to its formatted columns

webapp/src/theme/components.ts
skamril: Alright

webapp/src/utils/numberSelectionsUtils.ts
```diff
@@ -0,0 +1,184 @@
...
+ * @param selections - An array of numbers and/or ranges.
+ * @returns A sorted array of unique numbers.
+ */
+export function selectionsToNumbers(selections: NumberOrRange[]) {
+  const numbersSet = new Set<number>();
+
+  selections.forEach((selection) => {
+    if (Array.isArray(selection)) {
+      const [start, end] = selection;
+      R.range(start, end + 1).forEach((num) => numbersSet.add(num));
```
hdinia, line 40: callback passed to `forEach()` should not return a value

```ts
 R.range(start, end + 1).forEach((num) => {
        numbersSet.add(num);
 });
```

webapp/src/components/common/DataGridForm.tsx
```diff
@@ -69,14 +69,19 @@ export interface DataGridFormProps<TData extends Data = Data, SubmitReturnValue
...
-    event?: React.BaseSyntheticEvent,
+    event: React.FormEvent<HTMLFormElement>,
   ) => void | Promise<SubmitReturnValue>;
   onSubmitSuccessful?: (data: SubmitHandlerPlus<TData>, submitResult: SubmitReturnValue) => void;
   onDataChange?: (data: TData) => void;
   onStateChange?: (state: DataGridFormState) => void;
+  id?: string;
   sx?: SxProps<Theme>;
-  extraActions?: React.ReactNode;
+  extraActions?: React.ReactNode | ((state: { canSubmit: boolean }) => React.ReactNode);
```
hdinia, line 79: having `extraActions` be either a `ReactNode` or a function creates unnecessary complexity at the render site: -> `{typeof extraActions === "function" ? extraActions({ canSubmit }) : extraActions}`

---

Solution 1:
I suggest you define it as always a function and update existing locations 

---

Solution 2:
normalize it at the top of the component, so you only need to check once

```ts

function DataGridForm<TData extends Data>({
  defaultData,
  columns,
  allowedFillDirections = "vertical",
  onRowAppended,
  trailingRowOptions,
  enableColumnResize,
  rowMarkers: rowMarkersFromProps,
  onSubmit,
  onSubmitSuccessful,
  onDataChange,
  onStateChange,
  sx,
  id,
  extraActions: extraActionsFromProps,
  apiRef,
  submitButtonText,
  submitButtonIcon = <SaveIcon />,
  hideSubmitButton = false,
  disableErrorSnackbar = false,
}: DataGridFormProps<TData>) {
  const [isSubmitting, setIsSubmitting] = useState(false);
  const [savedData, setSavedData] = useState(defaultData);
  const [error, setError] = useState<string | null>(null);
  const [{ present: data }, { set: setData, undo, redo, canUndo, canRedo }] = useUndo(defaultData);

  // Normalize extraActions to always be a function
  const extraActions = useMemo(
    () => typeof extraActionsFromProps === "function" 
      ? extraActionsFromProps 
      : () => extraActionsFromProps,
    [extraActionsFromProps]
  );
```

then call it 

```ts
  {extraActions && (
            <Box sx={{ marginLeft: "auto", display: "flex", alignItems: "center", gap: 1 }}>
              {extraActions({ canSubmit })}
            </Box>
          )}
```

---

Solution 3: split into 2 props

keep both use cases but with separate, clearly named props:

```ts
  extraActions?: React.ReactNode;
  renderExtraActions?: (state: { canSubmit: boolean }) => React.ReactNode;
```

then in the render:

```ts
          {(extraActions || renderExtraActions) && (
            <Box sx={{ marginLeft: "auto", display: "flex", alignItems: "center", gap: 1 }}>
              {renderExtraActions ? renderExtraActions({ canSubmit }) : extraActions}
            </Box>
          )}

hdinia (changes requested):

webapp/src/utils/numberSelectionsUtils.ts
skamril: It's not important, by default all functions return `undefined`

hdinia: The `forEach` method is designed for side effects only it ignores return values. When you return a value from a `forEach` callback, it suggests you might have meant to use `map` instead, which could be confusing.

hdinia: https://eslint.org/docs/latest/rules/array-callback-return

hdinia (approved):

PR #2667, opened 2025-08-18
perf(raw): avoid double tree parsing when getting file content

sylvlecl:
Improves over https://github.com/AntaresSimulatorTeam/AntaREST/pull/2660

- we now correctly get the unconsumed URL part and use it
- the PR also fixes some of the many typing issues in inode related code


antarest/study/service.py
```diff
@@ -2420,18 +2419,9 @@ def get_raw_content(self, uuid: str, path: str, depth: int, formatted: bool) ->
...
-        # in the specific case of folders, we can assume there is no
-        # "remaining parts" in the URL.
-        # This is a technical debt because this is clearly an implicit contract.
-        # Should be solved by returning remaining parts after a search, for ex.
-        if isinstance(node, FolderNode):
-            return node.get(url=[], depth=depth)
-
-        # Else we have to restart from the root to guarantee the right behaviour
-        return file_study.tree.get(url, depth=depth, formatted=formatted)
+        return node.get(url=relative_url, depth=depth, formatted=formatted)
```
sylvlecl, line 2389: This is the main goal of the PR. We don't get from the root a second time anymore and the code is more clear.

antarest/study/storage/rawstudy/model/filesystem/matrix/matrix.py
```diff
@@ -77,50 +81,15 @@ def __init__(
...
-
-        if expanded:
-            link_content = self.matrix_mapper.get_link_content(self)
-            if link_content is not None:
-                return link_content
-            return self.get_lazy_content()
-
-        return cast("str | G", self.load(url, depth, expanded, formatted))
+    ) -> str | bytes | JSON:
+        return self._get(url, depth, expanded, formatted)
```
TheoPascoli: Now that this method only calls `_get`, can we remove this method and rename `_get` to `get` ?

sylvlecl: Indeed, I removed it and in other node implementations too, well spotted.

antarest/study/storage/rawstudy/model/filesystem/folder_node.py
```diff
@@ -60,41 +60,26 @@ def _forward_get(
...
+                key: children[key].get(
+                    sub_url,
+                    depth=depth,
+                    expanded=False,
+                    formatted=formatted,
+                )
+                for key in names
+            }
 
+    def _expand_get(self, depth: int = -1, formatted: bool = True) -> JSON:
```
MartinBelthle: Seems we could just write `(self, depth: int, formatted: bool)` but I don't know if we want to

antarest/study/service.py
```diff
@@ -2374,18 +2373,9 @@ def get_raw_content(self, uuid: str, path: str, depth: int, formatted: bool) ->
         assert_permission(study, StudyPermissionType.READ)
         file_study = self.get_file_study(study)
         url = [item for item in path.split("/") if item]
-        node = file_study.tree.get_node(url)
+        node, relative_url = file_study.tree.get_node_and_remainder(url)
 
         if isinstance(node, MatrixNode):
             return node.parse_as_dataframe()
```
MartinBelthle, line 2387: Perhaps we should add a comment here saying why we're doing this particular case

MartinBelthle (dismissed):

MartinBelthle:
Also minor but I think the PR should be named `perf` instead of `fix` as we're not really fixing an issue

PR #2669, opened 2025-08-18
fix(ui-optimization): add new options for 'exportMps' field

skamril:
options available since v8.3

ANT-3569


webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
```diff
@@ -12,93 +12,81 @@
...
-export const toBooleanIfNeeded = R.cond([
-  [R.equals("true"), R.T],
-  [R.equals("false"), R.F],
-  [R.T, R.identity],
-]);
+export const formatValuesForForm = (
+  values: OptimizationFormFields,
+  studyVersion: number,
+): FormattedOptimizationFormFields =>
+  R.evolve(
```
hdinia: `R.evolve` adds unnecessary complexity for simple object transformations, you can use plain JS

skamril: For me the code is easier with this util. The JSDoc for `R.evolve` is available.
Then I use it in the function below too, so by coherence I used it here also.

```typescript
const { transmissionCapacities, exportMps, ...rest } = values;

return {
  ...rest,
  transmissionCapacities:
    typeof transmissionCapacities === "boolean"
      ? booleanToString(transmissionCapacities)
      : transmissionCapacities,
  exportMps:
    typeof exportMps === "boolean" && studyVersion >= 830 ?  
      value ? "both-optims" : "none"
      : value
}
```

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
```diff
@@ -12,93 +12,81 @@
...
+] as const;
+
+export const SIMPLEX_OPTIMIZATION_RANGE_OPTIONS: SimplexOptimizationRange[] = [
+  "day",
+  "week",
+] as const;
+
+export const LEGACY_TRANSMISSION_CAPACITIES_OPTIONS: Array<
+  BooleanToStringOrIdentity<LegacyTransmissionCapacities>
+> = ["true", "false", "infinite"] as const;
```
hdinia: `BooleanToStringOrIdentity` adds unnecessary complexity for a simple string array. `as const` adds immutability and a plain literal type 
Type will be inferred as: readonly ["true", "false", "infinite"]
Which is assignable to: ("true" | "false" | "infinite")[]

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
```diff
@@ -12,93 +12,81 @@
...
-): Promise<void> {
-  return client.put(makeRequestURL(studyId), values);
+export const UNFEASIBLE_PROBLEM_BEHAVIOR_OPTIONS: UnfeasibleProblemBehavior[] = [
+  "warning-dry",
+  "warning-verbose",
+  "error-dry",
+  "error-verbose",
+] as const;
+
+export const SIMPLEX_OPTIMIZATION_RANGE_OPTIONS: SimplexOptimizationRange[] = [
```
hdinia: if using `as const` the explicit types may not be needed here for options, any reason to add them ?

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
```diff
@@ -12,93 +12,81 @@
...
+  extends Omit<OptimizationFormFields, "transmissionCapacities"> {
+  transmissionCapacities: BooleanToStringOrIdentity<TransmissionCapacities>;
 }
 
-export const toBooleanIfNeeded = R.cond([
-  [R.equals("true"), R.T],
-  [R.equals("false"), R.F],
-  [R.T, R.identity],
-]);
+export const formatValuesForForm = (
```
hdinia: you can use common adapter pattern naming usually starting by "adapt" + object + Response or Request

`formatValuesForForm` -> `adaptOptimizationResponse`
`formatValuesForApi` -> `adaptOptimizationRequest`

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/index.tsx
```diff
@@ -12,26 +12,55 @@
...
+  const handleSubmit = async ({
+    dirtyValues,
+  }: SubmitHandlerPlus<FormattedOptimizationFormFields>) => {
+    const updatedValues = await setOptimizationFormFields({
+      studyId: study.id,
+      values: formatValuesForApi(dirtyValues),
+    });
+
+    return formatValuesForForm(updatedValues, studyVersion);
   };
```
hdinia: adapters should live at API layer, use them in the request/response api functions instead (see the other comment about naming)

hdinia: example for the get

```ts
export async function getOptimizationFormFields({ 
  studyId,
  studyVersion 
}: BaseOptimizationParams & {studyVersion: number}) {
  const { data } = await client.get<OptimizationFormFields>(format(URL, { studyId }));
  
  // Adapt API response to consistent internal format
  return adaptOptimizationResponse(data, studyVersion);
}

File: webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts (hunk 4 above)
Comment (hdinia) [reply webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts]: (see my next comment about those adapter functions)

File: webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts [hunk 6]
```diff
@@ -12,93 +12,81 @@
...
+  "none",
+  "optim-1",
+  "optim-2",
+  "both-optims",
+] as const;
+
+export interface FormattedOptimizationFormFields
+  extends Omit<OptimizationFormFields, "transmissionCapacities"> {
+  transmissionCapacities: BooleanToStringOrIdentity<TransmissionCapacities>;
 }
```
hdinia: This interface creates a parallel type system that adds complexity without providing real value

remove `FormattedOptimizationFormFields` + `BooleanToStringOrIdentity` (complex type transformation) and use `OptimizationFormFields` -> Always strings in forms

Just transform the data at API layer, not the types

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
skamril: Personally I didn't like ternary inside a ternary, it's difficult to read.
But like I said I used it here because I used it in the other adapter (you cannot do spread like this one), so the two use the same util.

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
skamril: The goal is to keep a strong link with `LegacyTransmissionCapacities`.
If I write `("true" | "false" | "infinite")[]` directly, it is no longer tied to `LegacyTransmissionCapacities`.

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
skamril: I will change it by this:

const OPTIONS: readonly Options[] = [...];

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
skamril: `transmissionCapacities` is a select with boolean and string values, boolean are not allowed in a select.
I need the transformation only for the select not everywhere.

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/index.tsx
skamril: Since this adapter is dedicated to form handling, I didn’t apply it here to avoid mixing concerns between API response shaping and form-specific logic.

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
skamril: Have you a link about this convention?

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
hdinia: It's not a matter of personal preference here, otherwise I would not comment. `R.evolve` adds mental effort to understand what this API does, and for such a simple case, it's counterproductive.

hdinia: This feedback is optional because for now we dont have coding guidelines about this topic. But we must setup a ticket to do them ASAP and this should be clear that if it's not strictly required we must avoid using extra dependencies to do simple stuff, especially if it's just about avoiding a particular syntax

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
hdinia: `as const` for this use case is simple and type safe, like we do for other views. If you want to check types you can refer to the main interface(`OptimizationFormFields`)

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
hdinia: keep `as const` and remove the types annotations

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
hdinia: You talk about the adapter pattern, or just about the naming ?

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/index.tsx
hdinia: The API you use is specific to the Form values as the naming suggest (`getOptimizationFormFields`), and data transformations should never occur in the UI layer.

Even if later this API data is needed elsewhere (unlikely), you just create a different function that does not transform  for the UI form, and serves a general purpose only.

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
hdinia: I know but the point here is you could simplify this, especially `BooleanToStringOrIdentity`, if you remove it from the options array annotation, you could replace it by only `string` here, or create a derived type which is simpler

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
hdinia: For the naming their is no fixed rule, but a common choice to start an adapter function is "adaptSomething" or "transformSomething", their is nothing wrong with the name you choose but we should stick to a particular rule, for us to know we deal with an adapter

hdinia: And about the pattern in general, you should now about it it's very common when dealing with API's, here's some ressources if it can help: 
[API Layer & Data Transformations](https://profy.dev/article/react-architecture-api-layer-and-data-transformations)
[Adapter in TypeScript](https://refactoring.guru/design-patterns/adapter/typescript/example)

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/index.tsx
skamril: "as the naming suggest " => I will rename it. 

"just create a different function that does not transform for the UI form" => Or I keep this one and create another for the UI part based on it (this the adapter pattern)

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
skamril: For the first link: we can clearly see that the adapter is implemented separately, rather than being directly embedded in the base function. That’s essentially the same logic being applied here.

For the second link: it adapts the structure of the response to make it easier for the frontend to consume. Since this is a structural adjustment (not a data transformation), it can be applied globally to provide consistency everywhere. That’s not really relevant in my case.

skamril: For the first link: we can clearly see that the adapter is implemented separately, rather than being directly embedded in the base function. That’s essentially the same logic being applied here.

For the second link: it adapts the structure of the response to make it easier for the frontend to consume. Since this is a structural adjustment (not a data transformation), it can be embedded in the base function to provide consistency everywhere. That’s not really relevant in my case.

hdinia (dismissed):

webapp/src/components/App/Singlestudy/explore/Configuration/Optimization/utils.ts
hdinia: Yes it's a very basic example, but you got the main point : UI layer -> no responsibility about data transformations, we can keep that in the `adapters.ts` file, and handle it at API layer, like you did in the changes it's good

hdinia (approved):

PR #2670, opened 2025-08-19
chore(study)!: remove unused validation feature

sylvlecl:
This is a proposition to remove dead code which aimed at validating the content of a file study.

Reasons are:
 - it is not used and not maintained (lacks tests)
 - it is often poorly implemented (checking only types)
 - it duplicates a lot of validation implemented at th pydantic model level
 - it should be re-implemented at the DAO / import level
 - it's quite a lot of code :)


TheoPascoli (approved):
-1000 awesome ! The best PRs are the one removing code ;)

PR #2672, opened 2025-08-19
feat(linkbdd): implement bdd for link

TheoPascoli:
**To discuss** 

The hybrid dao is useful to perform the migration from files to bdd step by step and to have some models stored in bdd while the others are still stored in the files.

The migration is perform "on the fly" while using the app. If you import a study, all objects with bdd implementation are saved in bdd. 
When exporting a study, everything that is in bdd is saved in the files, then zipped.

When using an old study, when you try to get a link, it wont be found in bdd, so it will fetch in files, then save them in bdd for the next times. So you will fetch old models in file only one time per old study. 

In the end, when everything is migrated, the hybrid dao is renamed to bdddao for example.

Note : some tests has been modified poorly, just to make them pass ... To change later


antarest/study/dao/db/link.py
```diff
@@ -0,0 +1,59 @@
...
+from antarest.core.persistence import Base
+from antarest.study.business.model.link_model import (
+    AssetType,
+    LinkStyle,
+    TransmissionCapacity,
+)
+from antarest.study.model import Study
+
+
+class Link(Base):
```
sylvlecl, line 33: Since the DB objects will anyway be "hidden" behind the DAO layer, I think we'd be better off not using the ORM layer of sqlalchemy, but the core API only.

I had done some testing some time ago that I will try to find, where the plain core API was indeed faster than the ORM layer.

alembic/versions/e316540c7288_add_link_to_bdd.py
```diff
@@ -0,0 +1,59 @@
...
+        sa.Column('link_width', sa.Float(), nullable=False),
+        sa.Column(
+            'link_style',
+            sa.Enum('DOT', 'PLAIN', 'DASH', 'DOT_DASH', 'OTHER', name='linkstyle'),
+            nullable=False
+        ),
+        sa.Column('filter_synthesis', sa.String(), nullable=False),
+        sa.Column('filter_year_by_year', sa.String(), nullable=False),
+        sa.ForeignKeyConstraint(['study_id'], ['study.id'], ondelete='CASCADE'),
+        sa.PrimaryKeyConstraint('study_id', 'area1', 'area2')
```
sylvlecl, line 54: Not sure for links, but note that for other objects, it will be important to have in the primary key an ID separate from the name. This way, we'll be able to implement renaming easily, for example.

antarest/study/dao/hybrid_dao.py
```diff
@@ -0,0 +1,76 @@
...
+from sqlalchemy.orm import Session
+from typing_extensions import override
+
+from antarest.study.business.model.link_model import Link
+from antarest.study.dao.db_link_dao import DbLinkDao
+from antarest.study.dao.file.file_study_dao import FileStudyTreeDao
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+
+
+class HybridDao(FileStudyTreeDao):
```
sylvlecl, line 24: We need to decide the implementation strategy, in particular if we want this kind of hybrid storage or not:
to be honest I think it will make things more complex. But otherwise, we are more or less forced to implement things in a given order, starting with areas.

As discussed:
- one of the first thing we need to do, is also to implement an import feature where we can choose the target DAO (like, simply adding an option in the import API). This way we will be able to use the feature in the app
- we should probably also have a "feature flag" in the configuration to enable or not that feature, until it's completely implemented

sylvlecl (commented):
Looks promising, quite natural to implement!

PR #2673, opened 2025-08-19
feat(matrix_references): finished matrix references and unit testing

mehdiwahada:


antarest/matrixstore/web.py
```diff
@@ -97,6 +99,36 @@ def get(id: str) -> MatrixDTO:
             data=df.to_numpy().tolist(),
         )
 
+    @bp.get(
+        "/matrix/_references/{disk_usage}", tags=[APITag.matrix], description="Fetching a list of matrices statistics"
+    )
+    def get_matrices_references(disk_usage: bool) -> dict[str, MatrixReferencesDTO]:
+        logger.info("Fetching matrices references")
+        used_matrices = service.get_used_matrices()
```
TheoPascoli: All this business logic should be done inside the service. The "web" layer is just the entrypoint of the API, it gets the HTTP request, forward it to the service, then return the response. No business logic should be done here.
It is better because : 
- easier to test (no need to simulate http request to test this)
- reusable: sometimes we may might want to reuse some of the methods

As the others methods here, you should put most of the code here inside the service, and then call the service

antarest/matrixstore/web.py
```diff
@@ -97,6 +99,36 @@ def get(id: str) -> MatrixDTO:
...
+
+        references_dto: dict[str, MatrixReferencesDTO] = {}
+
+        for matrix in used_matrices:
+            refs = []
+            matrix_size = None
+            matrix_id = matrix.matrix_id
+            if disk_usage:
+                matrix_size = service.get(matrix_id).__sizeof__()
```
TheoPascoli: I'm not really a fan of using `__` methods. As an alternative, you could write : 
```
import sys
matrix_size = sys.getsizeof(service.get(matrix_id))
```
Let me know what you think

antarest/matrixstore/web.py
```diff
@@ -97,6 +99,36 @@ def get(id: str) -> MatrixDTO:
             data=df.to_numpy().tolist(),
         )
 
+    @bp.get(
+        "/matrix/_references/{disk_usage}", tags=[APITag.matrix], description="Fetching a list of matrices statistics"
```
sylvlecl: disk_usage should be a query parameter instead of a path parameter, you will need to annotate the function argument with `Query` (you can have a look at other endpoints for examples).

antarest/matrixstore/web.py
```diff
@@ -97,6 +99,36 @@ def get(id: str) -> MatrixDTO:
...
+                matrix_size = service.get(matrix_id).__sizeof__()
+
+            description = matrix.use_description
+
+            ref_dto = MatrixDescriptionDTO.model_validate(MatrixDescriptionDTO(description=description))
+
+            refs.append(ref_dto)
+
+            if matrix_id not in references_dto:
+                refs_dto = MatrixReferencesDTO.model_validate({"refs": refs, "disk_usage": matrix_size})
```
sylvlecl: Here you may use the constructor instead of using a dict and model_validate:
`MatrixReferencesDTO(refs=refs, disk_usage=matrix_size)`

TheoPascoli: I don't think you should include disk_usage in the return object if the boolean is set to False. 
The response should not include None value.
For example, you can add a header to the FastAPI request like that : 
```
 @bp.get(
        "/matrix/_references/{disk_usage}", tags=[APITag.matrix], description="Fetching a list of matrices statistics", response_model_exclude_none=True
    )
```
Maybe there is also a possibility to do it at at the model level. sometimes we do something like model_validate(exclude_none=True).

antarest/matrixstore/web.py
```diff
@@ -97,6 +99,36 @@ def get(id: str) -> MatrixDTO:
...
+        for matrix in used_matrices:
+            refs = []
+            matrix_size = None
+            matrix_id = matrix.matrix_id
+            if disk_usage:
+                matrix_size = service.get(matrix_id).__sizeof__()
+
+            description = matrix.use_description
+
+            ref_dto = MatrixDescriptionDTO.model_validate(MatrixDescriptionDTO(description=description))
```
sylvlecl: you can just use
```python
ref_dto = MatrixDescriptionDTO(description=description)
```

antarest/matrixstore/web.py
```diff
@@ -97,6 +99,36 @@ def get(id: str) -> MatrixDTO:
...
+
+            refs.append(ref_dto)
```
sylvlecl: we'd better move it inside the `if` statement, it's not useful if the references object already exists

antarest/matrixstore/web.py
```diff
@@ -97,6 +99,36 @@ def get(id: str) -> MatrixDTO:
...
+        "/matrix/_references/{disk_usage}", tags=[APITag.matrix], description="Fetching a list of matrices statistics"
+    )
+    def get_matrices_references(disk_usage: bool) -> dict[str, MatrixReferencesDTO]:
+        logger.info("Fetching matrices references")
+        used_matrices = service.get_used_matrices()
+
+        references_dto: dict[str, MatrixReferencesDTO] = {}
+
+        for matrix in used_matrices:
+            refs = []
```
sylvlecl: we'd better move it inside the `if` statement at line 124, it's not useful if the references object already exists.

Same for the lines regarding the disk_usage below.

TheoPascoli (changes requested):
Hello Mehdi, good work overall, just a few things to change. Let me know if you have questions

antarest/matrixstore/web.py
sylvlecl: actually here, what we want is really the size taken by the content of the matrix on disk, which is not what you get here.

To get that information, we could achieve it in 2 different ways:
- when the matrix is created, we get its size from disk and save it to the database. Then we can get it in the `MatrixMetadataDTO` from the matrix service. This is a good approach because we only compute the size once.
- a probably simpler approach is to add a function to the matrix content repository to get the disk usage of a matrix, since it knows where the matrices are stored on disk, and call that function here for each matrix

I propose that we start with the simpler approach.

Also, we'd better move that method to the matrix_service, because it's getting a little too complex for the API definition in web.py which should remain very basic.

sylvlecl (changes requested):
Nice, a few things to change though:

- some suggestions about implementation details
- the disk_usage part needs to be changed to get the actual size of the file on disk

antarest/matrixstore/web.py
```diff
@@ -97,6 +98,23 @@ def get(id: str) -> MatrixDTO:
...
+        tags=[APITag.matrix],
+        description="Fetching a list of matrices statistics",
+    )
+    def get_matrices_references(
+        disk_usage: bool = Query(
+            ...,
+            alias="disk_usage",
+            description="Determine if the disk usage should be displayed",
+            title="Disk Usage",
+            response_model_exclude_none=True,
```
TheoPascoli: I still have null value when setting `disk_usage` to false. You should move this line below tags and description

tests/matrixstore/test_web.py
```diff
@@ -123,3 +123,86 @@ def test_import() -> None:
...
+        "constant_id_1": matrix_reference_two,
+        "test": matrix_reference_three,
+    }
+
+    service.get_used_matrices.return_value = matrix_references
+
+    service.get_matrices_references.return_value = matrix_dict
+
+    res = client.get("/v1/matrix/_references/", headers=create_auth_token(app), params={"disk_usage": False})
+    assert res.status_code == 200
```
TheoPascoli, line 192: Can you also test the content of the http response ? To be sure that there is no disk_usage: null when it is set to false

antarest/matrixstore/service.py
```diff
@@ -578,3 +593,30 @@ def get_matrix_usage(self) -> list[MatrixReference]:
...
+            matrix_size = None
+            matrix_id = matrix.matrix_id
+
+            description = matrix.use_description
+
+            ref_dto = MatrixDescriptionDTO(description=description)
+
+            if matrix_id not in references_dto:
+                refs = []
+                data = self.get(matrix_id)
```
TheoPascoli: The matrix_id **is** the matrix hash, you dont have to load the matrix to calcultate its hash, you can use the matrix_id

antarest/matrixstore/service.py
```diff
@@ -578,3 +593,30 @@ def get_matrix_usage(self) -> list[MatrixReference]:
...
+
+    @override
+    def get_matrices_references(self, disk_usage: bool) -> dict[str, MatrixReferencesDTO]:
+        used_matrices = self.get_used_matrices()
+        references_dto: dict[str, MatrixReferencesDTO] = {}
+        for matrix in used_matrices:
+            matrix_size = None
+            matrix_id = matrix.matrix_id
+
+            description = matrix.use_description
```
TheoPascoli: You use this only one time, you can directly put `matrix.use_description` in the description

antarest/matrixstore/service.py
```diff
@@ -578,3 +593,30 @@ def get_matrix_usage(self) -> list[MatrixReference]:
...
+
+            ref_dto = MatrixDescriptionDTO(description=description)
+
+            if matrix_id not in references_dto:
+                refs = []
```
TheoPascoli: You dont have to instanciate a list here because there can only be one element in it at the same time. 
1. You remove `refs = []`
2. You remove `refs.append(ref_dto)`
3. You write `refs_dto = MatrixReferencesDTO(refs=[ref_dto], disk_usage=matrix_size) `(that creates a list with one single element in it)

TheoPascoli (changes requested):

antarest/matrixstore/repository.py
```diff
@@ -336,3 +338,10 @@ def delete(self, matrix_hash: str) -> None:
         # Abandoned lock files are deleted here to maintain consistent behavior.
         lock_file = matrix_path.with_suffix(".tsv.lock")
         lock_file.unlink(missing_ok=True)
+
+    def get_matrix_disk_usage(self, matrix_hash: str) -> int:
+        for internal_format in InternalMatrixFormat:
```
sylvlecl, line 347: we'd better extract a small method `_find_matrix_path` that will be used also by other functions that already perfom the same lookup

antarest/matrixstore/web.py
```diff
@@ -97,6 +98,20 @@ def get(id: str) -> MatrixDTO:
             data=df.to_numpy().tolist(),
         )
 
+    @bp.get(
+        "/matrix/_references/",
+        tags=[APITag.matrix],
+        description="Fetching a list of matrices statistics",
+        response_model_exclude_none=True,
+    )
+    def get_matrices_references(
```
sylvlecl, line 107: Small stuff: this should be restricted to admin users

tests/matrixstore/test_web.py
```diff
@@ -123,3 +123,103 @@ def test_import() -> None:
...
+                {"description": f"Referenced by command {command_id} of study study1"},
+                {"description": "Referenced by raw study study_id"},
+            ]
+        },
+        "constant_id_1": {"refs": [{"description": "Constant matrix"}]},
+        "test": {"refs": [{"description": "Referenced by dataset data_1"}]},
+    }
+    assert res.json() == data__
+
+    assert "disk_usage" not in res.json()[command_id]
```
sylvlecl: You don't need those assertions since we already check the exact content of the response

tests/matrixstore/test_web.py
```diff
@@ -123,3 +123,103 @@ def test_import() -> None:
...
+    assert "disk_usage" not in res.json()["constant_id_1"]
+    assert "disk_usage" not in res.json()["test"]
+
+    res_dict_test = creating_json_res_dict(res)
```
sylvlecl: Same, I don't think we need all the following code since we have already checked the full result

tests/matrixstore/test_web.py
```diff
@@ -123,3 +123,103 @@ def test_import() -> None:
...
+    app = create_app(service)
+    client = TestClient(app)
+    res = client.get("/v1/matrix/_references/", headers=create_auth_token(app), params={"disk_usage": True})
+
+    res_dict_test = creating_json_res_dict(res)
+
+    sorted_matrix_dict = dict(sorted(matrix_dict.items()))
+    sorted_res_matrix_dict = dict(sorted(res_dict_test.items()))
+    assert res.status_code == 200
+    assert sorted_res_matrix_dict == sorted_matrix_dict
```
sylvlecl, line 174: I would be in favor of hard coding the expected result like below, it's less complicated to understand

sylvlecl (changes requested):
Last minor stuff then it's OK !
In particular we need to restrict that feature to admin users, like the listing of all matrices (the GET previously implemented)

sylvlecl (approved):

PR #2677, opened 2025-08-21
feat(layer): storage isolation

TheoPascoli:
I moved layer related methods to a new layer_manager. This manager implements the DAO pattern.
I created 3 commands :
- CreateLayer
- RemoveLayer
- UpdateLayerName

There is still something to do regarding layers, it is the updates of all the areas associated with a given layer. However, this will be done when the Area has been refactored with the new models etc. 

UpdateLayerName might become UpdateLayer if we can implement the araes update in it.


antarest/study/dao/api/layer_dao.py
```diff
@@ -0,0 +1,31 @@
...
+
+class ReadOnlyLayerDao(ABC):
+    @abstractmethod
+    def get_layers(self) -> Sequence[Layer]:
+        raise NotImplementedError()
+
+
+class LayerDao(ReadOnlyLayerDao):
+    @abstractmethod
+    def save_layers(self, layer: Layer) -> None:
```
sylvlecl: typo: we should not have plural here

antarest/study/dao/file/file_study_layer_dao.py
```diff
@@ -0,0 +1,86 @@
...
+        layers = file_study.tree.get(["layers", "layers", "layers"])
+
+        if layer_id not in layers:
+            raise LayerNotFound
+
+        del layers[layer_id]
+
+        file_study.tree.save(layers, ["layers", "layers", "layers"])
+
+    def update_layer_name(self, layer: LayerUpdate) -> None:
```
sylvlecl: (not sure it's used?) So far, the design we have implemented does not implements updates at the DAO level:
the DAO reads and saves whole objects, the update feature is handled outside by commands and managers.
I think we'd better stick with it for now.

antarest/study/storage/variantstudy/model/command/update_layer_name.py
```diff
@@ -0,0 +1,55 @@
...
+
+from antarest.study.business.model.layer_model import LayerUpdate, update_layer_name
+from antarest.study.dao.api.study_dao import StudyDao
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+class UpdateLayerName(ICommand):
```
sylvlecl: I think we should use just `UpdateLayer` since later we'll add the possibility of updating also areas (I think ?)

antarest/study/storage/variantstudy/model/command/remove_layer.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # Overloaded metadata
+    # ===================
+
+    command_name: CommandName = CommandName.REMOVE_LAYER
+
+    # Command parameters
+    # ==================
+
+    parameters: Layer
```
sylvlecl: here just an ID is enough, we don't need other data to remove it

tests/integration/test_integration.py
```diff
@@ -635,16 +635,22 @@ def test_area_management(client: TestClient, admin_access_token: str) -> None:
...
     res = client.get(f"/v1/studies/{study_id}/layers")
     assert res.json() == [
-        LayerInfoDTO(id="0", name="All", areas=["area 1", "area 2"]).model_dump(),
+        Layer(id="0", name="All", areas=["area 1", "area 2"]).model_dump(),
     ]
 
     # Try to delete a non-existing layer
     res = client.delete(f"/v1/studies/{study_id}/layers/1")
-    assert res.status_code == HTTPStatus.NOT_FOUND
+    assert res.json() == {
```
sylvlecl: Indeed in other "managers", we duplicate controls before applying the command in order to get that kind of proper status code, not a 500, which is not even raised for variants since the command will not be applied instantly.

The duplication is not great and we could probably find a way to reduce it, for example by having a valdiation method in commands, but I think we should keep the behaviour.

sylvlecl (changes requested):
A few suggestions + we should try to keep proper error status (404 and so on) before the commands are actually applied.

antarest/study/dao/file/file_study_layer_dao.py
TheoPascoli: Indeed, I forgot to remove it

antarest/study/storage/variantstudy/model/command/update_layer_name.py
TheoPascoli: I agree with you, that is what I have written in the PR description. I will change it now

tests/integration/test_integration.py
TheoPascoli: Yes Done

tests/variantstudy/test_command_factory.py
```diff
@@ -100,6 +100,24 @@
...
+    pytest.param(
+        CommandDTO(
+            action=CommandName.CREATE_LAYER.value,
+            args={"parameters": {"name": "layer_name"}},
+            study_version=STUDY_VERSION_8_8,
+        ),
+        None,
+        id="create_layer",
+    ),
+    pytest.param(
```
sylvlecl, line 112: We should add an update_layer too

antarest/study/storage/variantstudy/model/command/update_layer_name.py
```diff
@@ -0,0 +1,55 @@
...
+
+from antarest.study.business.model.layer_model import LayerUpdate, update_layer_name
+from antarest.study.dao.api.study_dao import StudyDao
+from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+class UpdateLayer(ICommand):
```
sylvlecl: we should update the file name too

antarest/study/storage/variantstudy/model/command/common.py
```diff
@@ -46,6 +46,9 @@ class CommandName(Enum):
     UPDATE_AREAS_PROPERTIES = "update_areas_properties"
     UPDATE_AREA_UI = "update_area_ui"
     REMOVE_AREA = "remove_area"
+    CREATE_LAYER = "create_layer"
+    REMOVE_LAYER = "remove_layer"
+    UPDATE_LAYER = "update_layer_name"
```
sylvlecl: we should rename here too

sylvlecl (changes requested):
Very minor last comments

sylvlecl (approved):

PR #2679, opened 2025-08-22
refactor(scenario_builder)!: improve scenario builder implementation

sylvlecl:
Improves multiple aspects of scenario builder implementation while maintaining
compat with the current API:

 - share implementation between the 2 endpoints "by type" and complete
 - removes irrelevant use of dataframes:
   **speeds up reading by 10x on a large study** (getting thermal scenarios from more than 10s to 1s)
 - clearly separates business model, serialized form, and views for the API
 - better typing and documentation in the API
 - addition of unit tests and integration tests
 - migrate command `UpdateScenarioBuilder` to store only the delta, and store it
   as the business model, not the INI file model.
   **previously, the whole file was stored in DB for any small modification, which would clearly cause problems**

**breaking change**:
the endpoint `/studies/<id>/config/scenariobuilder` does not allow anymore to remove a ruleset by providing an empty dictionary.
The previous behaviour is too inconsistent to be understandable:
- a not empty dictionary is an update (not a replacement)
- but an empty dictionary is a removal (same as a replacement).

This needs to be changed.


antarest/study/business/model/scenario_builder_model.py
```diff
@@ -0,0 +1,323 @@
...
+    SOLAR = "solar"
+    LINK = "ntc"
+    RENEWABLE = "renewable"
+    BINDING_CONSTRAINTS = "bindingConstraints"
+    HYDRO_INITIAL_LEVEL = "hydroInitialLevels"
+    HYDRO_FINAL_LEVEL = "hydroFinalLevels"
+    HYDRO_GENERATION_POWER = "hydroGenerationPower"
+    SHORT_TERM_STORAGE_INFLOWS = "shortTermStorageInflows"
+
+    @override
```
MartinBelthle: Minor but I think the whole point of using StrEnum instead of Enum is to not have to write this

antarest/study/business/model/scenario_builder_model.py
```diff
@@ -0,0 +1,323 @@
...
+    @property
+    def storage_ids(self) -> Mapping[str, Iterable[str]]:
+        return self._storages
+
+    @property
+    def link_ids(self) -> Iterable[str]:
+        return [f"{a1} / {a2}" for a1, a2 in self._links.keys()]
+
+
+def study_index(file_study: FileStudyTree) -> StudyIndex:
```
MartinBelthle: We could just give the areas here and it would remove the usage of the `FileStudyTree` inside a `xxx_model.py` file

antarest/study/business/model/scenario_builder_model.py
```diff
@@ -0,0 +1,323 @@
...
+    """
+    Creates a ruleset initialized with random ("") for all items and years of a study.
+
+    Optionally, you may choose to initialize only certain scenario types.
+    """
+    if scenario_types is None:
+        scenario_types = set(ScenarioType)
+
+    return Ruleset(
+        load=_create_scenarios_mapping(names=index.area_ids, years=years)
```
MartinBelthle, line 222: Minor but I find it hard to read. We could use #fmt: off and at the end #fmt: on to write every field inside one line only

antarest/study/business/model/scenario_builder_model.py
```diff
@@ -0,0 +1,323 @@
...
+    _update_simple_mapping(base.solar, update.solar)
+    _update_simple_mapping(base.wind, update.wind)
+    _update_simple_mapping(base.binding_constraints, update.binding_constraints)
+    _update_double_mapping(base.renewable, update.renewable)
+    _update_double_mapping(base.storage_inflows, update.storage_inflows)
+    _update_simple_mapping(base.ntc, update.ntc)
+
+
+def update_rulesets(base: Rulesets, update: Rulesets) -> None:
+    lower_case_base = {k.lower(): v for k, v in base.items()}
```
MartinBelthle, line 295: Instead of having to do the .lower() here, shouldn't we use a LowerCaseStr inside the RuleSets typing ?

antarest/study/business/model/scenario_builder_model.py
```diff
@@ -0,0 +1,323 @@
...
+        links=((a1, a2) for a1 in areas for a2 in file_study.config.get_links(a1)),
+        bc_groups=file_study.config.get_binding_constraint_groups(),
+        thermals={a: file_study.config.get_thermal_ids(a) for a in areas},
+        renewables={a: file_study.config.get_renewable_ids(a) for a in areas},
+        storages={a: file_study.config.get_st_storage_ids(a) for a in areas},
+    )
+
+
+def _create_scenarios_mapping(names: Iterable[str], years: list[str]) -> AreaScenarios:
+    return {n: {y: RANDOM for y in years} for n in names}
```
MartinBelthle: You could write this instead to satifsy Sonar: `return {n: dict.fromkeys(years, RANDOM) for n in names}`

MartinBelthle:
You didn't modify the `scenariobuilder.py` file but i find it super weird. I think the method `_get_filtering_kwargs` shouldn't be private as it's also defined inside this file. And also I don't understand why we're doing the line `url[:] = [section]`. It even has a comment to explain but i don't understand

antarest/study/business/scenario_builder_management.py
```diff
@@ -142,135 +70,59 @@ def _get_active_ruleset_name(file_study: FileStudy, default_ruleset: str = "Defa
     return active_ruleset
 
 
-def _build_ruleset(file_study: FileStudy, symbol: str = "") -> RulesetMatrices:
+def _read_ruleset(file_study: FileStudy, scenario_type: ScenarioType) -> Ruleset:
     ruleset_name = _get_active_ruleset_name(file_study)
```
MartinBelthle, line 75: We could start to use the DAO easily, for instance, replacing the method `_get_nb_years`, by `study.get_study_dao().get_general_config().nb_years`

For the active scenario we'll have to put the info inside the GeneralConfig model we do not handle it yet :/

antarest/study/storage/variantstudy/model/command/update_scenario_builder.py
```diff
@@ -70,35 +89,26 @@ def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = N
...
+        study_data.tree.save(serialize_rulesets(rulesets), url)  # type: ignore
         return command_succeeded(message="Scenario builder updated successfully")
 
     @override
     def to_dto(self) -> CommandDTO:
         return CommandDTO(
-            action=CommandName.UPDATE_SCENARIO_BUILDER.value, args={"data": self.data}, study_version=self.study_version
+            action=CommandName.UPDATE_SCENARIO_BUILDER.value,
+            version=self._SERIALIZATION_VERSION,
+            args={
```
MartinBelthle, line 112: When testing on my local env the args are like this:
```
{"args":{"data":{"Default Ruleset":{"load":{"area_1":{"1":52},"area_2":{"3":4}},"thermal":{},"hydro":{}, .....}}
```
, we should only have: 
```
{"args":{"data":{"Default Ruleset":{"load":{"area_1":{"1":52},"area_2":{"3":4}}}}
```

I'm afraid it will be a bit hard as we cannot differentitate someone putting {} to remove all values and someone not setting this value :/ We probably have to introduce a new class with None ?

antarest/study/storage/variantstudy/model/command/update_scenario_builder.py
```diff
@@ -70,35 +89,26 @@ def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = N
...
-            if section:
-                curr_section = curr_cfg.setdefault(section_name, {})
-                for key, value in section.items():
-                    if isinstance(value, (int, float)) and not np.isnan(value):
-                        curr_section[key] = value
-                    else:
-                        curr_section.pop(key, None)
-            else:
-                curr_cfg.pop(section_name, None)
+        rulesets = parse_rulesets(study_data.tree.get(url))
```
MartinBelthle, line 96: Here we're reading all rulesets even if we only want to update one, we could go faster by using only the keys we want

antarest/study/storage/variantstudy/model/command/update_scenario_builder.py
```diff
@@ -70,35 +89,26 @@ def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = N
...
-            curr_cfg.setdefault(active_rules_scenario, {})
-
-        # Ensure keys are sorted in each section (to improve reading performance).
-        for section_name, section in curr_cfg.items():
-            curr_cfg[section_name] = dict(sorted(section.items()))
+        if active_rules_scenario and active_rules_scenario.lower() not in {k.lower() for k in rulesets.keys()}:
+            rulesets[active_rules_scenario] = Ruleset()
 
-        study_data.tree.save(curr_cfg, url)  # type: ignore
+        study_data.tree.save(serialize_rulesets(rulesets), url)  # type: ignore
```
MartinBelthle: We could remove the `type: ignore` by using dict instead of mapping everywhere but I don't know if we want to do so

antarest/study/storage/rawstudy/model/filesystem/config/scenario_builder.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def _serialize_links(section: dict[str, RuleValue], scenario_type: ScenarioType, data: AreaScenarios | None) -> None:
+    if not data:
+        return
+    symbol = _SCENARIO_TYPE_SYMBOLS[scenario_type]
+    for link, scenario_link in data.items():
+        for year, value in scenario_link.items():
+            area1, area2 = link.split(" / ")
+            section[f"{symbol},{area1},{area2},{year}"] = value
```
MartinBelthle: I struggled to understand why you weren't writing all values and just the one you updated. It's because you only instantiate the values when returning them to the front-end. If no test exist to ensure this we should add one.

antarest/study/web/views/scenario_builder_views.py
```diff
@@ -0,0 +1,73 @@
...
+    ntc: AreaScenarios | None = Field(alias="ntc", default=None)
+    renewable: AreaItemsScenarios | None = Field(alias="r", default=None)
+    binding_constraints: AreaScenarios | None = Field(alias="bc", default=None)
+    storage_inflows: AreaItemsScenarios | None = Field(alias="sts", default=None)
+
+    @classmethod
+    def from_model(cls, model: Ruleset) -> "RulesetView":
+        field_values = {f: getattr(model, f) for f in Ruleset.model_fields.keys()}
+        non_empty = {k: v for k, v in field_values.items() if v}
+        return RulesetView.model_construct(**non_empty)
```
MartinBelthle, line 59: Why using model_construct instead of model_validate ?

antarest/study/web/views/scenario_builder_views.py
```diff
@@ -0,0 +1,73 @@
...
+    wind: AreaScenarios | None = Field(alias="w", default=None)
+    solar: AreaScenarios | None = Field(alias="s", default=None)
+    ntc: AreaScenarios | None = Field(alias="ntc", default=None)
+    renewable: AreaItemsScenarios | None = Field(alias="r", default=None)
+    binding_constraints: AreaScenarios | None = Field(alias="bc", default=None)
+    storage_inflows: AreaItemsScenarios | None = Field(alias="sts", default=None)
+
+    @classmethod
+    def from_model(cls, model: Ruleset) -> "RulesetView":
+        field_values = {f: getattr(model, f) for f in Ruleset.model_fields.keys()}
```
MartinBelthle: Why not using model_dump ?

tests/integration/conftest.py
```diff
@@ -110,6 +128,13 @@ def admin_access_token_fixture(client: TestClient) -> str:
     return t.cast(str, credentials["access_token"])
 
 
+@pytest.fixture
+def admin_client(client: TestClient, admin_access_token: str) -> TestClient:
```
MartinBelthle, line 132: Nice

tests/storage/repository/filesystem/config/test_scenario_builder.py
```diff
@@ -0,0 +1,123 @@
...
+    rules = {
+        "s,be,1": 2,
+        "s,be,2": 1,
+    }
+    ruleset = parse_ruleset(rules)
+    assert ruleset == Ruleset(solar={"be": {"1": 2, "2": 1}})
+    assert serialize_ruleset(ruleset) == rules
+
+
+def test_ruleset_parsing_wing():
```
MartinBelthle: typo: wind

MartinBelthle (changes requested):
We should add UTs for the update_scenario_builder command

antarest/study/web/views/scenario_builder_views.py
sylvlecl: Just to not perform validation twice, for performance purpose: fastapi already performs pydantic validation when it gets/returns data.

antarest/study/web/views/scenario_builder_views.py
sylvlecl: To avoid the full conversion to dictionary, I just want to get the python objects as they are.

antarest/study/business/model/scenario_builder_model.py
sylvlecl: You're right, the dependency should be the other direction, I will move this creation method to the file study related folder.

sylvlecl:
> You didn't modify the `scenariobuilder.py` file but i find it super weird. I think the method `_get_filtering_kwargs` shouldn't be private as it's also defined inside this file. And also I don't understand why we're doing the line `url[:] = [section]`. It even has a comment to explain but i don't understand

Indeed. I will keep the function because it's indeed useful for performance, but will simplify it:
- remove this weird assignment which modifies the argument, there is no reason to do so. **Edit:** actually it does not work well, some further processing in ini file node indeed relies on it ... Will keep it for now
- keep only the case where URL size is 2, the other levels are not used for now. If we want to improve performances further, we should re-implement it differently, notably without inserting everything in a dictionary which is costly and not useful

Simply removing this filter, on my test case, slows down the load part retrieval from 400 ms to more than 1s

antarest/study/business/model/scenario_builder_model.py
sylvlecl: The current behaviour keeps the original case, I prefer not to change it.
Only the matching is performed ignoring the case, but if in the file we have "Default Ruleset", we keep it as is, not changing to "default ruleset".

But I agree that having that "micro-management" is not good. However I think we should re-work it separately, for example using custom objects which contain both the name and the ID, but which compare equals based on the ID. This way we would have better guarantees than with the LowerCaseId type, which only guarantees lower casing if the dictionary is first validated with pydantic.

antarest/study/business/scenario_builder_management.py
sylvlecl: If that's OK I prefer to handle it in the future DAO PR

antarest/study/storage/rawstudy/model/filesystem/config/scenario_builder.py
sylvlecl: What do you mean? actually everything is written to file except the "" values (random), which must not be written.

But actually I think the "| None " are not needed, I will remove them, I think that's what misled you.

antarest/study/storage/variantstudy/model/command/update_scenario_builder.py
sylvlecl: In any case we need to read everything to rewrite it to file, but it's true that we don't always need to convert it to the Ruleset object.
However I prefer not to optimize it, it will make the logic more complex for a use case which does not really exist today (having multiple large rulesets).

antarest/study/storage/variantstudy/model/command/update_scenario_builder.py
sylvlecl: Done

antarest/study/storage/variantstudy/model/command/update_scenario_builder.py
sylvlecl: Indeed, I did not differentiate RulesetUpdate and Ruleset.
It would probably be better. Not sure I want to do this in that PR :/ 

Note that `{}` does not imply to remove all values, it means that we don't update anything.

sylvlecl: So finally I added the `RulesetUpdate` class to differentiate it.
It makes some parts of the code a little bit more complex because it allows `None`, so we need to handle both cases in some parts of the code, but the typing is more clear.

antarest/study/business/model/scenario_builder_model.py
```diff
@@ -0,0 +1,308 @@
...
+    storage_inflows: AreaItemsScenarios | None = None
+
+    def get(self, scenario_type: ScenarioType) -> AnyScenarios | None:
+        return _get_by_type(self, scenario_type)
+
+    def set(self, scenario_type: ScenarioType, scenarios: AnyScenarios | None) -> None:
+        _set_by_type(self, scenario_type, scenarios)
+
+
+R = TypeVar("R", bound=Ruleset | RulesetUpdate)
```
MartinBelthle: We could just use an alias `RulesetType: TypeAlias = Ruleset | RulesetUpdate`

antarest/study/business/model/scenario_builder_model.py
```diff
@@ -0,0 +1,308 @@
...
+        case ScenarioType.BINDING_CONSTRAINTS:
+            return self.binding_constraints
+        case ScenarioType.LINK:
+            return self.ntc
+        case _:
+            raise ValueError(f"Unknown scenario type {scenario_type}")
+
+
+def _set_by_type(self: R, scenario_type: ScenarioType, scenarios: AnyScenarios | None) -> None:
+    match scenario_type:
```
MartinBelthle, line 172: What happens if scenarios is None here ? Seems we shouldn't enter this function with optional scenarios

antarest/study/storage/rawstudy/model/filesystem/config/scenario_builder.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def _to_percent(value: RuleValue) -> RuleValue:
+    if isinstance(value, (int, float)):
+        return int(100 * value)
+    return value
+
+
+@overload
+def _parse_ruleset(ruleset_data: RulesetFileData, cls: type[Ruleset]) -> Ruleset: ...
```
MartinBelthle, line 158: Why do you use overload here ?

antarest/study/storage/rawstudy/model/filesystem/config/scenario_builder.py
```diff
@@ -0,0 +1,240 @@
...
+    sections = {}
+    for ruleset_name, ruleset in rulesets.items():
+        sections[ruleset_name] = serialize_ruleset(ruleset)
+    return sections
+
+
+T = TypeVar("T")
+
+
+def _check_not_none(value: T | None) -> T:
```
MartinBelthle, line 133: We could just use Any but perhaps you perfer with the typevar

antarest/study/storage/rawstudy/model/filesystem/config/scenario_builder.py
```diff
@@ -0,0 +1,240 @@
...
+
+def _add_value_simple(values: AreaScenarios, key: str, year: str, value: RuleValue) -> None:
+    if value is None:
+        value = RANDOM
+    values.setdefault(key, {})[year] = value
+
+
+def _add_value_double(values: AreaItemsScenarios, key1: str, key2: str, year: str, value: RuleValue) -> None:
+    if values is None:
+        raise ValueError("Scenario mapping should be initialized.")
```
MartinBelthle: How could we reach this code as values is type as a dict ?

tests/storage/repository/filesystem/config/test_scenario_builder.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def test_ruleset_parsing_load() -> None:
+    rules = {
+        "l,be,1": 2,
+        "l,be,2": 1,
+    }
+    ruleset = parse_ruleset(rules)
+    assert ruleset == Ruleset(load={"be": {"1": 2, "2": 1}})
+    assert serialize_ruleset(ruleset) == rules
```
MartinBelthle, line 28: To me we should add a check in this file to ensure serializing a `""` does not write the value

MartinBelthle (changes requested):
Final round after that I'm good

antarest/study/business/model/scenario_builder_model.py
sylvlecl: Done

antarest/study/business/model/scenario_builder_model.py
sylvlecl: Typing guarantees that we don't, for the `Ruleset` object, but it's allowed for updates

antarest/study/storage/rawstudy/model/filesystem/config/scenario_builder.py
sylvlecl: To guarantee that the class argument has same type as return type. I tried to use a generic instead, but could not make it work :/

antarest/study/storage/rawstudy/model/filesystem/config/scenario_builder.py
sylvlecl: Having the typevar preserves type checking, Any does not (if you use Any, the result of the function can be passed to any other function).

tests/storage/repository/filesystem/config/test_scenario_builder.py
sylvlecl: Indeed, and it was not the case actually.

antarest/study/business/model/scenario_builder_model.py
```diff
@@ -0,0 +1,302 @@
...
+    solar: AreaScenarios | None = None
+    ntc: LinkScenarios | None = None
+    renewable: AreaItemsScenarios | None = None
+    binding_constraints: BcGroupScenarios | None = None
+    storage_inflows: AreaItemsScenarios | None = None
+
+    def get(self, scenario_type: ScenarioType) -> AnyScenarios | None:
+        return _get_by_type(self, scenario_type)
+
+    def set(self, scenario_type: ScenarioType, scenarios: AnyScenarios | None) -> None:
```
MartinBelthle: Last comment: seems we could type here that we only expect `AnyScenarios`. If so, it would simplify the code

PR #2680, opened 2025-08-22
feat(solver): add new solver options for linear and quadratic

TheoPascoli:
# # 3623

### DIFF
--- a/antarest/launcher/adapters/local_launcher/local_launcher.py
+++ b/antarest/launcher/adapters/local_launcher/local_launcher.py
@@ -193,16 +193,14 @@ class LocalLauncher(AbstractLauncher):
         elif "coin" in options:
             linear_solver = "coin"
 
-        if version >= STUDY_VERSION_9_3:
-            quadratic_solver = "sirius"
-            if linear_solver:
-                simulator_args += [f"--linear-solver={linear_solver}"]
-            simulator_args += [f"--quadratic-solver={quadratic_solver}"]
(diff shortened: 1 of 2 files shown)


antarest/launcher/adapters/local_launcher/local_launcher.py
```diff
@@ -187,16 +187,22 @@ def _parse_launcher_options(
...
+
+        if version >= STUDY_VERSION_9_3:
+            quadratic_solver = "sirius"
+            if linear_solver:
+                simulator_args += [f"--linear-solver={linear_solver}"]
+            simulator_args += [f"--quadratic-solver={quadratic_solver}"]
+        elif linear_solver:
             if version >= STUDY_VERSION_9_2:
-                simulator_args += [f"--solver={solver}"]
+                simulator_args += [f"--solver={linear_solver}"]
```
sylvlecl: Actually 9.2 has been changed to `--linear-solver` too, we need to adapt this (will be simpler since 9.2 and 9.3 are actually consistent)

sylvlecl (commented):
Finally only `--linear-solver` is supported, in 9.2 as well: `--solver` was never released

antarest/launcher/adapters/local_launcher/local_launcher.py
TheoPascoli: I did the modifications, thanks

sylvlecl (approved):

PR #2681, opened 2025-08-25
feat(scenariobuilder): handle storages additional constraints

sylvlecl:


MartinBelthle (commented):
Looks good to me. We could also add the front-end part inside this PR as it's really small

*NB*: Apparently it's a bit complicated as we introduce a new level in the front so this will be handled separatly

PR #2682, opened 2025-08-25
feat(ui-matrix-filter): enable granular removal of filter values via separate chips

hdinia:
## Overview

The `ListFilterControl` component has been enhanced to display separate, deletable chips for each value or range of values instead of a single concatenated chip.

### Before

- When entering values like "3" and then "6-9", the component would display a single chip: `"3, 6-9"`
- The entire chip could only be cleared using the "Clear All" button
- Individual values or ranges could not be removed

### After

- When entering values like "3" and then "6-9", the component now displays:
    - Chip 1: `"3"` with its own delete button (×)
    - Chip 2: `"6-9"` with its own delete button (×)
- Each chip can be individually deleted by clicking its delete button
- The "Clear All" button still removes all chips at once

### DIFF
--- a/webapp/src/components/common/Matrix/components/MatrixFilter/components/ListFilterControl.tsx
+++ b/webapp/src/components/common/Matrix/components/MatrixFilter/components/ListFilterControl.tsx
@@ -40,7 +40,14 @@ import {
   TYPOGRAPHY_STYLES,
 } from "../styles";
 import type { FilterOperatorType } from "../types";
-import { parseRangeInput } from "../utils";
+import {
+  compactSelections,
+  selectionsToNumbers,
+  selectionToString,
+  stringToSelections,
+  type NumberOrRange,
+} from "../../../../../../utils/numberSelectionsUtils";
+import { buildKey } from "@/utils/reactUtils";
 
 const isValidFilterOperator = (value: string): value is FilterOperatorType => {
   return Object.values(FILTER_OPERATORS).includes(value as FilterOperatorType);
@@ -54,7 +61,7 @@ interface ListFilterControlProps {
   onKeyPress: (event: React.KeyboardEvent) => void;
   onAddValue: () => void;
   onAddValues?: (values: number[]) => void;
-  onRemoveValue: (value: number) => void;
+  onRemoveValues: (values: number[]) => void;
   onOperatorChange?: (operator: FilterOperatorType) => void;
   onClearAll?: () => void;
   placeholder?: string;
@@ -69,6 +76,7 @@ function ListFilterControl({
   onKeyPress,
   onAddValue,
   onAddValues,
+  onRemoveValues,
   onOperatorChange,
   onClearAll,
   placeholder,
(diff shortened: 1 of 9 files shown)


webapp/src/components/common/Matrix/components/MatrixFilter/components/ListFilterControl.tsx
```diff
@@ -263,13 +255,19 @@ function ListFilterControl({
...
-          />
+          <Box sx={{ display: "flex", flexWrap: "wrap", gap: DESIGN_TOKENS.spacing.xs }}>
+            {formattedRanges.map((range, index) => (
+              <Chip
+                key={buildKey(range.label, index)}
+                label={range.label}
+                size="small"
+                color="primary"
+                disabled={disabled}
+                onDelete={disabled ? undefined : () => handleRemoveRange(range.values)}
```
skamril: `handleRemoveRange` is not necessary:

`() => onRemoveValues(range.values)`

webapp/src/components/common/Matrix/components/MatrixFilter/components/ListFilterControl.tsx
```diff
@@ -263,13 +255,19 @@ function ListFilterControl({
...
-            label={formattedValues}
-            size="small"
-            color="primary"
-            disabled={disabled}
-            sx={CHIP_SELECTOR_STYLES.dense}
-          />
+          <Box sx={{ display: "flex", flexWrap: "wrap", gap: DESIGN_TOKENS.spacing.xs }}>
+            {formattedRanges.map((range, index) => (
+              <Chip
+                key={buildKey(range.label, index)}
```
skamril: `range.label` is enough here

webapp/src/components/common/Matrix/components/MatrixFilter/components/ListFilterControl.tsx
```diff
@@ -69,53 +76,33 @@ function ListFilterControl({
...
-        lastRange[1] = curr;
-      } else {
-        // Start a new range
-        acc.push([curr, curr]);
-      }
+    const compacted = compactSelections(selectedValues);
 
-      return acc;
-    }, []);
+    return compacted.map((selection: NumberOrRange) => {
```
skamril, line 96:
```typescript
return compacted.map((selection) => ({
  label: selectionToString(selection),
  values: selectionToNumbers(selection),
}));
```

But you can just return `compacted` and use `selectionToString` is the JSX.

```typescript
const formattedRanges = useMemo(() => compactSelections(selectedValues), [selectedValues]);
```

(`formattedRanges` type will be `NumberOrRange[]`)

skamril (changes requested):

webapp/src/components/common/Matrix/components/MatrixFilter/components/ListFilterControl.tsx
hdinia: It's a bit more concise but I prefer the pre-computed approach to avoid transformations in the JSX

webapp/src/components/common/Matrix/components/MatrixFilter/components/ListFilterControl.tsx
hdinia: done

webapp/src/components/common/Matrix/components/MatrixFilter/components/ListFilterControl.tsx
hdinia: done

skamril (dismissed):

skamril (approved):

PR #2683, opened 2025-08-26
refactor(playlist)!: introduce dao and new class design

MartinBelthle:
**Breaking change:**
Remove unused GET and PUT `/playlist/config` endpoints


antarest/study/dao/file/file_study_playlist_config_dao.py
```diff
@@ -0,0 +1,93 @@
...
+            else:
+                deactivated_years.append(ini_year)
+            if model.weight != 1:
+                # Only write weights if they differ from the default value
+                weights.append(f"{ini_year},{model.weight}")
+
+        content: dict[str, Any] = {"playlist_year_weight": weights}
+
+        all_years = len(playlist.years)
+        # Determine what's lighter to write inside the file
```
sylvlecl, line 83: What about removing that logic on write ?
it's really unnecessarily complicated in my opinion.

sylvlecl (commented):
Looks good, I think we can simplify the write to not use anymore the "reverse" serialization.

TheoPascoli (approved):

antarest/study/dao/file/file_study_playlist_config_dao.py
MartinBelthle: We could change this but to me it's okay to keep it as it is.

As we have to be able to parse such data, I don't think it's that complicated to understand.
Also it's really lighter when reading the file when you have 1000 years in your study, you would have 1000 lines it starts to get heavy :/ and we open it quite often.

@TheoPascoli what do you think ?

sylvlecl: Ok let's keep it as is, it's working anyway

sylvlecl (approved):

PR #2684, opened 2025-08-27
fix(ui-districts): update erase areas

smailio:
When a ditrict has areas, if we update a district with the update district dialog, it will erase its areas. It's not easy to notice because the update happens on the backend but the front is not refreshed so we still see the areas in the table but if you check with the debug mode you'll see that the areas have been erased. 

image


hdinia (approved):
Good catch thanks

PR #2686, opened 2025-08-28
refactor(districts)!: introduce dao

smailio:
breaking change:

1. PUT /studies/{uuid}/districts/{district_id} doesn't accept extra parameters anymore, request will be rejected if it contains other fields than those defined in DistrictUpdate.
2. UpdateDistrict and CreateDistrict commands now accept a parameter argument., which is an object that wraps the parameters for the command. We can still pass parameters directly but we need to also pass an extra parameter "version" set to 1. 
3. The parameter filter_items was renamed to areas
4. the parameter base_filter was renamed to apply_filter
5. the fields set in FileStudyTreeConfigDTO and FileStudyTreeConfig were renamed to districts 

Other changes : 

1.  District object was introduced
2. DistrictSet was renamed DistrictFileData, also this class is not referenced outside of the file where it's defined. FileStudyTreeConfig  now relies on District instead of DistrictDTO. 

Bug fixes : 

1. updating a district attribute no longer erase areas
2. updating a district attribute no longer overwrite the add-all apply filter 
3. we handle studies with an "add-all" apply filter and a "+" attribute. As well as studies wieh "remove-all" and "-". 
4. caption are now optional. If a study has no caption on disk, loading the district won't fail anymore.


antarest/study/business/district_manager.py
```diff
@@ -55,37 +40,38 @@ class DistrictManager:
...
-                DistrictInfoDTO(
-                    id=district_id,
-                    name=district.name,
-                    areas=district.get_areas(all_areas),
-                    output=district.output,
-                    comments=file_study.tree.get(["input", "areas", "sets", district_id]).get("comments", ""),
-                )
-            )
-        return districts
+        # file_study = study.get_files()
```
sylvlecl: I guess this comment should be removed

antarest/study/business/district_manager.py
```diff
@@ -141,25 +117,26 @@ def update_district(
...
-        if invalid_areas := areas - all_areas:
+
+        invalid_areas = study_dao.get_invalid_areas(district_update.areas)
+        if invalid_areas:
             raise AreaNotFound(*invalid_areas)
+
         command = UpdateDistrict(
             id=district_id,
             base_filter=DistrictBaseFilter.remove_all,
-            filter_items=dto.areas or [],
```
smailio, line 160: this `or []` causes a bug, when the user only want to update a property it would erase the areas attached to this district

antarest/study/dao/api/district_dao.py
```diff
@@ -0,0 +1,88 @@
...
+            The (unordered) list of Data Transfer Objects (DTO) representing districts.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def district_exists(self, district_id: str) -> bool:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_invalid_areas(self, areas: list[str]) -> list[str]:
```
smailio: maybe I should already move this function in area management ? The plan as of now is to wait for Area DAO.

antarest/study/storage/rawstudy/model/filesystem/config/files.py
```diff
@@ -223,11 +223,12 @@ def _parse_sets(root: Path) -> Dict[str, DistrictSet]:
         file_type=FileType.MULTI_INI,
         multi_ini_keys=["+", "-"],
     )
+
     return {
         transform_name_to_id(name): DistrictSet(
-            areas=item.get("-" if item.get("apply-filter", "remove-all") == "add-all" else "+"),
```
smailio, line 228: I didn't have to update this code, but I find this line unecesseraly confusing and it made me lose time, so I introduced the district_set_sign function which is a straighforward implementation that doesn't require any mental gymnastics.

tests/study/business/test_district_manager.py
```diff
@@ -44,55 +41,74 @@ def manager(command_context: CommandContext) -> DistrictManager:
...
+                        "comments": "dummy",
+                        "+": ["n1", "n2", "n3"],
+                    },
+                },
+            }
         }
+    }
+    study = FileStudyInterface(empty_study_880)
+    study.file_study.tree.save(district_ini_content)
+    study.file_study.config.sets = {
```
smailio: I don't know if I could easily generate this config automatically instead of setting it manually.

tests/study/business/test_district_manager.py
```diff
@@ -102,182 +118,70 @@ def test_get_districts(self, manager: DistrictManager):
...
-            tree=Mock(spec=FileStudyTree),
-        )
-        study = create_study_interface(file_study, version=STUDY_VERSION_8_8)
-
-        # run
-        dto = DistrictUpdateDTO(output=True, comments="", areas=[])
+    def test_create_district__nominal(self, manager: DistrictManager, study_with_sets: FileStudy):
+        with patch.object(study_with_sets, "add_commands", wraps=study_with_sets.add_commands) as add_commands_mock:
+            dto = DistrictCreation(
+                name="D4",
```
smailio, line 139: In the original test it was D1, but D1 already exist in the fixture so this will raise, that's why I changed it to D4 for this test.

antarest/study/storage/variantstudy/model/command/create_district.py
```diff
@@ -50,7 +45,7 @@ class CreateDistrict(ICommand):
 
     name: str
     base_filter: Optional[DistrictBaseFilter] = None
-    filter_items: Optional[List[str]] = None
```
smailio, line 53: Maybe filter_items made sens back in the time, but as far as I can read in the code this field always contains area. This was unecessary confusing and I lost some time because of it so I'd rather have a more explicit name.

tests/integration/test_integration.py
```diff
@@ -673,7 +673,6 @@ def test_area_management(client: TestClient, admin_access_token: str) -> None:
     res = client.put(
         f"/v1/studies/{study_id}/districts/district%201",
         json={
-            "name": "District 1",
```
smailio, line 665: This is what the breaking change is about. We don't allow to update the name of a district (except through debug mode)  yet this endpoint accept "extra" fields, this allows the user to pass a name field that will just be ignored by the endpoint, which makes no sens and is confusing.

antarest/study/dao/file/file_study_district_dao.py
```diff
@@ -0,0 +1,143 @@
...
+    @override
+    def get_districts(self) -> Sequence[District]:
+        """
+        Returns all districts of the study.
+        """
+        file_study = self.get_file_study()
+        path = DISTRICTS_PATH
+        try:
+            # may raise KeyError if the path is missing
+            districts = file_study.tree.get(path)
```
smailio, line 43: Maybe, instead of reading districts from the tree, I should be reading districts from config, as they are already parsed and stored in config. I provided an alternative implementation in refactor/3464-introduce-dao-districts-v2

antarest/study/storage/variantstudy/model/command/update_district.py
```diff
@@ -42,52 +43,50 @@ class UpdateDistrict(ICommand):
...
-                "caption": sets[district_id]["caption"],
-                "apply-filter": apply_filter,
-                item_key: self.filter_items,
-                "output": study_data.config.sets[district_id].output,
-                "comments": self.comments,
-            },
-            ["input", "areas", "sets", district_id],
+        updated_district = district.model_copy(
+            update=self.parameters.model_dump(
+                mode="json", exclude_none=True, by_alias=True, include={"areas", "output", "comments"}
```
smailio: remove by_alias

antarest/study/storage/variantstudy/model/command/update_district.py
```diff
@@ -42,52 +43,50 @@ class UpdateDistrict(ICommand):
...
+
+    class Config:
+        populate_by_name = True
 
     @override
-    def _apply(self, study_data: FileStudy, listener: Optional[ICommandListener] = None) -> CommandOutput:
-        if self.id not in study_data.config.sets:
+    def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        if not study_data.district_exists(self.id):
             return command_failed(message=f"District '{self.id}' does not exist and should be created")
```
smailio, line 75: add area check

antarest/study/storage/variantstudy/model/command/update_district.py
```diff
@@ -42,52 +43,50 @@ class UpdateDistrict(ICommand):
...
-            ["input", "areas", "sets", district_id],
+        updated_district = district.model_copy(
+            update=self.parameters.model_dump(
+                mode="json", exclude_none=True, by_alias=True, include={"areas", "output", "comments"}
+            )
         )
 
+        try:
+            study_data.save_district(updated_district, self.parameters.base_filter)
+        except AreaNotFound as e:
```
smailio: remove try catch

antarest/study/storage/rawstudy/model/filesystem/config/district.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @classmethod
+    def from_model(cls, district: District, district_base_filter: Optional[DistrictBaseFilter]) -> "DistrictSet":
+        base_filter = district_base_filter or DistrictBaseFilter.remove_all
+        inverted_set = base_filter == DistrictBaseFilter.add_all
+        return DistrictSet.model_validate(
+            {
+                **district.model_dump(include={"name", "output", "comments"}),
+                "inverted_set": inverted_set,
+                "areas": district.areas,
```
smailio: move into include

antarest/study/storage/rawstudy/model/filesystem/config/district.py
```diff
@@ -0,0 +1,85 @@
...
+    Object linked to /inputs/sets.ini information
+    """
+
+    ALL: List[str] = ["hourly", "daily", "weekly", "monthly", "annual"]
+    name: Optional[str] = None
+    inverted_set: bool = False
+    areas: Optional[List[str]] = None
+    output: bool = True
+    filters_synthesis: List[str] = ALL
+    filters_year: List[str] = ALL
```
smailio: remove if I can

tests/variantstudy/model/command/test_manage_district.py
```diff
@@ -108,6 +114,7 @@ def test_manage_district(empty_study_810: FileStudy, command_context: CommandCon
     set_config = sets_config.get("empty district without output")
     assert not set_config["output"]
 
+    # case where district already exists
     output_d3 = create_district3_command.apply(
         study_data=study_dao,
```
smailio, line 124: add test for area not found

antarest/study/dao/api/district_dao.py
```diff
@@ -0,0 +1,94 @@
...
+
+        Returns:
+            The (unordered) list of Data Transfer Objects (DTO) representing districts.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_district(self, district_id: str) -> District:
+        """
+        Get the list of districts defined in this study.
```
sylvlecl: not exact ;)

antarest/study/business/district_manager.py
```diff
@@ -55,82 +40,59 @@ class DistrictManager:
...
-        if district_id in file_study.config.sets:
+        study_dao = study.get_study_dao()
+        district_id = transform_name_to_id(district_creation.name)
+        if study_dao.district_exists(district_id):
             raise DistrictAlreadyExist(district_id)
-        areas = frozenset(dto.areas or [])
-        all_areas = frozenset(file_study.config.areas)
-        if invalid_areas := areas - all_areas:
+
+        invalid_areas = study_dao.get_invalid_areas_in_district(district_creation.areas or [])
```
sylvlecl, line 80: Instead of having a dedicated method in the DAO, this should just call something such as "study_dao.area_exists".
Unfortunately the AreaDao is not yet implemented, but maybe we could start it with just that method ?
Else we risk to forget to remove that one.

antarest/study/dao/api/district_dao.py
```diff
@@ -0,0 +1,94 @@
...
+    @abstractmethod
+    def district_exists(self, district_id: str) -> bool:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_invalid_areas_in_district(self, areas: list[str]) -> list[str]:
+        """
+        Check all areas exists in the study
+        """
+        # TODO this method should be moved to the area DAO when we'll implement it
```
sylvlecl, line 57: Ah, OK this was indeed just a temporary, nice :)

antarest/study/business/model/district_model.py
```diff
@@ -0,0 +1,132 @@
...
+
+    #: Name of the district (this name is also used as a unique identifier).
+    name: str
+    #: Indicates whether this district is used in the output (usually all
+    #: districts are visible, but the user can decide to hide some of them).
+    output: Optional[bool] = True
+    #: User-defined comments.
+    comments: Optional[str] = ""
+    #: List of areas that will be grouped in the district.
+    areas: Optional[List[str]] = []
```
TheoPascoli: You must use `Field(default_factory=list)` instead of `[]` otherwise your list will be shared across all instances of this class

antarest/study/dao/file/file_study_district_dao.py
```diff
@@ -0,0 +1,135 @@
...
+            # may raise KeyError if "list" is missing
+        except KeyError:
+            raise DistrictConfigNotFound(str(path))
+
+        all_areas = list(file_study.config.areas)
+        return [
+            DistrictSet.from_data(district_data).to_model(district_id, all_areas)
```
sylvlecl: I would slightly prefer for consistency sake to have a method `parse_districts` and have `FileData` appear in the name of `DistrictSet`, and be used only in the file where it's defined.
For all other objects we have that clear boundary:
 - class `...FileData` is the almost 1-to-1 image of data as formatted in files
 - it's actually only used in the file where it belongs, as an intermediate to validate file content and to create our business model
 - the transformation of file data to model is called `parse_...` and is the only thing used outside of that file (together with `serialize_...`)

antarest/study/storage/variantstudy/model/command/update_district.py
```diff
@@ -42,52 +42,51 @@ class UpdateDistrict(ICommand):
...
-                "caption": sets[district_id]["caption"],
-                "apply-filter": apply_filter,
-                item_key: self.filter_items,
-                "output": study_data.config.sets[district_id].output,
-                "comments": self.comments,
-            },
-            ["input", "areas", "sets", district_id],
+        district = study_data.get_district(self.id)
+
+        updated_district = district.model_copy(
```
sylvlecl: For consistency sake we'd better have an `update_district` method in the `district_model` module, to encapsulate that

antarest/study/dao/file/file_study_district_dao.py
```diff
@@ -0,0 +1,135 @@
...
+            raise AreaNotFound(*invalid_areas)
+
+        district_id = transform_name_to_id(district.name)
+
+        # Update the in-memory config
+        study_data.config.sets[district_id] = DistrictSet.from_model(district, district_base_filter)
+
+        # Persist the change in the filesystem
+        item_key = district_set_sign_from_base_filter(district_base_filter)
+        apply_filter = district_base_filter if district_base_filter else DistrictBaseFilter.remove_all
```
TheoPascoli: If a district was created with the base filter `add_all` and the user only update fields like `comments` for exemple, the `add_all` will be replaced with `remove_all`, completely inverting the logic here

TheoPascoli (changes requested):

antarest/study/business/model/district_model.py
sylvlecl: Note that I think that pydantic protects you against this (by copying the default value), but it's probably a better habit to use default factory since it does not work for plain dataclasses for example.
https://stackoverflow.com/questions/63793662/how-to-give-a-pydantic-list-field-a-default-value

antarest/study/business/model/district_model.py
```diff
@@ -0,0 +1,132 @@
...
+    #: Base filter for the district.
+    base_filter: Optional[DistrictBaseFilter] = None
```
sylvlecl: The fact that we use those new models in the API actually adds feature to the API:
we can now choose between including or excluding the specified areas, which is something allowed by the file format.

It makes the API and the DAO a little asymmetric because when we update data we can use that feature, but when we get it we cannot get that information. It looks OK for now, but maybe some day we'll want to be able to get that information. At that point we'll need to differentiate the district definition, and its "instantiation" (which areas are actually part of it).

Now I think those 2 classes are not very clear:
the field `areas` looks like we define the areas that are part of that district, but if "base filter" is "add-all", it's the opposite.
It looks to me that we should at least do some renaming to make it more clear:
- maybe renaming "district" to "district definition" in the class names, to differentiate it from "district" which contain the actual areas ?
- renaming base_filter to something like "type" with values "include" or "exclude"

antarest/study/business/model/district_model.py
```diff
@@ -0,0 +1,132 @@
...
+
+from pydantic import field_validator
+from pydantic.alias_generators import to_camel
+
+from antarest.core.serde import AntaresBaseModel
+
+
+class DistrictBaseFilter(Enum):
+    add_all = "add-all"
+    remove_all = "remove-all"
```
sylvlecl, line 23: I know it's not related to your refactoring, but actually, in files, the real value is just "all" (it filters out all).
"remove-all" does not seem to exist.

antarest/study/storage/rawstudy/model/filesystem/config/district.py
```diff
@@ -0,0 +1,88 @@
...
+
+def district_set_sign_from_data(item: dict[str, Any]) -> str:
+    if "apply-filter" not in item:
+        return "+"
+    base_filter = DistrictBaseFilter(item["apply-filter"])
+    return district_set_sign_from_base_filter(base_filter)
+
+
+def district_set_sign_from_base_filter(district_base_filter: Optional[DistrictBaseFilter]) -> Literal["+", "-"]:
+    if district_base_filter is None:
```
sylvlecl: you could use a match here (does not bring much though :) )

antarest/study/storage/rawstudy/model/filesystem/config/district.py
```diff
@@ -0,0 +1,88 @@
...
+        return DistrictSet.model_validate(
+            {
+                **district.model_dump(include={"name", "output", "comments", "areas"}),
+                "inverted_set": inverted_set,
+            }
+        )
+
+    @classmethod
+    def from_data(cls, data: Any) -> "DistrictSet":
+        inverted_set = district_set_sign_from_data(data) == "-"
```
sylvlecl: Here too I know you did not change the behaviour, but it's actually buggy, strictly speaking, compared to the simulator behaviour.
In the file, the district is actually defined by a sequence of rules applied one by one. Those rules are:
- apply-filter which either adds all areas or removes them all
- \+ which adds one area
- \- which removes one area

We could for example have and add-all followed by some + followed by some - (does not make much sense, sure, but it can happen, strictly speaking).
Or even an add-all or filter all at the end which overwrites the previous rules.

We should at least add a check to verify we are in one of the cases we support:
- either an "add-all" followed by only -
- or an "all" or nothing followed by only +

antarest/study/dao/api/district_dao.py
```diff
@@ -0,0 +1,94 @@
...
+        raise NotImplementedError()
+
+
+class DistrictDao(ReadOnlyDistrictDao):
+    @abstractmethod
+    def save_district(self, district: District, district_base_filter: Optional[DistrictBaseFilter]) -> None:
```
sylvlecl: The "base_filter" should not be optional, we'd better force to be explicit or having an explicit default value.

Anyway, here I think we should gather the whole data into one object, maybe `DistrictDefinition`.

This is what we actually want to store, apparently.
**The role of the DAO will be to implement CRUD operations on that object, only that one.**

**THEN, the interpretation of that `DistrictDefinition` to get an instantiated `District` should be implemented in the business layer (manager class).**

antarest/study/dao/api/district_dao.py
```diff
@@ -0,0 +1,94 @@
...
+            If the base filter is "add_all" it means that this distrit will contain all existing areas except the ones in district.areas .
+
+            Otherwise if the base filter is "remove_all" it means the district contains only the areas in district.areas .
+
+        Args:
+            district: District object note that the areas field will contains exactly the area that will be stored.
+            district_base_filter: indicate whether this district include all areas except those given in district or only those .
+
+        Raises:
+            DistrictNotFound: exception raised when district is not found in the study.
```
sylvlecl: I don't think this can happen since it will be created, in that case ?

sylvlecl (changes requested):
My comments may be a little confusing, but the main take-aways are:

1. We need to separate the `DistrictDefinition` which is actually stored, and hence should appear in the DAO interface, from the `District` which is a plain set of areas created from that definition by a business level logic (and should not appear in the DAO)

2. Our interpretation of the file format is actually a little limited compared to the possibilities it offers: we must ensure the content of the files are actually compatible with our interpretation, or raise otherwise


Then a few comments about consistency with other objects.

antarest/study/business/model/district_model.py
smailio: If I do this it messes with mypy

antarest/study/business/model/district_model.py
smailio: what do you mean apply-filter takes two values "add-all" and "remove-all" ?

antarest/study/storage/variantstudy/model/command/update_district.py
```diff
@@ -42,51 +42,46 @@ class UpdateDistrict(ICommand):
...
-            },
-            ["input", "areas", "sets", district_id],
-        )
+        district = study_data.get_district(self.id)
+
+        updated_district = update_district(district, self.parameters)
+
+        study_data.save_district(updated_district)
 
         return command_succeeded(message=self.id)
```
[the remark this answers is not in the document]
smailio: I don't understand this, if someone does pelase enlighten me :)

webapp/src/components/App/Singlestudy/explore/Modelization/Map/MapConfig/Districts/UpdateDistrictDialog.tsx
```diff
@@ -65,14 +65,12 @@ function UpdateDistrictDialog(props: Props) {
...
   const handleSubmit = async (data: SubmitHandlerPlus<typeof defaultValues>) => {
     const { districtId, output, comments } = data.values;
-    const districtAreas = districtsById[districtId].areas;
     dispatch(
       updateStudyMapDistrict({
         studyId: study.id,
         districtId,
         output,
         comments,
-        areas: districtAreas,
```
smailio, line 75: This was a "fix" that I reverted. 

There was a bug : when someone update the comment of a district (and/or the output) it would erase all the areas. 

To fix this, I updated the dialog so that when someone updates a comment (and/or output), the front would send along the list of areas in the district. 

The issue with this fix is that it does send the area, but it doesn't send the apply filter. Before this PR, when the backend received areas but no apply filter was specified, it would use the default apply filter, which is "remove-all". Therefore, merely updating a comment could change the apply filter of the district. 

Now, the fix is on the backend side. When I don't receive areas, I don't erase areas (in the backend, I consider that a user didn't send areas, and sending empty areas are two different things).

webapp/src/components/App/Singlestudy/explore/Modelization/Map/MapConfig/Districts/index.tsx
```diff
@@ -85,6 +85,9 @@ function Districts() {
           output: districtsById[districtId].output,
           comments: districtsById[districtId].comments,
           areas: areasByDistrict[districtId],
+          // remove-all is hard coded here
+          // The front always send updates with apply filter set to remove-all
+          applyFilter: "remove-all",
```
hdinia, line 90: `applyFilter` is never used by the front, why not simply remove it ? And make it default as `"remove-all"` on the api if this is needed for some reason

webapp/src/types/types.ts
```diff
@@ -358,12 +358,12 @@ export interface Area {
 }
 
 export interface Set {
+  id: string;
```
hdinia: you can add `comments?: string`

webapp/src/components/App/Singlestudy/explore/Modelization/Map/MapConfig/Districts/index.tsx
smailio: It's always "remove-all" only for the front. Other users can send a command set to "add-all".

hdinia: OK makes sense if this api is not dedicated to the frontend only

smailio: Here's a case where it won't work. We've already discussed this on Teams, but I'll write it here for future readers. 

Let's say we have a district with apply-filter set to "add-all". 

Let's say by default, when no apply filter is provided, the backend defaults to "remove-all". 

Case where we want to update only a comment, then we would erase the existing "add-all" and we'd write "remove-all", which is not what we wanted; we only wanted to update the comment. 

You may say "ok, but then just default to "remove-all" except it the filter was already set to "add-all". 

Case the front wants to update the areas this time. The issue is that the front always assumes the apply filter of the district is "remove-all". When the front sends areas, it always means "no areas in this district except those in this list I sent you". So if a district was set to "add-all" we wouldn't overwrite it, and it wouldn't work in this case. 

The only solution I see is for the front to be explicit in its request, so the front sends areas along with the apply-filter.

hdinia (dismissed):
Good for the frontend part

antarest/study/business/model/district_model.py
sylvlecl: Just ignore my comment, I've been wrong, the correct value is indeed "remove-all".

antarest/study/dao/file/file_study_district_dao.py
```diff
@@ -0,0 +1,147 @@
...
+DISTRICTS_PATH = ["input", "areas", "sets"]
+
+
+class FileStudyDistrictDao(DistrictDao):
+    @abstractmethod
+    def get_file_study(self) -> FileStudy:
+        pass
+
+    @override
+    def get_districts(self) -> Sequence[DistrictDTO]:
```
sylvlecl: The transformation of `District` to `DistrictDTO` is not the responsibility of the DAO, it must be done at the business level (manager).

The DAO must only implement simple CRUD operations (actually get, save, delete) for what we store, which is the `District`, not the `DistrictDTO`.

The `DistrictDTO` class must not appear at all in that class.

When we implement a new DAO, we must not need to re-implement the `District` -> `DistrictDTO` transformation, which is business logic, not data storage related logic.

antarest/study/business/model/district_model.py
```diff
@@ -0,0 +1,216 @@
...
+    output: Optional[bool] = True
+    #: User-defined comments.
+    comments: Optional[str] = ""
+    #: List of areas that will be grouped in the district.
+    areas: Optional[List[str]] = []
+    #: Base filter for the district.
+    apply_filter: Optional[DistrictApplyFilter] = None
+
+
+class DistrictDTO(AntaresBaseModel):
```
sylvlecl, line 100: It does not seem an appropriate name to me:
the difference between `District` and `DistrictDTO` is not a technical difference, it's a business difference:
- `District` is the definition or specification of the district, which contain some build logic (include or exclude areas)
- `DistrictDTO` is the district "instantiated" on the study based on that definition, with an actual list of areas that compose the district

Could be `DistrictAreas` or `InstantiatedDistrict` or `MaterializedDistrict` ?

antarest/study/business/model/district_model.py
```diff
@@ -0,0 +1,216 @@
...
+    def validate_areas(cls, areas: Any) -> Optional[List[str]]:
+        if areas is None:
+            return areas
+        return list(set(areas))
+
+    #: Name of the district (this name is also used as a unique identifier).
+    name: str
+    #: Indicates whether this district is used in the output (usually all
+    #: districts are visible, but the user can decide to hide some of them).
+    output: Optional[bool] = True
```
sylvlecl: Default values are already defined in the `District` class, we should not duplicate them here (otherwise if one day we change on, we need to think also to change it here).
We should use `None` everywhere instead.

webapp/src/components/App/Singlestudy/explore/Modelization/Map/MapConfig/Districts/index.tsx
sylvlecl, line 90: Note that this changes the behaviour when you update an area:
for example, for the district "all areas", if you remove one area from the district, then add new areas in the study, they will not be part of the district anymore.

We'll not clarify this in that PR though

sylvlecl (changes requested):
Still reviewing, a few comments about the separation of concerns between DAO and higher layers:
the DAO must only expose data that we want to store --> here it means only `District` class, not the other ones

antarest/study/business/model/district_model.py
```diff
@@ -0,0 +1,216 @@
...
+    output: bool = True
+    #: User-defined comments.
+    comments: str = ""
+    #: List of areas that will be grouped in the district.
+    areas: List[str] = []
+    #: Name of the district (this name is also used as a unique identifier).
+    name: str
+
+
+class District(AntaresBaseModel):
```
sylvlecl, line 129: We should have some consistency validation about that data:
we cannot have bother `add_areas` and `subtract_areas`, and their presence also depends on the `apply-filter` field.

antarest/study/business/model/district_model.py
```diff
@@ -0,0 +1,216 @@
...
+            return areas
+        return list(set(areas))
+
+    #: Indicates whether this district is used in the output (usually all
+    #: districts are visible, but the user can decide to hide some of them).
+    output: Optional[bool] = None
+    #: User-defined comments.
+    comments: Optional[str] = None
+    #: List of areas that will be grouped in the district.
+    areas: Optional[List[str]] = None
```
sylvlecl, line 57: It seems a little weird to me that we don't have the same approach in the `District` class and in the update and creation classes:
in the first one, we have both "add" and "subtract", while in the other 2 we only have areas which are interpreted depending on the filter.

We should probably harmonize this (probably with just one list to avoid consistency issues ?).

Maybe the best model for the areas + filter would be to use a [discriminated union](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions).
However I don't want to make this PR overly complicated, it's not a part of the API which is currently important.

webapp/src/types/types.ts
```diff
@@ -357,13 +357,14 @@ export interface Area {
   filters_year: string[];
 }
 
-export interface Set {
+export interface District {
```
sylvlecl, line 360: we'd better have only what's useful here, which seems to be id and name

antarest/study/storage/rawstudy/model/filesystem/config/model.py
```diff
@@ -174,8 +155,8 @@ def __init__(
         study_id: str,
         version: StudyVersion,
         output_path: Optional[Path] = None,
+        sets: Optional[Dict[str, District]] = None,
```
sylvlecl: We'd better have only what's useful here, which seems to be id name and output: we can have a `DistrictConfig` class for that

sylvlecl (commented):
few more comments

antarest/study/business/model/district_model.py
smailio: For me, both classes represent the same concept a district. They both hold some parameters, and a set of areas. The difference is that District also contains the mechanics of how areas are constructed ( add_areas, subtract_areas, to save storage space, apply_filter to decide if future areas should be included automatically).

DistrictDTO strips away those internal mechanics and exposes only the data that the consumer needs. For me that's the role of a DTO. Also I'd rather stretch our DTO convention rather that introduce a new convention like MaterializedXXX.

webapp/src/components/App/Singlestudy/explore/Modelization/Map/MapConfig/Districts/index.tsx
smailio: It's true but this was already the case.

antarest/study/business/model/district_model.py
smailio: It is weird, for me it was very confusing but as this was a DAO refactoring PR I tried to introduce the DAO whil sticking as much as possible to the original design. 

For now in update when apply_filter is set to add_all, then DistrictUpdate.areas means "I want these areas in District.subtract_areas" otherwise it means "I want these areas in District.add_areas". I just added a comment so it's a bit les confusing. 

I’d be open to harmonizing this logic. Based on my own experience of how confusing this was to work with, I think it would significantly improve readability and maintainability.

webapp/src/types/types.ts
smailio: Front team wants this interface to match what's on the backend. Also this code will probably be removed soon as it's only used for the export dialog which is currently broken as we discussed wednesday.

sylvlecl (approved):

PR #2691, opened 2025-08-28
fix(scenario): fix scenario builder API regressions

sylvlecl:
Fixes 3 regressions of https://github.com/AntaresSimulatorTeam/AntaREST/pull/2679 : 

- MC year number must start at 0
- fixes a KeyError for studies <= 9.2 which have short term storages
- hydro levels must not be rounded as integers, float values are allowed

Also fixes existing issues in additional constraint management:
- the study config was not correctly updated when creating / removing storages
- constraint files must be removed when a storage is removed

### DIFF
--- a/antarest/study/business/model/sts_model.py
+++ b/antarest/study/business/model/sts_model.py
@@ -292,7 +292,8 @@ class STStorageAdditionalConstraint(AntaresBaseModel):
     @model_validator(mode="before")
     @classmethod
     def set_id(cls, data: dict[str, Any]) -> dict[str, Any]:
-        data["id"] = transform_name_to_id(data["name"])
+        if isinstance(data, dict) and "id" not in data and "name" in data:
+            data["id"] = transform_name_to_id(data["name"])
         return data
 
     id: str
--- a/antarest/study/dao/file/file_study_st_storage_dao.py
+++ b/antarest/study/dao/file/file_study_st_storage_dao.py
@@ -261,6 +261,9 @@ class FileStudySTStorageDao(STStorageDao, ABC):
             ["input", "st-storage", "clusters", area_id, "list", storage_id],
             ["input", "st-storage", "series", area_id, storage_id],
         ]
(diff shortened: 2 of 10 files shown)


antarest/study/business/model/sts_model.py
```diff
@@ -292,7 +292,8 @@ class STStorageAdditionalConstraint(AntaresBaseModel):
     @model_validator(mode="before")
     @classmethod
     def set_id(cls, data: dict[str, Any]) -> dict[str, Any]:
-        data["id"] = transform_name_to_id(data["name"])
+        if isinstance(data, dict) and "id" not in data and "name" in data:
```
MartinBelthle, line 295: I don't understand why you made that change

antarest/study/dao/file/file_study_st_storage_dao.py
```diff
@@ -261,6 +261,9 @@ def delete_st_storage(self, area_id: str, storage: STStorage) -> None:
             ["input", "st-storage", "clusters", area_id, "list", storage_id],
             ["input", "st-storage", "series", area_id, storage_id],
         ]
+        if study_data.config.version >= STUDY_VERSION_9_2:
```
MartinBelthle, line 264: Arf I forgot this, nice

MartinBelthle (approved):

antarest/study/business/model/sts_model.py
sylvlecl: model validation is not necessarily performed on a dictionary, anything may be passed to model_validate:
- it could be an invalid set of data (anything like tuple, string, ...)
- it could be an already validated STStorageAdditionalCOnstraint object (for example when we build the config DTO from the config)

So it's better to do this processing only when input data is actually a dict.

Actually if you look at th change in test_config_files, we ended up with a wrong ID in that test.
That's why I made this change. However now when I remove it, the test still passes, so I guess I fixed something else that needed to be fixed which now makes it less necessary.
But still think it's better to have it for the reasons above.

PR #2692, opened 2025-08-28
refactor(user): introduce dao for user commands

MartinBelthle:
This PR also allows users to create a file and giving its content. This way we avoid using the command `UpdateRawFile`


antarest/study/business/model/user_model.py
```diff
@@ -0,0 +1,39 @@
...
+
+from antarest.core.serde import AntaresBaseModel
+
+
+class ResourceType(StrEnum):
+    FILE = "file"
+    FOLDER = "folder"
+
+
+class CreateUserResourceData(AntaresBaseModel):
```
TheoPascoli: Can you write the verb at the end ? It looks like a method like

TheoPascoli (changes requested):
Looks good

TheoPascoli (approved):

PR #2693, opened 2025-08-29
refactor(ui-common):  improve `SplitView` prop naming and fix decimal precision in localStorage

hdinia:
## Issues

**Prop Naming Issue**: The `id` prop name was misleading and could trigger linter warnings when conflicting with the standard HTML `id` attribute. React components should avoid prop names that shadow DOM properties.

**Precision Issue**: When users drag the splitter, `react-split` calculates percentages with high floating-point precision (e.g., `33.33333333333333`). These values were stored directly in localStorage, creating messy entries and potential serialization issues.

**Inconsistent Defaults**: No standardized default split ratio, leading to inconsistent UX across different instances.

## Changes
- **Renamed `id` → `splitId`**: 
  - Follows React naming conventions (component-specific props)
  - Avoids HTML attribute conflicts and linter warnings
  - More descriptive of its actual purpose

- **Round stored sizes to whole numbers**:
  - Converts `[33.33333333333333, 66.7777777777777]` → `[33,66]`
  - Cleaner localStorage entries
  - Maintains sufficient precision for UI layout

- **Standardized default to `[10, 90]`**:
  - Common sidebar/main content pattern
  - Consistent UX across the application
  - Avoid prop repetitions

### DIFF
--- a/webapp/src/components/App/Singlestudy/explore/Debug/index.tsx
+++ b/webapp/src/components/App/Singlestudy/explore/Debug/index.tsx
@@ -112,7 +112,7 @@ function Debug() {
             response={treeDataResponse}
             ifPending={() =>
               Array.from({ length: 3 }).map((_, index) => (
-                <Skeleton key={buildKey(index)} height={32} />
+                <Skeleton key={buildKey("skeleton", index)} height={32} />
               ))
             }
             ifFulfilled={(data) => (
--- a/webapp/src/components/App/Singlestudy/explore/Debug/index.tsx
+++ b/webapp/src/components/App/Singlestudy/explore/Debug/index.tsx
@@ -104,7 +104,7 @@ function Debug() {
 
   return (
     <DebugContext.Provider value={contextValue}>
-      <SplitView id="debug" sizes={[20, 80]} minSize={150}>
+      <SplitView storageId="debug" sizes={[20, 80]} minSize={150}>
         <Box sx={{ p: 1, overflow: "auto", position: "relative" }}>
           <UsePromiseCond
             keepLastResolvedOnReload
--- a/webapp/src/components/App/Singlestudy/explore/Debug/index.tsx
+++ b/webapp/src/components/App/Singlestudy/explore/Debug/index.tsx
@@ -18,6 +18,7 @@ import { useMemo, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useOutletContext, useSearchParams } from "react-router-dom";
 import { useUpdateEffect } from "react-use";
+import { buildKey } from "@/utils/reactUtils";
 import usePromiseWithSnackbarError from "../../../../../hooks/usePromiseWithSnackbarError";
 import type { StudyMetadata } from "../../../../../types/types";
 import SplitView from "../../../../common/SplitView";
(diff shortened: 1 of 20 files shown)


webapp/src/components/App/Singlestudy/explore/Debug/index.tsx
```diff
@@ -104,13 +105,15 @@ function Debug() {
...
-      <SplitView id="debug" sizes={[20, 80]} minSize={150}>
+      <SplitView splitId="debug">
         <Box sx={{ p: 1, overflow: "auto", position: "relative" }}>
           <UsePromiseCond
             keepLastResolvedOnReload
             response={treeDataResponse}
             ifPending={() =>
-              Array.from({ length: 3 }).map((_, index) => <Skeleton key={index} height={32} />)
+              Array.from({ length: 3 }).map((_, index) => (
+                <Skeleton key={buildKey(index)} height={32} />
```
skamril: `key={buildKey(index)}` is the same as `key={index}`. However, there is no issue to use `index` as key for Skeleton.

webapp/src/components/common/SplitView/index.tsx
```diff
@@ -87,7 +87,11 @@ function SplitView({
       direction={direction}
       sizes={activeSizes ?? defaultSizes}
       minSize={minSize}
-      onDragEnd={setActiveSizes} // Update sizes on drag end.
+      onDragEnd={(sizes) => {
+        // Round to whole numbers to avoid excessive precision
+        const roundedSizes = sizes.map((size) => Math.round(size));
```
skamril: Better to round in the useEffect: `storage.setItem(localStorageKey, activeSizes.map(Math.round));`.
This way, we avoid altering the state itself, keep rounding logic isolated to persistence only, and prevent the very small visual change in the view caused by rounding.

webapp/src/components/common/SplitView/index.tsx
```diff
@@ -55,16 +55,16 @@ function isValidSizes(sizes: unknown): sizes is number[] {
  * @returns A React component displaying a split layout view with resizable panes.
  */
 function SplitView({
-  id,
+  splitId,
   children,
   direction = "horizontal",
-  sizes,
+  sizes = [10, 90], // Most common split ratio: small sidebar (10%) + main content area (90%)
```
skamril, line 61: It's small for the left panel of certain views. You need to test with a 15" size.

skamril (changes requested):

hdinia:
The direct index usage triggers linter error and should not be used even for a trivial case such as Skeleton.

webapp/src/components/App/Singlestudy/explore/Debug/index.tsx
hdinia: The direct index usage triggers linter error and should not be used even for a trivial case such as Skeleton.

webapp/src/components/common/SplitView/index.tsx
hdinia: Done.

webapp/src/components/common/SplitView/index.tsx
hdinia: I tested it and it's more than enough on laptop screens. It's resizable so users can adjust as needed.

webapp/src/components/App/Singlestudy/explore/Debug/index.tsx
skamril: `buildKey(index) == index`

Using `key={index}` can cause problems if the list order change — when an item is inserted, removed, or the list is reordered.

skamril (approved):

skamril (approved):

PR #2695, opened 2025-09-01
feat(area)!: rename some models

TheoPascoli:
**Breaking changes**:
- `type` argument in `GET /api/v1/studies/<id>/areas` is now ignored, the endpoint always returns **areas only**.
  We don't get anymore the districts (in particular "all areas"), when the argument is missing.
- `type` argument in the body of `POST /api/v1/studies/<id>/areas` is now ignored, the endpoint always creates an area
- properties names expected by the table mode endpoints are now consistent with the `GET /api/v1/studies/{uuid}/areas/{area_id}/properties/form` endpoint:
   - nonDispatchablePower -> nonDispatchPower
   - dispatchableHydroPower -> dispatchHydroPower
   - otherDispatchablePower -> otherDispatchPower
   - averageUnsuppliedEnergyCost -> energyCostUnsupplied
   - averageSpilledEnergyCost -> energyCostSpilled
   - filterYearByYear -> filterByYear
- in consequence, existing tables stored in the browser of users will break and need to be re-created


antarest/study/storage/rawstudy/model/filesystem/config/model.py
```diff
@@ -73,9 +73,9 @@ def validation(cls, values: MutableMapping[str, Any]) -> MutableMapping[str, Any
         return values
 
 
-class Area(AntaresBaseModel, extra="forbid"):
+class AreaConfig(AntaresBaseModel, extra="forbid"):
```
sylvlecl, line 76: Nice, it will be more clear this way

antarest/study/storage/rawstudy/model/filesystem/config/area.py
```diff
@@ -384,91 +153,63 @@ def to_config(self) -> Dict[str, Dict[str, Any]]:
...
-                                             'spread-unsupplied-energy-cost': 'yes'}},
-     'ui': {'style': {'color_rgb': (0, 128, 256)}}}
-
-    >>> AreaFolder.validate(data)
-    Traceback (most recent call last):
-      ...
-    pydantic.error_wrappers.ValidationError: 1 validation error for AreaFolder
-    optimization -> nodal optimization -> spread-unsupplied-energy-cost
-      value is not a valid float (type=type_error.float)
+class AdequacyPathFileData(AntaresBaseModel):
```
sylvlecl: There is (and was already there) a typo:
it should be  `AdequacyPat c hFileData`

antarest/study/storage/rawstudy/model/filesystem/config/area.py
```diff
@@ -384,91 +153,63 @@ def to_config(self) -> Dict[str, Dict[str, Any]]:
...
     """
+    Only available if study version >= 830.
+    """
+
+    model_config = ConfigDict(alias_generator=to_kebab_case, populate_by_name=True, extra="forbid")
+
+    class AdequacyPathSection(AntaresBaseModel):
```
sylvlecl: Same typo: c is missing in patch

antarest/study/business/model/area_model.py
```diff
@@ -16,29 +16,26 @@
...
+    thermals: Optional[List[ThermalCluster]] = None
 
 
-class AreaInfoDTO(AreaCreationDTO):
-    id: str
-    thermals: Optional[List[ThermalCluster]] = None
+class AreaCreation(AntaresBaseModel):
+    name: str
+    type: AreaType
+    set: Optional[List[str]] = None
```
sylvlecl: Does not seem to be used, let's remove it

antarest/study/business/model/area_model.py
```diff
@@ -16,29 +16,26 @@
...
     set: Optional[List[str]] = None
+    thermals: Optional[List[ThermalCluster]] = None
 
 
-class AreaInfoDTO(AreaCreationDTO):
-    id: str
-    thermals: Optional[List[ThermalCluster]] = None
+class AreaCreation(AntaresBaseModel):
+    name: str
+    type: AreaType
```
sylvlecl: Does not seem to be used either, we can remove it.
However this will break the API, since it was required so far ... So we could have another class in the web layer that still accepts the type field, mark it as [deprecated](https://docs.pydantic.dev/latest/concepts/fields/#deprecated-fields), and turn it into our "clean" internal class.

antarest/study/business/model/area_model.py
```diff
@@ -16,29 +16,26 @@
...
 class AreaType(enum.Enum):
     AREA = "AREA"
     DISTRICT = "DISTRICT"
 
 
-class AreaCreationDTO(AntaresBaseModel):
+class Area(AntaresBaseModel):
+    id: str
     name: str
     type: AreaType
```
sylvlecl: We should try to remove it, there is already a `District` representation, this duplication does not make sense.
But we need to be careful about the impact on the API, of course.

I checked that:
- R packages don't use it
- front only uses the AREA value
- antares-craft only uses the AREA value

I think we should:
- remove it from our internal representation.
- in the GET /areas:
  - make the "type" query parameter deprecated an only return areas
  - return areas object where type is added, but with the only allowed value AREA, marked as deprecated

antarest/study/web/study_data_blueprint.py
```diff
@@ -154,9 +154,9 @@ def create_study_data_routes(study_service: StudyService, config: Config) -> API
         "/studies/{uuid}/areas",
         tags=[APITag.study_data],
         summary="Get all areas basic info",
-        response_model=List[AreaInfoDTO] | Dict[str, Any],
+        response_model=List[Area] | Dict[str, Any],
     )
-    def get_areas(uuid: str, type: AreaType = Query(None), ui: bool = False) -> List[AreaInfoDTO] | Dict[str, Any]:
+    def get_areas(uuid: str, type: AreaType = Query(None), ui: bool = False) -> List[Area] | Dict[str, Any]:
```
sylvlecl: As mentioned above, I propose to mark type as deprecated (in the `Query` annotation).

Also, I propose to change the responde model to only `list[Area***]`: in this list we need to add the `type` to avoid breaking the API, unfortunately.

antarest/study/web/study_data_blueprint.py
```diff
@@ -154,9 +154,9 @@ def create_study_data_routes(study_service: StudyService, config: Config) -> API
         "/studies/{uuid}/areas",
         tags=[APITag.study_data],
         summary="Get all areas basic info",
-        response_model=List[AreaInfoDTO] | Dict[str, Any],
+        response_model=List[Area] | Dict[str, Any],
```
sylvlecl: response model seems redundant here

antarest/study/web/study_data_blueprint.py
```diff
@@ -171,9 +171,9 @@ def get_links(uuid: str) -> List[Link]:
         "/studies/{uuid}/areas",
         tags=[APITag.study_data],
         summary="Create a new area",
-        response_model=AreaInfoDTO,
+        response_model=Area,
     )
-    def create_area(uuid: str, area_creation_info: AreaCreationDTO) -> Any:
+    def create_area(uuid: str, area_creation_info: AreaCreation) -> Any:
```
sylvlecl: should return Area

antarest/study/web/study_data_blueprint.py
```diff
@@ -171,9 +171,9 @@ def get_links(uuid: str) -> List[Link]:
         "/studies/{uuid}/areas",
         tags=[APITag.study_data],
         summary="Create a new area",
-        response_model=AreaInfoDTO,
+        response_model=Area,
```
sylvlecl: Will be redundant if we change the return type

antarest/study/web/study_data_blueprint.py
```diff
@@ -1102,7 +1102,7 @@ def get_allocation_matrix(uuid: str) -> AllocationMatrix:
         """
         study = study_service.check_study_access(uuid, StudyPermissionType.READ)
         all_areas = cast(
-            List[AreaInfoDTO],  # because `ui=False`
+            List[Area],
             study_service.get_all_areas(uuid, area_type=AreaType.AREA, ui=False),
```
sylvlecl: Let's take this opportunity to clean up the typing:
we should split the `get_all_areas` function in 2 depending on whether we ask for UI info or not. Then we can have a proper return type in those 2 methods.

Another possibility is to add `@overload` signatures to the method (see method _parse_ruleset for an example), but it does not seem that relevant here, the code is really an if/else in the implementation of the method, clearly we should just have 2 methods.

antarest/study/web/study_data_blueprint.py
```diff
@@ -1319,7 +1319,7 @@ def set_correlation_form_fields(
         """
         study = study_service.check_study_access(uuid, StudyPermissionType.WRITE)
         all_areas = cast(
```
sylvlecl: All those casts need to be removed once we have cleaned up the typing of `get_all_areas`

antarest/study/service.py
```diff
@@ -1645,7 +1645,7 @@ def get_all_areas(
         uuid: str,
         area_type: Optional[AreaType],
         ui: bool,
-    ) -> List[AreaInfoDTO] | Dict[str, Any]:
+    ) -> List[Area] | Dict[str, Any]:
```
sylvlecl: Here we should separate the methods in 2 to have proper typing I think, as proposed in previous comment

antarest/study/business/model/area_properties_model.py
```diff
@@ -114,22 +115,25 @@ def validation(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+                values[filter_name] = options
+                if filter_name == "filterByYear":
+                    values.setdefault("filterYearByYear", options)
+                    values.pop("filterByYear", None)
         return values
 
 
 @all_optional_model
-class AreaFileData(AntaresBaseModel, extra="forbid", populate_by_name=True):
+class AreaPropertiesFileData(AntaresBaseModel, extra="forbid", populate_by_name=True):
```
sylvlecl: Actually we should move that out from the business model modeul, it should be moved to the filesystem representation module, together with other "file data" classes.
Also we should remove the annotation to explicitly make fields optionals, if relevant.

antarest/study/business/model/area_properties_model.py
```diff
@@ -19,8 +19,8 @@
 from antarest.study.business.all_optional_meta import all_optional_model
 from antarest.study.storage.rawstudy.model.filesystem.config.area import (
     AdequacyPatchMode,
-    AdequacyPathProperties,
-    OptimizationProperties,
+    AdequacyPathFileData,
```
sylvlecl: There should be no dependency in that direction (model to "file data"):
only the filesystem layer should know about the models, to create them. Should be fixed by moving the file data class out from this module.

antarest/study/business/table_mode_management.py
```diff
@@ -109,7 +110,15 @@ def __init__(
...
         if table_type == TableModeType.AREA:
             areas_map = self._area_manager.get_all_area_props(study)
-            data = {area_id: area.model_dump(mode="json", by_alias=True) for area_id, area in areas_map.items()}
+            data = {}
+            for area_id, area in areas_map.items():
+                area_dict = area.model_dump(by_alias=True)
+                area_dict["filterSynthesis"] = decode_filter(area.filter_synthesis)
+                area_dict["filterByYear"] = decode_filter(area.filter_by_year)
+                data[area_id] = area_dict
+            if study.version < STUDY_VERSION_8_3:
```
sylvlecl: Do we really need this here, shouldn't it be already absent from the `Area` object up front? maybe just with an `exclude_none` above ? but maybe that's the problem, this will exclude too much stuff ?

antarest/study/business/table_mode_management.py
```diff
@@ -208,10 +217,18 @@ def update_table_data(
...
+            area_props_by_ids = {key: AreaPropertiesUpdate(**values) for key, values in data.items()}
             areas_map = self._area_manager.update_areas_props(study, area_props_by_ids)
-            data = {area_id: area.model_dump(by_alias=True, exclude_none=True) for area_id, area in areas_map.items()}
+            data = {}
+            for area_id, area in areas_map.items():
+                area_dict = area.model_dump(by_alias=True, exclude_none=True)
+                area_dict["filterSynthesis"] = decode_filter(area.filter_synthesis)
+                area_dict["filterByYear"] = decode_filter(area.filter_by_year)
+                data[area_id] = area_dict
+            if study.version < STUDY_VERSION_8_3:
```
sylvlecl: Same question here, is it really needed, in particular with the exclude_none above ?

antarest/study/business/model/area_model.py
sylvlecl: An alternative maybe easier, is to keep it here, allow only the AREA value, and mark it as deprecated

sylvlecl (commented):
Nice, I added some suggestions to try to push the clean up a little further, in particular the handling of the `type` attribute in `Area` which does not make sense anymore in my opinion.

It's also possible to postpone that to a future work, though.

antarest/study/business/model/area_properties_model.py
TheoPascoli: I moved it. However, no field are optionnals so I removed the annotation completely

antarest/study/web/study_data_blueprint.py
TheoPascoli: I marked it as deprecated and changed the response model

antarest/study/business/table_mode_management.py
TheoPascoli: get_area_properties always set adequacy_patch_mode, even for <8.3 studies, so it is not droped. I can remove this logic by passing the version number to `get_all_area_props` and similar paths and make the model conditional which is a larger refacto. So for now the excplicit pop is still needed in both read and update to keep <8.3 studies clean.

sylvlecl: I see, that's an issue that it's present in the result of `get_all_area_props`, we should have the same handling as in `ThermalCluster` etc: new props should be `Optional`, and `None` for older versions.

If not in that PR, we need to do it in the next one for consistency sake.

antarest/study/business/area_management.py
```diff
@@ -152,47 +152,29 @@ def update_areas_props(
...
+    def get_all_areas(self, study: StudyInterface, area_type: Optional[AreaType] = None) -> List[Area]:
+        """Retrieve all physical areas of a raw study.
 
-        Args:
-            study: The raw study object.
-            area_type: The type of area. Retrieves areas and districts if `None`.
-
-        Returns:
-            A list of area/district information.
+        The optional ``area_type`` argument is kept for backward compatibility but
```
sylvlecl: If feasible, I would propose to completely remove that argument from our internal methods, so that it really only remains for backwards compatibility at the API level.

antarest/study/web/study_data_blueprint.py
```diff
@@ -149,17 +149,30 @@ def create_study_data_routes(study_service: StudyService, config: Config) -> API
...
+    def get_areas(
+        uuid: str,
+        type: Optional[AreaType] = Query(default=None, deprecated=True),
+        ui: bool = Query(default=False),
+    ) -> List[AreaResponse] | Dict[str, Any]:
+        logger.info(f"Fetching area list (type={type}, ui={ui}) for study {uuid}")
+        if ui:
+            return study_service.get_all_areas_ui_info(uuid)
+
+        areas = study_service.get_all_areas(uuid, type)
```
sylvlecl: I think we should not even pass the type parameter

sylvlecl (commented):
nice, last comment about area type, I think we can completely remove it from internal methods arguments

MartinBelthle (commented):
You should add the `!` in your PR title and explain what breaking change you did.
It will help me change the `antares-craft` code :)

antarest/study/business/table_mode_management.py
sylvlecl: I chanegd the model so that it's `None` as soon as we read it, on versions older than 8.3.

antarest/study/business/area_management.py
sylvlecl: Done

antarest/study/web/study_data_blueprint.py
sylvlecl: Done

sylvlecl (approved):
Great cleanup !

PR #2696, opened 2025-09-02
feat(ui): use custom dialog instead of window.confirm for unsaved/ongoing submit warning

skamril:
add `useFormCloseConfirm` to show a confirmation dialog when leaving a view with unsaved changes or during submit, even outside react-router


hdinia (approved):

hdinia (changes requested):
Do you have any particular reason to use Redux here ? For this use case it seems not relevant 

Form confirmation dialogs are:

- Ephemeral: exist only during the user’s interaction
- Component scoped: relevant only to the specific form instance
- Non shareable: no need for global access

Redux can pollute global state, clutter the store with temporary UI details, add unnecessary complexity, why not using a simple hook with local state ?

webapp/src/hooks/useFormCloseConfirm.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { getFormState } from "@/redux/selectors";
+import { useCallback, useEffect } from "react";
+import useConfirm from "./useConfirm";
+
+function useFormCloseConfirm() {
+  const { showConfirm, isPending, yes, no } = useConfirm();
+  const { status, closeDialogStatus } = useAppSelector(getFormState);
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
```
smailio, line 44: maybe useeffect isn't necessary here ?

webapp/src/hooks/useFormCloseProtection.ts
```diff
@@ -0,0 +1,47 @@
...
+import useBlocker from "./useBlocker";
+import useFormCloseConfirm from "./useFormCloseConfirm";
+
+export interface UseFormCloseProtectionParams {
+  isSubmitting: boolean;
+  isDirty: boolean;
+}
+
+function useFormCloseProtection({ isSubmitting, isDirty }: UseFormCloseProtectionParams) {
+  const dispatch = useAppDispatch();
```
smailio, line 42: add doc

webapp/src/hooks/useFormCloseConfirm.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import { setFormCloseDialogStatus } from "@/redux/ducks/ui";
+import useAppDispatch from "@/redux/hooks/useAppDispatch";
+import useAppSelector from "@/redux/hooks/useAppSelector";
+import { getFormState } from "@/redux/selectors";
+import { useCallback, useEffect } from "react";
+import useConfirm from "./useConfirm";
+
+function useFormCloseConfirm() {
+  const { showConfirm, isPending, yes, no } = useConfirm();
```
smailio, line 40: add doc

hdinia (approved):

PR #2697, opened 2025-09-03
feat(ui-matrix): add separate enable flags for `Matrix` filters and resize

hdinia:
### Changes Made:

  1. Added two new optional props to Matrix component:
    - `enableFilters?`: boolean - Controls whether the filter feature is available
    - `enableResize?`: boolean - Controls whether the resize feature is available
    
  2. Set defaults:
    - Both props default to isTimeSeries value, maintaining current behavior
    - This means existing code won't break and timeseries matrices get both features by default
    
  3. Updated `MatrixActions` component:
    - Added the new props to the interface
    - Changed the conditional rendering logic to use separate enable flags
    - Each feature can now be enabled/disabled independently

  Usage Examples:

```tsx
  // Current behavior (unchanged) - both features enabled for timeseries
  <Matrix url="/path" isTimeSeries />

  // With filters but no resize 
  <Matrix url="/path" isTimeSeries={false} enableFilters />

  // With resize but no filters 
  <Matrix url="/path" isTimeSeries={false} enableResize />

  // Disable both features even for timeseries
  <Matrix url="/path" isTimeSeries enableFilters={false} enableResize={false} />
  ```

### DIFF
--- a/webapp/src/components/common/Matrix/index.tsx
+++ b/webapp/src/components/common/Matrix/index.tsx
@@ -12,12 +12,12 @@
  * This file is part of the Antares project.
  */
 
-import MatrixUpload from "@/components/common/Matrix/components/MatrixUpload";
 import GridOffIcon from "@mui/icons-material/GridOff";
 import { Box, Skeleton, Tooltip } from "@mui/material";
 import { useMemo, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useOutletContext } from "react-router";
+import MatrixUpload from "@/components/common/Matrix/components/MatrixUpload";
 import type { StudyMetadata } from "../../../types/types";
 import type { fetchMatrixFn } from "../../App/Singlestudy/explore/Modelization/Areas/Hydro/utils";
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: webapp/src/components/common/Matrix/index.tsx [hunk 1]
```diff
@@ -37,6 +37,10 @@ interface MatrixProps {
   title?: string;
   customRowHeaders?: string[];
   dateTimeColumn?: boolean;
+  /**
+   * Enables time series features like temporal filters, resize, and calculated columns.
+   * Note: For specific cases (e.g., hourly time series that shouldn't resize), use enableFilters/enableResize to override.
+   */
   isTimeSeries?: boolean;
```
skamril, line 44: Tip:

```js
/**
 * Shortcut prop equivalent to enabling filters, resize, 
 * and calculated columns all at once.
 *
 * Setting `isTimeSeries` is the same as using:
 *   enableFilters = true
 *   enableResize = true
 *   enableCalculations = true
 */
isTimeSeries?: boolean;
```

skamril (approved):

PR #2700, opened 2025-09-05
feat(editor): updating editor name when modifying file studies

mehdiwahada:
I have implemented a method that is updating the editor's name when the user is modifying studies. The method is used inside the add_commands from RawStudyInterface.

It is a continuity of a task where the editor field was changed when a study is imported or created.

You can test it by creating a study and modify it (you can add / update comments, add areas, update their metadatas and delete them -> example in test_integration.py)

### DIFF
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -427,9 +427,8 @@ class RawStudyInterface(StudyInterface):
         self._update_editor(file_study)
 
     def _update_editor(self, file_study: FileStudy) -> None:
-        user_id = get_user_impersonator()
-        user = self._user_service.get_identity(user_id)
-        if user and user.name:
+        user = self._user_service.get_identity(get_user_impersonator())
+        if user:
             user_name = user.name or ""
             study_antares = file_study.tree.get(["study", "antares"])
             study_antares["editor"] = user.name
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -372,10 +372,12 @@ class RawStudyInterface(StudyInterface):
         self,
         raw_service: RawStudyService,
         variant_service: VariantStudyService,
+        service: "StudyService",
         study: RawStudy,
     ):
         self._raw_study_service = raw_service
         self._variant_study_service = variant_service
+        self._study_service = service
         self._study = study
         self._cached_file_study: Optional[FileStudy] = None
         self._version = StudyVersion.parse(self._study.version)
@@ -405,7 +407,7 @@ class RawStudyInterface(StudyInterface):
         study = self._study
         should_invalidate_cache = False
         file_study = self.get_files()
-        update_editor_name(file_study, get_user_id())
+        update_editor_name(file_study, self._study_service.get_user_name())
         for command in commands:
             result = command.apply(FileStudyTreeDao(file_study), listener)
             if result.should_invalidate_cache:
@@ -431,8 +433,9 @@ class VariantStudyInterface(StudyInterface):
     to the variant.
     """
 
-    def __init__(self, variant_service: VariantStudyService, study: VariantStudy):
+    def __init__(self, variant_service: VariantStudyService, service: "StudyService", study: VariantStudy):
         self._variant_service = variant_service
+        self.study_service = service
         self._study = study
         self._version = StudyVersion.parse(self._study.version)
 
@@ -458,7 +461,7 @@ class VariantStudyInterface(StudyInterface):
     def add_commands(self, commands: Sequence[ICommand], listener: Optional[ICommandListener] = None) -> None:
         # get current user if not in session, otherwise get session user
         file_study = self.get_files()
-        update_editor_name(file_study, get_user_id())
+        update_editor_name(file_study, self.study_service.get_user_name())
         self._variant_service.append_commands(self._study.id, transform_command_to_dto(commands, force_aggregate=True))
 
 
@@ -808,12 +811,14 @@ class StudyService:
         if isinstance(study, VariantStudy):
             return VariantStudyInterface(
                 self.storage_service.variant_study_service,
+                self,
                 study,
             )
         elif isinstance(study, RawStudy):
             return RawStudyInterface(
                 self.storage_service.raw_study_service,
                 self.storage_service.variant_study_service,
+                self,
                 study,
             )
         else:
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -429,11 +429,13 @@ class RawStudyInterface(StudyInterface):
             user = self._user_service.get_identity(jwt_user.id)
             if user and user.name:
                 study_antares = file_study.tree.get(["study", "antares"])
-                study_db = self._repository.get(self._study.id)
                 study_antares["editor"] = user.name
-                study_db.additional_data.editor = user.name
                 file_study.tree.save(study_antares, ["study", "antares"])
-                self._repository.save(study_db)
+                study_db = self._repository.get(self._study.id)
+
+                if study_db:
+                    study_db.additional_data.editor = user.name
+                    self._repository.save(study_db)
 
 
 class VariantStudyInterface(StudyInterface):
--- a/antarest/study/service.py
+++ b/antarest/study/service.py
@@ -424,21 +424,21 @@ class RawStudyInterface(StudyInterface):
             data = FileStudyTreeConfigDTO.from_build_config(file_study.config).model_dump()
             update_cache(self._raw_study_service.cache, study.id, data)
         self._variant_study_service.on_parent_change(study.id)
-        self.update_editor(file_study)
+        self._update_editor(file_study)
 
-    def update_editor(self, file_study: FileStudy) -> None:
-        jwt_user = get_user_impersonator()
-        if jwt_user:
-            user = self._user_service.get_identity(jwt_user)
-            if user and user.name:
-                study_antares = file_study.tree.get(["study", "antares"])
-                study_antares["editor"] = user.name
-                file_study.tree.save(study_antares, ["study", "antares"])
-                study_db = self._repository.get(self._study.id)
-
-                if study_db and study_db.additional_data:
-                    study_db.additional_data.editor = user.name
-                    self._repository.save(study_db)
+    def _update_editor(self, file_study: FileStudy) -> None:
+        user_id = get_user_impersonator()
+        user = self._user_service.get_identity(user_id)
+        if user and user.name:
+            user_name = user.name or ""
+            study_antares = file_study.tree.get(["study", "antares"])
+            study_antares["editor"] = user.name
+            file_study.tree.save(study_antares, ["study", "antares"])
+            if not self._study.additional_data:
+                self._study.additional_data = StudyAdditionalData(author=user_name, editor=user_name)
+            else:
+                self._study.additional_data.editor = user_name
+            self._repository.save(self._study)
 
 
 class VariantStudyInterface(StudyInterface):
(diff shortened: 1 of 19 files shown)


antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -117,6 +117,26 @@ def __init__(
         self.generator = VariantCommandGenerator(self.study_factory)
         CommandMatrixUsageProvider(variant_study_repo=repository, command_factory=command_factory)
 
+    def _update_editor(self, study: VariantStudy) -> None:
+        user_id = get_user_id()
```
sylvlecl: you can use `require_user().id` instead, and remove the if condition below together with the cast.
We must have a logged in user when we get here.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -117,6 +117,26 @@ def __init__(
...
+
+        if user_id:
+            user_name = self._get_user_name_from_id(int(user_id))
+            if user_name:
```
sylvlecl: we don't need to check this

antarest/study/storage/variantstudy/variant_study_service.py
sylvlecl: Actually even better: we should use `get_user_impersonator()` instead.

The current user id may be the ID of a token: using the "impersonator" ensures that we get the ID of the real user, not the ID of a token which belongs to a real user.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -117,6 +117,26 @@ def __init__(
...
+
+    def _update_db_study_editor(self, study_id: str) -> None:
+        user_id = get_user_id()
+        study_db = self.repository.get(study_id)
+
+        if user_id and study_db:
+            user_name = self._get_user_name_from_id(int(user_id))
+
+            if user_name:
+                study_db.additional_data.editor = user_name
```
sylvlecl: You already did that in the other method, do we really need both ? 
Can't we just add the save in the other method ?

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -236,6 +256,8 @@ def append_commands(self, study_id: str, commands: List[CommandDTO]) -> List[str
             for i, command in enumerate(validated_commands)
         ]
         study.commands.extend(new_commands)
+        self._update_editor(study)
+        self._update_db_study_editor(study.id)
```
sylvlecl: Seems to me we can remove those duplicate method calls and only keep the `_update_editor`

antarest/study/storage/variantstudy/variant_study_service.py
sylvlecl, line 1: About the handling of editor for variants:
now we correctly update the field in database, good thing.

But, we don't update the `study.(antares` file when we generate the corresponding snapshots.
This should be done in the `update_antares_info` method and for consistency sake we should also read the editor at the start of the generation in `snapshot_generator` module, method `_read_additional_data`.


Also, it seems that in variant creation, we don't correctly set the editor name from the parent study (see line 670).

antarest/study/service.py
```diff
@@ -419,6 +424,18 @@ def add_commands(self, commands: Sequence[ICommand], listener: Optional[ICommand
             data = FileStudyTreeConfigDTO.from_build_config(file_study.config).model_dump()
             update_cache(self._raw_study_service.cache, study.id, data)
         self._variant_study_service.on_parent_change(study.id)
+        jwt_user = get_current_user()
```
sylvlecl: As for variants, we should use `get_user_impersonator` instead here.

Also, you should extract all that part in a dedicated `update_editor` method so that this method remains understandable

tests/integration/test_integration.py
```diff
@@ -1890,3 +1890,122 @@ def test_links_deletion_with_binding_constraints(
     # delete the link
     res = client.delete(f"/v1/studies/{internal_study_id}/links/area_1/area_2")
     assert res.status_code == 200, res.json()
+
+
+def test_update_with_editor(client: TestClient, admin_access_token: str):
```
sylvlecl, line 1896: We should do the same test with a variant study, you can probably do a parameterized test for this: maybe you can use the `PreparerProxy` as in other tests

tests/variantstudy/model/test_variant_model.py
```diff
@@ -312,3 +313,75 @@ def check_orm_operations(orm_execute_state):
...
+
+        with current_user_context(jwt_user):
+            variant_study = variant_study_service.create_variant_study(root_study_id, "new_variant_1")
+        study_version = StudyVersion.parse(variant_study.version)
+        saved_id = variant_study.id
+        study = variant_study_service.repository.get(saved_id)
+        assert study is not None
+        assert study.id == saved_id
+        assert study.parent_id == root_study_id
+        assert study.additional_data.author == "john.doe"
```
sylvlecl, line 351: we should have an assertion on the editor too

tests/integration/test_integration.py
```diff
@@ -1890,3 +1890,122 @@ def test_links_deletion_with_binding_constraints(
...
+    client.headers = {"Authorization": f"Bearer {admin_access_token}"}
+
+    # 1. Create a group and two users
+    group_name = "test_copy_group"
+    res = client.post("/v1/groups", json={"name": group_name})
+    group_id = res.json()["id"]
+
+    client.post("/v1/users", json={"name": "creator_2", "password": "password123"})
+    client.post("/v1/users", json={"name": "editor_2", "password": "password456"})
```
sylvlecl, line 1905: we should have a test with a "bot" user too (a token associated to a user), to check that the editor is set to the impersonator (the associated user, not the token)

sylvlecl (changes requested):
A few changes requested to:
- handle editor in places where it's missing (variant files, variant creation)
- improve code quality (remove some duplication ...)
- ensure that the editor is correctly set to the real user, even when a "bot" token is used for authentication

tests/study/test_repository.py
```diff
@@ -1250,3 +1250,31 @@ def test_count_studies__general_case(
     # test that the expected studies are returned
     if expected_ids is not None:
         assert count == len(expected_ids)
+
+@with_db_context
```
sylvlecl: It does not test anything, I will remove it

tests/variantstudy/model/test_variant_model.py
```diff
@@ -312,3 +313,76 @@ def check_orm_operations(orm_execute_state):
...
+        assert study.additional_data.editor == "Unknown"
```
sylvlecl: Here the editor should be the one who created the variant study

tests/variantstudy/model/test_variant_model.py
```diff
@@ -312,3 +313,76 @@ def check_orm_operations(orm_execute_state):
...
+
+        # creating area by the author, making him the editor of the study
+        command_1 = CommandDTO(action="create_area", args={"area_name": "area_be"}, study_version=study_version)
+        command_2 = CommandDTO(action="create_area", args={"area_name": "area_fr"}, study_version=study_version)
```
sylvlecl, line 356: to keep the test simple we can have only one command

tests/study/test_repository.py
```diff
@@ -1250,3 +1250,31 @@ def test_count_studies__general_case(
...
+def test_editor_study(
+    db_session: Session,
+) -> None:
+    test_workspace = "test-repository"
+    #icache: Mock = Mock(spec=ICache)
```
sylvlecl: comments to be removed

antarest/study/service.py
```diff
@@ -419,6 +424,21 @@ def add_commands(self, commands: Sequence[ICommand], listener: Optional[ICommand
             data = FileStudyTreeConfigDTO.from_build_config(file_study.config).model_dump()
             update_cache(self._raw_study_service.cache, study.id, data)
         self._variant_study_service.on_parent_change(study.id)
+        self.update_editor(file_study)
+
+    def update_editor(self, file_study: FileStudy) -> None:
```
sylvlecl: we should make it private

antarest/study/service.py
```diff
@@ -419,6 +424,21 @@ def add_commands(self, commands: Sequence[ICommand], listener: Optional[ICommand
...
+        jwt_user = get_user_impersonator()
+        if jwt_user:
```
sylvlecl: we should remove the check, and rename the variable, it's not anymore a JWT user

antarest/study/service.py
```diff
@@ -419,6 +424,21 @@ def add_commands(self, commands: Sequence[ICommand], listener: Optional[ICommand
...
+            user = self._user_service.get_identity(jwt_user)
+            if user and user.name:
+                study_antares = file_study.tree.get(["study", "antares"])
+                study_antares["editor"] = user.name
+                file_study.tree.save(study_antares, ["study", "antares"])
+                study_db = self._repository.get(self._study.id)
+
+                if study_db and study_db.additional_data:
```
sylvlecl: - we should raise if study_db does not exist
- if additional_data does not exist, we should initialize it

antarest/study/storage/utils.py
```diff
@@ -83,20 +83,24 @@ def update_antares_info(metadata: Study, study_tree: FileStudyTree, update_autho
     study_data_info = study_tree.get(["study"])
     antares_info = study_data_info["antares"]
     author = "Unknown"
+    editor = "Unknown"
 
     if metadata.additional_data:
         author = metadata.additional_data.author
+        editor = metadata.additional_data.editor
+        # if metadata.additional_data.editor is None:
```
sylvlecl: comment to be removed

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -344,6 +350,8 @@ def update_command(self, study_id: str, command_id: str, command: CommandDTO) ->
         if index >= 0:
             study.commands[index].command = validated_commands[0].action
             study.commands[index].args = to_json_string(validated_commands[0].args)
+            self._update_editor(study)
+            # self._update_db_study_editor(study.id)
```
sylvlecl: comments to be removed

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -117,18 +116,15 @@ def __init__(
...
-        """
-        user_obj: Identity | None = db.session.get(Identity, user_id)
-        if user_obj is None:
-            return "Unnamed"
-        return str(user_obj.name)
+    def _update_editor(self, study: VariantStudy) -> None:
+        user_id = get_user_impersonator()
+        study_db = self.repository.get(study.id)
+
+        if user_id and study_db:
```
sylvlecl: user_id is never None, we should remove the check

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -117,18 +116,15 @@ def __init__(
...
-            user_id: user id (user must exist)
-        Returns: String representing the user's name
-        """
-        user_obj: Identity | None = db.session.get(Identity, user_id)
-        if user_obj is None:
-            return "Unnamed"
-        return str(user_obj.name)
+    def _update_editor(self, study: VariantStudy) -> None:
+        user_id = get_user_impersonator()
+        study_db = self.repository.get(study.id)
```
sylvlecl: actually we already have the study, we don't need to get it again

sylvlecl (commented):
Some comments that I will handle tomorrow

sylvlecl:
I just rebased on release branch.

sylvlecl (dismissed):
Thanks!

PR #2708, opened 2025-09-16
feat(sts): add new field `allow-overflow` for v9.3+ studies

MartinBelthle:
[ANT-3753] and [ANT-3752]


smailio (approved):
Code is straightforward and tested.

PR #2710, opened 2025-09-16
feat(ui-studies): add folder navigation to individual study views

hdinia:
ANT-3177

### DIFF
--- a/webapp/src/components/App/Singlestudy/NavHeader/Actions.tsx
+++ b/webapp/src/components/App/Singlestudy/NavHeader/Actions.tsx
@@ -96,14 +96,8 @@ function Actions({ study, parentStudy, variantNb, isExplorer }: Props) {
         }}
       >
         <Box sx={{ flex: 1 }}>
-          <Breadcrumb
-            studyId={study.id}
-            studyName={study.name}
-            workspace={study.workspace}
-            folder={study.folder}
-          />
+          <Breadcrumb study={study} />
         </Box>
-        <Divider flexItem orientation="vertical" />
         <Box sx={{ overflow: "auto" }}>
           <CustomScrollbar>
             <Box
@@ -140,6 +134,8 @@ function Actions({ study, parentStudy, variantNb, isExplorer }: Props) {
             onClick={handleUnarchive}
             startIcon={<UnarchiveOutlinedIcon />}
             variant="contained"
+            size="extra-small"
+            sx={{ px: 1 }}
           >
             {t("global.unarchive")}
           </Button>
--- a/webapp/src/components/App/Singlestudy/NavHeader/Actions.tsx
+++ b/webapp/src/components/App/Singlestudy/NavHeader/Actions.tsx
@@ -12,24 +12,23 @@
  * This file is part of the Antares project.
  */
 
-import CustomScrollbar from "@/components/common/CustomScrollbar";
-import useEnqueueErrorSnackbar from "@/hooks/useEnqueueErrorSnackbar";
-import { unarchiveStudy } from "@/services/api/study";
-import { type StudyMetadata } from "@/types/types";
-import { toError } from "@/utils/fnUtils";
-import ArrowBackIcon from "@mui/icons-material/ArrowBack";
 import ContentCopyIcon from "@mui/icons-material/ContentCopy";
 import HistoryOutlinedIcon from "@mui/icons-material/HistoryOutlined";
 import MoreVertIcon from "@mui/icons-material/MoreVert";
 import UnarchiveOutlinedIcon from "@mui/icons-material/UnarchiveOutlined";
-import { Box, Button, Chip, Divider, IconButton, Tooltip, Typography } from "@mui/material";
+import { Box, Button, Chip, Divider, IconButton, Tooltip } from "@mui/material";
 import { enqueueSnackbar } from "notistack";
 import { useState } from "react";
 import { useTranslation } from "react-i18next";
-import { useNavigate } from "react-router";
+import CustomScrollbar from "@/components/common/CustomScrollbar";
+import useEnqueueErrorSnackbar from "@/hooks/useEnqueueErrorSnackbar";
+import { unarchiveStudy } from "@/services/api/study";
+import type { StudyMetadata } from "@/types/types";
+import { toError } from "@/utils/fnUtils";
 import FavoriteStudyToggle from "../../shared/studies/FavoriteStudyToggle";
 import StudyActionsMenu from "../../shared/studies/StudyActionsMenu";
 import CommandsDrawer from "../CommandsDrawer";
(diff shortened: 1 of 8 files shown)


webapp/src/components/App/Singlestudy/NavHeader/Actions.tsx
```diff
@@ -102,28 +92,27 @@ function Actions({ study, parentStudy, variantNb, isExplorer }: Props) {
...
         }}
       >
-        <Button onClick={handleClickBack} startIcon={<ArrowBackIcon />} color="secondary">
-          {isExplorer ? t("button.back") : t("global.studies")}
-        </Button>
-        <Divider flexItem orientation="vertical" />
-        <Box sx={{ flex: 1, overflow: "auto" }}>
+        <Box sx={{ flex: 1 }}>
+          <Breadcrumb
+            studyId={study.id}
```
skamril: You can give directly the study to reduce props `study={study}`

webapp/src/components/App/Singlestudy/NavHeader/Breadcrumb.tsx
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+import { Box, Breadcrumbs } from "@mui/material";
+import { useNavigate } from "react-router";
+import { updateStudyFilters } from "@/redux/ducks/studies";
+import useAppDispatch from "@/redux/hooks/useAppDispatch";
+import BreadcrumbLink from "./BreadcrumbLink";
+import { buildBreadcrumbPath } from "./utils";
+
+interface BreadcrumbProps {
```
skamril: Facultative: Props when the type is not exported

webapp/src/components/App/Singlestudy/NavHeader/Breadcrumb.tsx
```diff
@@ -0,0 +1,81 @@
...
+  studyId: string;
+  studyName: string;
+  workspace: string;
+  folder?: string;
+}
+
+function Breadcrumb({ studyId, studyName, workspace, folder }: BreadcrumbProps) {
+  const dispatch = useAppDispatch();
+  const navigate = useNavigate();
+  const pathHierarchy = buildBreadcrumbPath({
```
skamril: Hierarchy is not clear, name it like the function `breadcrumbPath` or `pathSegments`

webapp/src/components/App/Singlestudy/NavHeader/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+export function buildBreadcrumbPath({
+  studyName,
+  folderPath,
+  workspaceName,
+}: BuildBreadcrumbPathParams) {
+  const pathSegments = folderPath ? folderPath.split("/").filter(Boolean) : [];
+
+  // Build complete hierarchy including workspace (same pattern as StudyTree)
+  const hierarchy = [workspaceName, ...pathSegments];
```
skamril: `hierarchy` not clear

webapp/src/components/App/Singlestudy/NavHeader/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,69 @@
+/**
```
skamril: Facultative: create a Breadcrumb folder

Breadcrumb/ 
- index.tsx
- BreadcrumbLink.tsx

webapp/src/components/App/Singlestudy/NavHeader/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+import HomeIcon from "@mui/icons-material/Home";
+import { Link, Tooltip } from "@mui/material";
+import { useLayoutEffect, useRef, useState } from "react";
+
+interface BreadcrumbLinkProps {
```
skamril: Same for type name `Props`

webapp/src/components/App/Singlestudy/NavHeader/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,69 @@
...
+  "&:hover": {
+    color: "info.main",
+  },
+} as const;
+
+function BreadcrumbLink({ folderName, isFirstSegment, onClick }: BreadcrumbLinkProps) {
+  const textRef = useRef<HTMLSpanElement>(null);
+  const [isTruncated, setIsTruncated] = useState(false);
+
+  useLayoutEffect(() => {
```
skamril: `useEffect` works here.

> useLayoutEffect can hurt performance. Prefer [useEffect](https://react.dev/reference/react/useEffect) when possible.

https://react.dev/reference/react/useLayoutEffect

skamril (changes requested):

webapp/src/components/App/Singlestudy/NavHeader/Actions.tsx
hdinia: ok I will see

webapp/src/components/App/Singlestudy/NavHeader/Breadcrumb.tsx
hdinia: their is no difference if exported or not the common React convention is `ComponentNameProps` for both exported and non-exported

webapp/src/components/App/Singlestudy/NavHeader/BreadcrumbLink.tsx
hdinia: Thanks for the feedback, yes it can affect performance but only if used incorrectly, in that case `useLayoutEffect` is indeed the correct choice, or at least not incorrect 
This is a lightweight operation measuring DOM elements, and the effect runs only once, for those kind of DOM dependent operations its better to use it, because `useEffect` can cause visual flickering as it's executed after the browser paint

webapp/src/components/App/Singlestudy/NavHeader/BreadcrumbLink.tsx
hdinia: ok

webapp/src/components/App/Singlestudy/NavHeader/Breadcrumb.tsx
hdinia: will change to breadcrumbPath

webapp/src/components/App/Singlestudy/NavHeader/BreadcrumbLink.tsx
skamril: Here I test with useEffect there is no issue, because you use it only for the tooltip

skamril (approved):

PR #2711, opened 2025-09-17
feat(ui-launcher): update the dialog

skamril:
- add cluster field
- add cluster metrics
- move fields
- remove adequacy patch fields

### DIFF
--- a/webapp/src/components/App/shared/studies/dialogs/LaunchStudyDialog.tsx
+++ b/webapp/src/components/App/shared/studies/dialogs/LaunchStudyDialog.tsx
@@ -81,7 +81,36 @@ function LaunchStudyDialog(props: Props) {
 
   const launcherEngines = usePromiseWithSnackbarError(getLauncherEngines, {
     errorMessage: t("study.error.launcherEngines"),
+    onDataChange: (data) => {
+      if (data && options.launcher_id && !data.engines.includes(options.launcher_id)) {
+        setOptions((prevOptions) => ({
+          ...prevOptions,
+          launcher_id: data.engines[0],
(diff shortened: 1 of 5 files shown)


webapp/src/services/api/study.ts
```diff
@@ -262,6 +262,11 @@ interface LauncherMetrics {
   status: string;
 }
 
+export const getLauncherEngines = async () => {
```
smailio: no return types and no interface to represent this response ?

webapp/src/components/App/shared/studies/dialogs/LaunchStudyDialog.tsx
```diff
@@ -470,6 +394,140 @@ function LaunchStudyDialog(props: Props) {
...
+            response={launcherTimeLimit}
+            ifFulfilled={(timeLimit) => (
+              <TextField
+                label={t("study.timeLimit")}
+                type="number"
+                variant="outlined"
+                required
+                value={options.time_limit ? options.time_limit / 3600 : timeLimit.defaultValue}
+                onChange={(e) => {
+                  const newValue = parseInt(e.target.value, 10);
```
smailio: isn't this field supposed to become non editable ?

webapp/src/types/types.ts
```diff
@@ -150,6 +150,8 @@ export interface LaunchOptions {
   output_suffix?: string;
```
smailio, line 150: is the adequacy_patch field still usefull ?

smailio (dismissed):

hdinia (approved):

PR #2718, opened 2025-09-18
fix(output)!: allow aggregation with columns mismatch

MartinBelthle:
**Breaking change**
Drops the support of HDF5 for output aggregation and matrix export

**Feature**
Supports parquet format for output aggregation and matrix export

**Fix**
- https://github.com/AntaresSimulatorTeam/antares_craft/issues/264
- If 2 areas/links did not share the same columns and we specifically asked for a column only present in the second one, we used to stop and return an empty response. Now we iterate over all files to find the specified column

**Performance**:
- CSV format before: 18.8 seconds 
- CSV format now : 21.5s
- Parquet format with different columns: 14.8s
- Parquet format with same columns: 12.2s
- **Take-away**: 
     - We should use the parquet format.
     - If possible we should request data for one area at a time or specify which columns interest us. This way we'll take even greater advantage of performance improvements


antarest/core/serde/matrix_export.py
```diff
@@ -57,27 +57,45 @@ class DataframeStreamWriter(Protocol):
...
 def _write_dataframes_stream_csv(path: Path, sep: str, decimal: str, dataframes: Iterator[pd.DataFrame]) -> None:
     headers = True
     append = False
-    for df in _checked_dataframes_generator(dataframes):
-        df.to_csv(path, mode="a" if append else "w", sep=sep, decimal=decimal, index=False, header=headers)
+    for df, columns_to_add in _checked_dataframes_generator(dataframes):
+        if not columns_to_add:
+            df.to_csv(path, mode="a" if append else "w", sep=sep, decimal=decimal, index=False, header=headers)
+        else:
+            current_df = pd.read_csv(path, sep=sep, decimal=decimal, header=0)
```
sylvlecl: Wa can't afford to do this, this will be larger than memory.
The whole point of streaming the dataframes to the file (using generators instead of lists) is to avoid to load the whole data in memory.

We either need to:
1. make a first pass on all the files to detect the actual columns
2. update columns on the fly like you did, but in that case we should rewrite the file with the new columns by streaming the data in chunks instead of reading the whole file in memory
3. store data with different columns in different files, then concatenate them at the end in one file

The solutions will have different performance profiles:
- 1 will slow down aggregation for cases where all files have identical columns
- 2 will be very slow if many files have different columns, causing to re-write the file multiple times, but will not slow down the case where all files have identical columns
- 3 could be the best "mix" between those 2 solutions

Note that this raises a functional question also:
one goal of the aggregation is to flatten the result table, for example in the cluster file we have a set of X columns per cluster, and in the aggregation we flatten them as one line per cluster.
I suspect here this is a little similar, with a set of columns by group of STS for example: do we want to flatten that in one line per group of STS too ?

sylvlecl (changes requested):
See comment: we cannot read the whole file in memory, it will be larger than memory in many real world cases

antarest/core/serde/matrix_export.py
```diff
@@ -135,8 +113,8 @@ def media_type(self) -> str:
                 return "text/tab-separated-values"
             case TableExportFormat.CSV | TableExportFormat.CSV_SEMICOLON:
                 return "text/csv"
-            case TableExportFormat.HDF5:
-                return "application/x-hdf5"
+            case TableExportFormat.PARQUET:
+                return "application/octet-stream"
```
sylvlecl: We should use [application/vnd.apache.parquet](https://www.iana.org/assignments/media-types/application/vnd.apache.parquet)

antarest/core/serde/parquet_writer.py
```diff
@@ -0,0 +1,84 @@
...
+            file_path = path / file_name
+            file_counter += 1
+
+            new_writer = _parquet_writer(file_path, table.schema)
+            writers[df_cols] = new_writer
+
+        writers[df_cols].write_table(table)
+
+    # Close all writers
+    for writer in writers.values():
```
sylvlecl: we should probably do that in a `finally` to ensure we close them also when an exception occurs

antarest/study/storage/df_download.py
```diff
@@ -113,3 +103,25 @@ def _export_file(
...
+        # No need to reindex as all dataframes have the same columns
+        should_reindex = False
+
+    stream_writer = export_format.get_stream_writer()
+    new_index = []
+    if should_reindex:
+        # Sort the index for reproducibility
+        new_index = list(all_cols)
+        new_index.sort(key=lambda x: (x.isupper(), x.lower()))
+    stream_writer(file_download_path, yield_parquet_dataframes(tmp_path, all_df_names, new_index))
```
sylvlecl: Note that it would probably be more efficient to work with pyarrow tables here instead of going through dataframes again. However this would need more work to refactor existing writers, not worth it I think.

antarest/study/storage/df_download.py
```diff
@@ -113,3 +103,25 @@ def _export_file(
         },
         media_type=media_type,
     )
+
+
+def export_df_chunks(
+    tmp_path: Path, file_download_path: Path, df_chunks: Iterator[pd.DataFrame], export_format: TableExportFormat
+) -> None:
+    all_df_names, all_cols = write_dataframes_in_parquet_format(tmp_path, df_chunks)
```
sylvlecl: we need to explain somewhere **why** we have implemented that workflow:
1. we cannot hold everything in memory
2. dataframes are not required to have same columns

antarest/study/storage/df_download.py
```diff
@@ -113,3 +103,25 @@ def _export_file(
...
+            return
+        # No need to reindex as all dataframes have the same columns
+        should_reindex = False
+
+    stream_writer = export_format.get_stream_writer()
+    new_index = []
+    if should_reindex:
+        # Sort the index for reproducibility
+        new_index = list(all_cols)
+        new_index.sort(key=lambda x: (x.isupper(), x.lower()))
```
sylvlecl: It's kind of weird that we don't get the same column order when ONE area has additional columns.
Is there really a reproducibility issue ? don't we go over the files in the same order at every execution ?

antarest/core/serde/parquet_writer.py
```diff
@@ -0,0 +1,84 @@
...
+import pandas as pd
+import pyarrow as pa
+from pyarrow.parquet import ParquetFile, ParquetWriter
+
+
+def _parquet_writer(output_file: Path, schema: pa.Schema) -> ParquetWriter:
+    return ParquetWriter(output_file, schema, compression="zstd", data_page_version="2.0")
+
+
+def write_dataframes_in_parquet_format(path: Path, dataframes: Iterator[pd.DataFrame]) -> tuple[set[str], list[str]]:
```
sylvlecl: We need to document in docstring (or through a dataclass) what's the content of the return, and what the method is doing: what is the path we receive and what it's used for.
Otherwise we'll need to read the code and guess what's done which is not obvious.

We could rename the method to something more meaningful, like `write_by_column_sets`, a little ugly but it more or less says what it does

antarest/core/serde/parquet_writer.py
```diff
@@ -0,0 +1,84 @@
...
+        writer.write_table(first_table)
+        for df in dataframes:
+            if df.empty:
+                continue
+
+            table = pa.Table.from_pandas(df)
+            writer.write_table(table)
+
+
+def yield_parquet_dataframes(folder_path: Path, all_df_names: set[str], new_index: list[str]) -> Iterator[pd.DataFrame]:
```
sylvlecl: We should document a little what are the arguments.

This method has a strong link to the one which does the writing (basically it will never be used on something else than the result of the writing method), maybe it could be a method of an object returned by that one, to encapsulate it.

tests/core/serde/test_matrix_export.py
```diff
@@ -15,41 +15,7 @@
 import pandas as pd
 import pytest
 
-from antarest.core.serde.matrix_export import TableExportFormat, _checked_dataframes_generator
-
-
-def test_checked_stream_removes_empty_dataframes():
```
sylvlecl, line 21: We should add unit tests here for the new methods with various cases:

dataframes with same columns, successive dataframes with different columns (like first with A, second with B, third with C), and  assert something on the order of columns if we define an ordering, etc

sylvlecl (changes requested):
mainly suggestions for understandability of the added workflow, + we'd better add a few unit tests for the new methods, to cover more cases

antarest/study/storage/df_download.py
MartinBelthle: I'll let it as it is for now, we could do it in a next PR if we see it improves perfs

antarest/study/storage/df_download.py
```diff
@@ -113,3 +106,31 @@ def _export_file(
...
+    But we cannot perform a classic concatenation as we cannot hold all the dataframes in memory.
+    So first, we're writing them in parquet files according to their columns.
+
+    If only one file is created, it means all dataframes shared the same columns.
+    If the user asked for `parquet` format, there's nothing more to do, we can use the created file as the response.
+
+    Else, we'll have to iterate over written file(s), reading in chunks to avoid using too much memory and transforming them in the requested format.
+    If there's several files, we also have to reindex the dataframes to fill missing columns and to share the same columns order.
+    """
+    all_df_names, all_cols = write_dataframes_in_parquet_format_by_column_sets(tmp_path, df_chunks)
```
sylvlecl: It seems those files are not cleaned up, we need to create a temp directory for them

antarest/core/serde/parquet_writer.py
```diff
@@ -0,0 +1,133 @@
...
+
+                if should_write_new_file:
+                    file_name = f"file{file_counter}.parquet"
+                    filenames.append(file_name)
+                    file_path = path / file_name
+                    file_counter += 1
+
+                    table = pa.Table.from_pandas(df)
+                    current_writer = _parquet_writer(file_path, table.schema)
+                    writers[current_writer] = table.schema
```
sylvlecl: needs renaming. Actually we don't need anymore to keep a dict, we only need the current one, which is the schema of the current writer

sylvlecl (approved):

sylvlecl (approved):

PR #2719, opened 2025-09-19
refactor(scenario_builder): introduce dao

smailio:


antarest/study/dao/api/scenario_builder_dao.py
```diff
@@ -0,0 +1,61 @@
...
+    def get_scenario_by_type(self, scenario_type: ScenarioType) -> AnyScenarios:
+        """
+        Get a scenario, from the active ruleset, by its type (name).
+        """
+        raise NotImplementedError()
+
+
+class ScenarioBuilderDao(ReadOnlyScenarioBuilderDao):
+    @abstractmethod
+    def save_scenario_builder(self, ruleset_update: RulesetsUpdate) -> None:
```
sylvlecl: In other DAOs, we have only implemented "replacements", not updates, updates being handled at the business level.

For consistency sake, I think we should stick with that. Otherwise we should rename the method.

To be honest, maybe in the future we'll have some benefit to have an update at the DAO level, because the data can be quite large, and so being able to not read everything and save everything back could be interesting.
But we can decide that later.

sylvlecl (changes requested):
Thanks!
One thing to change I think: for consistency sake, the "save" operation should be a real save taking a full `Ruleset`, not an update.

sylvlecl (approved):

PR #2720, opened 2025-09-22
fix(study): bug deleting raw study with variants

TheoPascoli:


antarest/study/service.py
```diff
@@ -1275,18 +1275,28 @@ def delete_study(self, uuid: str, children: bool) -> None:
...
-                    )
-                    return
-            else:
+        variant_service = self.storage_service.variant_study_service
+
+        if variant_service.has_children(study):
+            if not children:
                 raise StudyDeletionNotAllowed(study.id, "Study has variant children")
 
+            if isinstance(study, VariantStudy):
```
sylvlecl: Shouldn't we just remove that check ? (and adapt the walk_children function if needed) Indeed raw studies may have children too, we should be able to "walk" their children

TheoPascoli: Yes ! Now, walk_children accepts any Study. I added a include_parent = False boolean so the raw study does not execute the callback on himself the root. The raw study is deleted after.

sylvlecl (approved):
Nice thanks

PR #2725, opened 2025-09-23
fix(upgrade): allow version to be given as 9.2 instead of 920

MartinBelthle:
[ANT-3850]


sylvlecl (approved):
Looks good to me.
Not very clear why the bug did not happen systematically, maybe sometimes we store the old format in database and sometimes the new one ?

MartinBelthle:
> Looks good to me. Not very clear why the bug did not happen systematically, maybe sometimes we store the old format in database and sometimes the new one ?

Not sure either but the error message that Brahim had corresponds to the message we had due to this issue

PR #2729, opened 2025-09-24
fix(ui-study): variant not visible after creation

skamril:
ANT-3896

### DIFF
--- a/webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
+++ b/webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
@@ -36,6 +36,7 @@ interface Props {
 function CreateVariantDialog({ parentId, open, variantTree, onClose }: Props) {
   const [t] = useTranslation();
   const navigate = useNavigate();
+  const defaultValues = { name: "", sourceId: parentId };
 
   const studiesSource = useMemo(() => createListFromTree(variantTree), [variantTree]);
 
@@ -45,8 +46,6 @@ function CreateVariantDialog({ parentId, open, variantTree, onClose }: Props) {
   );
 
   const existingNames = useMemo(() => studiesSource.map(({ name }) => name), [studiesSource]);
-
-  const defaultValues = { name: "", sourceId: parentId };
 
   ////////////////////////////////////////////////////////////////
   // Event Handlers
--- a/webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
+++ b/webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
@@ -15,12 +15,12 @@
 import SelectFE from "@/components/common/fieldEditors/SelectFE";
 import { validateString } from "@/utils/validation/string";
 import AddCircleIcon from "@mui/icons-material/AddCircle";
-import { useEffect, useMemo, useState } from "react";
+import { useMemo } from "react";
 import { useTranslation } from "react-i18next";
 import { useNavigate } from "react-router";
 import { createVariant } from "../../../../../services/api/variant";
 import { createListFromTree } from "../../../../../services/utils";
-import type { GenericInfo, VariantTree } from "../../../../../types/types";
+import type { VariantTree } from "../../../../../types/types";
 import FormDialog from "../../../../common/dialogs/FormDialog";
 import StringFE from "../../../../common/fieldEditors/StringFE";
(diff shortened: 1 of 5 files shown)


webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
```diff
@@ -36,14 +36,17 @@ interface Props {
...
+    () => studiesSource.map(({ id, name }) => ({ value: id, label: name })),
+    [studiesSource],
+  );
 
-  useEffect(() => {
-    setSourceList(createListFromTree(variantTree));
-  }, [variantTree]);
+  const existingNames = useMemo(() => studiesSource.map(({ name }) => name), [studiesSource]);
+
+  const defaultValues = { name: "", sourceId: parentId };
```
hdinia: suggestion: you can put it on top (above hooks) and add `as const`. Or outside the component

webapp/src/hooks/useFormCloseConfirm.tsx
hdinia, line 1: It seems like an over-engineered way to disable the form confirm if the state is dirty. Is that the only goal with this change or I missed something else ? 

1. If the goal was to preserve the form confirm + `navigate();` via `react-router-dom` -> this fix may be incomplete, I suggest you revise it.

2.  If we only need to prevent form confirmation for specific cases, a simple `disableFormConfirm` prop passed from `FormDialog` to `Form` would be cleaner and more explicit.

3. Why navigate automatically at all? The variant appears on the same page after creation, users could easily click to access it if needed, avoiding this complexity entirely.

webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
```diff
@@ -36,14 +36,17 @@ interface Props {
...
+
+  const studySourceOptions = useMemo(
+    () => studiesSource.map(({ id, name }) => ({ value: id, label: name })),
+    [studiesSource],
+  );
 
-  useEffect(() => {
-    setSourceList(createListFromTree(variantTree));
-  }, [variantTree]);
+  const existingNames = useMemo(() => studiesSource.map(({ name }) => name), [studiesSource]);
```
hdinia, line 48: suggestion: make it more specific like `existingVariantNames`

webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
```diff
@@ -36,14 +36,17 @@ interface Props {
...
-  const [sourceList, setSourceList] = useState<GenericInfo[]>([]);
-  const defaultValues = { name: "", sourceId: parentId };
 
-  const existingVariants = useMemo(() => sourceList.map((variant) => variant.name), [sourceList]);
+  const studiesSource = useMemo(() => createListFromTree(variantTree), [variantTree]);
+
+  const studySourceOptions = useMemo(
+    () => studiesSource.map(({ id, name }) => ({ value: id, label: name })),
+    [studiesSource],
+  );
```
hdinia, line 46: suggestion: 
`studiesSource` -> `variantList`

`studySourceOptions` -> `variantOptions`

webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
skamril: There is the name of the reference study also

webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
skamril: There is also the reference study in the list

webapp/src/components/App/Singlestudy/HomeView/InformationView/CreateVariantDialog.tsx
skamril: There is `parentId` from the props.

webapp/src/hooks/useFormCloseConfirm.tsx
skamril: 1. What's your solution?
2. We only need to prevent the confirmation dialog if `navigate` is called inside `onSubmitSuccessful`, not completely.
3. It's better to select the new element added to a list. Then it's a general fix, a view transition can be made in `onSubmitSuccessful` on another view.

hdinia (approved):

PR #2730, opened 2025-09-24
feat(ui-studies): move favorites studies from Studies sidenav to main app sidebar

hdinia:


webapp/src/components/App/Tasks/index.tsx
```diff
@@ -94,9 +93,6 @@ function JobsListing() {
   const init = async (fetchOnlyLatest = true) => {
     setLoaded(false);
     try {
-      if (studies.length === 0) {
```
skamril, line 97: Can you add `studies` as dependency of `const jobsMemo = useMemo<TaskView[]>`

skamril (changes requested):

skamril (approved):

PR #2735, opened 2025-09-26
fix(commands): use impersonator instead of id for `user_id` attribute

MartinBelthle:
Will fix [ANT-3906] for new commands, not for existing ones

### DIFF
--- a/antarest/login/utils.py
+++ b/antarest/login/utils.py
@@ -37,6 +37,13 @@ def get_user_id() -> str:
     return str(escape(str(user.id))) if user else "Unknown"
 
 
+def get_user_impersonator() -> int:
+    user = get_current_user()
+    if not user:
+        raise MustBeAuthenticatedError()
+    return user.impersonator
+
+
 @contextlib.contextmanager
 def current_user_context(token: Optional[JWTUser]) -> Iterator[JWTUser | None]:
     reset_token = _current_user.set(token)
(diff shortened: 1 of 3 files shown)


tests/integration/test_integration_token_end_to_end.py
```diff
@@ -146,52 +124,45 @@ def test_nominal_case_of_an_api_user(client: TestClient, admin_access_token: str
...
+    res = client.post(f"/v1/studies/{variant_id}/commands", json=commands)
     assert res.status_code == 200
 
     # Check if the author's name and date of update are retrieved with commands created by a bot
-    commands_res = client.get(f"/v1/studies/{variant_id}/commands", headers=bot_headers)
+    commands_res = client.get(f"/v1/studies/{variant_id}/commands")
 
     for command in commands_res.json():
-        assert command["user_name"] == "admin_bot"
+        assert command["user_name"] == "admin"  # Should be the user's name not the token name
```
MartinBelthle, line 159: It's the only change in the file, the rest is just cleaning

antarest/login/utils.py
```diff
@@ -37,6 +37,13 @@ def get_user_id() -> str:
     return str(escape(str(user.id))) if user else "Unknown"
 
 
+def get_user_impersonator() -> int:
+    user = get_current_user()
+    if not user:
```
sylvlecl: there is an existing shortcut for this `require_current_user`

sylvlecl (dismissed):
Very minor comment, OK otherwise

PR #2738, opened 2025-09-29
fix(launcher): remove duplicate launcher id argument and fix API call

sylvlecl:
- launcher argument was present as both a query parameter and in the body of the submission
- the frontend used the body parameter, which was actually not used by the backend

### DIFF
--- a/webapp/src/services/api/study.ts
+++ b/webapp/src/services/api/study.ts
@@ -251,9 +251,9 @@ export const launchStudy = async (
   version: string | undefined = undefined,
   launcher: string | undefined = undefined,
 ): Promise<string> => {
-  const versionArg = version ? `?version=${version}` : "";
-  const launcherArg = launcher ? `?launcher=${launcher}` : "";
-  const res = await client.post(`/v1/launcher/run/${sid}${versionArg}${launcherArg}`, options);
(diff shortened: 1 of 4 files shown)


TheoPascoli (dismissed):

webapp/src/services/api/study.ts
```diff
@@ -249,9 +249,11 @@ export const launchStudy = async (
   sid: string,
   options: LaunchOptions = {},
   version: string | undefined = undefined,
+  launcher: string | undefined = undefined,
 ): Promise<string> => {
   const versionArg = version ? `?version=${version}` : "";
-  const res = await client.post(`/v1/launcher/run/${sid}${versionArg}`, options);
+  const launcherArg = launcher ? `?launcher=${launcher}` : "";
+  const res = await client.post(`/v1/launcher/run/${sid}${versionArg}${launcherArg}`, options);
```
skamril:
```typescript
const res =  await client.post<Storage>(
    `/v1/launcher/run/${sid}${versionArg}`,
    options,
    { params: { launcher } },
);
```

skamril (changes requested):

skamril (approved):

PR #2739, opened 2025-09-30
feat(ui-studies): make search field manual

skamril:
ANT-3234

### DIFF
--- a/webapp/src/hooks/useFormCloseProtection.ts
+++ b/webapp/src/hooks/useFormCloseProtection.ts
@@ -14,13 +14,15 @@
 
 import { setFormStatus } from "@/redux/ducks/ui";
 import useAppDispatch from "@/redux/hooks/useAppDispatch";
-import { useUnmount, useUpdateEffect } from "react-use";
+import { useEffect } from "react";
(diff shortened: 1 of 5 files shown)


webapp/src/hooks/useFormCloseProtection.ts
```diff
@@ -14,13 +14,15 @@
...
-import { useUnmount, useUpdateEffect } from "react-use";
+import { useEffect } from "react";
+import { usePrevious, useUnmount } from "react-use";
 import useBlocker from "./useBlocker";
 import useFormCloseConfirm from "./useFormCloseConfirm";
 
 export interface UseFormCloseProtectionParams {
   isSubmitting: boolean;
   isDirty: boolean;
+  disableHook?: boolean;
```
hdinia, line 25: suggestion: disabled

hdinia (approved):

webapp/src/hooks/useFormCloseProtection.ts
skamril: `disabled` doesn't make it clear whether we're talking about the hook or the form.

PR #2740, opened 2025-10-01
refactor(area-properties): implement dao interface

TheoPascoli:


antarest/study/dao/api/area_properties_dao.py
```diff
@@ -0,0 +1,29 @@
...
+from abc import ABC, abstractmethod
+
+from antarest.study.business.model.area_properties_model import AreaProperties
+
+
+class ReadOnlyAreaPropertiesDao(ABC):
+    @abstractmethod
+    def get_area_properties(
+        self,
+        area_id: str,
```
MartinBelthle: Really minor, you can delete the comma at the end of the line to put all in one line.

You can do the same at other places in your code

antarest/study/dao/file/file_study_area_properties_dao.py
```diff
@@ -0,0 +1,93 @@
...
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+from abc import ABC, abstractmethod
+
+from typing_extensions import override
+
+from antarest.study.business.model.area_properties_model import (
+    AreaProperties,
+    get_adequacy_patch_path,
```
MartinBelthle: To me these 3 methods belong inside this file as they are only meant to be used by the fs

antarest/study/dao/file/file_study_area_properties_dao.py
```diff
@@ -0,0 +1,93 @@
...
+    get_optimization_path,
+    get_thermal_path,
+)
+from antarest.study.dao.api.area_properties_dao import AreaPropertiesDao
+from antarest.study.model import STUDY_VERSION_8_3
+from antarest.study.storage.rawstudy.model.filesystem.config.area import (
+    AdequacyPatchFileData,
+    AreaPropertiesFileData,
+    OptimizationFileData,
+    ThermalAreasProperties,
```
MartinBelthle: I think we should rename also add `FileData` at the end of the class: `ThermalAreasProperties`

antarest/study/dao/file/file_study_area_properties_dao.py
```diff
@@ -0,0 +1,93 @@
...
+        pass
+
+    @override
+    def get_area_properties(
+        self,
+        area_id: str,
+    ) -> AreaProperties:
+        file_study = self.get_file_study()
+
+        current_thermal_props = file_study.tree.get(get_thermal_path())
```
smailio, line 58: this get_thermal_path function should be moved in this dao file I think. The path is related to the file system implementaion not the buisness model.

MartinBelthle (changes requested):
Nice work :)

MartinBelthle:
Also, minor but you should probably rename your PR `refactor(...)` instead of `feat(...)`

antarest/study/dao/file/file_study_area_properties_dao.py
```diff
@@ -0,0 +1,153 @@
...
+            )
+            area_map[area_id] = props_data.get_area_properties(area_id, file_study.config.version)
+
+        return area_map
+
+    @override
+    def save_area_properties(self, area_id: str, area_properties: AreaProperties) -> None:
+        file_study = self.get_file_study()
+
+        thermal_props = ThermalAreasFileData(**file_study.tree.get(get_thermal_path()))
```
smailio: Usually we use model validate but it seems it's almost identical to just use the __init__. Also usually instead of creating new object directly in the dao we interact with the fs module using parse_ and serialize_ functions.

smailio (changes requested):
Neat PR

antarest/study/dao/file/file_study_area_properties_dao.py
```diff
@@ -0,0 +1,153 @@
...
+            optimization_properties=OptimizationFileData(**current_optim_properties),
+            adequacy_patch_properties=AdequacyPatchFileData(**current_adequacy_patch),
+        )
+
+        return properties.get_area_properties(area_id, file_study.config.version)
+
+    @override
+    def get_all_area_properties(self) -> dict[str, AreaProperties]:
+        """
+        Retrieves all areas of a study.
```
MartinBelthle: This docstring should be updated

antarest/study/dao/file/file_study_area_properties_dao.py
```diff
@@ -0,0 +1,153 @@
...
+            )
+            area_map[area_id] = props_data.get_area_properties(area_id, file_study.config.version)
+
+        return area_map
+
+    @override
+    def save_area_properties(self, area_id: str, area_properties: AreaProperties) -> None:
+        file_study = self.get_file_study()
+
+        thermal_props = ThermalAreasFileData.model_validate(file_study.tree.get(get_thermal_path()))
```
MartinBelthle, line 126: Seems that here you should not have to read the existing data as you're giving a full object as a parameter

antarest/study/dao/file/file_study_area_properties_dao.py
```diff
@@ -0,0 +1,153 @@
...
+    AdequacyPatchFileData,
+    AreaFileData,
+    AreaPropertiesFileData,
+    OptimizationFileData,
+    ThermalAreasFileData,
+)
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+
+ALL_AREAS_PATH = "input/areas"
+THERMAL_AREAS_PATH = "input/thermal/areas"
```
MartinBelthle: This looks like we could simplify and just use the `get_thermal_path` method

MartinBelthle (commented):
Little comments only

MartinBelthle (approved):

PR #2741, opened 2025-10-01
fix(launcher): fix arguments handling in local launcher

sylvlecl:
- adds compatibility check for launcher arguments
- fixes presolve option handling for versions < 9.2
- fixes arguments formatting for subprocess: options and their arguments are separate command line arguments

Also fixes encoding-related errors when reading logs:
on windows in particular, error messages from the solver may have a locale-dependent encoding.
Not clear if the solver as a whole uses that same encoding, so we just assume utf-8 and replace
invalid characters with a replacement char...


.github/workflows/deploy.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - "master"
       - "hotfix/**"
+      - "fix/fix-local-launcher-options"
```
sylvlecl: Remove it before merging

tests/launcher/test_local_launcher.py
```diff
@@ -143,9 +142,9 @@ def test_parse_launcher_arguments(launcher_config: LocalConfig, xpress_env):
         for version in [solver_version_8_8, solver_version_9_2, solver_version_9_3]:
             sim_args, _ = local_launcher._parse_launcher_options(launcher_parameters, version)
             if version == solver_version_8_8:
-                assert sim_args == ["--use-ortools", f"--ortools-solver={solver}"]
+                assert sim_args == ["--use-ortools", "--ortools-solver", f"{solver}"]
```
MartinBelthle: Minor: I think f"{solver}" could be solver

.github/workflows/deploy.yml
MartinBelthle: As you said, should be removed

tests/launcher/test_local_launcher.py
```diff
@@ -166,14 +182,40 @@ def test_parse_launcher_arguments(launcher_config: LocalConfig, xpress_env):
         sim_args, _ = local_launcher._parse_launcher_options(launcher_parameters, version)
         arg = "lp" if version == solver_version_9_2 else "linear"
         assert sim_args == [
```
MartinBelthle, line 184: Not sure to see where you test this : `fixes arguments formatting for subprocess:` as the test seems to be the same as before

MartinBelthle (dismissed):

tests/launcher/test_local_launcher.py
sylvlecl: Tests have been adapted to check that arguments are correctly separated from their options, which is enough:
the problem was that `--linear-solver-param-optim-1="PRESOLVE 2 THREADS 4"` was provided as 1 argument, including the quotes. Unlike from a shell, the solver seemed unable to understand this.
Now we pass `--linear-solver-param-optim-1` and `PRESOLVE 2 THREADS 4` separately, without quotes, which fixes that issue.

MartinBelthle: Okay I see

PR #2742, opened 2025-10-02
feat(study)!: add new endpoint for `antares-craft` to read a study

MartinBelthle:
Introduces a new endpoint, GET `/{uuid}/data`
Allows antares-craft to read the study around 60% faster

Also add hydro allocation inside the endpoint

**Breaking change:**
 - Removal of the endpoint GET `/{uuid}/hydro` that was only used by antares-craft


tests/integration/assets/STA-mini.zip
MartinBelthle, line 1: I changed the content of the xpansion settings file as it was in a very old format no longer readable by the app

antarest/study/service.py
```diff
@@ -206,6 +222,45 @@ def assert_raw(study: Study) -> RawStudy:
     return study
 
 
+class AntaresCraftStudySettings(AntaresBaseModel):
```
sylvlecl: We should not refer to antares-craft here, it should remain as much as possible a client like any other. This endpoint could also be used by any other client.

antarest/study/service.py
```diff
@@ -206,6 +222,45 @@ def assert_raw(study: Study) -> RawStudy:
...
+class AntaresCraftStudy(AntaresBaseModel):
+    """
+    Study representation used by antares-craft inside its reading method.
+    """
+
+    name: str
+    version: StudyVersionStr
+    path: str | None
+
+    area_properties: dict[str, AreaProperties]
```
sylvlecl: It could feel more natural to have container objects for everything in an area, instead of having dicts, for example:
```json
"areas" : [
    {
       "id": "fr",
       "name": "FR",
       "properties": "...",
       "ui": "...",
       "thermal_clusters": "...",
       "renewable_clusters": "...",
    }
]
```

antarest/study/service.py
```diff
@@ -206,6 +224,54 @@ def assert_raw(study: Study) -> RawStudy:
...
+    name: str
+    version: StudyVersionStr
+    path: str | None
+
+    areas: list[StudyAreasDTO]
+    links: list[Link]
+    binding_constraints: list[BindingConstraint]
+    settings: StudySettingsDTO
+    xpansion: StudyXpansionDTO | None
+    outputs: dict[str, Simulation]
```
sylvlecl: For consistency sake we could use a list here too, seem the key is redundant with the name inside `Simulation`

antarest/study/service.py
```diff
@@ -206,6 +224,54 @@ def assert_raw(study: Study) -> RawStudy:
...
+
+
+class StudyDataDTO(AntaresBaseModel):
+    """
+    DTO representing data of the whole study.
+    """
+
+    name: str
+    version: StudyVersionStr
+    path: str | None
```
sylvlecl: Seems weird to me to have this here, it's not actual study data

antarest/study/service.py
```diff
@@ -206,6 +224,54 @@ def assert_raw(study: Study) -> RawStudy:
...
+
+
+class StudyXpansionDTO(AntaresBaseModel):
+    settings: XpansionSettings
+    candidates: list[XpansionCandidate]
+    constraint: bytes
+
+
+class StudyShortTermStorageDTO(AntaresBaseModel):
+    cluster: STStorage
```
sylvlecl: We should not name it cluster, it's not a cluster.
In any case, it would be better to have something like that with the name and id and other properties at the first level:
```json
{
  "id": "...",
  "name": "...",
  "injectionNominalCapacity": 100,
  ...,
  "constraints": [

  ]
}
```

antarest/study/service.py
```diff
@@ -206,6 +224,54 @@ def assert_raw(study: Study) -> RawStudy:
...
+    adequacy_patch: AdequacyPatchParameters
+    optimization: OptimizationPreferences
+    thematic_trimming: ThematicTrimming
+    playlist: Playlist
+
+
+class StudyXpansionDTO(AntaresBaseModel):
+    settings: XpansionSettings
+    candidates: list[XpansionCandidate]
+    constraint: bytes
```
sylvlecl: Can't we have a real model instead of bytes here ?
I don't know what's the actual content of that, but we should as much as possible have business-meaningful objects here, not raw data that are dependent on implementation details (exact format of the data etc).

I think it can be multiple constraint**s**, no ?

antarest/study/service.py
```diff
@@ -206,6 +224,54 @@ def assert_raw(study: Study) -> RawStudy:
...
+
+
+class StudyShortTermStorageDTO(AntaresBaseModel):
+    cluster: STStorage
+    constraints: list[STStorageAdditionalConstraint]
+
+
+class StudyAreasDTO(AntaresBaseModel):
+    id: str
+    properties: AreaProperties
```
sylvlecl: We need to have the name too

antarest/study/service.py
```diff
@@ -206,6 +224,54 @@ def assert_raw(study: Study) -> RawStudy:
...
+    ui: dict[str, Any]
```
sylvlecl: Can we have a pydantic model here too ?

sylvlecl (commented):
More suggestions about the model

antarest/study/service.py
MartinBelthle: I'll wait for the DAO to be created for this as for now we only have a dict, no Pydantic object

antarest/study/service.py
MartinBelthle: We need the information inside antares-craft. I could rename it `folder`

antarest/study/service.py
MartinBelthle: I'll have to use the `AreaInfo` model that I don't like at all to access this information. And antares-craft doesn't really use names, the only way you can access an area is by doing `study.get_areas()[area_id]`

sylvlecl: But you still want to be able to get the name from antares-craft, no ?

sylvlecl: PS: we can just add the name to this model, not using `AreaInfo`

antarest/study/service.py
sylvlecl: hmmm ok then, there's already some "metadata" here anyway. Indeed we could rename it to folder, path seems a little more ambiguous (could mean the full path of the study or something like that)

antarest/study/service.py
```diff
@@ -207,6 +228,75 @@ def assert_raw(study: Study) -> RawStudy:
     return study
 
 
+class StudySettingsDTO(AntaresBaseModel):
```
sylvlecl: We'd better move those models to a dedicated module, maybe alongside other  models modules (`area_model` etc)

sylvlecl (changes requested):
Minor comment

antarest/study/service.py
MartinBelthle: Yes you can access the area name in antares-craft. What I meant is that I don't think it's useful, people will use the id due to the user API. But I'll add the name as this endpoint could be used by other people in the future

sylvlecl (approved):

PR #2743, opened 2025-10-02
feat(study_editor): added study author and editor in study metadata dto

mehdiwahada:
I added the author and editor field in the StudyMetadataDTO class


antarest/study/storage/abstract_storage_service.py
```diff
@@ -80,6 +80,8 @@ def get_study_information(
             id=study.id,
             name=study.name,
             version=study.version,
+            author=study.author,
```
sylvlecl: We should take the field author from additional_data.
This one from study seems never used, we should remove it, but not in that PR.

antarest/study/model.py
```diff
@@ -515,6 +517,7 @@ def transform_horizon_to_str(cls, val: str | int | None) -> Optional[str]:
 class StudyMetadataPatchDTO(AntaresBaseModel):
     name: Optional[str] = None
     author: Optional[str] = None
+    editor: Optional[str] = None
```
sylvlecl: We don't need to add it here: 
this object is used in the API to allow clients to update some study metadata, but they should not be able to modify the editor directly.
It's the application responsibility to update the editor depending on actions that are performed in the application (this is the goal of your other PR :) ).

sylvlecl (changes requested):
A few changes requested

antarest/study/model.py
```diff
@@ -493,6 +493,8 @@ class StudyMetadataDTO(AntaresBaseModel):
     id: str
     name: str
     version: StudyVersionInt
+    author: Optional[str] = None
+    editor: Optional[str] = None
```
hdinia, line 497: We already have the owner in the study metadata, so what’s the purpose of the “author” field? It seems redundant, especially since the frontend only needs the editor. Is “author” used for something else?

sylvlecl: It's not exactly redundant:
for disk studies, the author is defined in study files and can be  any name, including people that don't have a corresponding application user. The owner of a study is the owner of the study in the application, and is always an application user.

Anyway, indeed it has not been asked to display it in the frontend, but it makes sense to make it available from the API point of view, since we indeed have that information in the data.

sylvlecl (approved):

antarest/study/storage/abstract_storage_service.py
```diff
@@ -80,6 +80,8 @@ def get_study_information(
             id=study.id,
             name=study.name,
             version=study.version,
+            author=study.additional_data.author,
```
sylvlecl: Actually here we cannot assume that study.additional_data is not None, be cause for all studies this will not be the case.

Don't know why it does not break any test :s 

You can use the `additional_data` defined above instead.

Actually, mypy should catch those errors, for this we need to fix typing in the `Study` class definition like so, so that mypy understands that additional_data can be `None`:
```python
    additional_data: Mapped[StudyAdditionalData | None] = relationship(
        StudyAdditionalData,
        uselist=False,
        cascade="all, delete, delete-orphan",
    )
```

sylvlecl (changes requested):
One more change requested, I did not see it at first sorry: let's take this opportunity to fix the typing of `Study.additional_data`

antarest/study/model.py
hdinia: OK I understand better the difference thanks

sylvlecl (approved):

PR #2745, opened 2025-10-03
fix(worker): fix unarchive worker build

sylvlecl:
The unarchive worker for windows does not work anymore:
 - we use an outdated pyinstaller version to build it
 - it doesn't call `multiprocessing.freeze_support()`, which causes an endless loop on startup

Also fixes event parsing in redis backend, which was not working anymore.


sylvlecl:
Replaced by https://github.com/AntaresSimulatorTeam/AntaREST/pull/2748

PR #2747, opened 2025-10-06
refactor(area-ui)!: adjust models and update logic for area UI properties

TheoPascoli:
Breaking change: 
- Layers related field (layerX, layerY, layerColor) are not accepted anymore in the `PUT "/studies/{uuid}/areas/{area_id}/ui"` endpoint

### DIFF
--- a/antarest/study/storage/variantstudy/model/command/update_area_ui.py
+++ b/antarest/study/storage/variantstudy/model/command/update_area_ui.py
@@ -14,8 +14,7 @@ import typing as t
 
 from typing_extensions import override
 
-from antarest.study.business.model.area_model import AreaUI, AreaUIUpdate, update_area_ui
-from antarest.study.storage.rawstudy.model.filesystem.config.area import AreaUIFileData, AreaUIStyle
+from antarest.study.business.model.area_model import AreaUIUpdate
 from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
 from antarest.study.storage.variantstudy.model.command.common import CommandName, CommandOutput, command_succeeded
 from antarest.study.storage.variantstudy.model.command.icommand import ICommand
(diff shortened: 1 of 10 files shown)


antarest/study/storage/variantstudy/model/command/update_area_ui.py
```diff
@@ -36,21 +36,32 @@ class UpdateAreaUI(ICommand):
     # ==================
 
     area_id: str
-    area_ui: UpdateAreaUi
     layer: str
+    parameters: AreaUIUpdate
```
sylvlecl, line 43: If we change the model here, we need to ensure that we can still read the old ones that are present in database.
In order to do this, we need to increment the version to 2, and parse the input (in a model validator) depending on the version (see other commands for this).

We then need to add compatibility tests in `test_command_factory.py`

antarest/study/web/study_data_blueprint.py
```diff
@@ -213,7 +213,7 @@ def update_link(uuid: str, area_from: str, area_to: str, link_update_dto: LinkUp
         summary="Update area information",
         response_model=None,
     )
-    def update_area_ui(uuid: str, area_id: str, area_ui: UpdateAreaUi, layer: str = "0") -> Any:
+    def update_area_ui(uuid: str, area_id: str, area_ui: AreaUIUpdate, layer: str = "0") -> Any:
```
sylvlecl, line 216: Seems legit but we need to document the breaking change:
the body does not accept anymore layers data

sylvlecl (changes requested):
Few comments about breaking changes (API and commands stored in database).

antarest/study/storage/variantstudy/model/command/update_area_ui.py
TheoPascoli: Done

PR #2748, opened 2025-10-06
fix(worker): fix archive worker startup

sylvlecl:
The unarchive worker for windows does not work anymore:

 - we use an outdated pyinstaller version to build it
 - it doesn't call `multiprocessing.freeze_support()`, which causes an endless loop on startup

Also fixes event parsing in redis backend, which was not working anymore.

### DIFF
--- a/scripts/package_worker.sh
+++ b/scripts/package_worker.sh
@@ -13,5 +13,3 @@ echo "INFO: Generating the Worker Application..."
 pushd ${PROJECT_DIR}
 pyinstaller --distpath ${DIST_DIR} AntaresWebWorker.spec
 popd
-
-chmod +x "${DIST_DIR}/AntaresWebWorker"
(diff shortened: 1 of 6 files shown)


scripts/package_worker.sh
```diff
@@ -13,5 +13,3 @@ echo "INFO: Generating the Worker Application..."
 pushd ${PROJECT_DIR}
 pyinstaller --distpath ${DIST_DIR} AntaresWebWorker.spec
 popd
-
-chmod +x "${DIST_DIR}/AntaresWebWorker"
```
MartinBelthle, line 17: Why did you remove this ? It was useless ?

MartinBelthle (dismissed):

scripts/package_worker.sh
sylvlecl: Indeed, under linux it's already created with execute permissions, under windows it does not exist.

PR #2753, opened 2025-10-07
refactor(hydro): introduce DAO for allocation

MartinBelthle:
Also fix an issue where we had the name written in the file instead of the id. GET endpoints used to fail, now it's good.


antarest/study/storage/variantstudy/model/command/replace_hydro_allocation.py
TheoPascoli, line 1: Double copyright

antarest/study/business/allocation_management.py
```diff
@@ -10,111 +10,13 @@
...
-        """
-
-        array = np.array(data)
-        new_values = values if isinstance(values, dict) else values.data
-        rows = len(new_values.get("index", []))
-        cols = len(new_values.get("columns", []))
-
-        if array.size == 0:
-            raise ValueError("allocation matrix must not be empty")
-        if array.shape != (rows, cols):
```
TheoPascoli, line 102: I cant find those validations in the new version. They are not needed anymore ? this one and the 2 below

TheoPascoli:
Nice ! Just one question about the data validations

antarest/study/storage/variantstudy/model/command/replace_hydro_allocation.py
MartinBelthle: Indeed

antarest/study/business/allocation_management.py
MartinBelthle: The 2 below are in the HydroAllocation class so we don't need to reperform the validation inside this class as it's only built from the other one

And this one is also guaranteed via construction as we build the class from a dataframe

TheoPascoli (approved):

PR #2756, opened 2025-10-08
refactor(area): implement dao

TheoPascoli:


antarest/study/business/area_management.py
```diff
@@ -52,17 +52,7 @@ def __init__(
...
-            Area(
-                id=area_id,
-                name=area.name,
-                thermals=self._get_clusters(file_study, area_id),
-            )
-            for area_id, area in cfg_areas.items()
-        ]
+        return study.get_study_dao().get_all_areas()
 
     def get_all_areas_ui_info(self, study: StudyInterface) -> Dict[str, Any]:
```
MartinBelthle: This should return a Pydantic object dict[str, AreaUI]

antarest/study/business/area_management.py
```diff
@@ -187,11 +121,46 @@ def delete_area(self, study: StudyInterface, area_id: str) -> None:
...
+            A mapping of ALL area IDs to area properties.
+        """
+        old_areas_by_ids = self.get_all_area_properties(study)
+        new_areas_by_ids = dict(old_areas_by_ids)
+
+        areas_properties: Dict[str, AreaPropertiesUpdate] = {}
+
+        for area_id, update_area in properties.items():
+            old_area = old_areas_by_ids[area_id]
+            update_data = update_area.model_dump(exclude_none=True)
```
MartinBelthle: Could use `update_area_properties` here I think

antarest/study/business/area_management.py
```diff
@@ -77,73 +67,17 @@ def get_all_areas_ui_info(self, study: StudyInterface) -> Dict[str, Any]:
...
+        command = UpdateLayerAreas(
+            layer_id=layer_id,
+            area_ids=areas,
+            command_context=self._command_context,
+            study_version=study.version,
+        )
+        study.add_commands([command])
 
     def create_area(self, study: StudyInterface, area_creation_info: AreaCreation) -> Area:
         file_study = study.get_files()
```
MartinBelthle: We shouldn't have to do that here, we should use the DAO

antarest/study/dao/api/area_dao.py
```diff
@@ -0,0 +1,106 @@
...
+        Args:
+            area_id: The area identifier to delete.
+
+        Raises:
+            AreaNotFound: If the area does not exist.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def save_area_ui(self, area_id: str, layer: str, area_ui_update: AreaUIUpdate) -> None:
```
MartinBelthle: Shouldn't this be AreaUI instead of AreaUIUpdate ?

antarest/study/dao/api/area_dao.py
```diff
@@ -0,0 +1,106 @@
...
+# This file is part of the Antares project.
+from abc import ABC, abstractmethod
+from typing import Any, Dict, List
+
+from antarest.study.business.model.area_model import Area, AreaUIUpdate
+
+
+class ReadOnlyAreaDao(ABC):
+    @abstractmethod
+    def get_all_areas(self) -> List[Area]:
```
MartinBelthle: To me this `Area` object should be renamed as it's only used at some places in the code and is super weird

antarest/study/dao/api/study_dao.py
```diff
@@ -10,12 +10,13 @@
 #
 # This file is part of the Antares project.
 from abc import abstractmethod
-from typing import Dict, Sequence
+from typing import Any, Dict, List, Sequence
```
MartinBelthle: Minor: could use `list` and `dict` to avoid importing

antarest/study/dao/memory/in_memory_study_dao.py
```diff
@@ -820,3 +822,62 @@ def get_scenario_by_type(self, scenario_type: ScenarioType) -> AnyScenarios:
     @override
     def save_scenario_builder(self, rulesets: Rulesets) -> None:
         self.rulesets = rulesets
+
+    @override
+    def get_all_areas(self) -> List[Area]:
+        # For in-memory DAO, we only store area names, not full Area objects
```
MartinBelthle: I'm a bit confused cause you didn't store Ui and Properties properly as we did for other objects but you performed some try catch that we never did. Is there a particular reason ?

antarest/study/dao/file/file_study_area_dao.py
```diff
@@ -0,0 +1,601 @@
...
+from antarest.study.storage.rawstudy.model.filesystem.config.thermal import parse_thermal_cluster
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+
+
+class FileStudyAreaDao(AreaDao):
+    @abstractmethod
+    def get_file_study(self) -> FileStudy:
+        pass
+
+    def _get_thermal_clusters(self, area_id: str) -> List[ThermalCluster]:
```
MartinBelthle: We could use the thermal dao instead of this

antarest/study/dao/file/file_study_area_dao.py
```diff
@@ -0,0 +1,596 @@
...
+        """Helper method to build the complete area data structure."""
+        # Import here to avoid circular import
+        from antarest.study.storage.variantstudy.model.command.common import FilteringOptions
+        from antarest.study.storage.variantstudy.model.command.create_area import NodalOptimization
+
+        study_data = self.get_file_study()
+
+        # Generate thermal areas ini
+        thermal_areas_ini = study_data.tree.get(["input", "thermal", "areas"])
+        thermal_areas_ini.setdefault("unserverdenergycost", {})[area_id] = NodalOptimization.UNSERVERDDENERGYCOST
```
MartinBelthle: This NodalOptimization class shouldn't exist we have all we need in the AreaProperties I believe

antarest/study/storage/variantstudy/model/command/create_area.py
```diff
@@ -81,200 +62,12 @@ class CreateArea(ICommand):
     metadata: Dict[str, str] = Field(default_factory=dict, description="Area metadata: country and tag list")
```
MartinBelthle: We should drop this field (means we have to introduce a command version :/)

antarest/study/business/area_management.py
```diff
@@ -77,73 +67,17 @@ def get_all_areas_ui_info(self, study: StudyInterface) -> Dict[str, Any]:
...
-        study.add_commands(commands)
+        command = UpdateLayerAreas(
+            layer_id=layer_id,
+            area_ids=areas,
+            command_context=self._command_context,
+            study_version=study.version,
+        )
+        study.add_commands([command])
 
     def create_area(self, study: StudyInterface, area_creation_info: AreaCreation) -> Area:
```
MartinBelthle: We should remove the AreaCreation object and just use it inside the blueprint if we need no breaking change

antarest/study/storage/variantstudy/model/command/create_area.py
```diff
@@ -81,200 +62,12 @@ class CreateArea(ICommand):
...
-        # of the diagonal (always equal to 1).
-
-        study_data.tree.save(new_area_data)
-
-        return command_succeeded(message=f"Area '{self.area_name}' created")
+    def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        try:
+            study_data.save_area(self.area_name, self.command_context)
+            return command_succeeded(message=f"Area '{self.area_name}' created")
+        except ValueError as e:
```
MartinBelthle: We never perform this try catch, this case is handled in the `add_commands`

antarest/study/storage/variantstudy/model/command/remove_area.py
```diff
@@ -43,202 +37,13 @@ class RemoveArea(ICommand):
...
-        self.remove_from_config(study_data.config)
-
-        new_area_data: JSON = {"input": {"areas": {"list": [area.name for area in study_data.config.areas.values()]}}}
-        study_data.tree.save(new_area_data)
-
-        return command_succeeded(message=f"Area '{self.id}' deleted")
+    def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        try:
+            study_data.delete_area(self.id)
+            return command_succeeded(message=f"Area '{self.id}' deleted")
```
MartinBelthle: Same here. Also minor: you can remove commas in the `to_dto` to make it take less place

antarest/study/storage/variantstudy/model/command/replace_layer_areas.py
```diff
@@ -0,0 +1,58 @@
...
+    layer_id: str
+    area_ids: List[str]
+
+    @override
+    def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener] = None) -> CommandOutput:
+        study_data.save_layer_areas(self.layer_id, self.area_ids)
+        return command_succeeded(message=f"Layer '{self.layer_id}' areas updated")
+
+    @override
+    def to_dto(self) -> CommandDTO:
```
MartinBelthle, line 47: Same minor comment on comma

MartinBelthle (changes requested):
First comments, did not read the entire code

tests/variantstudy/model/command/test_replace_layer_areas.py
```diff
@@ -0,0 +1,210 @@
...
+            command_context=command_context,
+            study_version=empty_study.config.version,
+        )
+        output = update_command1.apply(study_data=empty_study)
+        assert output.status
+
+        # Remove one area from the layer
+        update_command2 = UpdateLayerAreas(
+            layer_id="1",
+            area_ids=["area1", "area3"],  # area2 removed
```
MartinBelthle, line 115: Based on this logic, the command should be named `Replace`

antarest/study/business/area_management.py
TheoPascoli: AreaUI represents data for only the layer 0. 
This method returns every layers
I can introduce a model AreaUIComplete with all layers if you want

antarest/study/business/area_management.py
TheoPascoli: Can you explain why ?

antarest/study/storage/variantstudy/model/command/replace_layer_areas.py
TheoPascoli: ruff puts the comma back

antarest/study/dao/api/area_dao.py
TheoPascoli: I renamed it areainfo. Tell me what you think

tests/variantstudy/model/command/test_create_area.py
```diff
@@ -164,7 +164,25 @@ def test_apply(
...
+                "area_name": "test_area",
+                "metadata": {"country": "FR", "tag": "test"},
+                "command_context": command_context,
+                "study_version": study_version,
+            }
+        )
+
+        # The metadata should be dropped and not stored
+        assert command.area_name == "test_area"
+        assert not hasattr(command, "metadata") or not command.model_dump().get("metadata")
```
MartinBelthle: minor: instead of model_dump we should probably check the `to_dto()`

tests/variantstudy/model/command/test_create_area.py
```diff
@@ -164,7 +164,25 @@ def test_apply(
...
 
         create_area_command: ICommand = CreateArea(
-            area_name=area_name, command_context=command_context, metadata={}, study_version=study_version
+            area_name=area_name, command_context=command_context, study_version=study_version
         )
         output = create_area_command.apply(study_data=empty_study)
         assert not output.status
+
+    def test_backward_compatibility_with_metadata(self, command_context: CommandContext) -> None:
+        """Test that version 1 format (with metadata field) can still be loaded."""
```
MartinBelthle: We usually perform this type of test inside the `test_command_factory.py`

antarest/study/storage/variantstudy/model/command/update_area_ui.py
```diff
@@ -89,31 +89,23 @@ def _validate_parameters(cls, values: dict[str, Any], info: ValidationInfo) -> d
...
-            current_area["layerColor"][self.layer] = f"{r}, {g}, {b}"
-            if layer_int == 0:
-                current_area["ui"]["color_r"] = r
-                current_area["ui"]["color_g"] = g
-                current_area["ui"]["color_b"] = b
-
-        study_data.tree.save(current_area, ["input", "areas", self.area_id, "ui"])
+    def _apply_dao(self, study_data: StudyDao, listener: t.Optional[ICommandListener] = None) -> CommandOutput:
+        # Get current UI info for all areas
+        all_areas_ui = study_data.get_all_areas_ui_info()
```
MartinBelthle: This is not performant at all as we're opening all files just for one area. We should introduce a method just to get the Ui of one area (antares-craft also needs this for the same performance issue)

antarest/study/storage/variantstudy/model/command/update_area_ui.py
```diff
@@ -89,31 +89,23 @@ def _validate_parameters(cls, values: dict[str, Any], info: ValidationInfo) -> d
...
+        current_ui = all_areas_ui.get(self.area_id)
+
+        # If area doesn't exist in UI info, use defaults
```
MartinBelthle: How is this possible ?

antarest/study/storage/variantstudy/model/command/update_area_ui.py
```diff
@@ -89,31 +89,23 @@ def _validate_parameters(cls, values: dict[str, Any], info: ValidationInfo) -> d
...
+        if current_ui is None:
+            current_ui = AreaUI(x=0, y=0, color_rgb=(230, 108, 44))
```
MartinBelthle: This makes me realize that we should put these default values inside the AreaUi class

antarest/study/storage/variantstudy/model/command/update_area_ui.py
```diff
@@ -89,31 +89,23 @@ def _validate_parameters(cls, values: dict[str, Any], info: ValidationInfo) -> d
...
+
+        # Apply the update: merge current values with new values
+        area_ui = AreaUI(
```
MartinBelthle: Introduce a method `update_area_ui` to do the same as other parts of the code ?

antarest/study/storage/variantstudy/model/command/replace_layer_areas.py
MartinBelthle: I have this on my env
```python
return CommandDTO(
    action=CommandName.REPLACE_LAYER_AREAS.value,
    args={"layer_id": self.layer_id, "area_ids": self.area_ids},
    study_version=self.study_version,
)
```

antarest/study/storage/variantstudy/model/command/create_area.py
```diff
@@ -10,44 +10,25 @@
...
-    if unserverdenergycost is not None:
-        new_areas.setdefault("unserverdenergycost", {})[area_id] = unserverdenergycost
-    if spilledenergycost is not None:
-        new_areas.setdefault("spilledenergycost", {})[area_id] = spilledenergycost
-    return new_areas
-
-
+# Constants kept here for backward compatibility with NodalOptimization import
+# These are now used in the DAO layer
 class NodalOptimization:
```
MartinBelthle: You can remove this class now

antarest/study/business/area_management.py
```diff
@@ -50,19 +51,9 @@ def __init__(
         """
         self._command_context = command_context
 
-    def get_all_areas(self, study: StudyInterface) -> List[Area]:
+    def get_all_areas(self, study: StudyInterface) -> List[AreaInfo]:
```
MartinBelthle: Minor: I would rename it `get_all_areas_info`

antarest/study/dao/api/area_dao.py
```diff
@@ -0,0 +1,106 @@
...
+# This file is part of the Antares project.
+from abc import ABC, abstractmethod
+from typing import Any, Dict, List
+
+from antarest.study.business.model.area_model import AreaInfo, AreaUI
+
+
+class ReadOnlyAreaDao(ABC):
+    @abstractmethod
+    def get_all_areas(self) -> List[AreaInfo]:
```
MartinBelthle: Same here, would add info

antarest/study/dao/api/area_dao.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        Retrieve all physical areas of a study.
+
+        Returns:
+            The list of areas with their basic information.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_all_areas_ui_info(self) -> Dict[str, Any]:
```
MartinBelthle: Still think we need a Pydantic model instead of the Any even if it means introducing a Pydantic object just for this

antarest/study/dao/api/area_dao.py
```diff
@@ -0,0 +1,106 @@
...
+class AreaDao(ReadOnlyAreaDao):
+    """
+    DAO for area operations.
+
+    Note: Write operations (create, update, delete) are handled through the DAO,
+    which abstracts the storage implementation (filesystem, database, etc.).
+    """
+
+    @abstractmethod
+    def save_area(self, area_name: str, command_context: Any) -> None:
```
MartinBelthle: I don't like having the command context here even if I see why you need it. Don't see another way around :/

antarest/study/dao/file/file_study_area_dao.py
```diff
@@ -0,0 +1,584 @@
...
+            null_matrix=null_matrix,
+        )
+
+        # Save to filesystem
+        study_data.tree.save(new_area_data)
+
+    def _build_area_data_structure(
+        self,
+        area_id: str,
+        area_name: str,
```
MartinBelthle: Seems you don't need `area_name`

MartinBelthle (changes requested):

antarest/study/dao/api/area_dao.py
TheoPascoli: I'm not sure how

sylvlecl: Indeed the DAO must not know about commands, the dependency is the other way.

If all we need is an object containing the "constant matrices", which is something we could need somewhere else in the DAO, we can add that dependency only to the implementation of the DAO, in its constructor.

antarest/study/business/model/area_model.py
```diff
@@ -61,3 +66,49 @@ class AreaUIUpdate(AntaresBaseModel):
...
+
+class AreaUIData(AntaresBaseModel):
+    """
+    UI data for a single area containing base UI properties and layer-specific data.
+    This represents the complete UI configuration for an area across all layers.
+    """
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    ui: Dict[str, int | str] = Field(
```
sylvlecl, line 96: That model is so ugly, with data for one layer distributed in multiple dicts ...

Ideally, we should have a model which is basically a list of `AreaUI`.
It's a little weird that the DAO exposes the same information under different formats (in AreaUI and in AreaUIData).

We'd better have only one or the other.

However, that's something we could improve later to not postpone too much that PR

antarest/study/storage/variantstudy/model/command/create_area.py
```diff
@@ -67,219 +36,35 @@ class CreateArea(ICommand):
     # ===================
 
     command_name: CommandName = CommandName.CREATE_AREA
+    # version 2: remove unused `metadata` field
```
sylvlecl, line 39: nice

sylvlecl (changes requested):
Looks good overall, we need to change the dependency to command context though!

The different data models for areas seem to have quite alot of overlap too, in particular ui-related ones, it's really tempting to try to refactor them. However we'd probably better not mix it with this PR

sylvlecl (dismissed):

sylvlecl (dismissed):

sylvlecl:
Just pushed a commit to use the new models in the API layer, so that they are reflected in swagger doc

PR #2757, opened 2025-10-08
feat(ui-launcher): add `adequacyCriterions` field

skamril:
ANT-3726


webapp/src/components/App/shared/studies/dialogs/StudyLaunchDialog/Fields.tsx
```diff
@@ -93,6 +93,16 @@ function Fields() {
     return true;
   };
 
+  const validateAdequacyCriterions = (
+    value: FormValues["adequacyCriterions"],
+    { sensitivityMode }: FormValues,
+  ) => {
+    if (value && sensitivityMode) {
+      return "'Reliability criterion' and 'Sensitivity mode' cannot be both enabled.";
```
hdinia: you can use the i18n key here

webapp/public/locales/en/main.json
```diff
@@ -307,6 +307,8 @@
   "launcher.field.otherOptions.error.xpress": "'xpress' option is available from version 8.3",
   "launcher.field.xpress": "Xpress (>= 8.3)",
   "launcher.field.autoUnzip": "Automatically unzip",
+  "launcher.field.adequacyCriterions": "Reliability criterion",
```
sylvlecl: The official wording seems to be "Reliability standard", see for example in https://www.entsoe.eu/eraa/2023/report/ERAA_2023_v2_Executive_Report.pdf

webapp/src/components/App/shared/studies/dialogs/StudyLaunchDialog/Fields.tsx
```diff
@@ -117,6 +127,18 @@ function Fields() {
...
+      setValue("sensitivityMode", false);
+    }
+  };
+
+  const handleSensitivityModeChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (event.target.checked && getValues("adequacyCriterions")) {
+      setValue("adequacyCriterions", false);
+    }
+  };
+
```
hdinia, line 144: This is a bad pattern, those fields can be active 1 at the time, so use only a `disabled` based on each other, or a radio group (better for this use case)


For the disabled pattern, first you must add `isAdequacyCriterionsEnabled` to the watched fields 

Then:

```ts
 {isXpansionEnabled && (
          <>
            <SwitchFE
              label={t("launcher.field.adequacyCriterions")}
              name="adequacyCriterions"
              control={control}
              disabled={isSensitivityModeEnabled}
              sx={{ minWidth: 250 }}
            />
            {isSingleStudy && (
              <>
                <SwitchFE
                  label={t("launcher.field.sensitivityMode")}
                  name="sensitivityMode"
                  control={control}
                  disabled={isAdequacyCriterionsEnabled}
                />
                {isSensitivityModeEnabled && (
                  <>
                    <Fieldset.Break />
                    <SelectFE
                      label={t("global.output")}
                      name="output"
                      options={outputOptions}
                      control={control}
                      rules={{ required: t("form.field.required") }}
                      sx={{ flex: 1 }}
                    />
                  </>
                )}
              </>
            )}
          </>
        )}
```

you can remove the handlers (`handleAdequacyCriterionsChange` and `handleSensitivityModeChange`), and the validation that will no longer be necessary too

webapp/src/components/App/shared/studies/dialogs/StudyLaunchDialog/Fields.tsx
```diff
@@ -168,22 +190,40 @@ function Fields() {
 
       <Fieldset legend={t("launcher.legend.xpansion")}>
         <SwitchFE label={t("global.enable")} name="xpansion" control={control} />
-        {isSingleStudy && (
+        {isXpansionEnabled && (
```
hdinia: I think we can show them but disabled when `xpansion` is false, it's better for the UX to see all the options without having to check things

webapp/src/components/App/shared/studies/dialogs/StudyLaunchDialog/index.tsx
```diff
@@ -41,11 +41,18 @@ function StudyLaunchDialog({ open, onClose, studyIds }: Props) {
...
+      // and then disables Xpansion before submitting the form.
+      xpansion: values.xpansion
+        ? {
+            enabled: true,
+            adequacy_criterions: values.adequacyCriterions,
+            sensitivity_mode: values.sensitivityMode,
+            // `output_id` has to be provided only if `sensitivity_mode` is enabled.
+            output_id: values.sensitivityMode ? values.output : undefined,
+          }
+        : { enabled: false },
```
hdinia, line 56: This is confusing, it should be handled before the submit, on a event handler that reset the fields

You can add this in the Fields component:

```ts
  const handleXpansionChange = (event: React.ChangeEvent<HTMLInputElement>) => {
    if (!event.target.checked) {
      // Clear xpansion-related fields when disabling xpansion
      setValues({
        adequacyCriterions: false,
        sensitivityMode: false,
        output: "",
      });
    }
  }; 
```


Then keep it like this:

```ts

 const handleSubmit = ({ values }: SubmitHandlerPlus<FormValues>) => {
    const options: LaunchOptions = {
      output_suffix: values.name,
      other_options: values.otherOptions,
      auto_unzip: values.autoUnzip,
      xpansion: {
        enabled: values.xpansion,
        adequacy_criterions: values.adequacyCriterions,
        sensitivity_mode: values.sensitivityMode,
        // `output_id` has to be provided only if `sensitivity_mode` is enabled.
        output_id: values.sensitivityMode ? values.output : undefined,
      },
      nb_cpu: values.nbCores,
    };

    return Promise.all(
      studyIds.map((id) => launchStudy(id, options, values.version, values.launcher)),
    );
  };

```

hdinia: Also since it's a major refactor, you should eliminate the snake case from the view (you can change the case at the api methods level)

hdinia (changes requested):

webapp/public/locales/en/main.json
skamril: Ok I take it from the ticket

webapp/src/components/App/shared/studies/dialogs/StudyLaunchDialog/Fields.tsx
skamril: Reported by copilot

webapp/src/components/App/shared/studies/dialogs/StudyLaunchDialog/Fields.tsx
skamril: They want to do like this, and this reduce the size of the dialog.

webapp/src/components/App/shared/studies/dialogs/StudyLaunchDialog/index.tsx
skamril: I want to keep the last value defined if Xpansion is activated again, this is why I am not reseting them.

skamril: I have not refactored the API part, study.ts is legacy code that should be moved to the `api` folder later when there is time.

webapp/src/components/App/shared/studies/dialogs/StudyLaunchDialog/Fields.tsx
skamril: I don't understand why it's a bad pattern, I prefer this one personally.

I am not using Radio Group because the fields have no direct link, they just can't be active at the same time.

hdinia: It's not matter of preferences, fields switching each other is bug-prone, it's the form state that controls fields, not the other way around normally

When you have 2 boolean fields that depend on each other for activation, the most simple here is to use `disabled`  or even better a radio group


React's philosophy: Data flows down, events flow up

```
      Form State
         ↓
    ┌────┴────┐
    ↓         ↓
  Field A   Field B
```

But this pattern creates bidirectional coupling:
```
      Form State
         ↕
    ┌────┼────┐
    ↓    ↕    ↓
  Field A ←→ Field B
```

Fields are reaching sideways to control each other instead of letting the form state be the single source of truth.

hdinia: here with only 2 it can be acceptable, but it's not ideal for maintainability etc if more fields are added later it can become a mess

webapp/src/components/App/shared/studies/dialogs/StudyLaunchDialog/index.tsx
hdinia: ok

hdinia (dismissed):

hdinia (approved):

PR #2759, opened 2025-10-09
refactor(hydro)!: introduce DAO for hydro correlation

MartinBelthle:
**Breaking changes:**
- Remove the parameter `columns` from the GET `/correlation/matrix` endpoint
- Remove the PUT `/correlation/matrix` endpoint

### DIFF
--- a/antarest/study/business/model/hydro_correlation_model.py
+++ b/antarest/study/business/model/hydro_correlation_model.py
@@ -59,6 +59,12 @@ class HydroCorrelationMatrix(AntaresBaseModel, extra="forbid", populate_by_name=
         if self.index != self.columns:
             raise ValueError("correlation matrix must have the same rows and columns")
 
+        if np.any((self.data < -1) | np.any(self.data > 1)):
+            raise ValueError("coefficients must be between -1 and 1")
+
+        if np.any(np.isnan(self.data)):
+            raise ValueError("correlation matrix must not contain NaN coefficients")
+
         n = len(self.index)
         if self.data.shape != (n, n):
             raise ValueError(f"correlation matrix must have shape ({n}×{n})")
--- a/antarest/study/business/model/hydro_correlation_model.py
+++ b/antarest/study/business/model/hydro_correlation_model.py
@@ -94,6 +94,8 @@ class HydroCorrelationMatrix(AntaresBaseModel, extra="forbid", populate_by_name=
     def to_hydro_correlations(self) -> dict[str, HydroCorrelation]:
         correlations_dict = {}
         for k, area_id in enumerate(self.index):
-            correlations = [HydroCorrelationArea(area_id=self.index[k], coefficient=n) for n in self.data[k]]
+            correlations = [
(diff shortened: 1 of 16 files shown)


antarest/study/business/model/hydro_correlation_model.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @model_validator(mode="after")
+    def check_model(self) -> "HydroCorrelationMatrix":
+        if self.data.size == 0:
+            raise ValueError("correlation matrix must not be empty")
+
+        if self.index != self.columns:
+            raise ValueError("correlation matrix must have the same rows and columns")
+
+        if np.any((self.data < -1) | np.any(self.data > 1)):
```
TheoPascoli: second np.any is not necessary

antarest/study/business/model/hydro_correlation_model.py
```diff
@@ -0,0 +1,124 @@
...
+            self.data[index][other_index] = self.data[other_index][index] = correlation.coefficient / 100
+            modified_indexes.add(other_index)
+
+        # Reset existing values that were not modified
+        for other_index in range(len(self.index)):
+            if other_index not in modified_indexes:
+                self.data[index][other_index] = self.data[other_index][index] = 0
+
+        # Round trip to validate the data
+        HydroCorrelationMatrix.from_hydro_correlations(self.to_hydro_correlations())
```
TheoPascoli: Instead of doing this round trip, you could validate the matrix directly instead of rebuilding it. 
You explicitely check the symmetry, the diagonal and the coefficient range

antarest/study/business/model/hydro_correlation_model.py
```diff
@@ -0,0 +1,124 @@
...
+            raise ValueError("correlation matrix must not contain NaN coefficients")
+
+        n = len(self.index)
+        if self.data.shape != (n, n):
+            raise ValueError(f"correlation matrix must have shape ({n}×{n})")
+
+        if not np.array_equal(self.data, self.data.T):
+            raise ValueError("correlation matrix is not symmetric")
+
+        if np.any(np.diag(self.data) != 1):
```
TheoPascoli: I'm not sure != 1 with floats is accurate. idk honestly. there is another method with numpy : 
`if not np.allclose(np.diag(self.data), 1.0):` that would not catch values like 1.000001. Maybe its better

TheoPascoli (changes requested):

antarest/study/business/model/hydro_correlation_model.py
MartinBelthle: Indeed, probably safer, I'll use it

antarest/study/business/model/hydro_correlation_model.py
MartinBelthle: Indeed

antarest/study/business/model/hydro_correlation_model.py
MartinBelthle: Indeed. I introduced a private method that performs the validation and I use it inside the model_validator and here

TheoPascoli (approved):

PR #2763, opened 2025-10-10
feat(launcher): add support for solver presets

smailio:


alembic/versions/d87f410934f1_create_launcher_config_table.py
```diff
@@ -0,0 +1,40 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "launcher_configuration",
+        sa.Column("id", sa.String(length=36), nullable=False),
+        sa.Column("name", sa.String(), nullable=False),
+        sa.Column("linear_solver", sa.String(), nullable=False),
+        sa.Column("min_antares_version", sa.String(), nullable=True),
+        sa.Column("max_antares_version", sa.String(), nullable=True),
+        sa.Column("linear_solver_param_optim_1", sa.JSON(), nullable=True),
+        sa.Column("linear_solver_param_optim_2", sa.JSON(), nullable=True),
```
smailio: I'm not a 100% confident in this JSON type, it works on sqlite. When I want this object from DB I need to parse it. I don't know if it behaves the same with postrgres. What do you think ?

antarest/launcher/model.py
```diff
@@ -27,6 +32,21 @@
...
 
+SolverParams = List[Tuple[str, str]]
+
+ALLOWED_LAUNCHER_CONFIG_PARAM_PATTERN = re.compile(
+    r"^[a-zA-Z0-9_]+$|^\d+\.\d+$"
+)  # alphanumeric, underscore, or decimal number
+MIN_SUPPORTED_VERSION_FOR_OPTIM_PARAMS = StudyVersion(9, 2)
+
+
+class BadLauncherConfigInput(HTTPException):
```
smailio: Not sure about the HTTPException in a model.py file

antarest/launcher/model.py
```diff
@@ -324,3 +344,250 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
         description="The status of the launcher: 'SUCCESS' or 'FAILED'",
         title="Launcher Status",
     )
+
+
+class LauncherConfigDTO(AntaresBaseModel):
+    id: Optional[str] = None
+    name: str
+    linear_solver: str
```
smailio, line 378: should I limit this to some values ? It seems to me it's dynamic (available solvers depends on the instance of the app currently running)

antarest/launcher/model.py
```diff
@@ -324,3 +344,256 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+
+        return self
+
+
+def make_other_options_from_launcher_config(config: LauncherConfigDTO) -> str:
+    """
+    Generate an 'other_options' string from a LauncherConfigDTO. This will be passed to the antares launcher script
+
+    Example output:
+        --solver=xpress --nobasis1 --param-optim1="THREADS 4 PRESOLVE 1" --param-optim2="MIPRELSTOP 0.01"
```
smailio: I'm not sure about this. 

Should it instead be : 
`xpress nobasis1 param-optim1="THREADS 4 PRESOLVE 1" param-optim2="MIPRELSTOP 0.01"`

I read code all over the project and my conclusion so far is that it doesn't matter. I choosed this syntax because it makes it looks like CLI options wich is what this field will be used for.

antarest/launcher/model.py
```diff
@@ -324,3 +344,256 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+            use_optim_1_basis_optim_2=self.use_optim_1_basis_optim_2,
+        )
+
+    @classmethod
+    def from_dto(cls, dto: LauncherConfigDTO) -> "LauncherConfigModel":
+        data = dto.model_dump(exclude_none=True)
+        if dto.min_antares_version is not None:
+            data["min_antares_version"] = f"{dto.min_antares_version:2d}"
+        if dto.max_antares_version is not None:
+            data["max_antares_version"] = f"{dto.max_antares_version:2d}"
```
smailio: I used the `:2d` format because that's what I saw elsewhere in the code but I'm not sure.

antarest/launcher/service.py
```diff
@@ -238,13 +263,24 @@ def run_study(
...
     ) -> str:
         job_uuid = self._generate_new_id()
         logger.info(f"New study launch (study={study_uuid}, job_id={job_uuid})")
         study_info = self.study_service.get_study_information(uuid=study_uuid)
         solver_version = SolverVersion.parse(study_version or study_info.version)
 
+        if launcher_configuration_id is not None:
+            launcher_config = self.get_launcher_config(launcher_configuration_id)
+            if not is_launcher_config_compatible(
+                launcher_config, StudyVersion.parse(study_version or study_info.version)
```
smailio: I'm not sure about whether I should check the compatibility with study_version given as an argument to this endpoint. Should I just check against study_info.version ? After all the user can put whatever version as an input to this endpoint.

alembic/versions/d87f410934f1_create_launcher_config_table.py
sylvlecl: Not sure either:
I know that postrgesql has a JSON type which even allows to execute queries on it, but here I think we'd better keep things simple and juste use a string type, that we will parse/serialize in our application.

antarest/launcher/model.py
sylvlecl: Yes I think it's better to keep it this way indeed.
We'd better not assume what are the solvers available in the installed version of antares-simulator.

antarest/launcher/model.py
sylvlecl: Exactly, unfortunately for now we need to use the syntax in your github comment.
It's an implicit convention ... It has some added value: in the launch script, we adapt the exact syntax for antares-simulator, depending on its version (some options can be renamed, not supported, etc).

In order to have a cleaner version, we would need to make some changes in antares-launcher to have more explicit fields than this "other_options", but that will be out of the scope of that first implementation.

antarest/launcher/model.py
sylvlecl: That's right, it allows to get version formatted as "9.3" for example (not 930, not "9.3.0", etc).
Yes, the syntax is awful :/

antarest/launcher/service.py
sylvlecl: Actually, compatibility should be checked against the `solver_version`, not the study version.
The allowed arguments depend on the version of the solver, not the version of the study.

antarest/launcher/service.py
```diff
@@ -238,13 +263,24 @@ def run_study(
...
+            ):
+                raise IncompatibleLauncherConfig("Launcher configuration is not compatible with study version")
+            if launcher_parameters.other_options:
```
sylvlecl, line 259: I think that in that case, what the user would want is to concatenate with those "other_options". However, It's actually not that easy to implement, I think, because we should do some merging of options, etc ...

I think in a first version we should keep it this way.

antarest/launcher/model.py
```diff
@@ -324,3 +344,256 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+        "linear_solver_param",
+        "linear_solver_param_optim_1",
+        "linear_solver_param_optim_2",
+    )
+    def validate_solver_params(cls, sp: Optional[SolverParams]) -> Optional[SolverParams]:
+        if not sp:
+            return sp
+        for k, v in sp:
+            if not ALLOWED_LAUNCHER_CONFIG_PARAM_PATTERN.match(k):
+                raise BadLauncherConfigInput(
```
sylvlecl: In pydantic validator, we should stick with `ValueError`, see https://docs.pydantic.dev/latest/errors/errors/ 

This will already be converted to a 400.

antarest/launcher/model.py
```diff
@@ -324,3 +344,256 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
         description="The status of the launcher: 'SUCCESS' or 'FAILED'",
         title="Launcher Status",
     )
+
+
+class LauncherConfigDTO(AntaresBaseModel):
+    id: Optional[str] = None
+    name: str
+    linear_solver: str
+    min_antares_version: Optional[StudyVersion] = None
```
sylvlecl: This should be `SolverVersion` instead.

Also we could use an annotated type such as `StudyVersionStr` to handle parsing and serialization from the usual format "9.3".

antarest/launcher/model.py
```diff
@@ -324,3 +344,256 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
         description="The status of the launcher: 'SUCCESS' or 'FAILED'",
         title="Launcher Status",
     )
+
+
+class LauncherConfigDTO(AntaresBaseModel):
```
sylvlecl: Naming:
I think "launch**er** config" is a little awkward, because it's not a configuration of the launcher, it a configuration for a particular "run" or "launch" of the simulation.

antarest/launcher/service.py
```diff
@@ -238,13 +263,24 @@ def run_study(
         study_uuid: str,
         launcher: str,
         launcher_parameters: LauncherParametersDTO,
+        launcher_configuration_id: Optional[str] = None,
```
sylvlecl: I see, only the translation of the config into other options is saved in database, which will avoid any issue related to removed configurations still referenced in database etc, nice.

Maybe at some point we will want to remember which was the selected configuration, but looks ok to me for now.

antarest/launcher/model.py
```diff
@@ -324,3 +344,256 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+            linear_solver_param_optim_1=param_optim_1,
+            linear_solver_param_optim_2=param_optim_2,
+            linear_solver_param=param,
+            use_optim_1_basis_next_week=self.use_optim_1_basis_next_week,
+            use_optim_1_basis_optim_2=self.use_optim_1_basis_optim_2,
+        )
+
+    @classmethod
+    def from_dto(cls, dto: LauncherConfigDTO) -> "LauncherConfigModel":
+        data = dto.model_dump(exclude_none=True)
```
sylvlecl: Why going through a dump, it's easier to use the model directly, isn't it ?

antarest/launcher/model.py
```diff
@@ -324,3 +344,256 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+    launcher_params.other_options = new_other_options
+
+
+class LauncherConfigModel(Base):
+    __tablename__ = "launcher_configuration"
+
+    id = mapped_column(String(36), primary_key=True)
+    name = mapped_column(String)
+    linear_solver = mapped_column(String)
+    min_antares_version = mapped_column(JSON, nullable=True)
```
sylvlecl: For the versions, we should have strings such as "9.2" in database.
Maybe sqlalchemy has some support, like pydantic, to try to parse it automatically to a `SolverVersion`, but that's not necessary.

antarest/launcher/model.py
```diff
@@ -324,3 +344,256 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+    max_antares_version = mapped_column(JSON, nullable=True)
+    linear_solver_param_optim_1 = mapped_column(JSON, nullable=True)
```
sylvlecl: As commented in alembic, it's probably simpler to use a string here that we will parse.

antarest/launcher/model.py
```diff
@@ -324,3 +344,256 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+        if self.max_antares_version is not None:
+            try:
+                max_version = StudyVersion.parse(self.max_antares_version)
+            except Exception as e:
+                raise ValueError(f"Failed to parse max_antares_version '{self.max_antares_version}': {e}") from e
+
+        param_optim_1 = None
+        if self.linear_solver_param_optim_1 is not None:
+            try:
+                param_optim_1 = (
```
sylvlecl: I guess we should just know that we have a string here (and parse it indeed).

sylvlecl (commented):
Thanks: some comments, did not reviewed everything yet.

antarest/launcher/model.py
smailio: Isn't it better to use SolverVersion ? And then serialize it using an explicit format like this `f"config.min_antares_version:2d"` ? Otherwise if the SolverVersion has a patch it would serlaize to `major.minor.patch`

tests/integration/test_integration.py
```diff
@@ -311,6 +312,49 @@ def test_main(client: TestClient, admin_access_token: str) -> None:
...
+    }
+    create_launch_config_res = client.post(
+        "/v1/launcher/configurations",
+        headers={"Authorization": f"Bearer {fred_credentials['access_token']}"},
+        json=create_launch_config_payload,
+    )
+    assert create_launch_config_res.status_code == 200
+    create_launch_config__data = create_launch_config_res.json()
+
+    import time
```
smailio: please help, can't make this test pass without this sleep. Currently trying to debut but It doesn't make any sense for me yet, it's probably something silly if you notice tell me.

alembic/versions/d87f410934f1_create_launcher_config_table.py
```diff
@@ -0,0 +1,52 @@
...
+
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "d87f410934f1"
+down_revision = "d2942741ae68"
```
MartinBelthle: You'll have to check this down_revision as yesterday it changed and Theo's PR intorduced a migration too.

antarest/launcher/main.py
```diff
@@ -39,15 +39,17 @@ def build_launcher(
     service_launcher: Optional[LauncherService] = None,
 ) -> Optional[LauncherService]:
     if not service_launcher:
-        repository = JobResultRepository()
+        job_repository = JobResultRepository()
+        launcher_config_repository = LaunchConfigRepository()
```
MartinBelthle: Minor: I would rename it `LauncherConfigRepository`. Same for the other `Launchxxx` class. Probably ask for someone else opinion as it's a bit long to rename everything

antarest/launcher/web.py
```diff
@@ -182,4 +186,53 @@ def get_solver_versions(solver: Optional[str] = None) -> List[str]:
         logger.info(f"Fetching the list of solver versions for the '{solver}' configuration")
         return service.get_solver_versions(solver)
 
+    @bp.post(
+        "/configurations",
+        tags=[APITag.launcher],
+        summary="Create a new launcher configuration",
+        response_model=LaunchConfigDTO,
```
MartinBelthle: Minor: no need for this `response_model` as you put it as the method response

antarest/launcher/service.py
```diff
@@ -654,3 +687,70 @@ def get_launch_progress(self, job_id: str) -> float:
...
+            return [config.to_dto() for config in configs]
+
+    def update_launcher_config(
+        self, configuration_id: str, launch_config_update: LaunchConfigUpdate
+    ) -> LaunchConfigDTO:
+        """
+        Update an existing launcher configuration using LauncherParametersDTO.
+        """
+        user = get_current_user()
+        if not user or not user.is_site_admin():
```
MartinBelthle: Minor: You could write: `user = require_current_user()` to simplify a bit

antarest/launcher/service.py
```diff
@@ -654,3 +687,70 @@ def get_launch_progress(self, job_id: str) -> float:
...
+            if not config:
+                raise ValueError("Failed to update launcher configuration")
+            return config.to_dto()
+
+    def delete_launch_config(self, configuration_id: str) -> None:
+        """
+        Delete a launcher configuration by its ID. Only site administrators can delete configs.
+        """
+        user = get_current_user()
+        if not user or not user.is_site_admin():
```
MartinBelthle: Same remark

antarest/launcher/model.py
```diff
@@ -27,6 +30,18 @@
 from antarest.core.serde.json import from_json
 from antarest.login.model import Identity, UserInfo
 
+SolverParams = List[Tuple[str, str]]
+
+ALLOWED_LAUNCHER_CONFIG_PARAM_PATTERN = re.compile(
+    r"^[a-zA-Z0-9_]+$|^\d+\.\d+$"
+)  # alphanumeric, underscore, or decimal number
+MIN_SUPPORTED_VERSION_FOR_OPTIM_PARAMS = StudyVersion(9, 2)
```
MartinBelthle: Minor: You could use `STUDY_VERSION_9_2`

antarest/launcher/model.py
```diff
@@ -27,6 +30,18 @@
...
+
+
+class BadLaunchConfigInput(ValueError):
```
MartinBelthle: You can put this inside exception.py (same goes for other exceptions you introduced)
Also it seems you always call these exceptions with a mesage so you probably don't need to introduce one here

antarest/launcher/model.py
```diff
@@ -324,3 +339,255 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+    @field_validator(
+        "linear_solver_param",
+        "linear_solver_param_optim_1",
+        "linear_solver_param_optim_2",
+    )
+    def validate_solver_params(cls, sp: Optional[SolverParams]) -> Optional[SolverParams]:
+        if not sp:
+            return sp
+        for k, v in sp:
+            if not ALLOWED_LAUNCHER_CONFIG_PARAM_PATTERN.match(k):
```
MartinBelthle: You could probably use a BeforeValidator and a AnnotatedType for `SolverParams` to simplify this a bit

antarest/launcher/model.py
```diff
@@ -324,3 +339,255 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+            f" use_optim_1_basis_next_week={self.use_optim_1_basis_next_week!r},"
+            f" use_optim_1_basis_optim_2={self.use_optim_1_basis_optim_2!r})>"
+        )
+
+
+def apply_update_launcher_config(
+    launcher_config: LaunchConfigModel,
+    launcher_config_update: LaunchConfigUpdate,
+) -> LaunchConfigModel:
+    return LaunchConfigModel.from_dto(
```
MartinBelthle: This is really hard to understand. You shoul declare intermediate vairables IMO

antarest/launcher/repository.py
```diff
@@ -89,3 +89,45 @@ def delete_by_study_id(self, study_id: str) -> None:
         stmt = delete(JobResult).where(JobResult.study_id == study_id)
         db.session.execute(stmt)
         db.session.commit()
+
+
+class LaunchConfigRepository:
+    def __init__(self) -> None:
```
MartinBelthle: is this needed, what happens if we remove the `__init__ ` ?

tests/integration/test_integration.py
```diff
@@ -311,6 +313,59 @@ def test_main(client: TestClient, admin_access_token: str) -> None:
         "owner": {"id": fred_id, "name": "Fred"},
     }
 
+    # wait for job completion, otherwise next tests may interfere as we can't have multiple runs on the same study
```
MartinBelthle: To me you should put this inside the new test you've created, this file is already large enough.
And you won't have to do this weird thing at the beginning of the test

MartinBelthle (changes requested):
Nice, just a few remarks

antarest/launcher/web.py
```diff
@@ -182,4 +186,49 @@ def get_solver_versions(solver: Optional[str] = None) -> List[str]:
         logger.info(f"Fetching the list of solver versions for the '{solver}' configuration")
         return service.get_solver_versions(solver)
 
+    @bp.post(
+        "/solver-presets",
+        tags=[APITag.launcher],
+        summary="Create new solver presets",
+    )
+    def create_solver_presets(solver_presets_creation: SolverPresetsDTO) -> SolverPresetsDTO:
```
MartinBelthle: To me we should also have a `SolverPresetCreation` to avoid having to check when creating a preset if you gave an id, check that id doesn't exist. To me the creation model shouldn't hold an id, we should create it ourselves and this way avoid this issue

antarest/launcher/repository.py
```diff
@@ -89,3 +89,42 @@ def delete_by_study_id(self, study_id: str) -> None:
...
+        return db.session.get(SolverPresetsModel, id)
+
+    def get_all(self) -> List[SolverPresetsModel]:
+        logger.debug("Retrieving all SolverPresetsModel")
+        stmt = select(SolverPresetsModel)
+        return list(db.session.scalars(stmt).all())
+
+    def delete(self, id: str) -> None:
+        logger.debug(f"Deleting SolverPresetsModel {id}")
+        config = db.session.get(SolverPresetsModel, id)
```
MartinBelthle: This performs 2 DB calls ? Shouldn't we do something like 
```
stmt = delete(SolverPresetsModel).where(SolverPresetsModel.id.in_([id]))
session.execute(stmt)
session.commit()
```

antarest/launcher/repository.py
```diff
@@ -89,3 +89,42 @@ def delete_by_study_id(self, study_id: str) -> None:
         stmt = delete(JobResult).where(JobResult.study_id == study_id)
         db.session.execute(stmt)
         db.session.commit()
+
+
+class SolverPresetsRepository:
+    def exists(self, id: str) -> bool:
```
MartinBelthle: No need for this if you create specific classes

antarest/launcher/model.py
```diff
@@ -324,3 +335,260 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+    linear_solver: str
+    min_antares_version: Optional[SolverVersion] = None
+    max_antares_version: Optional[SolverVersion] = None
+    linear_solver_param_optim_1: Optional[SolverParams] = None
+    linear_solver_param_optim_2: Optional[SolverParams] = None
+    linear_solver_param: Optional[SolverParams] = None
+    use_optim_1_basis_next_week: bool = True
+    use_optim_1_basis_optim_2: bool = True
+
+    @field_validator("name")
```
MartinBelthle: Seems you could type name as ItemName instead of having this field validator

MartinBelthle (changes requested):

antarest/launcher/web.py
```diff
@@ -182,4 +187,49 @@ def get_solver_versions(solver: Optional[str] = None) -> List[str]:
...
+        "/solver-presets/{solver_presets_id}",
+        tags=[APITag.launcher],
+        summary="Retrieve solver presets by ID",
+    )
+    def get_solver_presets(solver_presets_id: str) -> SolverPresetsDTO:
+        logger.info(f"Retrieving solver presets for ID {solver_presets_id}")
+        return service.get_solver_presets(solver_presets_id)
+
+    @bp.get(
+        "/solver-presets/",
```
sylvlecl: we should remove the trailing slash

antarest/launcher/model.py
```diff
@@ -324,3 +336,266 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+        if combined2:
+            options.append(f'param-optim2="{combined2}"')
+
+        return " ".join(options)
+
+
+class SolverPresetsCreation(AntaresBaseModel):
+    name: ItemName
+    linear_solver: str
+    min_antares_version: Optional[SolverVersion] = None
```
sylvlecl: We must expect the version as a string in the API --> should use an annotated class such `StudyVersionStr` but for the solver, with only 2 digits.

Also we should take car that the JSON schema and example appears correcly in the swagger doc.
See for now the example ooks like this: 
```json
{
  "name": "xpress + presolve",
  "linear_solver": "xpress",
  "min_antares_version": {
    "major": 0,
    "minor": 0,
    "patch": 0
  },
 ...
}
```

Instead we should have `"min_antares_version": "9.2"` for example.

I think that we can achieve this by using an appropriate annotation on the `SolverVersionStr` if we create it

antarest/core/exceptions.py
```diff
@@ -849,6 +849,11 @@ def __init__(self, output_id: str, message: str) -> None:
         super().__init__(HTTPStatus.UNPROCESSABLE_ENTITY, msg)
 
 
+class BadSolverPresetsInput(ValueError):
```
MartinBelthle: You don't use this exception anymore

antarest/launcher/model.py
```diff
@@ -324,3 +336,266 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
         description="The status of the launcher: 'SUCCESS' or 'FAILED'",
         title="Launcher Status",
     )
+
+
+class SolverPresetsDTO(AntaresBaseModel):
```
MartinBelthle: To me this should be called `SolverPresets` as you didn't put `DTO` at the end of the creation and update model

antarest/launcher/model.py
```diff
@@ -11,21 +11,33 @@
...
 from typing_extensions import override
 
 from antarest.core.persistence import Base
 from antarest.core.serde import AntaresBaseModel
 from antarest.core.serde.json import from_json
 from antarest.login.model import Identity, UserInfo
+from antarest.study.model import STUDY_VERSION_9_2
+from antarest.study.storage.rawstudy.model.filesystem.config.validation import ItemName
+
+SolverParams = List[Tuple[str, str]]
```
sylvlecl: From the API point of view, it seems a little weird to use a tuple here, because in JSON it translates to a list, therefore it looks like we could have any number of items:

```json
  "linear_solver_param_optim_1": [
    [ "PRESOLVE", "1"]
  ],
```

It would feel more natural to have either a json mapping, or a list of items:
```json
  "linear_solver_param_optim_1": {
     "PRESOLVE": "1"
  },
```
or 
```json
  "linear_solver_param_optim_1": [
     {"key": "PRESOLVE", "value": "1"}
  ],

File: antarest/launcher/service.py [hunk 37]
```diff
@@ -654,3 +687,71 @@ def get_launch_progress(self, job_id: str) -> float:
...
+
+        with db():
+            solver_presets = self.solver_presets_repository.get(configuration_id)
+            if not solver_presets:
+                raise SolverPresetsNotFound(configuration_id)
+            # Update only the fields that are provided in the update DTO
+            updated_solver_presets = apply_update_solver_presets(solver_presets, solver_presets_update)
+            config = self.solver_presets_repository.save(updated_solver_presets)
+            if not config:
+                raise ValueError("Failed to update solver presets")
```
MartinBelthle: Minor but I don't think we need this check, you didn't performed it line 701

MartinBelthle (changes requested):
Last round of review for me, after this we're good to go :)

antarest/launcher/model.py
```diff
@@ -324,3 +336,266 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+    min_antares_version: Optional[SolverVersion] = None
+    max_antares_version: Optional[SolverVersion] = None
+    linear_solver_param_optim_1: Optional[SolverParams] = None
+    linear_solver_param_optim_2: Optional[SolverParams] = None
+    linear_solver_param: Optional[SolverParams] = None
+    use_optim_1_basis_next_week: Optional[bool] = None
+    use_optim_1_basis_optim_2: Optional[bool] = None
+
+
+def overwrite_params_other_options_with_config(
```
MartinBelthle: Also you don't use this function

antarest/launcher/model.py
```diff
@@ -324,3 +336,266 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
+            raise ValueError(
+                "linear_solver_param_optim_2 is not supported before Antares version 9.2 "
+                f"(got {self.min_antares_version})"
+            )
+
+        return self
+
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def other_options(self) -> str:
```
sylvlecl: I am a little afraid of exposing it as a field of that class:

- it's quite nice because it will allow for example to display in the web app the equivalent options as they would have been defined in the "other options" field
- on the other side:
  - it's likely that at some point we'll change the way we pass those options to the launcher (in a more structured way) --> at that time we'll need to make a breaking change here
  - it could be not clear, in the API, that it's only an alternative representation of other fields: in the swagger doc, the docstring appears which provides some information though

sylvlecl (changes requested):
A very few comments: we need to change the solver version format in the API, other ones are not blockers :)

MartinBelthle (commented):
Good for me, I'll let Sylvain continue as he made a review

antarest/launcher/model.py
```diff
@@ -324,3 +352,310 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
...
         title="Launcher Status",
     )
+
+
+class SolverPresets(AntaresBaseModel):
+    class Config:
+        extra = "forbid"
+
+        @staticmethod
+        def json_schema_extra(schema: MutableMapping[str, Any]) -> None:
```
sylvlecl, line 362: :heart:

antarest/launcher/service.py
```diff
@@ -654,3 +687,67 @@ def get_launch_progress(self, job_id: str) -> float:
             "progress": 0
         }
         return launch_progress_json.get("progress", 0)
+
+    def create_solver_presets(self, solver_presets_creation: SolverPresetsCreation) -> SolverPresets:
+        """
+        Create a new solver presets.
+        """
+        with db():
```
sylvlecl: In general we don't need to require a session this way in services: each HTTP request already gets a DB session through the DB middleware.
Here, this will create a second session different from the one created by the middleware.

We only need to explicitly require a DB session in methods that are executed outside of requests, such as in tasks etc.

sylvlecl (approved):

PR #2770, opened 2025-10-13
feat(output): add endpoint to retrieve metadata on available columns

MartinBelthle:
Benchmark performed on my local environment on a real BP study:
- 566 files to open (212 mc-ind and 354 mc-all) 
- Took around 9 seconds for the first call and less than 0.1 second for other calls once the info was in the DB.


antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,132 @@
...
+
+                        variables[mc_root_key]["areas"].append(areas_dict)
+
+                # Links
+                links_folder = mc_path / "links"
+                if links_folder.exists():
+                    for link_name in sorted(links_folder.iterdir()):
+                        area1, area2 = link_name.name.split("-")
+                        links_dict: dict[str, Any] = {"area_1_name": area1, "area_2_name": area2}
+                        parent_path = areas_folder / link_name
```
TheoPascoli: links_folder instead of area_folder ?

TheoPascoli (changes requested):
Hello, I just read the method get_variables_metadata, I have a few points:

  Main concerns:
  - 6 indentation levels is too much in my opinion, makes it hard to follow the logic
  - Links and areas treatment are almost identical, this should be refactored to avoid duplication
  - The conditions and mappings are hard to follow

The logic is correct overall, but I found it a bit hard to follow

antarest/study/business/output_variables_management.py
MartinBelthle: Indeed, don't understand why it didn't break my test

MartinBelthle:
I tried to put more code in common, reducing code complexity from 52 to 35. It's still a bit hard to read because the algorithm is quite complex. I'm open to other code propositions to simply the code further

antarest/study/storage/output_service.py
```diff
@@ -545,3 +549,36 @@ def aggregate_output_task(notifier: ITaskNotifier) -> TaskResult:
...
+        # Fetches the data inside the FS
+        output_path = self._storage.get_output_path(study, output_id)
+        model = self._output_variables_manager.get_variables_metadata(output_path)
+
+        # Save the model inside DB for next calls
+        with db():
+            db_model = OutputVariables(
+                study_id=study_id,
+                output_id=output_id,
+                variables_metadata_version=1,
```
TheoPascoli: This version number should not be hardcoded in my opinion, you can extract it as a constant. And also use this constant as a default value in the model

antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,133 @@
...
+    MCAllLinksQueryFile,
+    MCIndAreasQueryFile,
+    MCIndLinksQueryFile,
+    MCRoot,
+    QueryFileType,
+)
+from antarest.study.storage.output_model import OutputVariablesMetadata
+
+
+class OutputVariablesManager:
```
TheoPascoli: This whole class could benefit from more documentation, docstring, and explaination about the parsing logic

antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,133 @@
...
+            else:
+                name_to_consider = " ".join([col[0], col[2]])
+            new_cols.append(name_to_consider.upper().strip())
+
+        return {c: set("0") for c in new_cols}
+
+    def _get_filtered_files(
+        self, mc_root: MCRoot, parent_path: Path, file_type_class: type[QueryFileType]
+    ) -> Iterator[tuple[dict[str, set[str]], QueryFileType]]:
+        all_files = [d.name for d in parent_path.iterdir()]
```
TheoPascoli: iterdir() can fail if the path is not a directory or inaccessible. Maybe you can add an error check

TheoPascoli (changes requested):
Looks good, there is still some duplication for areas and links but its fine.
In our documentation, there is a mention of compression for variables metadata, is it something that we want to do now ? Maybe we could add some JSON compression with gzip

antarest/study/business/output_variables_management.py
MartinBelthle: Even if it can indeed fail generally, we're only calling this method when we're sure the path exists and is a folder

antarest/study/storage/output_service.py
MartinBelthle: I moved this code inside a method of the DB class, tell me what you think about it.
Also, I prefer to type the version as `non_nullable` to avoid future weird behaviors

antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,174 @@
...
+    MCIndAreasQueryFile,
+    MCIndLinksQueryFile,
+    MCRoot,
+    QueryFileType,
+)
+from antarest.study.storage.output_model import OutputVariablesMetadata
+from antarest.study.storage.rawstudy.model.filesystem.matrix.matrix import MatrixFrequency
+
+
+class OutputVariablesManager:
```
sylvlecl: We could have only free functions here, not a class.
It's useful to have a class when you have some state to encapsulate, but it's not the case here

antarest/study/storage/output_model.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class OutputVariables(Base):
+    __tablename__ = "output_variables"
+    __table_args__ = (PrimaryKeyConstraint("study_id", "output_id"),)
+
+    study_id: Mapped[str] = mapped_column(String(36), ForeignKey("study.id"), nullable=False)
+    output_id: Mapped[str] = mapped_column(String, nullable=False)
+    variables_metadata_version: Mapped[int] = mapped_column(Integer, nullable=False)
+    variables_metadata: Mapped[str] = mapped_column(String, nullable=False)
```
sylvlecl: We should store it compressed, it will be quite large for large studies, with a lot of repetitions.

(another possibility would be to change slightly the model to detect common columns and not repeat them, but not sure it's a better approach )

antarest/study/storage/output_service.py
```diff
@@ -69,6 +72,7 @@ def __init__(
         self._task_service = task_service
         self._file_transfer_manager = file_transfer_manager
         self._event_bus = event_bus
+        self._output_variables_manager = OutputVariablesManager()
```
sylvlecl: See other comment, we don't need this object, we can call the function straight

antarest/study/business/output_variables_management.py
sylvlecl, line 1: I think we could start having a dedicated package for outputs-related features, which are really a separate object, compared to input data.
This will be only one module of this package.

We can also move the "aggregator" inside it.

antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,174 @@
...
+        Args:
+            - file_path: path of the output file
+            - mc_root: mc-ind or mc-all
+            - freq: MatrixFrequency of the given file.
+            - file_type: query file type class.
+
+        Returns:
+            - A mapping containing the dataframe headers
+        """
+        body = AggregatorManager.parse_output_file(file_path, freq, 0)
```
sylvlecl: it's not meaningful to have it as a method of the aggregator "mamanger", it shuld be a free function in another module

antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,174 @@
...
+            else:
+                file_type1, file_type2, freq = splitted_file_name
+                file_type = f"{file_type1}-{file_type2}"
+            file_dict[file_type_class(file_type)] = MatrixFrequency(freq)
+        return file_dict
+
+    @staticmethod
+    def _read_headers_only(
+        file_path: Path, mc_root: MCRoot, freq: MatrixFrequency, file_type: QueryFileType
+    ) -> dict[str, set[str]]:
```
sylvlecl: Wed'better have a dataclass for the return type, to clarify what it is, for example:
```python
@dataclass(frozen=True)
class ColumnHeader:
    name: str
    sub_columns_names: list[str]
```
and have a `list[ColumnHeader]` as the return type

antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,174 @@
...
+            "details": "thermal_clusters",
+            "details-res": "renewable_clusters",
+            "details-STstorage": "short_term_storages",
+            "values": "variables",
+            "id": "variables",
+        }
+
+        existing_paths = []
+        if mc_ind_path.exists():
+            first_mc_year = next(mc_ind_path.iterdir()).name
```
sylvlecl: we'd better ensure by sorting or by seraching for an explicit name that we always parse the same folder. It may be safer also to match the pattern for years

Question: are we sure that there is no folder for years that are excluded from the playlist ?

antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        Returns the headers of all output files located in the parent_path.
+        For each header, also returns it corresponding file type.
+        """
+        filtered_files = self._filter_files_with_same_prefix(parent_path, file_type_class)
+        for file_type, freq in filtered_files.items():
+            file_path = parent_path / f"{file_type}-{freq.value}.txt"
+            yield self._read_headers_only(file_path, mc_root, freq, file_type), file_type
+
+    def get_variables_metadata(self, output_path: Path) -> OutputVariablesMetadata:
```
sylvlecl: maybe `extract_variables_metadata` to make it clear it's not a plain getter

antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,174 @@
...
+            for col in body.columns:
+                cols_mapping.setdefault(col[0], set()).add(col[1])
+            return cols_mapping
+
+        new_cols: dict[str, set[str]] = {}
+        for col in body.columns:
+            if mc_root == MCRoot.MC_IND:
+                name_to_consider = col[0] if file_type.value == "values" else " ".join(col)
+            else:
+                name_to_consider = " ".join([col[0], col[2]])
```
sylvlecl: I don't think we should concatenated it here, there is a loss of information.
We'd better keep the statistic name separate in the model, for the mc-all part.

Maybe, in a first implementation, we can just keep the variable name and not the list of statistics for that variable ?

(Also `name_to_consider` really does not bring useful information to the reader)

antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,174 @@
...
+        if "details" in file_type.value:
+            cols_mapping: dict[str, set[str]] = {}
+            for col in body.columns:
+                cols_mapping.setdefault(col[0], set()).add(col[1])
+            return cols_mapping
+
+        new_cols: dict[str, set[str]] = {}
+        for col in body.columns:
+            if mc_root == MCRoot.MC_IND:
+                name_to_consider = col[0] if file_type.value == "values" else " ".join(col)
```
sylvlecl: Not sure in which case we join here:
the "details" case is already handled above ?

antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        body = AggregatorManager.parse_output_file(file_path, freq, 0)
+
+        if "details" in file_type.value:
+            cols_mapping: dict[str, set[str]] = {}
+            for col in body.columns:
+                cols_mapping.setdefault(col[0], set()).add(col[1])
+            return cols_mapping
+
+        new_cols: dict[str, set[str]] = {}
```
sylvlecl: why "new_cols" ?

antarest/study/business/output_variables_management.py
```diff
@@ -0,0 +1,174 @@
...
+        for col in body.columns:
+            if mc_root == MCRoot.MC_IND:
+                name_to_consider = col[0] if file_type.value == "values" else " ".join(col)
+            else:
+                name_to_consider = " ".join([col[0], col[2]])
+            new_cols[name_to_consider.upper().strip()] = set()
```
sylvlecl: Why do we apply upper case here but not for the details section ?

antarest/study/storage/output_model.py
```diff
@@ -0,0 +1,83 @@
...
+    Integer,
+    PrimaryKeyConstraint,
+    String,
+)
+from sqlalchemy.orm import Mapped, mapped_column
+
+from antarest.core.persistence import Base
+from antarest.core.serde import AntaresBaseModel
+
+Variables: TypeAlias = Annotated[list[str], BeforeValidator(lambda x: sorted(x))]
```
sylvlecl, line 29: Let's keep it this way, but note that it's not obvious that we want to sort here:
if in output data the variables are sorted by relevance for the user, it could be better to keep that sort order.

antarest/study/storage/output_model.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class AreaClusterVariables(AntaresBaseModel, extra="forbid", populate_by_name=True, alias_generator=to_camel):
+    name: str
+    variables: Variables = []
+
+
+ClusterVariables: TypeAlias = Annotated[
```
sylvlecl, line 37: cluster is not a perfect name, storages are not clusters in particular

sylvlecl (changes requested):
Looks good overall, still some comments and questions

antarest/study/business/output_variables_management.py
MartinBelthle: Tested with a 9.3 launcher, it didn't generate a folder for a year deactivated in the playlist
Also I added the sort even if I don't think it's necessary

antarest/study/business/output_variables_management.py
MartinBelthle: Good question. This code comes from the aggregation manager where we have the same issue, when asking for values we have  `NP COST` and when asking for details the col name is `NP Cost` ...

antarest/study/business/output_variables_management.py
MartinBelthle: Indeed I don't think we can enter the `else`

antarest/study/business/output_variables_management.py
MartinBelthle: As discussed we'll keep it like this for the moment even if it's not ideal

antarest/study/storage/output_model.py
MartinBelthle: I agree but I didn't find a more suitable name :/

sylvlecl (commented):
Approved, but let's rename to variables-list as discussed with the team, sorry for the late change once more.

**MB**: Will do, no worry it's a very little change

PR #2772, opened 2025-10-14
feat(CR23):  RTE-i Implement CR23 

dusanparipovic:
# Summary – Multiple Maximum Generation and Pumping for hydro 

This change request introduces new matrices and frontend logic to enhance hydro configuration management.

---

### Added Four New Matrix Files
- `input/hydro/series/{areaId}/maxHourlyGenPower`  
- `input/hydro/series/{areaId}/maxHourlyPumpPower`  
- `input/hydro/common/capacity/maxDailyGenEnergy_{areaId}`  
- `input/hydro/common/capacity/maxDailyPumpEnergy_{areaId}`  

### Default Values
- **maxDailyGenEnergy** and **maxDailyPumpEnergy**: matrices of size **365×1**, default value **24**  
- **maxHourlyGenPower** and **maxHourlyPumpPower**: initialized as **empty matrices**

### Frontend Update
- Parameters are conditionally displayed when the user enables the related feature via **Advanced Parameters**.
## Note 
- This PR uses custom made antares-study-version library [Link](https://github.com/AntaresSimulatorTeam/antares-study-version/pull/51)


webapp/public/locales/en/main.json
```diff
@@ -504,6 +494,8 @@
   "study.configuration.advancedParameters.spatialTimeSeriesCorrelation": "Spatial time-series correlation",
   "study.configuration.advancedParameters.otherPreferences": "Other preferences",
   "study.configuration.advancedParameters.windTimeSeriesGeneration": "Wind time-series generation",
+  "study.configuration.advancedParameters.compatibility": "Compatibility",
```
skamril, line 497: Missing french translations

skamril (changes requested):

webapp/src/components/App/Singlestudy/explore/Configuration/AdvancedParameters/utils.ts
```diff
@@ -119,6 +125,8 @@ export interface AdvancedParamsFormFields {
   seedUnsuppliedEnergyCosts: number;
   sheddingPolicy: string;
   unitCommitmentMode: string;
+  hydroPmax: string;
```
skamril:
```typescript
// Since v9.2
hydroPmax?: `${HydroPMax}`;
```

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Hydro/index.tsx
```diff
@@ -12,21 +12,28 @@
...
 import TabWrapper from "../../../TabWrapper";
+import { getAdvancedParamsFormFields } from "../../../Configuration/AdvancedParameters/utils";
 
 function Hydro() {
   const { study } = useOutletContext<{ study: StudyMetadata }>();
   const areaId = useAppSelector(getCurrentAreaId);
   const studyVersion = Number(study.version);
+  const [hydroPmax, setHydroPmax] = useState<string>("");
+
+  useEffect(() => {
```
skamril:
```typescript
const { data: hydroPmax = "" } = usePromise(async () => {
  const values = await getAdvancedParamsFormFields(study.id);
  return values.hydroPmax;
}, [study.id])
```

`useState` can be removed.

skamril (changes requested):

webapp/public/locales/en/main.json
dusanparipovic: Added french translation

webapp/src/components/App/Singlestudy/explore/Modelization/Areas/Hydro/index.tsx
dusanparipovic: Removed. Learned from this,thank you!

webapp/src/components/App/Singlestudy/explore/Configuration/AdvancedParameters/utils.ts
dusanparipovic: Refactored.

MartinBelthle:
I'll explain the app behavior we want and if there are unclear things or things you see in a different way, you can answer in this PR.

We would like to support both hydro formats, your one and the existing one as the Simulator is capable of handling both cases.
So the PR concerning the study upgrade should not be necessary.

There would be a view inside the front-end with a toggle to choose which hydro format you want your study to be in.
At study creation, it will always be the current one but you could use this view to change this.

When using this toggle, it will trigger a command in the back that will change the hydro modelization for all areas in the study.
This command will have a listener to notify the front of the modification advancement and the front will freeze the view as long as the modification is not done. This is really similar to what's done today for the thermal timeseries generation.

dusanparipovic:
> I'll explain the app behavior we want and if there are unclear things or things you see in a different way, you can answer in this PR.
> 
> We would like to support both hydro formats, your one and the existing one as the Simulator is capable of handling both cases. So the PR concerning the study upgrade should not be necessary.
> 
> There would be a view inside the front-end with a toggle to choose which hydro format you want your study to be in. At study creation, it will always be the current one but you could use this view to change this.
> 
> When using this toggle, it will trigger a command in the back that will change the hydro modelization for all areas in the study. This command will have a listener to notify the front of the modification advancement and the front will freeze the view as long as the modification is not done. This is really similar to what's done today for the thermal timeseries generation.

Hello Martin, thanks for the review.
I have a couple of points I’d like to clarify with you to make sure I’ve correctly understood your comments before digging into the refactoring. I would also suggest a short 15–30 min call to align, if you’re available this week.
If I understand correctly, the front-end toggle that determines which hydro format is used would trigger the following back-end actions:
Take daily data from

- `input/hydro/common/capacity/maxpower_[area]`

Generate the hourly-format files:

- `hydro/series/[area]/maxHourlyGenPower.txt`
- `hydro/series/[area]/maxHourlyPumpPower.txt`
- `input/hydro/common/capacity/maxDailyGenEnergy_[area].txt`
- `input/hydro/common/capacity/maxDailyPumpEnergy_[area].txt`

These additional files (fully supported by Antares Simulator starting from v9.2) would not be pre-generated by default in the create_area function.
On the front-end side, things are clear to me: we simply display one format or the other depending on the selected option.
Of couse we must enable user to input the hourly data directly via GUI, that will be written back to appropriate txt file.
The main questions I have are:

1. What should happen if the user switches the toggle back to daily?

- Do we delete all the hourly-related files?

2. If the user then switches back to hourly, do we regenerate them again?
3. How should we handle the case where a user imports a study created locally with hourly hydro files into AntaresWeb?

- In the current state, this could lead to errors—what would be the expected behavior here?

Could you please confirm whether I understood the intended actions correctly, and how you see these cases being handled?
As mentioned, a short call might be the quickest way to clarify this.

MartinBelthle:
Yes if the user switches from a format to another, each time we will regenerate the files we need.

If a user creates locally a study with hourly hydro files, we should accept it at the import (currently we raise, you should remove this function in your development)

Not sure to understand what could lead to errors. If the matrix don't match the flag daily/hourly ? If you're talking about that, we could change the current `checks_antares_web_compatibility` to raise in such case

dusanparipovic:
> Yes if the user switches from a format to another, each time we will regenerate the files we need.
> 
> If a user creates locally a study with hourly hydro files, we should accept it at the import (currently we raise, you should remove this function in your development)
> 
> Not sure to understand what could lead to errors. If the matrix don't match the flag daily/hourly ? If you're talking about that, we could change the current `checks_antares_web_compatibility` to raise in such case



> Yes if the user switches from a format to another, each time we will regenerate the files we need.
> 
> If a user creates locally a study with hourly hydro files, we should accept it at the import (currently we raise, you should remove this function in your development)
> 
> Not sure to understand what could lead to errors. If the matrix don't match the flag daily/hourly ? If you're talking about that, we could change the current `checks_antares_web_compatibility` to raise in such case

Thanks for the clarification, that helps a lot.
I have one more scenario I’d like to validate.
Imagine a study with **100 areas**. The user switches the toggle to hourly, the hourly files are generated, and everything is fine. Later, the user realises that one area is missing and adds a new area.
By default, this new area will **not have the hourly hydro files**, since they are not generated in create_area.
In that case, the only workaround today would be for the user to:

- switch back to **daily** (which deletes all hourly data, potentially after a lot of preparation)
- then switch again to **hourly** to regenerate everything.

This feels a bit risky from a **user-experience** perspective, especially if the hourly data required significant effort to prepare.
Related to that, I’m also slightly concerned about **performance**: regenerating all hydro data back and forth for large studies could be quite time-consuming.
Would it make sense to:

- have `create_area` be aware of the current hydro format and generate the appropriate files directly?

I’d be interested to hear your thoughts on how you see this handled.

MartinBelthle:
Exactly, the `save_area` method inside FileStudyAreaDao should be aware of the hydro format. We thought it could be good to store this info inside the `FileStudyTreeConfig` class as this class can also be used by the front-end to know how to display the matrices

And you could also use it inside the `build` method of the `InputHydroxxx` node

dusanparipovic:
> Exactly, the `save_area` method inside FileStudyAreaDao should be aware of the hydro format. We thought it could be good to store this info inside the `FileStudyTreeConfig` class as this class can also be used by the front-end to know how to display the matrices
> 
> And you could also use it inside the `build` method of the `InputHydroxxx` node

Thank you for the clarification. I’ll work on this as soon as possible and open a new PR, since the current one is significantly out of date compared to the dev branch.

PR #2773, opened 2025-10-14
chore(datetime): clean deprecated method `utc_now`

TheoPascoli:


antarest/core/filetransfer/service.py
```diff
@@ -78,7 +78,7 @@ def request_download(
             ready=False,
             path=str(tmpfile),
             owner=owner.impersonator if owner is not None else None,
-            expiration_date=datetime.datetime.utcnow()
+            expiration_date=datetime.datetime.now(datetime.timezone.utc)
```
MartinBelthle: Why using `.replace(tzinfo=None)` some places in the code but not everywhere ? See the 2 lines you modified in this file

TheoPascoli: I did an error-driven development, if it does not raise an error its good, sometimes it works without, sometimes it does not. 
Technically, the date that were later compared would not work without this, because of some sqlalchemy format, I'm not sure why. 
Anyway, I added this replace thing everywhere. 
Thanks for pointing it out

MartinBelthle (approved):
According to chatGPT, `datetime.utcnow()` and `datetime.now(timezone.utc).replace(tzinfo=None)` are equivalent but `datetime.utcnow()` is deprecated so seems good to me

PR #2774, opened 2025-10-14
feat(directory): implement directory management backend

TheoPascoli:
## New endpoints
- GET    /v1/directories                 # List all directories

- POST   /v1/directories               # Create a directory

- PATCH    /v1/directories/{id}         # update name or parent

- DELETE /v1/directories/{id}      # Delete directory if no study inside

 ## Database migration
- Create a new table `directory`
- Add a `directory_id` inside the study table
- All studies are parsed: every study that is in a specific folder will trigger a creation of corresponding directories and will be assigned a `directory_id`

 ## Rollback plan if needed
Just have to delete the new `directory` table and the new field in the `study` table

## Transition from folder to directories
folder paths are built from the directory while the front-end dont use the new directories endpoint. At some point, we want to remove this. I have wrote a removal guide that will be needed when the front-end is done (I have it locally).


alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,210 @@
...
+    Args:
+        folder: A POSIX-style path like "project1/subfolder"
+
+    Returns:
+        A list of directory names: ["project1", "subfolder"]
+    """
+    if not folder:
+        return []
+    path = PurePosixPath(folder)
+    return list(path.parts)
```
MartinBelthle: Minor: Why not just using the tuple ?

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,210 @@
...
+        ),
+        sa.ForeignKeyConstraint(["owner_id"], ["identities.id"]),
+    )
+
+    # Create indexes
+    op.create_index("ix_directory_parent_id", "directory", ["parent_id"])
+    op.create_index("ix_directory_owner_id", "directory", ["owner_id"])
+
+    # Step 2: Create directory_group junction table
+    op.create_table(
```
MartinBelthle: Is this easy to get all groups of a given directory id with this architecture ?

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,210 @@
...
+    op.create_index("ix_directory_group_directory_id", "directory_group", ["directory_id"])
+    op.create_index("ix_directory_group_group_id", "directory_group", ["group_id"])
+
+    # Step 3: Add directory_id column to study table
+    with op.batch_alter_table("study", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("directory_id", sa.String(36), nullable=True))
+        batch_op.create_index("ix_study_directory_id", ["directory_id"])
+
+    dialect_name: str = op.get_context().dialect.name
+    if dialect_name == "postgresql":
```
MartinBelthle: Why performing this only with postgre ?
Seems that sqlite doesn't support dropping foreign keys but we could juste ignore it in the downgrade no ?

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,210 @@
...
+            )
+
+    # Step 4: Data migration - Build directory tree from study.folder
+    connection: Connection = op.get_bind()
+
+    # Cache for created directories: path -> directory_id
+    directory_cache: Dict[str, str] = {}
+
+    # Get all studies with folder information
+    studies = connection.execute(
```
MartinBelthle: Anis did the same in his last alembic migration and we said it would be better to use ORM language instead of SQL to fit with existing migrations

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,210 @@
...
+            JOIN rawstudy r ON s.id = r.id
+            WHERE r.workspace = 'default' AND s.folder IS NOT NULL AND s.folder != ''
+        """)
+    ).fetchall()
+
+    for study_id, folder, owner_id in studies:
+        if not folder:
+            continue
+
+        path_parts = parse_folder_path(folder)
```
MartinBelthle, line 126: Due to previous line you don't need to tpye it as optional in your method

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,210 @@
...
+            SELECT s.id, s.folder, s.owner_id
+            FROM study s
+            JOIN rawstudy r ON s.id = r.id
+            WHERE r.workspace = 'default' AND s.folder IS NOT NULL AND s.folder != ''
+        """)
+    ).fetchall()
+
+    for study_id, folder, owner_id in studies:
+        if not folder:
+            continue
```
MartinBelthle: Due to your SQL request yo ucan't have this case I think

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,210 @@
...
+        final_dir_id = None
+
+        # Create each directory in the path if it doesn't exist
+        for dir_name in directory_parts:
+            current_path = f"{current_path}/{dir_name}" if current_path else dir_name
+
+            if current_path not in directory_cache:
+                # Create new directory
+                dir_id = str(uuid.uuid4())
+                connection.execute(
```
MartinBelthle: Will be very long in prod we should insert everything in one time,

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,210 @@
...
+    if dialect_name == "postgresql":
+        with op.batch_alter_table("study", schema=None) as batch_op:
+            batch_op.drop_constraint("fk_study_directory_id", type_="foreignkey")
+
+    with op.batch_alter_table("study", schema=None) as batch_op:
+        batch_op.drop_index("ix_study_directory_id")
+        batch_op.drop_column("directory_id")
+
+    # Step 2: Drop directory_group table
+    op.drop_index("ix_directory_group_group_id", table_name="directory_group")
```
MartinBelthle: Couldn't we directly remove the table and it will automatically remove the index ?

antarest/study/web/directory_blueprint.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @bp.get(
+        "/directories",
+        tags=[APITag.study_management],
+        summary="List directories",
+    )
+    def list_directories() -> List[DirectoryMetadata]:
+        logger.info("Listing directories for current user")
+        access_permissions = AccessPermissions.for_current_user()
+        return directory_service.list_directories(access_permissions)
```
MartinBelthle: To me these permissions should be handled in the service, the blueprint shouldn't give them

antarest/study/model.py
```diff
@@ -160,6 +161,115 @@ def __repr__(self) -> str:  # pragma: no cover
...
+
+    directory_id: Mapped[str] = mapped_column(
+        String(36), ForeignKey("directory.id", ondelete="CASCADE"), index=True, nullable=False
+    )
+    group_id: Mapped[str] = mapped_column(
+        String(36), ForeignKey("groups.id", ondelete="CASCADE"), index=True, nullable=False
+    )
+
+    @override
+    def __str__(self) -> str:  # pragma: no cover
```
MartinBelthle: Is this really necessary to introduce these eq, str, ... methods ? I find them really ugly

antarest/study/model.py
```diff
@@ -689,3 +805,52 @@ def to_dto(self) -> MatrixAggregationResultDTO:
...
+class DirectoryUpdate(AntaresBaseModel):
+    name: Optional[str] = Field(None, min_length=1, max_length=255)
+    parent_id: Optional[str] = None
+    groups: Optional[List[str]] = None  # List of group IDs to share with (replaces existing groups)
+
+    model_config = ConfigDict(populate_by_name=True, alias_generator=to_camel)
+
+    @field_validator("name")
+    @classmethod
+    def validate_name(cls, v: Optional[str]) -> Optional[str]:
```
MartinBelthle, line 799: You could put this in a method as the 2 validators look really similar

antarest/study/repository.py
```diff
@@ -458,3 +462,161 @@ def has_children(self, uuid: str) -> bool:
...
+        return self._session
+
+    def save(self, directory: Directory) -> Directory:
+        session = self.session
+        directory.groups = [session.merge(g) for g in directory.groups]
+        if directory.owner:
+            directory.owner = session.merge(directory.owner)
+        session.add(directory)
+        session.commit()
+        session.refresh(directory)
```
MartinBelthle: Why using refresh ?

antarest/study/web/studies_blueprint.py
```diff
@@ -248,13 +248,18 @@ def edit_comments(uuid: str, data: CommentsDto) -> None:
         summary="Import Study",
         response_model=str,
     )
-    def import_study(study: UploadFile, groups: str = "") -> str:
+    def import_study(
+        study: UploadFile,
+        groups: str = "",
+        path: str = Query("", description="Directory path for the imported study (e.g., 'project/subfolder')"),
```
MartinBelthle: Why introducing this in this PR, IMO this PR is already big enough, we should do this later

MartinBelthle (changes requested):
Didn't read everything as it's quite big

alembic/versions/f1ff6904ff9f_add_directory_management.py
TheoPascoli: Yes why not

alembic/versions/f1ff6904ff9f_add_directory_management.py
TheoPascoli: I removed it yes. It was here for safety reasons but it looks like that it is not necessary

alembic/versions/f1ff6904ff9f_add_directory_management.py
TheoPascoli: ok lets go

alembic/versions/f1ff6904ff9f_add_directory_management.py
TheoPascoli: Yes

alembic/versions/f1ff6904ff9f_add_directory_management.py
TheoPascoli: You're right

antarest/study/web/directory_blueprint.py
TheoPascoli: I think this approach is better. The blueprint handles the authentication part (web layer), making it more easily testable. We can discuss it

antarest/study/model.py
TheoPascoli: It is more a personal taste and it was already implemented on other model like Study. It has pros and cons ...

antarest/study/repository.py
TheoPascoli: yea its an artefact of an old version

antarest/study/web/studies_blueprint.py
TheoPascoli: Its a few lines and its already done

alembic/versions/f1ff6904ff9f_add_directory_management.py
TheoPascoli: Yes it is, with this join table and the groups attribute in the directory model, it is implicit

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,239 @@
...
+                directories_to_insert.append({
+                    "id": dir_id,
+                    "name": dir_name,
+                    "parent_id": parent_id,
+                    "owner_id": owner_id,
+                })
+
+                # Copy groups from study to directory (only for leaf directory)
+                if current_path == "/".join(directory_parts):
+                    stmt = select(group_metadata_table.c.group_id).where(
```
MartinBelthle: Not sure to understand fully what you're doing here, but seems we have a problem.

We have 2 studies, one in dirA/dirB and the other in dirA/dirC. The first one is in group "B" and the second in group "C". The groups of dirA should be "B" and "C" and it seems that you're only adding the groups of the first study

Also, seems you're doing lots of requests as you're inside of your loop. I think we should do this once before the loop and put it in memory to avoid doing too much DB requests

antarest/study/web/directory_blueprint.py
MartinBelthle: I'll let @sylvlecl choose what he prefers

tests/integration/studies_blueprint/test_create_study.py
```diff
@@ -78,3 +78,91 @@ def test_create_study_with_different_names(
...
+            json={"name": "project"},
+            headers={"Authorization": f"Bearer {admin_access_token}"},
+        )
+        assert res.status_code == 201
+        project_dir_id = res.json()["id"]
+
+        res = client.post(
+            "/v1/directories",
+            json={"name": "subfolder", "parentId": project_dir_id},
+            headers={"Authorization": f"Bearer {admin_access_token}"},
```
MartinBelthle: You can put this in one line at the beginning of your tests instead of putting this everytime you're performing a call, cf. other tests we've written.

antarest/study/web/studies_blueprint.py
MartinBelthle: Okay, but I don't see any test concerning this feature and it's not mention in the PR description. It's a bit misleading as it has very little to do with the PR

antarest/study/directory_exceptions.py
```diff
@@ -0,0 +1,54 @@
...
+
+from http import HTTPStatus
+
+from fastapi import HTTPException
+
+
+class DirectoryNotFoundError(HTTPException):
+    """Raised when a directory is not found."""
+
+    def __init__(self, directory_id: str):
```
MartinBelthle, line 21: Minor:
I think this writing is clearer:
```python
class NotAManagedStudyException(HTTPException):
    def __init__(self, uuid: str) -> None:
        super().__init__(HTTPStatus.UNPROCESSABLE_ENTITY,f"Study {uuid} is not managed")
```

antarest/study/directory_service.py
```diff
@@ -0,0 +1,199 @@
...
+from antarest.study.model import (
+    Directory,
+    DirectoryCreation,
+    DirectoryMetadata,
+    DirectoryUpdate,
+    OwnerInfo,
+)
+from antarest.study.repository import AccessPermissions, DirectoryRepository, StudyMetadataRepository
+
+if TYPE_CHECKING:
```
MartinBelthle: No need for this TYPE_CHECKING i think (we don't have this in the output_service)

antarest/study/model.py
```diff
@@ -160,6 +161,115 @@ def __repr__(self) -> str:  # pragma: no cover
...
+            and other.name == self.name
+            and other.parent_id == self.parent_id
+            and other.owner_id == self.owner_id
+        )
+
+    @override
+    def __repr__(self) -> str:
+        return f'Directory(id="{self.id}", name="{self.name}", parent_id="{self.parent_id}")'
+
+    def get_path(self) -> str:
```
MartinBelthle: Not sure to see where we're using this method

antarest/study/web/studies_blueprint.py
```diff
@@ -319,6 +325,7 @@ def copy_study(
         groups: str = "",
         use_task: bool = True,
         destination_folder: str = "",
+        path: str = Query("", description="Directory path for the copied study (e.g., 'project/subfolder')"),
```
MartinBelthle: This shouldn't be needed, we already have the `destination_folder` attribute

antarest/study/web/studies_blueprint.py
```diff
@@ -374,13 +384,20 @@ def move_study(uuid: str, folder_dest: str) -> Any:
         summary="Create a new empty study",
         response_model=str,
     )
-    def create_study(name: str, version: StudyVersionStr | None = None, groups: str = "") -> Any:
+    def create_study(
+        name: str,
+        version: StudyVersionStr | None = None,
+        groups: str = "",
+        path: str = Query("", description="Directory path where the study will be created (e.g., 'project/subfolder')"),
```
MartinBelthle: Same here. Don't think it's directly related to the PR and we should rename this `directory` instead of path IMO

MartinBelthle (changes requested):
Still didn't read everything as it's quite big

TheoPascoli:
I did the corrections. I also removed all the code about the copy and the import. Thoses endpoints will be modified later if needed.

antarest/study/web/directory_blueprint.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @bp.post(
+        "/directories",
+        status_code=HTTPStatus.CREATED,
+        tags=[APITag.study_management],
+        summary="Create a new directory",
+    )
+    def create_directory(
+        data: DirectoryCreation,
```
MartinBelthle: You could put groups inside the DirectoryCreation. Or you should remove them from the DirectoryUpdate class as you currently can define groups at 2 different places in the update endpoint

antarest/study/web/studies_blueprint.py
MartinBelthle: This remark is still valid IMO

antarest/study/directory_service.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+class DirectoryService:
+    """Service for managing directories."""
+
+    def __init__(
+        self,
+        directory_repository: DirectoryRepository,
+        study_repository: StudyMetadataRepository,
+        study_service: "StudyService",
```
MartinBelthle: Minor: no need for the quotes

antarest/study/directory_exceptions.py
```diff
@@ -0,0 +1,54 @@
...
+        self.directory_id = directory_id
```
MartinBelthle: Do we need this ?

antarest/study/directory_exceptions.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __init__(self, name: str):
+        self.name = name
+        super().__init__(HTTPStatus.CONFLICT, f"A directory named '{name}' already exists in this location")
+
+
+class DirectoryPermissionError(HTTPException):
+    """Raised when user doesn't have permission to perform an operation on a directory."""
+
+    def __init__(self, message: str = "You don't have permission to perform this operation"):
```
MartinBelthle: Minor: You're using this exception twice and giving it a msg each time so we don't need this

antarest/study/directory_service.py
```diff
@@ -0,0 +1,197 @@
...
+
+        directory = Directory(
+            id=str(uuid.uuid4()),
+            name=data.name,
+            parent_id=data.parent_id,
+            owner_id=owner_id,
+        )
+
+        session = self.directory_repository.session
+        directory.groups = [session.get(Group, gid) for gid in group_ids if session.get(Group, gid)]
```
MartinBelthle: Can't we perform this with only 1 DB query ? Also do'nt think we need to introduce `self.directory_repository.session` we could just do a `with db():` and `db.session.`

antarest/study/directory_service.py
```diff
@@ -0,0 +1,197 @@
...
+
+        saved_directory = self.directory_repository.save(directory)
+        return self._to_dto(saved_directory)
```
MartinBelthle: Shouldn't we replace this by a Directory method ? I find it weird to read the way it is currently

antarest/study/directory_service.py
```diff
@@ -0,0 +1,197 @@
...
+        if data.name is not None and data.name != directory.name:
+            if self.directory_repository.has_duplicate_name(data.name, directory.parent_id):
+                raise DirectoryAlreadyExistsError(data.name)
+            directory.name = data.name
+
+        if data.parent_id is not None:
+            if self.directory_repository.check_cycle(directory_id, data.parent_id):
+                raise DirectoryCycleError()
+
+            if data.parent_id != "":
```
MartinBelthle: Not sure to understand what's the `""` is supposed to mean in this case

antarest/study/directory_service.py
```diff
@@ -0,0 +1,197 @@
...
+
+        # Check all subdirectories recursively
+        children = self.directory_repository.get_children(directory_id, AccessPermissions(is_admin=True))
+        for child in children:
+            if self._has_studies_recursive(child.id):
+                return True
+
+        return False
+
+    def _collect_directories_to_delete(self, directory_id: str, access_permissions: AccessPermissions) -> List[str]:
```
MartinBelthle: Seems we have an issue here. If there's a parent folder A that contains 2 folders B and C. The user has permissions on folder A and B but not C. If he wants to delete A, we should raise as it cannot delete C. But it seems the code doesn't see the folder C as it uses the user permissions

antarest/study/service.py
```diff
@@ -1202,15 +1214,75 @@ def copy_task(notifier: ITaskNotifier) -> TaskResult:
...
+
+        Returns:
+            Directory ID of the leaf directory, or None if path is empty
+        """
+        if not folder_path:
+            return None
+
+        # Parse path
+        path = PurePosixPath(folder_path)
+        path_parts = list(path.parts)
```
MartinBelthle: Minor: no need for  `list()`

antarest/study/service.py
```diff
@@ -1202,15 +1214,75 @@ def copy_task(notifier: ITaskNotifier) -> TaskResult:
...
+            if not existing_dir:
+                # Create missing directory
+                new_dir = Directory(
+                    id=str(uuid.uuid4()),
+                    name=dir_name,
+                    parent_id=parent_id,
+                    owner_id=owner_id,
+                )
+                # Add groups
+                new_dir.groups = [
```
MartinBelthle: Same remark here on DB calls

antarest/study/service.py
```diff
@@ -1202,15 +1214,75 @@ def copy_task(notifier: ITaskNotifier) -> TaskResult:
...
+            # Get or create directory from path, using study's owner and groups
+            group_ids = [g.id for g in study.groups]
+            owner_id = study.owner_id or 0
+            directory_id = self._get_directory_from_path(folder_dest, owner_id, group_ids)
+            study.directory_id = directory_id
         else:
             new_folder = None
+            study.directory_id = None
+
         study.folder = new_folder
```
MartinBelthle, line 1247: We shouldn't have this now that we don't use folder

MartinBelthle (changes requested):
Probably the last round of comments for me

MartinBelthle:
Also it has nothing to do with the PR excepts we're using the folder attribute, you can remove the useless `study_matcher` function in `utils.py`

antarest/study/web/studies_blueprint.py
TheoPascoli: It is not related to this PR yes but I think it is useful for the tests to be able to create studies in a given directory. Also I renamed the field

antarest/study/web/directory_blueprint.py
TheoPascoli: I moved the permissions inside the service

antarest/study/directory_exceptions.py
TheoPascoli: no lol

antarest/study/directory_service.py
TheoPascoli: I added some documentation

antarest/study/service.py
TheoPascoli: I want to keep it for the moment in case the migration does not work as expected. it also breaks some unit tests ...

antarest/study/directory_service.py
TheoPascoli: I removed this logic. We can only delete directories that has nothing inside, no study and no subdirectories. It was not even asked to do this

antarest/study/directory_service.py
TheoPascoli: yes

TheoPascoli:
I fixed everything except comments related to the alchemy script. I want a clear picture of all the changes before I start modifying it

antarest/study/directory_service.py
```diff
@@ -0,0 +1,167 @@
...
+            raise DirectoryAlreadyExistsError(data.name)
+
+        directory = Directory(
+            id=str(uuid.uuid4()),
+            name=data.name,
+            parent_id=data.parent_id,
+            owner_id=owner_id,
+        )
+
+        session = self.directory_repository.session
```
sylvlecl: we'd better use the group repository here, the repositories are supposed to hide the session as an implementation detail

antarest/study/directory_service.py
```diff
@@ -0,0 +1,167 @@
...
+
+        if not self.directory_repository.has_permission(directory, access_permissions):
+            raise UserHasNotPermissionError()
+
+        # Check if directory contains studies
+        if self._has_studies(directory_id):
+            raise DirectoryNotEmptyError("Cannot delete directory: it contains studies.")
+
+        # Check if directory contains subdirectories
+        children = self.directory_repository.get_children(directory_id, access_permissions)
```
sylvlecl: THis check should be done without limiting the access permissions, or we may end up deleting a directory which contains other people subdirs

antarest/study/web/directory_blueprint.py
```diff
@@ -0,0 +1,91 @@
...
+        status_code=HTTPStatus.CREATED,
+        tags=[APITag.study_management],
+        summary="Create a new directory",
+    )
+    def create_directory(data: DirectoryCreation) -> DirectoryMetadata:
+        logger.info(f"Creating directory '{data.name}'")
+
+        user = require_current_user()
+
+        group_ids = data.groups if data.groups is not None else [group.id for group in user.groups]
```
sylvlecl: I think it's better to encapsulate that logic inside the service

antarest/study/web/directory_blueprint.py
```diff
@@ -0,0 +1,91 @@
...
+
+        return directory_service.create_directory(data, user.id, group_ids)
```
sylvlecl: for the user ID, we need to get the "impersonator" actually:
here `user.id` may be the ID of a token, not the ID of the actual user.

Ou user model is not very clear here, it's a mistake we often make.

antarest/study/model.py
```diff
@@ -160,6 +161,111 @@ def __repr__(self) -> str:  # pragma: no cover
...
+        primary_key=True,
+        default=lambda: str(uuid.uuid4()),
+        unique=True,
+    )
+    name: Mapped[str] = mapped_column(String(255), nullable=False)
+    parent_id: Mapped[Optional[str]] = mapped_column(
+        String(36), ForeignKey("directory.id", name="fk_directory_parent_id"), nullable=True, index=True
+    )
+    owner_id: Mapped[Optional[int]] = mapped_column(Integer, ForeignKey(Identity.id), nullable=True, index=True)
+
```
sylvlecl, line 189: I think we will also need a "public mode", to keep the wording we use for studies, similar to the permissions for "others" under linux.

Otherwise, we will not be able to have directories visible by everyone.

antarest/study/model.py
```diff
@@ -689,3 +801,62 @@ def to_dto(self) -> MatrixAggregationResultDTO:
 class ReferenceStudy(AntaresBaseModel):
     version: str
     template_name: str
+
+
+class DirectoryMetadata(AntaresBaseModel):
```
sylvlecl, line 744: That module is getting a little messy, at some point we would need at least to separate DB-layer models and plain python models, and probably also to separate them "domaine-wise" (dirs and studies separate)

antarest/study/model.py
```diff
@@ -160,6 +161,111 @@ def __repr__(self) -> str:  # pragma: no cover
...
+    def to_metadata(self) -> "DirectoryMetadata":
+        """Convert this Directory entity to DirectoryMetadata DTO."""
+        owner_info = OwnerInfo(id=self.owner_id, name=self.owner.name if self.owner else "Unknown")
+
+        return DirectoryMetadata(
+            id=self.id,
+            name=self.name,
+            parent_id=self.parent_id,
+            owner=owner_info,
+            groups=[GroupDTO(id=g.id, name=g.name) for g in self.groups],
```
sylvlecl: you might use the `to_dto` method of groups

antarest/study/model.py
```diff
@@ -245,6 +351,8 @@ class Study(Base):
         folder: Where the study is located in the workspace, from the user point of view.
                 Note that generally speaking, this will not correspond to a valid folder on disk, this is only a logical
                 folder presented to the user, not the way we organize data internally.
+                This field is kept for backward compatibility but will be progressively replaced by directory_id.
+        directory_id: The ID of the directory containing this study. None for studies not in a directory.
```
sylvlecl: We can add that it's also used only for "managed" studies

antarest/study/web/directory_blueprint.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @bp.put(
```
sylvlecl: we'd better use `@patch` for such updates.
In the API world `PUT` is supposed to be used for replacement of resources more than for updates

antarest/study/web/directory_blueprint.py
```diff
@@ -0,0 +1,91 @@
...
+        "/directories/{directory_id}",
+        tags=[APITag.study_management],
+        summary="Update directory",
+    )
+    def update_directory(directory_id: str, data: DirectoryUpdate) -> DirectoryMetadata:
+        """
+        Update directory name, parent, or groups.
+
+        - **name**: New name for the directory (optional)
```
sylvlecl: we'd better have that doc in the model field, so that they are visible in API doc

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,239 @@
...
+    for study_id, folder, owner_id in studies:
+        path_parts = parse_folder_path(folder)
+
+        # Skip the last part (which is the study UUID itself, not a directory)
+        # Only create directories for the parent folders
+        if len(path_parts) <= 1:
+            # No parent directories, study is at root level
+            continue
+
+        directory_parts = path_parts[:-1]  # Exclude the last part (study UUID)
```
sylvlecl: Maybe we should check for cases where we don't have the study ID as the last part: I'm not 100% sure it's te case for all studies.

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,239 @@
...
+            )
+        )
+    )
+    studies = bind.execute(stmt).fetchall()
+
+    # First pass: build directory structure and collect groups
+    for study_id, folder, owner_id in studies:
+        path_parts = parse_folder_path(folder)
+
+        # Skip the last part (which is the study UUID itself, not a directory)
```
sylvlecl: See comment below, would be safier to check that the last part is indeed the study ID

alembic/versions/f1ff6904ff9f_add_directory_management.py
sylvlecl: If we want to make things simpler, and if we add the "public_mode", we may just initialize folder as public, which they are indeed today, and with admin owner.
Then users would be in charge of changing the ownership.

sylvlecl (changes requested):
Looks good overall, some questions/suggestions though, of course :)

antarest/study/directory_service.py
TheoPascoli: Yes, I delegated this to a group repository that hide the request details

antarest/study/directory_service.py
TheoPascoli: nice catch

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,202 @@
...
+from pathlib import PurePosixPath
+from typing import Dict
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.sql import and_, column, select, table
+
+# revision identifiers, used by Alembic.
+revision = "f1ff6904ff9f"
+down_revision = "d2942741ae68"
```
MartinBelthle: You should change the down revision as my PR that was merged yesterday contained an alembic migration too

alembic/versions/f1ff6904ff9f_add_directory_management.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+    Create directory tables and migrate existing folder data.
+    """
+    # Step 1: Create directory table
+    op.create_table(
+        "directory",
+        sa.Column("id", sa.String(36), primary_key=True),
+        sa.Column("name", sa.String(255), nullable=False),
+        sa.Column("parent_id", sa.String(36), nullable=True),
+        sa.Column(
```
MartinBelthle: As discussed no need for publicMode

antarest/login/repository.py
```diff
@@ -61,6 +61,13 @@ def get_all(self) -> List[Group]:
         groups: List[Group] = list(self.session.execute(select(Group)).scalars().all())
         return groups
 
+    def get_by_ids(self, ids: List[str]) -> List[Group]:
```
MartinBelthle: You don't seem to use this anymore

antarest/study/web/directory_blueprint.py
```diff
@@ -0,0 +1,81 @@
...
+        return directory_service.create_directory(data)
+
+    @bp.patch(
+        "/directories/{directory_id}",
+        tags=[APITag.study_management],
+        summary="Update directory",
+    )
+    def update_directory(directory_id: str, data: DirectoryUpdate) -> DirectoryMetadata:
+        """
+        Update directory name, parent, or public mode.
```
MartinBelthle: You can remove the mention of publicMode

antarest/study/directory_exceptions.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class DirectoryAlreadyExistsError(HTTPException):
+    """Raised when trying to create a directory with a duplicate name."""
+
+    def __init__(self, name: str):
+        super().__init__(HTTPStatus.CONFLICT, f"A directory named '{name}' already exists in this location")
+
+
+class DirectoryPermissionError(HTTPException):
```
MartinBelthle: Don't think this can happen anymore

antarest/study/directory_service.py
```diff
@@ -0,0 +1,184 @@
...
+    def list_directories(self) -> List[DirectoryMetadata]:
+        access_permissions = AccessPermissions.for_current_user()
+        directories = self.directory_repository.get_all(access_permissions)
+        return [directory.to_metadata() for directory in directories]
+
+    def create_directory(
+        self,
+        data: DirectoryCreation,
+    ) -> DirectoryMetadata:
+        access_permissions = AccessPermissions.for_current_user()
```
MartinBelthle: Should not need this no more and in evey other part of the code since we don't handle permissions in this PR

antarest/study/directory_service.py
```diff
@@ -0,0 +1,184 @@
...
+        if not self.directory_repository.has_permission(access_permissions):
+            raise UserHasNotPermissionError()
+
+        # Check if directory contains studies
+        if self._has_studies(directory_id):
+            raise DirectoryNotEmptyError("Cannot delete directory: it contains studies.")
+
+        # Check if directory contains subdirectories
+        if self.directory_repository.has_children_directories(directory_id):
+            raise DirectoryNotEmptyError("Cannot delete directory: it contains subdirectories.")
```
MartinBelthle: We should be able to do this if subdirectory does not contain any study

antarest/study/directory_service.py
```diff
@@ -0,0 +1,184 @@
...
+            directory_id: The root directory ID to check
+            access_permissions: The user's access permissions
+
+        Raises:
+            UserHasNotPermissionError: If user doesn't have permission on any study
+        """
+        # Check current directory
+        directory = self.directory_repository.get(directory_id)
+        if not directory:
+            raise DirectoryNotFoundError(directory_id)
```
MartinBelthle: Minor but you've already check this before calling this method

antarest/study/directory_service.py
```diff
@@ -0,0 +1,184 @@
...
+        if not self.directory_repository.has_permission(access_permissions):
+            raise UserHasNotPermissionError()
+
+        # Check all studies in the tree
+        studies = self.directory_repository.get_all_studies_in_tree(directory_id)
+        for study in studies:
+            # Check that user has at least READ permission on each study
+            try:
```
MartinBelthle: This try: except is already handled in the `assert_xxx` method

MartinBelthle (changes requested):

antarest/study/service.py
```diff
@@ -1206,17 +1216,65 @@ def copy_task(notifier: ITaskNotifier) -> TaskResult:
...
+            return None
+
+        # Parse path
+        path = PurePosixPath(folder_path)
+
+        if not path.parts:
+            return None
+
+        # Navigate/create directory hierarchy
+        parent_id = None
```
sylvlecl: Here we should go through methods of the directory services, we should not access the database table of another service directly

TheoPascoli:
Applied the relevant feedback, thanks for the review.

antarest/study/directory_service.py
```diff
@@ -0,0 +1,155 @@
...
+            directory_id: The root directory ID to check
+
+        Raises:
+            UserHasNotPermissionError: If user doesn't have permission on any study
+        """
+
+        # Check all studies in the tree
+        studies = self.directory_repository.get_all_studies_in_tree(directory_id)
+        for study in studies:
+            assert_permission(study, StudyPermissionType.READ)
```
MartinBelthle: To me this should be `PermissionType.WRITE` as it's kind of a modification

antarest/study/web/studies_blueprint.py
```diff
@@ -372,15 +372,23 @@ def move_study(uuid: str, folder_dest: str) -> Any:
...
         tags=[APITag.study_management],
         summary="Create a new empty study",
-        response_model=str,
     )
-    def create_study(name: str, version: StudyVersionStr | None = None, groups: str = "") -> Any:
+    def create_study(
+        name: str,
+        version: StudyVersionStr | None = None,
+        groups: str = "",
+        directory: str = Query(
```
sylvlecl, line 380: everywhere else in this API, we use the word `folder` and not `directory`. I tend to think we'd better keep consistent and use folder here too, except if there is a clear difference, but it's not obvious to me.

It raises the questions of being also consistent in other places, in particular in the /directories API :/

antarest/study/service.py
TheoPascoli: `directory_service` is now an argument of `study_service`. `study_service` now calls `directory_service` that calls the `directory_eepository` to know if the directory exists

antarest/study/repository.py
```diff
@@ -458,3 +462,164 @@ def has_children(self, uuid: str) -> bool:
...
+
+    def count_studies_in_tree(self, directory_id: str) -> int:
+        all_directory_ids = [directory_id]
+        descendants = self.get_all_descendant_directories(directory_id)
+        all_directory_ids.extend([d.id for d in descendants])
+
+        stmt = select(func.count(Study.id)).where(Study.directory_id.in_(all_directory_ids))
+        return int(self.session.scalar(stmt))
+
+    def get_directory_paths_bulk(self, directory_ids: List[str]) -> dict[str, str]:
```
sylvlecl, line 599: nice

antarest/study/repository.py
```diff
@@ -458,3 +462,164 @@ def has_children(self, uuid: str) -> bool:
...
+    def get_children_directories(self, directory_id: str) -> Sequence[Directory]:
+        stmt = select(Directory).where(Directory.parent_id == directory_id)
+        result = self.session.execute(stmt)
+        return list(result.scalars().all())
+
+    def get_all_descendant_directories(self, directory_id: str) -> Sequence[Directory]:
+        descendants: List[Directory] = []
+        to_process = [directory_id]
+
+        while to_process:
```
sylvlecl: We can probably use the same kind of recursive request you used in hte "bulk" method ?

antarest/study/service.py
```diff
@@ -709,8 +712,28 @@ def get_studies_information(
             pagination=pagination,
         )
         logger.info("Studies retrieved")
+
+        # Bulk optimization: pre-calculate all directory paths in one query
+        from collections import defaultdict
+
+        from antarest.study.directory_service import DirectoryService
```
sylvlecl: seems already imported

antarest/study/service.py
```diff
@@ -709,8 +712,28 @@ def get_studies_information(
...
+        from antarest.study.repository import DirectoryRepository
```
sylvlecl: is it here because it causes a circular import?

antarest/study/service.py
```diff
@@ -709,8 +712,28 @@ def get_studies_information(
...
+
+        directory_to_studies = defaultdict(list)
+        for study in matching_studies:
+            if hasattr(study, "directory_id") and study.directory_id is not None:
```
sylvlecl: the `hasattr` seems unnecessary ?

antarest/study/service.py
```diff
@@ -709,8 +712,28 @@ def get_studies_information(
...
+                directory_to_studies[study.directory_id].append(study.id)
+
+        # Get all folder paths in bulk (single CTE query)
+        folder_paths = {}
+        if directory_to_studies:
+            directory_service = DirectoryService(directory_repository=DirectoryRepository())
```
sylvlecl: we should have the directory service as a dependency of the study service instead (all those are singletons)

antarest/study/service.py
```diff
@@ -709,8 +712,28 @@ def get_studies_information(
...
+            folder_paths = directory_service.build_folder_paths_bulk(dict(directory_to_studies))
```
sylvlecl: Would it not be simpler to just get a mapping directory id -> directory path here?
Then for each study we just look into that dict

antarest/study/service.py
```diff
@@ -709,8 +712,28 @@ def get_studies_information(
...
+
+        # Build study metadata with pre-calculated paths
         for study in matching_studies:
-            study_metadata = self._try_get_studies_information(study)
+            folder_path = folder_paths.get(study.id) if hasattr(study, "directory_id") and study.directory_id else None
```
sylvlecl: `hasattr` seems not necessary, and we could just do a `folder_path = directory_paths[study.directory_id]` if we have the mapping directory id -> path

antarest/study/directory_service.py
```diff
@@ -0,0 +1,247 @@
...
+
+        Returns:
+            The reconstructed folder path in the format "parent/child/study-id"
+        """
+        path_parts: list[str] = []
+        current_id: str | None = directory_id
+
+        # Traverse up the directory hierarchy
+        while current_id is not None:
+            directory = self.repository.get_by_id(current_id)
```
sylvlecl: Here too we'd better use a recursive query

antarest/study/directory_service.py
```diff
@@ -0,0 +1,247 @@
...
+            path_parts.insert(0, directory.name)
+            current_id = directory.parent_id
+
+        # Build the complete path with study ID at the end
+        if path_parts:
+            return "/".join(path_parts) + f"/{study_id}"
+        else:
+            return study_id
+
+    def build_folder_paths_bulk(self, directory_to_study_map: dict[str, list[str]]) -> dict[str, str]:
```
sylvlecl: See comment in study service, this seems a little complicated, why don't we just get the path corresponding to the study ID from the mapping id -> path when we handle each study id the study service?

antarest/study/service.py
TheoPascoli: No, I imported it here so it's easier to remove once the migration from folder to directory is done

antarest/study/service.py
TheoPascoli: yes it was already a dependency actually

antarest/study/service.py
TheoPascoli: done

antarest/study/web/directory_blueprint.py
```diff
@@ -0,0 +1,79 @@
...
+def create_directory_routes(
+    directory_service: DirectoryService,
+    config: Config,
+) -> APIRouter:
+    auth = Auth(config)
+    bp = APIRouter(prefix="/v1", dependencies=[auth.required()])
+
+    @bp.get(
+        "/directories",
+        tags=[APITag.study_management],
```
sylvlecl: minor: I think we'd better have another tag for that API, so that it appears separate in API doc. Note: you may specify tags in the router definition

antarest/study/storage/abstract_storage_service.py
```diff
@@ -65,10 +65,12 @@ def __init__(
...
         study: Study,
+        folder_path: Optional[str] = None,
     ) -> StudyMetadataDTO:
         additional_data = study.additional_data or StudyAdditionalData()
         study_workspace = getattr(study, "workspace", DEFAULT_WORKSPACE_NAME)
-        folder: Optional[str] = None
+        folder: Optional[str] = folder_path
+
         if hasattr(study, "folder"):
             folder = study.folder
```
sylvlecl: Here we override the folder passed in argument with the one already present in the study --> we return to the API the old folder name instead of the new one computed from the database.

We probably miss some test for this!

sylvlecl (approved):

PR #2777, opened 2025-10-15
feat(user): create a blobstore for variant user resources

MartinBelthle:
[ANT-3907]

**NOTE**: With the current behavior for variant and raw studies we're writing data twice, once in the study and once in the blob store.
For variants, we have to as the snapshot will eventually be cleaned and so we have to keep the data in the blob store.
For raw studies, we don't have to. The files will be removed from the blob store the next time the garbage collector runs.

Also removes a useless loop inside the matrix garbage collector


sylvlecl:
I think we should adapt the script start.sh to allow multiple modules to be started in one process:
we start to have multiple background tasks (matrix gc, archival, bloc gc), we should be able to execute them in one process, in particular because each process adds up a non-negligible amount of memory which is sleeping most of the time.

antarest/blobstore/service.py
```diff
@@ -0,0 +1,72 @@
...
+from typing import Iterable, List
+
+from typing_extensions import override
+
+from antarest.blobstore.blob_usage_provider import IBlobUsageProvider
+from antarest.blobstore.model import BlobReference
+from antarest.blobstore.repository import BlobContentRepository
+
+
+class ISimpleBlobService(ABC):
```
sylvlecl: Just `IBlobService` will do

antarest/blobstore/service.py
```diff
@@ -0,0 +1,72 @@
...
+    @abstractmethod
+    def create(self, data: bytes) -> str:
```
sylvlecl: Would prefer just `put` or `save`

antarest/blobstore/repository.py
```diff
@@ -0,0 +1,48 @@
...
+class BlobContentRepository:
+    def __init__(self, bucket_dir: Path) -> None:
+        self.bucket_dir = bucket_dir
+        self.bucket_dir.mkdir(parents=True, exist_ok=True)
+
+    def get(self, blob_hash: str) -> bytes:
+        file_path = self.bucket_dir / blob_hash
+        if not file_path.exists():
+            raise BlobNotFound(blob_hash)
+        return (self.bucket_dir / blob_hash).read_bytes()
```
sylvlecl: you can re-use file_path

antarest/blobstore/repository.py
```diff
@@ -0,0 +1,48 @@
...
+import hashlib
+import logging
+from pathlib import Path
+
+from antarest.blobstore.exceptions import BlobNotFound
+
+logger = logging.getLogger(__name__)
+
+
+def compute_blob_hash(content: bytes) -> str:
```
sylvlecl, line 22: you may make it _private

sylvlecl (commented):
Looks good, minor comments

antarest/blobstore/service.py
MartinBelthle: I used `save`

antarest/blobstore/repository.py
MartinBelthle: I used it inside the tests and in the in memoery implem. Also the same method is not private inside the matrixstore so I prefer to keep things coherent between services

sylvlecl (commented):
Approved but I prefer to check if we need to change the operations documentation before deployment

sylvlecl (approved):

sylvlecl (approved):

PR #2780, opened 2025-10-17
feat(area): implement database support

TheoPascoli:
Overview

  Introduces a dual storage mode for Antares studies: FILESYSTEM (existing) or DATABASE (new). This is the first step toward database study storage

  Architecture

  - StorageMode enum: FILESYSTEM | DATABASE
  - New tables: area and area_ui store area data
  - DAO pattern: DatabaseStudyDao routes operations to DB, FileStudyTreeDao to filesystem
  - Dynamic routing: Commands are applied to the appropriate storage based on study.storage_mode

  Changes

  - Added storage_mode column to study table (Alembic migration)
  - Implemented DatabaseAreaDao with all CRUD operations for areas
  - Updated add_commands() to handle both storage modes
  - Added storage_mode parameter to POST /v1/studies API endpoint


alembic/versions/2572145aff71_add_storage_mode_and_area_tables.py
```diff
@@ -0,0 +1,85 @@
...
+            )
+        )
+        batch_op.create_index(batch_op.f("ix_study_storage_mode"), ["storage_mode"], unique=False)
+
+    # 2. Create area table
+    op.create_table(
+        "area",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("area_id", sa.String(length=255), nullable=False),
```
MartinBelthle, line 37: I think we should also add a column `area_name`

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,389 @@
...
+
+        stmt = select(area.c.area_id).where(area.c.study_id == study_id)
+        result = session.execute(stmt)
+
+        areas_info = []
+        for row in result:
+            area_id = row.area_id
+            # For now, we don't have the name stored in DB (it's in filesystem config)
+            # and thermals is not yet implemented
+            areas_info.append(AreaInfo(id=area_id, name=area_id, thermals=None))
```
MartinBelthle: We could use `area_name` if stored in DB here

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,389 @@
...
+
+        result: Dict[str, AreaUIData] = {}
+
+        for area_row in areas_result:
+            area_db_id = area_row.id
+            area_id = area_row.area_id
+
+            # Get all UI info for this area across all layers
+            stmt_ui = select(area_ui).where(area_ui.c.area_id == area_db_id)
+            ui_rows = session.execute(stmt_ui).fetchall()
```
MartinBelthle: This does 1 request by area, shouldn't we do one big request and then do the filters in memory ?

antarest/study/dao/database/models.py
```diff
@@ -0,0 +1,48 @@
...
+These tables store study data (areas, UI positions, etc.) in the database instead of the filesystem.
+"""
+
+from sqlalchemy import Column, ForeignKey, Integer, String, Table, UniqueConstraint
+
+from antarest.dbmodel import Base
+
+metadata = Base.metadata
+
+area = Table(
```
MartinBelthle: Why not using the `class XX(Base):
    __tablename__ = "xx"` as we use elsewhere in the code ? I'm probably missing something

antarest/study/model.py
```diff
@@ -239,6 +239,18 @@ class StudyContentStatus(enum.Enum):
     ERROR = "ERROR"
 
 
+class StorageMode(enum.StrEnum):
+    """
+    Storage mode for study data.
+
+    FILESYSTEM: Traditional file-based storage (default for backward compatibility)
+    DATABASE: Database-based storage for areas and related entities
```
MartinBelthle: You could be more global as this is only relative to your PR

antarest/study/service.py
```diff
@@ -879,16 +917,22 @@ def get_study_path(self, uuid: str) -> Path:
         return self.storage_service.get_storage(study).get_study_path(study)
 
     def create_study(
-        self, study_name: str, version: Optional[StudyVersion], group_ids: List[str], directory: str = ""
+        self,
+        study_name: str,
+        version: Optional[StudyVersion],
+        group_ids: List[str],
+        storage_mode: StorageMode,
```
MartinBelthle: To me you should keep the default value to FILESYSTEM here to avoid having to modify existing tests for instance

antarest/study/dao/database/models.py
```diff
@@ -0,0 +1,48 @@
...
+    "area",
+    metadata,
+    Column("id", Integer, primary_key=True, autoincrement=True),
+    Column("study_id", String(36), ForeignKey("study.id", ondelete="CASCADE"), nullable=False, index=True),
+    Column("area_id", String(255), nullable=False),
+    UniqueConstraint("study_id", "area_id", name="uq_area_study_id_area_id"),
+)
+
+# Relation: One area can have multiple UI configurations (one per layer)
+area_ui = Table(
```
MartinBelthle: Also we should probably have names to specifiy it's in DB and with camelcase such as AreaUiDB. it's just a proposition but i don't quite like that it's in lower case

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,389 @@
...
+        area_db_id = area_result.id
+
+        # Get UI for the specified layer
+        stmt_ui = select(area_ui).where((area_ui.c.area_id == area_db_id) & (area_ui.c.layer_id == layer))
+        ui_row = session.execute(stmt_ui).fetchone()
+
+        if ui_row:
+            return AreaUI(x=ui_row.x, y=ui_row.y, color_rgb=(ui_row.color_r, ui_row.color_g, ui_row.color_b))
+
+        # If layer not found, fall back to layer "0"
```
MartinBelthle: Is this waht we do in the FS implem ? If not we should not do this here i think

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,389 @@
...
+            # Check if area already exists
+            stmt_check = select(area.c.id).where((area.c.study_id == study_id) & (area.c.area_id == area_id))
+            existing = session.execute(stmt_check).fetchone()
+
+            if existing:
+                raise ValueError(f"Area '{area_name}' already exists and could not be created")
+
+            # Insert new area
+            stmt_area = insert(area).values(study_id=study_id, area_id=area_id)
+            # Cast to CursorResult to access inserted_primary_key attribute
```
MartinBelthle: I'm not familiar with this so I'm confused when reading this code.

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,389 @@
...
+            # (locally mypy infers CursorResult, but CI with type stubs requires explicit cast)
+            result = cast(CursorResult[Any], session.execute(stmt_area))  # type: ignore[redundant-cast, unused-ignore]
+            session.flush()
+
+            new_area_id = result.inserted_primary_key[0]
+
+            # Create default UI for layer "0"
+            stmt_ui = insert(area_ui).values(
```
MartinBelthle: These default values should only exist in the AreaUi model so we should use them here instead of hard-coding them

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,389 @@
...
+                raise AreaNotFound(area_id)
+
+            # Delete area (cascade will delete area_ui automatically)
+            stmt = delete(area).where((area.c.study_id == study_id) & (area.c.area_id == area_id))
+            session.execute(stmt)
+
+            # Single atomic commit
+            session.commit()
+        except Exception:
+            session.rollback()
```
MartinBelthle: Why always put rollback in case an exception occurs ? We don't do this elsewhere in the code, only in the DBSession class

antarest/study/service.py
```diff
@@ -407,34 +408,64 @@ def get_files(self) -> FileStudy:
...
+        from antarest.study.model import StorageMode
+
+        if self._study.storage_mode == StorageMode.DATABASE:
+            return DatabaseStudyDao(self._study.id, self._repository.session).read_only()
         return FileStudyTreeDao(self.get_files()).read_only()
 
     @override
     def add_commands(self, commands: Sequence[ICommand], listener: Optional[ICommandListener] = None) -> None:
+        from antarest.study.dao.database.database_study_dao import DatabaseStudyDao
+        from antarest.study.model import StorageMode
```
MartinBelthle: Why not putting these imports at the start of the file ? To avoid circular imports ?

antarest/study/service.py
```diff
@@ -407,34 +408,64 @@ def get_files(self) -> FileStudy:
...
+            # Invalidate cache if needed
+            if should_invalidate_cache:
+                remove_from_cache(self._raw_study_service.cache, study.id)
         else:
-            data = FileStudyTreeConfigDTO.from_build_config(file_study.config).model_dump()
-            update_cache(self._raw_study_service.cache, study.id, data)
+            # FILESYSTEM mode: apply to file system
+            file_study = self.get_files()
+
+            for command in commands:
```
MartinBelthle: We could put code in common for these 2 cases, seems doable

antarest/study/service.py
```diff
@@ -479,6 +511,11 @@ def get_files(self) -> FileStudy:
 
     @override
     def get_study_dao(self) -> ReadOnlyStudyDao:
+        from antarest.study.dao.database.database_study_dao import DatabaseStudyDao
+        from antarest.study.model import StorageMode
+
+        if self._study.storage_mode == StorageMode.DATABASE:
+            return DatabaseStudyDao(self._study.id, self._repository.session).read_only()
```
MartinBelthle: Discussed with Sylvain, you should simply use: `, db.session`. This way you can also remove the repository inside the __init__ you've introduced

antarest/study/service.py
```diff
@@ -407,34 +408,64 @@ def get_files(self) -> FileStudy:
...
-                raise CommandApplicationError(result.message)
-
-        # if commands that can't update the cache are applied, we need to invalidate it.
-        # Otherwise, we can update it.
-        if should_invalidate_cache:
-            remove_from_cache(self._raw_study_service.cache, study.id)
+        # Apply commands based on storage mode
+        if study.storage_mode == StorageMode.DATABASE:
+            # DATABASE mode: apply to database
+            database_dao = DatabaseStudyDao(study.id, self._repository.session)
```
MartinBelthle: Same here, should be `db.session`

antarest/study/dao/database/database_study_dao.py
```diff
@@ -0,0 +1,730 @@
...
+        """
+        Get study comments.
+
+        Note: Comments are not yet stored in database for database-mode studies.
+        This method returns an empty string for now.
+
+        Returns:
+            Empty string.
+        """
+        return ""
```
MartinBelthle: We should raise everywhere IMO

MartinBelthle (changes requested):
Nice work :), few remarks

antarest/study/dao/database/models.py
TheoPascoli: This is intentional. Since DB objects are hidden behind the DAO layer, we prefer using SQLAlchemy Core over ORM for better performance

antarest/study/dao/database/models.py
TheoPascoli: Convention is snake_case, we can change it it does not matter too much

antarest/study/dao/database/database_area_dao.py
TheoPascoli: both implementations have the same behavior: fall back to the default/layer 0 when the requested layer doesn't exist. The implementations are consistent.

alembic/versions/2572145aff71_add_storage_mode_and_area_tables.py
```diff
@@ -0,0 +1,89 @@
...
+        sa.Column("color_r", sa.Integer(), nullable=False),
+        sa.Column("color_g", sa.Integer(), nullable=False),
+        sa.Column("color_b", sa.Integer(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["area_id"],
+            ["area.id"],
+            name=op.f("fk_area_ui_area_id_area"),
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint("id", name=op.f("pk_area_ui")),
```
sylvlecl: I think it's important that we include the study_id in the primary key:
we will always work on a particular study, performing queries such as "retrieve all areas for that study", therefore it makes sense that the study_id be part of the primary key.
I think it will have performance benefits compared to a secondary index.
(on disk, all areas for one study will be stored "more contiguously" than if we had random IDs inependent of the study).

I did not benchmark this though

alembic/versions/2572145aff71_add_storage_mode_and_area_tables.py
```diff
@@ -0,0 +1,89 @@
...
+        sa.PrimaryKeyConstraint("id", name=op.f("pk_area")),
+        sa.UniqueConstraint("study_id", "area_id", name=op.f("uq_area_study_id_area_id")),
+    )
+    with op.batch_alter_table("area") as batch_op:
+        batch_op.create_index(batch_op.f("ix_area_study_id"), ["study_id"], unique=False)
+
+    # 3. Create area_ui table
+    op.create_table(
+        "area_ui",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
```
sylvlecl: I think this will not be necessary, the primary key can be just study_id + area_id

antarest/study/dao/database/models.py
sylvlecl: As for me, i would prefer to have CAPITAL_CASE here since it's a global, indeed we can also have a `TABLE` suffix to clarify what it is, like `AREA_UI_TABLE` and `AREA_TABLE`

antarest/study/model.py
```diff
@@ -240,6 +240,15 @@ class StudyContentStatus(enum.Enum):
     ERROR = "ERROR"
 
 
+class StorageMode(enum.StrEnum):
```
sylvlecl, line 243: Note for me:
we should be consistent between here and the output storage, where I used "type" I think. "mode" is OK to me too.

antarest/study/dao/database/models.py
```diff
@@ -0,0 +1,53 @@
...
+    metadata,
+    Column("id", Integer, primary_key=True, autoincrement=True),
+    Column("study_id", String(36), ForeignKey("study.id", ondelete="CASCADE"), nullable=False, index=True),
+    Column("area_id", String(255), nullable=False),
+    Column("area_name", String(255), nullable=False),
+    UniqueConstraint("study_id", "area_id", name="uq_area_study_id_area_id"),
+)
+
+# Relation: One area can have multiple UI configurations (one per layer)
+area_ui = Table(
```
sylvlecl: Here there is a design decision: we can wonder if it's worth it to have a table dedicated to the UI data, or if we'd better store the whole UI configuration of an area as a document inside the area table.

Seems to completely make sense to have used a separate table, since for example we can this way easily retrieve data for all areas for a specific layer.

antarest/study/web/studies_blueprint.py
```diff
@@ -349,16 +350,31 @@ def create_study(
         directory: str = Query(
             "", description="Directory path where the study will be created (e.g., 'project/subfolder')"
         ),
+        storage_mode: StorageMode = StorageMode.FILESYSTEM,
```
sylvlecl, line 353: Nice :)

sylvlecl: Later (not that PR), we can also add that option to the copy endpoint, which would use the "roundtrip" method recently introduced.

sylvlecl: I think we should add a configuration option the the application, to enable or not the use of the new storage mode.

Since the feature will be in dev for quite some time, it will be useful to not enable it in production systems until it has reached a suitable level of maturity.

antarest/study/dao/database/database_study_dao.py
```diff
@@ -0,0 +1,870 @@
...
+    """
+    Database implementation of StudyDao for area operations.
+
+    This DAO uses database storage for areas when storage_mode=DATABASE.
+    It uses composition with DatabaseAreaDao to provide area-related functionality.
+
+    Note: This is a minimal implementation focused on area operations only.
+    Some methods from StudyDao are not supported in database mode and will raise NotImplementedError.
+
+    Note: Write operations do NOT commit transactions. The caller (service layer)
```
sylvlecl, line 70: It's great to have documented that choice !
As for now, I don't have a clear opinion about whether it's the best choice or not.
What you have implemented has the advantage to be able to have a transaction that spans multiple method calls.
On the other side, it makes the DAO abstraction kind of "leaky" and requires more code and more "thinking" at the service level.

sylvlecl (commented):
Looks very good overall :) 
I think we should change the following things:
1. use study_id + area_id as the primary key, instead of an ID independant from the study
2. have a kind of "feature flag" in config, so that the use of the new storage mode is forbidden unless we explicitly opt-in, for now. In particular we should not enable it in prod systems until we reach some level of maturity

antarest/study/web/studies_blueprint.py
TheoPascoli: I added a `database_mode_enabled` field in the config. It returns 500 if we try to create a database study in production

alembic/versions/2572145aff71_add_storage_mode_and_area_tables.py
```diff
@@ -0,0 +1,81 @@
...
+    op.drop_table("area")
+
+    # Drop storage_mode column and index
+    with op.batch_alter_table("study") as batch_op:
+        batch_op.drop_index(batch_op.f("ix_study_storage_mode"))
+        batch_op.drop_column("storage_mode")
+
+    # Drop enum type (PostgreSQL only)
+    if op.get_context().dialect.name == "postgresql":
+        sa.Enum(name="storagemode").drop(op.get_bind(), checkfirst=True)
```
MartinBelthle, line 81: What does this do ? I'm curious

antarest/core/config.py
```diff
@@ -169,6 +169,7 @@ class StorageConfig:
     blobstore: Path = Path("./blobstore")
     blob_gc_sleeping_time: int = 86400
     blob_gc_dry_run: bool = False
+    database_mode_enabled: bool = False
```
MartinBelthle: We said we put this inside a category `study` I believe

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,446 @@
...
+            .select_from(
+                AREA_TABLE.join(
+                    AREA_UI_TABLE,
+                    (AREA_TABLE.c.study_id == AREA_UI_TABLE.c.study_id)
+                    & (AREA_TABLE.c.area_id == AREA_UI_TABLE.c.area_id),
+                )
+            )
+            .where(AREA_TABLE.c.study_id == study_id)
+        )
+        rows = session.execute(stmt).fetchall()
```
MartinBelthle: Why do we perform a `fetchall` here but not in the previous method ?

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,446 @@
...
+
+        if not rows:
+            return {}
+
+        # Group UI rows by area_id
+        ui_by_area: Dict[str, List[Any]] = {}
+        for row in rows:
+            area_id = row.area_id
+            if area_id not in ui_by_area:
```
MartinBelthle: Seems you could replace these 3 lines by 
`ui_by_area.setdefault(area_id, []).append(row)
`

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,446 @@
...
+            area_ui_data: The UI properties to save (x, y, color_rgb).
+
+        Raises:
+            AreaNotFound: If the area does not exist.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        # Check if area exists
+        stmt_area = select(AREA_TABLE.c.area_id).where(
```
MartinBelthle: You could put this logic inside a utility function as you've already duplicated it 3 times and it will be even more in the future

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,446 @@
...
+        valid_area_ids = {row.area_id for row in session.execute(stmt_areas).fetchall()}
+
+        # Get current areas in this study that have this layer
+        stmt_existing = select(AREA_UI_TABLE.c.area_id).where(
+            (AREA_UI_TABLE.c.study_id == study_id) & (AREA_UI_TABLE.c.layer_id == layer_id)
+        )
+        existing_area_ids = {row.area_id for row in session.execute(stmt_existing).fetchall()}
+
+        # Filter target area_ids to only include valid ones
+        target_area_ids = {aid for aid in area_ids if aid in valid_area_ids}
```
MartinBelthle: Shouldn't we raise if any given area inside `area_ids` does not exist ?

antarest/study/dao/database/models.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+from sqlalchemy import Column, ForeignKey, ForeignKeyConstraint, Integer, String, Table
+
+from antarest.dbmodel import Base
+
+metadata = Base.metadata
+
+# Default layer ID constant
+DEFAULT_LAYER_ID = "0"
```
MartinBelthle: This is also used in the filestudy dao. But for now we only use `"0"`. I think we should move this and use it everywhere

tests/integration/study_data_blueprint/test_area_database_mode.py
```diff
@@ -0,0 +1,419 @@
...
+DAO methods are implemented.
+"""
+
+import pytest
+from starlette.testclient import TestClient
+
+from tests.integration.prepare_proxy import PreparerProxy
+
+# Reason for xfail - CREATE_AREA command requires methods not yet implemented in DatabaseStudyDao
+DATABASE_MODE_INCOMPLETE = (
```
MartinBelthle, line 32: Nice to add this

antarest/study/dao/database/database_study_dao.py
```diff
@@ -0,0 +1,870 @@
...
+        raise NotImplementedError("This method is not yet implemented for database storage mode")
+
+    # Area operations - delegated to DatabaseAreaDao via composition
+    @override
+    def get_all_areas_info(self) -> List[AreaInfo]:
+        return self._area_dao.get_all_areas_info()
+
+    @override
+    def get_all_areas_ui_info(self) -> Dict[str, AreaUIData]:
+        return self._area_dao.get_all_areas_ui_info()
```
MartinBelthle: Why do you introduce such methods ? In the filesystem DAO the `file_study_dao` is almost empty

antarest/study/service.py
```diff
@@ -416,29 +420,43 @@ def get_study_dao(self) -> ReadOnlyStudyDao:
...
         if should_invalidate_cache:
             remove_from_cache(self._raw_study_service.cache, study.id)
-        else:
+        elif file_study is not None:
+            # Update cache only in FILESYSTEM mode when not invalidated
             data = FileStudyTreeConfigDTO.from_build_config(file_study.config).model_dump()
             update_cache(self._raw_study_service.cache, study.id, data)
+
+        # Update editor metadata (only in FILESYSTEM mode)
+        if file_study is not None:
```
MartinBelthle: I prefer we check this `if study.storage_mode == StorageMode.FILE_SYSTEM:` as it's straight forward and easier to understand. This way we could even remove your comment

antarest/study/service.py
```diff
@@ -416,29 +420,43 @@ def get_study_dao(self) -> ReadOnlyStudyDao:
...
             if not result.status:
                 raise CommandApplicationError(result.message)
 
-        # if commands that can't update the cache are applied, we need to invalidate it.
-        # Otherwise, we can update it.
+        # Handle cache invalidation
         if should_invalidate_cache:
             remove_from_cache(self._raw_study_service.cache, study.id)
-        else:
+        elif file_study is not None:
```
MartinBelthle: Same remark

MartinBelthle (changes requested):
Last round of minor comments and questions for me

antarest/study/service.py
sylvlecl: Agreed, it will be more explicit.

BUT I don't think this condition is right, actually: with the DB DAO, we will still want to update study metadata.
Only the file part will not be needed. This part should probably go into the DAO and be empty for the DB DAO. (I think I had said more or less the opposite in a previous PR by @MartinBelthle, but here we can see that it's needed for "backward compat", although the data inside the study is not the reference data which is used elsewhere in the app)

antarest/study/service.py
sylvlecl: Note that in the future we may still want to keep a synthesis of the study in cache (and invalidate here).
But we'll think of the implementation in future works.

alembic/versions/2572145aff71_add_storage_mode_and_area_tables.py
TheoPascoli: It's just to delete the enum from the database in case of downgrade. It is a postgres specific command. The checkfirst is just a security in case it does not exist, it wont crash.

antarest/core/config.py
TheoPascoli: Done

antarest/study/dao/database/database_area_dao.py
TheoPascoli: It was needed for the condition later `if not rows` that must be perfomed on a `Iterator` (by fetchall). It can not be tested on a `Result` (without fetchall). I removed both so its simpler

antarest/study/dao/database/database_area_dao.py
TheoPascoli: yes

antarest/study/dao/database/models.py
TheoPascoli: I moved it in area_model. Maybe Ill move it later once the layer models is created

antarest/study/dao/database/database_study_dao.py
TheoPascoli: Yes I made `database_study_dao` inherit `area_dao` like the filesystem

antarest/study/service.py
TheoPascoli: I moved the file-specific logic into the DAO:
  - Added `update_antares_file(editor, last_save)` abstract method in StudyDao
  - FileStudyTreeDao: updates study.antares file
  - DatabaseStudyDao:  no-op (empty implementation)

  The service now always calls _update_editor_and_lastsave(dao), which:
  1. Calls dao.update_antares_file() : does nothing in DB mode
  2. Updates Study.editor in DB : always runs

  No more if file_study is not None condition needed.

MartinBelthle: Nice :)

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,435 @@
...
+            AreaNotFound: If any area_id does not exist.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        # Get all areas and which ones already have this layer
+        stmt = (
+            select(
+                AREA_TABLE.c.area_id,
+                case((AREA_UI_TABLE.c.area_id.isnot(None), True), else_=False).label("has_layer"),
```
MartinBelthle, line 327: What does this do ? I'm curious as how could this be verified ? `AREA_UI_TABLE.c.area_id.isnot(None)`

antarest/study/dao/database/database_area_dao.py
```diff
@@ -0,0 +1,435 @@
...
+        for row in session.execute(stmt):
+            all_area_ids.add(row.area_id)
+            if row.has_layer:
+                areas_with_layer.add(row.area_id)
+
+        # Check for invalid area_ids
+        target_area_ids = set(area_ids)
+        invalid_ids = target_area_ids - all_area_ids
+        if invalid_ids:
+            raise AreaNotFound(next(iter(invalid_ids)))
```
MartinBelthle: Minor but we should raise with all names as the exception accepts them

antarest/study/dao/database/models.py
```diff
@@ -0,0 +1,52 @@
...
+)
+
+# Relation: One area can have multiple UI configurations (one per layer)
+AREA_UI_TABLE = Table(
+    "area_ui",
+    metadata,
+    Column("study_id", String(36), nullable=False, primary_key=True),
+    Column("area_id", String(255), nullable=False, primary_key=True),
+    Column("layer_id", String(10), nullable=False, primary_key=True, default=DEFAULT_LAYER_ID),
+    Column("x", Integer, nullable=False, default=0),
```
MartinBelthle: It's more a global question for future implems but shouldn't we declare these default values as variables to use inside the `AreaUi` class ? What's your opinion @sylvlecl ?

antarest/study/service.py
```diff
@@ -417,38 +421,50 @@ def get_study_dao(self) -> ReadOnlyStudyDao:
...
             if not result.status:
                 raise CommandApplicationError(result.message)
 
-        # if commands that can't update the cache are applied, we need to invalidate it.
-        # Otherwise, we can update it.
+        # Handle cache invalidation
         if should_invalidate_cache:
             remove_from_cache(self._raw_study_service.cache, study.id)
-        else:
+        elif study.storage_mode == StorageMode.FILESYSTEM and file_study is not None:
```
MartinBelthle: To me this should be 
```python
elif study.storage_moed == StorageMode.FILESYSTEM:
     assert file_study is not None (for mypy)
```
because file_study will never be None with StorageMode.FILESYSTEM

antarest/core/config.py
MartinBelthle: Indeed but the field is still flat in the StorageConfig object :/ Shouldn't we also change this ? What's your opinion @sylvlecl

antarest/study/dao/database/database_study_dao.py
MartinBelthle: Okay yes I see, we'll remove them when implementing other DAOs, thanks for the explanation

MartinBelthle (changes requested):
I didn't see where we handle the `commit()` in DB we talked about, could you show me where to look at ?

antarest/study/dao/database/database_area_dao.py
TheoPascoli: It's used with an outerjoin (line 330). When there's a match in AREA_UI_TABLE, area_id has a value. When there's no match (area doesn't have this layer yet), area_id is NULL. So isnot(None) checks if the join found a row = area already has this layer

antarest/core/config.py
sylvlecl: Agreed that it's better to have classes that mimic the file structure, like is currently done.
At some point we should refactor this to use pydantic instead of hand-coded parsing, which will actually enforce that structure.

antarest/study/dao/database/models.py
sylvlecl: You're right, when we can avoid it, it's better that the responsibility of the default value relies only on the application.

Otherwise, we will end up with the default values being defined in multiple places, which will be more fragile.

Actually I am not identifying good use cases for default values at the DB level, but I have a limited experience for those DB-related questions:
- for new tables, we'd better define it in the code, as mentioned above
- for new columns in existing tables, filling columns at migration time seems a better option for the reason above (not having the default value defined in multiple places)

sylvlecl (approved):

PR #2782, opened 2025-10-20
feat(output): add endpoint to retrieve time index for output matrices

TheoPascoli:


antarest/study/web/studies_blueprint.py
```diff
@@ -399,13 +400,41 @@ def get_study_synthesis(uuid: str) -> Any:
...
+
+        Args:
+        - `uuid`: The UUID of the study.
+        - `output_id`: The ID of the output simulation.
+        - `frequency`: The temporal granularity (hourly, daily, weekly, monthly, or annual).
+
+        Returns:
+        - MatrixIndex containing start_date, steps, first_week_size, and level.
+        """
+        study_id = sanitize_uuid(uuid)
```
MartinBelthle: We could also sanitize the `output_id` as we do elsewhere

tests/integration/studies_blueprint/test_study_matrix_index.py
```diff
@@ -120,3 +121,125 @@ def test_get_study_matrix_index(
...
+        "frequency,expected_level,expected_steps",
+        [
+            ("hourly", "hourly", 168),
+            ("daily", "daily", 7),
+            ("weekly", "weekly", 1),
+            ("monthly", "monthly", 1),
+            ("annual", "annual", 1),
+        ],
+    )
+    def test_get_output_time_index_parametrized(
```
MartinBelthle: Seems that your test is duplicated. The parametrized version takes longer to run so I would prefer to put this inside the test instead (just a personal opinion)

tests/integration/studies_blueprint/test_study_matrix_index.py
```diff
@@ -120,3 +121,125 @@ def test_get_study_matrix_index(
...
+        internal_study_id: str,
+        frequency: str,
+        expected_level: str,
+        expected_steps: int,
+    ) -> None:
+        """
+        Parametrized test for different frequencies.
+        """
+        client.headers = {"Authorization": f"Bearer {user_access_token}"}
+        output_id = "20201014-1427eco"
```
MartinBelthle: If it exists we should test with another output that has a different start date

MartinBelthle (changes requested):

antarest/study/web/studies_blueprint.py
```diff
@@ -399,13 +400,42 @@ def get_study_synthesis(uuid: str) -> Any:
...
         summary="Return study input matrix start date index",
-        response_model=MatrixIndex,
     )
-    def get_study_matrix_index(uuid: str, path: str = "") -> Any:
+    def get_study_matrix_index(uuid: str, path: str = "") -> MatrixIndex:
         study_id = sanitize_uuid(uuid)
         logger.info(f"Return the start date for input matrix '{study_id}'")
         return study_service.get_input_matrix_startdate(study_id, path)
 
+    @bp.get(
```
MartinBelthle: I just realized we should put this in the `output_blueprint` and use this `tags=[APITag.study_outputs]`

antarest/study/service.py
```diff
@@ -941,6 +941,23 @@ def get_input_matrix_startdate(self, study_id: str, path: Optional[str]) -> Matr
                 level = StudyDownloadLevelDTO(data_node.freq)
         return get_start_date(file_study, output_id, level)
 
+    def get_output_time_index(self, study_id: str, output_id: str, frequency: StudyDownloadLevelDTO) -> MatrixIndex:
+        """
```
MartinBelthle: Same here, this should go inside the `output_service`

tests/integration/studies_blueprint/test_study_matrix_index.py
MartinBelthle: We don't have this ?

MartinBelthle (changes requested):

tests/integration/studies_blueprint/test_study_matrix_index.py
TheoPascoli: I couldnt find one. I can create it though

MartinBelthle: No need if it doesn't

MartinBelthle (approved):

PR #2783, opened 2025-10-20
feat(metrics): introduce prometheus metrics

sylvlecl:
**Description**

This work introduces the possibility ro report various metrics, in prometheus format,
at the `/metrics/` endpoint.

The following metrics are available:
 - DB usage related metrics:
   - db_connections_duration_seconds (histogram)
   - db_connections_used (gauge)
   - db_connections_idle (gauge)
   - db_connections_events (counter)
   - db_transactions_current (gauge)
   - db_transactions_duration_seconds (histogram)
   - db_session_events (counters)
 - Tasks related metrics:
   - tasks_duration_seconds (histogram)
   - tasks_wait_seconds (histogram)
   - tasks_running (gauge)
   - tasks_pending (gauge)
 - HTTP requests related metrics:
   - http_requests_duration_seconds (histogram)

A `metrics` section is added to the configuration:
```yaml
metrics:
    prometheus:
        multiprocess: true  # if true, concatenates metrics of all workers,
                                      # env var PROMETHEUS_MULTIPROC_DIR must be defined
```

**Future possible work**
 - Enable to start an HTTP metrics server for background services (matrix GC etc)
 - Add metrics for "jobs" (simulations)
 - Use opentelemetry (but ecosystem does not seem that mature, in particular for multiprocess mode)
 - Add status labels for tasks
 - ...


antarest/core/metrics.py
```diff
@@ -0,0 +1,309 @@
...
+from antarest.core.config import Config
+from antarest.core.exceptions import ConfigurationError
+from antarest.core.tasks.service import TaskServiceListener
+
+logger = logging.getLogger(__name__)
+
+
+_PROMETHEUS_MULTIPROCESS_ENV_VAR = "PROMETHEUS_MULTIPROC_DIR"
+
+WORKER_ID = str(os.getpid())
```
MartinBelthle, line 44: How that works ?

antarest/core/metrics.py
```diff
@@ -0,0 +1,309 @@
...
+from typing_extensions import override
+
+from antarest.core.config import Config
+from antarest.core.exceptions import ConfigurationError
+from antarest.core.tasks.service import TaskServiceListener
+
+logger = logging.getLogger(__name__)
+
+
+_PROMETHEUS_MULTIPROCESS_ENV_VAR = "PROMETHEUS_MULTIPROC_DIR"
```
MartinBelthle, line 42: General comment: You introduced an "Interface" MetricsConfig inside the Config, a general class TaskMetricsRecorder but at some places you put everything as promotheus without hiding the fact that it's an implem. I think it should be hidden a bit more

docker-compose.yml
```diff
@@ -49,9 +49,13 @@ services:
     volumes:
       - ./resources/deploy/db:/var/lib/postgresql/data
     command: [ "postgres", "-c", "log_statement=all", "-c", "log_destination=stderr" ]
+    ports:
```
MartinBelthle: What happend before you specified the ports ?

requirements.txt
```diff
@@ -41,6 +41,7 @@ plyer~=2.0.0
 psycopg2-binary~=2.9.9
 pyarrow~=21.0.0
 py7zr~=0.20.6
+prometheus-client==0.23.1
```
MartinBelthle, line 44: Minor: Is this really needed inside requirements or should this be in requirements-dev ?

MartinBelthle (commented):
Questions rather than changes requested

MartinBelthle:
Also I think you PR title is wrongly formatted due to the capital `I`

docker-compose.yml
sylvlecl: I will remove that, it has nothing to do with the PR

requirements.txt
sylvlecl: It's indeed needed in the production code (all the metrics definition etc)

antarest/core/metrics.py
sylvlecl: Indeed here it's basically the implementation module for all metrics, using prometheus which is the only option for now.
For the specific case of tasks service, I added an interface so that the metrics-related code is in that implementation file, and not clutter the actual service code.

I chose to put a prometheus-specific sub-config because at some point we may want to add other possibilities, in particular with opentelemetry. I propose to decide that day if we want to add an abstraction layer: it may not be necessary because opentelemetry is already an abstraction over multiple metrics technologies.

TheoPascoli (dismissed):
Nice ! There is also Pydantic Logfire that is built on OpenTelemtry, maybe we can monitor some metrics with it as weel

sylvlecl:
> Nice ! There is also Pydantic Logfire that is built on OpenTelemtry, maybe we can monitor some metrics with it as weel

Indeed it will be worth to improve on this first work, in particular using opentelemetry and things built on it. logfire looks very feature-rich!
For now we don't have the backend to benefit from tracing, but we in the future or other people may benefit from this.

PR #2785, opened 2025-10-21
chore(test): clean mypy issues

TheoPascoli:
# what's left
  354 [arg-type] - Incompatible argument types
  179 [assignment] - Incompatible types in assignment
  90 [attr-defined] - Attribute not defined (mostly on mocks)
  70 [call-arg] - Missing or extra arguments in function calls
  48 [union-attr] - Attribute access on union with None
  41 [index] - Indexing issues
  27 [method-assign] - Invalid method assignment
  26 [dict-item] - Incompatible dictionary item types
  16 [operator] - Operator not supported for these types
  15 [call-overload] - No overload variant matches
  13 [list-item] - Incompatible list item types
  13 [comparison-overlap] - Non-overlapping comparison types
  7 [no-redef] - Name redefinition
  5 [no-any-return] - Function returns Any
  4 [type-var] - Type variable issues
  3 [return-value] - Incompatible return value type
  3 [empty-body] - Empty function body
  2 [valid-type] - Invalid type
  2 [override] - Method override issues
  1 [misc] - Miscellaneous error
  1 [func-returns-value] - Function should return a value


sylvlecl (approved):
Great first step

PR #2788, opened 2025-10-22
refactor(ui): migrate `Layers` and `Districts` to glide data grid

hdinia:


webapp/src/components/App/Singlestudy/explore/Modelization/Map/MapConfig/Districts/index.tsx
```diff
@@ -129,15 +137,16 @@ function Districts() {
...
-          <TableForm
-            key={JSON.stringify(defaultValues)}
-            defaultValues={defaultValues}
-            tableProps={{
-              columns,
-              colHeaders: (_, colName) => districtsById[colName].name,
-              selectionMode: "single",
-            }}
+          <DataGridForm
+            key={JSON.stringify(defaultData)}
```
hdinia, line 141: What you think of this @skamril ? It was like this before but It's not recommended 
With Glide's performances it should not be an issue on large tables I guess

skamril: For me it's not an issue

skamril (approved):

PR #2791, opened 2025-10-23
refactor(output): adapt `/variables` endpoint to use new code

MartinBelthle:
[ANT-4031]


antarest/study/storage/output_model.py
```diff
@@ -84,3 +84,26 @@ def from_model(study_id: str, output_id: str, variables_list: OutputVariablesLis
...
+
+    @staticmethod
+    def from_variables_list(variables_list: OutputVariablesList) -> "OutputVariablesInformation":
+        args = {}
+
+        # Areas
+        all_area_variables = []
+        for area in variables_list.mc_ind.areas:
+            all_area_variables.extend(area.variables)
+        args["area"] = sorted(set(all_area_variables))
```
TheoPascoli: You can initialize `all_area_variables` as a set so its clear you dont want duplicates and you dont have to allocate extra memory to free it right after.
Also, the `sorted` method, is it for the tests or there is a business need ?

antarest/study/storage/output_model.py
```diff
@@ -84,3 +84,26 @@ def from_model(study_id: str, output_id: str, variables_list: OutputVariablesLis
             variables_list_version=1,
             variables_list=compressed_content,
         )
+
+
+class OutputVariablesInformation(AntaresBaseModel, extra="forbid"):
```
TheoPascoli, line 89: Cool model

tests/integration/test_integration.py
```diff
@@ -98,8 +98,8 @@ def test_main(client: TestClient, admin_access_token: str) -> None:
         f"/v1/studies/{study_id}/outputs/20201014-1427eco/variables",
         headers={"Authorization": f"Bearer {george_credentials['access_token']}"},
     )
-    assert res.status_code == 417
-    assert res.json()["description"] == "Not a year by year simulation"
+    assert res.status_code == 200
+    assert res.json() == {"area": [], "link": []}
```
TheoPascoli, line 102: Why this change ? If this PR is a refactor, it should not change some behaviors

TheoPascoli (changes requested):

tests/integration/test_integration.py
MartinBelthle: The only change in the PR is to not raise when no `mc-ind` folder exists. I think it's better as we don't raise when no areas or no links are present

antarest/study/storage/output_model.py
MartinBelthle: The sorting is for the test and for reproducibility but it has no business logic

sylvlecl (approved):

PR #2796, opened 2025-10-24
feat(uv): use astral uv to manage project environment

TheoPascoli:


scripts/pre-start.sh
```diff
@@ -10,9 +10,9 @@ CUR_DIR=$(cd "$(dirname "$0")" && pwd)
...
 
 cd "$BASE_DIR"
-alembic upgrade head
+uv run alembic upgrade head
 cd -
 
 export PYTHONPATH=$BASE_DIR
-python3 "$BASE_DIR/antarest/tools/admin.py" clean-locks -c "$ANTAREST_CONF"
-python3 "$BASE_DIR/antarest/tools/admin.py" fix-tasks-status -c "$ANTAREST_CONF"
+uv run python "$BASE_DIR/antarest/tools/admin.py" clean-locks -c "$ANTAREST_CONF"
```
sylvlecl: I don't think we should update startup script to use uv, seems to me we should only use it in dev workflow

sylvlecl (commented):
Looks good to me!

sylvlecl (approved):

PR #2797, opened 2025-10-27
feat(aggregate): add custom file download expiration time

smailio:


antarest/study/web/output_blueprint.py
```diff
@@ -37,6 +37,7 @@
 logger = logging.getLogger(__name__)
 
 DEFAULT_EXPORT_FORMAT = Query(TableExportFormat.CSV, alias="format", description="Export format", title="Export Format")
+DEFAULT_DOWNLOAD_EXPIRATION_TIME = 60  # in minutes
```
sylvlecl, line 40: I think we'd better use seconds for consistency, although you named things explicitly which is good, it will still be awkward for the API user to have a timeout in minutes here while it's in seconds for other endpoints.

sylvlecl: Oh, ok, it's already in minutes in the file service, never mind you can ignore the comment.

sylvlecl (dismissed):

sylvlecl (approved):

PR #2799, opened 2025-10-27
fix(generaldata): allow reading with missing sections

MartinBelthle:
[ANT-3999]

### DIFF
--- a/antarest/study/dao/file/file_study_adequacy_patch_parameters_dao.py
+++ b/antarest/study/dao/file/file_study_adequacy_patch_parameters_dao.py
@@ -10,7 +10,7 @@
 #
 # This file is part of the Antares project.
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, cast
 
 from typing_extensions import override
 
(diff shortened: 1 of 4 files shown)


antarest/study/dao/file/file_study_adequacy_patch_parameters_dao.py
```diff
@@ -41,16 +38,21 @@ def get_impl(self) -> "FileStudyTreeDao":
...
+        data = self._get_adequacy_patch_parameters(file_study)
         new_content = serialize_adequacy_patch_parameters(file_study.config.version, parameters)
 
         # Include fields that are in the generaldata.ini file but not in the FileData class
         new_content.update({k: v for k, v in data.items() if k not in new_content})
 
-        file_study.tree.save(new_content, ADEQUACY_PATCH_PATH)
+        file_study.tree.save(new_content, ["settings", "generaldata", "adequacy patch"])
+
+    @staticmethod
```
sylvlecl: you can move it out of the class and make it a free function, it's weird to associate it with the class whereas it's not used

sylvlecl (dismissed):
Would prefer to have it as a free function, approved anyway

PR #2802, opened 2025-10-27
fix(tasks): remove irrelevant status updates

sylvlecl:
This PR removes 2 irrelevant task status updates:

- **timeouts:**
  the timeout status of a task was set when a **waiting client** hit a timeout,
  which does not make sense since it's not the task itself which timed out.
  In particular, the task execution continues and will anyway update the
  status at the end (either completed or failed)
- **cancels:**
  a task can actually be cancelled only when it has not started. Once it has
  started, the execution cannot be interrupted. Therefore, it makes not sense
  to update the status in that case, since it will be overriden at the end of the
  execution.

It also fixes the behaviour of task wait, which was not consistent:
in some cases we raised a `TimeoutError`, in others not.
This left the door open to weird behaviour, where one could consider a task
finished whereas it was not.


antarest/core/tasks/service.py
```diff
@@ -377,6 +371,17 @@ def _cancel_task(self, task_id: str, dispatch: bool = False) -> None:
                 )
             )
 
+    def _do_cancel_task(self, task_id: str) -> None:
+        task = self.repo.get_or_raise(task_id)
+        if task_id in self.tasks:
+            cancelled = self.tasks[task_id].cancel()
```
smailio, line 393: If `self.tasks[task_id].cancel()` returns False, this function just silently fails ?

sylvlecl: Well it's a no-op indeed.
It may be relevant to raise in that case so that the API does not return a 2**, good point.
Not sure what would be the best status code, maybe "409 Conflict" ?

smailio: I never used this but yes it seems appropriate.

sylvlecl: I propose to leave it as is just because it will be "difficult" to implement:
since the actual cancellation will not be executed by the worker which received the HTTP request, the request is actually an asynchronous one today. We don't wait for the result to check if it has indeed been cancelled.

I will just add a log.

antarest/core/tasks/service.py
TheoPascoli, line 1: `TaskStatus.TIMEOUT` is used no where in the code now. Should we keep it ?

antarest/core/tasks/service.py
```diff
@@ -377,6 +371,17 @@ def _cancel_task(self, task_id: str, dispatch: bool = False) -> None:
...
+            # Only udpating status when the task is actually cancelled.
```
TheoPascoli: small typo

antarest/core/tasks/service.py
```diff
@@ -424,17 +429,6 @@ def await_task(self, task_id: str, timeout_sec: int = DEFAULT_AWAIT_MAX_TIMEOUT)
                 time.sleep(2)
 
             logger.error(f"Timeout while awaiting task '{task_id}'")
-            stmt = (
```
TheoPascoli, line 427: When removing this, the http client does not receive error 408 for tasks that are not handled by the current worker. Is it intentionnal ? Maybe you can raise TimeoutError

TheoPascoli (changes requested):

antarest/core/tasks/service.py
sylvlecl: Yes, we need to keep it for backward compat with stuff in database

sylvlecl: (although it's very possible that there are none, we'd better be safe ... I'll add a comment on the value)

antarest/core/tasks/service.py
sylvlecl: Actually we already don't raise an error, so for example when a wait is requested through the API, we already don't return a 408, it seems to me :/

The behaviour is inconsistent between the 2 cases where the worker is the one executing the task or not: in the first case, we raise, in the other case, we don't ... 

So the method contract is not clear.

sylvlecl: I clarified the behaviour by raising a `TimeoutError` also in the case where we wait for another worker

TheoPascoli (dismissed):

PR #2804, opened 2025-10-27
fix(auto-archiver): fix bug `auto-archiver`

TheoPascoli:
Bug when retrieving study to archive.

### DIFF
--- a/antarest/study/storage/auto_archive_service.py
+++ b/antarest/study/storage/auto_archive_service.py
@@ -39,7 +39,6 @@ class AutoArchiveService(IService):
         self.output_service = output_service
         self.config = config
         self.sleep_cycle = self.config.storage.auto_archive_sleeping_time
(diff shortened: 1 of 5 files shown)


antarest/study/storage/auto_archive_service.py
```diff
@@ -62,7 +61,7 @@ def _try_archive_studies(self) -> None:
                 and last_activity < old_date
                 and (isinstance(study, VariantStudy) or not study.archived)
             ]
-        for study_id, is_raw_study in study_ids_to_archive[0 : self.max_parallel]:
+        for study_id, is_raw_study in study_ids_to_archive:
```
sylvlecl, line 64: To compensate for the removal of the parallel parameter, I think we m’d better just perform archivals sequentially, which means waiting for the archive task completion for each study

sylvlecl (approved):

PR #2808, opened 2025-10-28
feat(task_list_get_to_post): replacing post api method task_list by get

mehdiwahada:


antarest/core/tasks/web.py
```diff
@@ -41,10 +41,14 @@ def create_tasks_api(service: TaskJobService, config: Config) -> APIRouter:
     auth = Auth(config)
     bp = APIRouter(prefix="/v1", dependencies=[auth.required()])
 
-    @bp.post("/tasks", tags=[APITag.tasks])
+    @bp.post("/tasks", tags=[APITag.tasks], deprecated=True)
     def list_tasks(filter: TaskListFilter) -> Any:
         return service.list_tasks(filter)
 
+    @bp.get("/tasks", tags=[APITag.tasks])
+    def get_task_list(task_filter: TaskListFilter = Depends()) -> list[TaskDTO]:
```
sylvlecl: There is no reason to use the `Depends` annotation here, the filter would be received as a body parameter.

**BUT** actually, there is an issue that I had not identified:
the best practices in a **GET** endpoint is to have parameters not as a request body, but as request parameters.

sylvlecl (approved):

PR #2809, opened 2025-10-28
fix(rate-limit): migrate to redis.asyncio

insatomcat:
Problem
- Requests hitting the rate limiter crashed under Python 3.11 with uvloop due to aredis using
deprecated asyncio/uvloop APIs (e.g. create_connection(..., loop=...), leading to TypeError and
aredis ConnectionError).

Solution
- Keep Redis as the rate-limiting backend but switch to redis-py asyncio (redis.asyncio) which
supports Python 3.11.
- Update asgi-ratelimit to 0.10.0, which works with redis.asyncio, and require redis>=5.
- Initialize RedisBackend with a redis.asyncio client when available; keep a compatibility path
for older backends, and log+fallback to MemoryBackend only if initialization fails (so the app
doesn’t crash).

Files
- antarest/main.py: create redis.asyncio client and pass to RedisBackend with a
compatibility/legacy path; keep safe fallback.
- requirements.txt: bump asgi-ratelimit[redis] to 0.10.0; use redis>=5,<6, so using the current 6.4 is ok

Impact
- Production keeps using Redis for global, accurate rate limiting on Py3.11.
- No more crashes from aredis/uvloop incompatibility.
- MemoryBackend is only used as a last resort if Redis init fails (can be made hard-fail if
desired).

### DIFF
--- a/antarest/main.py
+++ b/antarest/main.py
@@ -261,14 +261,52 @@ def fastapi_app(
 
     # rate limiter
     auth_manager = Auth(config)
+    # Prefer Redis backend with redis-py asyncio on Py3.11; fall back to legacy signature and finally to memory
+    rate_limit_backend: Any
+    if config.redis is None:
+        rate_limit_backend = MemoryBackend()
+    else:
+        try:
+            # Use redis.asyncio client when available (compatible with Python 3.11)
+            try:
+                from redis.asyncio import Redis as AsyncRedis  # type: ignore
+
+                redis_client = AsyncRedis(
+                    host=config.redis.host,
+                    port=config.redis.port,
+                    db=1,
+                    password=config.redis.password,
+                )
+                try:
+                    # Newer asgi-ratelimit accepts a client instance
+                    rate_limit_backend = RedisBackend(redis=redis_client)  # type: ignore[arg-type]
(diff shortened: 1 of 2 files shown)


sylvlecl:
Thanks for identifying this issue and proposing a solution @insatomcat .

As for the solution, I tend to think that  we'd better move the responsibility of rate limiting to another service (for ex. nginx), instead of keeping stuck with a specific version of redis client "just" for the compat with "asgi-ratelimit".

insatomcat:
Thanks for the feedback! 🙏

I totally agree that moving rate limiting to another service like nginx would be a cleaner long-term approach.
That said, the current situation is that the feature is broken, and this patch restores functionality right away with minimal impact.

I think it makes sense to merge this fix as a first step — it unblocks users and buys time to implement the more robust solution later, without being mutually exclusive.

I’m not planning to work on the nginx-based approach myself, but this patch at least keeps things working in the meantime.

sylvlecl:
HI @insatomcat , we'll just remove that feature for now, see https://github.com/AntaresSimulatorTeam/AntaREST/pull/2848 
If needs arise, we'll reimplement it with more care.
Note that it was not even working with https.

PR #2816, opened 2025-10-29
test(ci): add more time for windows ci

TheoPascoli:
Windows CI often fails on some duration thresholds because runners can be pretty slow.

This PR generalizes the approach we already had for tasks waiting, where we multiply by 2
the durations just for windows. We now also do it only when in CI environment.


tests/integration/utils.py
```diff
@@ -31,7 +34,13 @@ def wait_for(predicate: Callable[[], bool], timeout: float = 10, sleep_time: flo
 
 
 IS_WINDOWS = os.name == "nt"
-TIMEOUT_MULTIPLIER = 2 if IS_WINDOWS else 1
+
+
+def is_windows_ci() -> bool:
```
sylvlecl, line 39: I think we can wrap everything in a `duration_threshold(0.2)`, so that we can just write everywhere in our tests:

```python
assert 0 <= duration <= duration_threshold(0.2), f"Duration is {duration} seconds"
```

sylvlecl: I pushed that

PR #2818, opened 2025-10-30
chore(api): misc cleanups

sylvlecl:
- define tags at the router level, not for each endpoint
- remove the use of `response_model` where irrelevant (and `-> Any`)
- raise some errors instead of returning them (probably did not work ?)
- remove some return for DELETE endpoints where we returned the ID that was received
  as input, which is not useful


antarest/study/web/output_blueprint.py
```diff
@@ -130,8 +127,8 @@ def get_output_time_index(
 
     @bp.post(
         "/studies/{study_id}/outputs/{output_id}/download",
-        tags=[APITag.study_outputs],
         summary="Get outputs data",
+        response_model=None,  # only pydantic models are supported as response model
```
MartinBelthle, line 137: Why not just removing this line ? We don't need it inside other endpoints that return a `Response`

tests/launcher/test_web.py
```diff
@@ -189,7 +189,20 @@ def test_get_job_log() -> None:
 @pytest.mark.unit_test
 def test_kill_job() -> None:
     service = Mock()
-    service.kill_job.return_value.to_dto.return_value = ""
+    service.kill_job.return_value.to_dto.return_value = JobResultDTO(
```
MartinBelthle, line 185: Why did this change ?

antarest/study/web/output_blueprint.py
sylvlecl: I think it's because we mix Response and a pydantic model, it makes `fastapi` raise an exception where it advises to use that solution :/

tests/launcher/test_web.py
sylvlecl: Before the PR, since we did not tell fastapi that the endpoint was returning a `JobResultDTO`, it did not validate the response.
Now it validates it, and in that test the service did not return a valid response (it did not respect the typing)

antarest/study/web/output_blueprint.py
MartinBelthle: Weird but okay :/

MartinBelthle (dismissed):

PR #2821, opened 2025-10-30
fix(launcher): fix launcher argument naming in /versions endpoint

sylvlecl:
The `/launcher/versions` endpoint allows to retrieve solver versions available for a given launcher.

This "launcher" argument is wrongly named as "solver", this PR fixes that.

### DIFF
--- a/antarest/launcher/web.py
+++ b/antarest/launcher/web.py
@@ -14,7 +14,7 @@ import logging
 from typing import List, Optional
 from uuid import UUID
 
-from fastapi import APIRouter
+from fastapi import APIRouter, Query
 
 from antarest.core.config import Config
 from antarest.core.filetransfer.model import FileDownloadTaskDTO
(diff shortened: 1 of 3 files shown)


tests/integration/launcher_blueprint/test_solver_versions.py
```diff
@@ -14,66 +14,60 @@
 from starlette.testclient import TestClient
 
 
-class TestSolverVersions:
-    """
-    The purpose of this unit test is to check the `/v1/launcher/versions` endpoint.
-    """
+def test_get_solver_versions(
```
sylvlecl, line 17: I just removed the unneeded class, please ignore whitespaces diffs when reviewing

tests/launcher/test_web.py
```diff
@@ -138,34 +137,32 @@ def test_get_solver_versions() -> None:
 
 
 @pytest.mark.parametrize(
-    "solver, status_code, expected",
+    "launcher",
     [
-        pytest.param("default", http.HTTPStatus.OK, ["1", "2", "3"], id="default"),
```
sylvlecl, line 143: 2 of the arguments were never changed, I removed them

tests/launcher/test_web.py
```diff
@@ -138,34 +137,32 @@ def test_get_solver_versions() -> None:
...
-        pytest.param("local", http.HTTPStatus.OK, ["1", "2", "3"], id="local"),
+        "default",
+        "slurm",
+        "local",
+    ],
+)
+@pytest.mark.parametrize(
+    "param_name",
+    [
+        "solver",  # deprecated
```
sylvlecl, line 150: We still test the old naming

skamril (dismissed):

antarest/launcher/web.py
```diff
@@ -157,18 +157,22 @@ def get_load(launcher_id: Optional[str] = None) -> LauncherLoadDTO:
 
     @bp.get(
         "/versions",
-        summary="Get list of supported solver versions",
+        summary="Get list of supported solver versions for the specified launcher",
     )
-    def get_solver_versions(solver: Optional[str] = None) -> List[str]:
+    def get_solver_versions(
+        launcher_id: str | None = None, solver: str | None = Query(deprecated=True, default=None)
```
smailio: fastapi recommends that "for new code and whenever possible, use Annotated as explained above. There are multiple advantages (explained below) and no disadvantages. 🍰"

https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#add-query-to-annotated-in-the-q-parameter

new code would be ; `solver: Annotated[str | None, Query(deprecated=True)] = None`

antarest/launcher/web.py
```diff
@@ -157,18 +157,22 @@ def get_load(launcher_id: Optional[str] = None) -> LauncherLoadDTO:
 
     @bp.get(
         "/versions",
-        summary="Get list of supported solver versions",
+        summary="Get list of supported solver versions for the specified launcher",
```
smailio, line 161: Maybe we can add example of returned values. Otherwise it's a bit confusing because solver versions also refers to triplet versions. The reader has to dig deeper in the code to understand what this is about.

smailio (changes requested):

antarest/launcher/web.py
sylvlecl: Indeed, I added an exampled in the return type definition

smailio (approved):

PR #2822, opened 2025-10-30
fix(build): add explicit type

smailio:
On windows mypy fails on  date_serializer.py

```sh
$ mypy antarest/study/storage/rawstudy/model/filesystem/matrix/date_serializer.py
antarest\study\storage\rawstudy\model\filesystem\matrix\date_serializer.py:155: error: Incompatible return value type (got "tuple[Index[int], DataFrame]", expected "tuple[Index[str], DataFrame]")  [return-value]
Found 1 error in 1 file (checked 1 source file)
```

### DIFF
--- a/antarest/study/storage/rawstudy/model/filesystem/matrix/date_serializer.py
+++ b/antarest/study/storage/rawstudy/model/filesystem/matrix/date_serializer.py
@@ -146,7 +146,7 @@ class DailyMatrixSerializer(IDateMatrixSerializer):
         df_date = df.iloc[:, 2:4]
         df_date.columns = pd.Index(["day", "month"])
         df_date["month"] = df_date["month"].map(IDateMatrixSerializer._MONTHS)
-        date = df_date["month"].astype(str) + "/" + df_date["day"].astype(str).str.zfill(2)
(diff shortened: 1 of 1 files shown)


antarest/study/storage/rawstudy/model/filesystem/matrix/date_serializer.py
```diff
@@ -146,7 +146,7 @@ def extract_date(self, df: pd.DataFrame) -> Tuple[pd.Index[str], pd.DataFrame]:
         df_date = df.iloc[:, 2:4]
         df_date.columns = pd.Index(["day", "month"])
         df_date["month"] = df_date["month"].map(IDateMatrixSerializer._MONTHS)
-        date = df_date["month"].astype(str) + "/" + df_date["day"].astype(str).str.zfill(2)
+        date: pd.Series[str] = df_date["month"].astype(str) + "/" + df_date["day"].astype(str).str.zfill(2)
```
smailio, line 149: without the explicit type mypy interprets this as "Any".

Then on linux by default mypy interprets it as a Serie of str and on windows as a serie of int, which is not compatible.

sylvlecl (approved):

PR #2823, opened 2025-10-30
feat(study): introduce roundtrip method 

MartinBelthle:
Just to give some perspective,
Benchmark performed on a local BP study that had its matrices in the matrixstore:

- **Areas** (hydro, thermals, renewables, short-term storages): 333 s
- **Links**: 74 s
- **Binding constraints**: 14 s
- **Scenario Builder**: 7 s
- **Total**: 428 s

Around 300 s are only allocated to saving matrices in TSV format

I reran the benchamrk today
New results:
- With denormalization: 285s (no idea why it dropped so much)
- Without denormalization: 41s


antarest/study/business/study_interface.py
```diff
@@ -138,4 +149,4 @@ def add_commands(self, commands: Sequence[ICommand], listener: Optional[ICommand
 
     @override
     def get_study_dao(self) -> ReadOnlyStudyDao:
-        return FileStudyTreeDao(self.file_study).read_only()
+        return FileStudyTreeDao(self.file_study, self._generator_matrix_constants, self._blob_service).read_only()  # type: ignore
```
TheoPascoli: This `type ignore` hide a typing problem. `FileStudyTreeDao` require non-`None` parameters. Is it wanted ?

antarest/study/dao/study_conversion/study_converter.py
```diff
@@ -0,0 +1,329 @@
...
+        for area_id in area_properties:
+            # Properties
+            self._new_dao.save_area_properties(area_id, area_properties[area_id])
+
+            # Ui
+            source_ui = areas_ui[area_id]
+            for layer, x in source_ui.layer_x.items():
+                y = source_ui.layer_y[layer]
+                color = source_ui.layer_color[layer]
+                color_rgb = (int(c) for c in color.strip(" ").split(","))
```
TheoPascoli: minor : its not a tuple, its a generator. You need to add tuple(...). Anyway pydantic does the conversion.

antarest/study/service.py
```diff
@@ -2581,3 +2590,30 @@ def get_study_data(self, uuid: str) -> StudyDataDTO:
         ##########################
 
         return StudyDataDTO.model_validate(obj)
+
+    def write_study_as_file_study(
```
TheoPascoli, line 2578: for information : this is only use in tests for the moment

antarest/study/dao/study_conversion/study_converter.py
```diff
@@ -0,0 +1,329 @@
...
+            self._new_dao.save_st_storage_lower_rule_curve(area_id, sts_id, low_id)
+
+            upper_id = self._matrix_service.create(self._source_dao.get_st_storage_upper_rule_curve(area_id, sts_id))
+            self._new_dao.save_st_storage_upper_rule_curve(area_id, sts_id, upper_id)
+
+            inflows_id = self._matrix_service.create(self._source_dao.get_st_storage_inflows(area_id, sts_id))
+            self._new_dao.save_st_storage_inflows(area_id, sts_id, inflows_id)
+
+            if self._study_version >= STUDY_VERSION_9_2:
+                self._new_dao.save_st_storage_additional_constraints(area_id, sts_id, constraints[sts_id])
```
TheoPascoli: `constraints[sts_id]` can raise a `KeyError`. You can handle it with `constraints.get(sts_id, []`

TheoPascoli (changes requested):
Nice work ! Just few questions. Sorry for the delay

antarest/study/service.py
MartinBelthle: Indeed it's because it will be a code we'll only use when exporting a DB study. For a file study, we'll just copy the files as it's significantly quicker

antarest/study/business/study_interface.py
MartinBelthle: Yep it requires non-None parameters for 2 specific methods. But for the vast majority of tests we don't need these methods so we don't give them when instanciating the FileStudyInterface it's easier. The only other way would be to always give Mock() inside tests but it's a bit painful to add everywhere. Let me know if you have other ideas

antarest/study/dao/study_conversion/study_converter.py
MartinBelthle: Indeed. I even checked if it was inside the dict as using the DAO method with an empty list will read and re-save the files so it's time consuming for no reason

antarest/study/dao/study_conversion/study_converter.py
MartinBelthle: I used tuple() as you suggested. Seems that my IDE still is not happy as i don't check there are exactly 3 values but who cares

antarest/study/business/study_interface.py
TheoPascoli: Make the parameters Optional in FileStudyTreeDao.__init__(), then add a check in the 2 methods that use them:
```

  if generator_matrix_constants is None:
      raise ValueError("generator_matrix_constants required for this operation")
```

  No more type: ignore, no Mock() needed in tests, clear error if dependency is missing.

antarest/study/dao/study_conversion/study_converter.py
TheoPascoli: Yes same for me. it works with 
```
r, g, b = (int(c) for c in color.strip(" ").split(","))
area_ui = AreaUI(x=x, y=y, color_rgb=(r, g, b))
```
still minor to me, as you want

antarest/study/business/study_interface.py
MartinBelthle: This behavior is the existing one before this PR. And Sylvain did not like it if I recall. @sylvlecl anything to say ?
I could also create a function inside tests that takes the study interface and instantiate them with mocks to just have to do it once instead of everywhere. This way nor more optionals and no type ignore. Because typing it as optional inside the FileStudyTreeDao even if you need it sometimes I'm not a big fan

antarest/study/dao/study_conversion/study_converter.py
MartinBelthle: I'll do this if i have to do another change in the PR

antarest/study/business/study_interface.py
sylvlecl: I agree that it should not be optional in the constructor.
The object needs it to behave correctly, as a whole.

It's more in tests that adaptations should be made, as you propose in your last comment @MartinBelthle . Indeed that's some boiler plate but it's the best solution.

MartinBelthle: I implemented this

TheoPascoli (approved):

PR #2824, opened 2025-10-30
fix(tasks): fix a memory leak

sylvlecl:
The dictionary we use for storing futures is never ever cleaned up.
It seems it has never been an issue in production, but it still can
be non-negligible, in particular because it probably stores a reference
to the function executed, which generally carry some state with them.

Removing the future from the dict at the end of the computation
should not hurt, since the status is updated in database, therefore
future "waiters" will see on first request that it has indeed completed.

### DIFF
--- a/antarest/core/tasks/service.py
+++ b/antarest/core/tasks/service.py
@@ -583,6 +583,10 @@ class TaskJobService(ITaskService):
             for listener in self._listeners:
                 listener.on_task_end(task_id, task_type, status)
 
+            # Task has been updated in database, we can safely remove it from running tasks
+            if task_id in self.tasks:
+                del self.tasks[task_id]
+
     def get_task_progress(self, task_id: str) -> Optional[int]:
         task = self.repo.get_or_raise(task_id)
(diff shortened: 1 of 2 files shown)


antarest/core/tasks/service.py
```diff
@@ -583,6 +583,10 @@ def _run_task(
             for listener in self._listeners:
                 listener.on_task_end(task_id, task_type, status)
 
+            # Task has been updated in database, we can safely remove it from running tasks
+            if task_id in self.tasks:
```
smailio, line 587: if the update above fail you catch the exception and merely log it so it's not that safe no ?

sylvlecl: Well in that case we are already not good because the status in database will never be updated anyway, so the tasks will indeed be in a "wrong" state.

That's probably another issue to fix indeed.

sylvlecl: (like we could store those failures in a list for retry later, but that will complexify things for that edge case)

smailio (approved):

PR #2825, opened 2025-10-31
feat(dao): add several missing methods

MartinBelthle:
Introduces the following methods that are needed for the PR https://github.com/AntaresSimulatorTeam/AntaREST/pull/2823:
- GET and SAVE hydro matrices
- GET link matrices
- GET user resources
- GET and SAVE load / reserves / misc_gen / solar /wind matrices

Also fixes [ANT-3992]


antarest/study/dao/file/file_study_hydro_dao.py
```diff
@@ -200,3 +202,99 @@ def save_hydro_allocation(self, area_id: str, allocation: HydroAllocation) -> No
...
+    def save_hydro_water_values(self, area_id: str, series_id: str) -> None:
+        file_study = self.get_file_study()
+        file_study.tree.save(series_id, ["input", "hydro", "common", "capacity", f"waterValues_{area_id}"])
+
+    @override
+    def save_hydro_mingen(self, area_id: str, series_id: str) -> None:
+        file_study = self.get_file_study()
+        file_study.tree.save(series_id, ["input", "hydro", "series", area_id, "mingen"])
+
+    def _get_hydro_matrix(self, url: list[str]) -> pd.DataFrame:
```
TheoPascoli: This could be a more global `get_matrix` method. The `Link` one is exactly the same

TheoPascoli (changes requested):

antarest/study/dao/file/file_study_area_dao.py
```diff
@@ -200,163 +193,49 @@ def _build_area_data_structure(
...
-        new_correlation = study_data.tree.get(["input", "hydro", "prepro", "correlation"])
-        new_correlation.setdefault("annual", {})
+        null_matrix = generator_matrix_constants.get_null_matrix()
+        null_scenario_matrix = generator_matrix_constants.get_null_scenario_matrix()
 
         new_area_data: JSON = {
             "input": {
                 "areas": {
                     "list": [area.name for area in config.areas.values()],
-                    area_id: {
```
TheoPascoli, line 231: We dont need all of this anymore ? It seems to work though

antarest/study/dao/file/file_study_dao.py
```diff
@@ -101,3 +103,8 @@ def get_comments(self) -> str:
     @override
     def save_comments(self, comments: str) -> None:
         self._file_study.tree.save({"settings": {"comments": comments.encode("utf-8")}})
+
+    def get_matrix(self, url: list[str]) -> pd.DataFrame:
+        node = self._file_study.tree.get_node(url)
```
TheoPascoli, line 113: I think you could add just a small docstring to this. Also, I'm not a big fan of assert, I think it could be better with a proper error handling with log message

antarest/study/dao/file/file_study_area_dao.py
```diff
@@ -434,6 +313,14 @@ def _delete_area_files(self, area_id: str, study_data: Any) -> None:
             study_data.tree.delete(["input", "hydro", "hydro", "leeway low", area_id])
             study_data.tree.delete(["input", "hydro", "hydro", "leeway up", area_id])
             study_data.tree.delete(["input", "hydro", "hydro", "pumping efficiency", area_id])
+            study_data.tree.delete(["input", "hydro", "hydro", "reservoir", area_id])
```
TheoPascoli, line 308: `file_study_area_dao.py` has a lot of repetitive `tree.delete()` calls that could be factorized into loops. Would make it cleaner and easier to maintain.
like 
```
to_delete = ["reservoir", ..., "pumping"]
for x in to_delete:
  file_study.tree.delete
```
Same for the `save`, you could split it in smaller methods. 
Tell me what you think

antarest/study/dao/file/file_study_area_dao.py
```diff
@@ -200,163 +193,49 @@ def _build_area_data_structure(
...
-                            "data": generator_matrix_constants.get_prepro_data(),
-                            "k": null_matrix,
-                            "settings": {},
-                            "translation": null_matrix,
-                        }
-                    },
-                    "series": {
-                        f"load_{area_id}": generator_matrix_constants.get_null_scenario_matrix(),
-                    },
+                    "prepro": _create_prepro_data(generator_matrix_constants, null_matrix, area_id),
```
TheoPascoli: Minor: this can be more refactored with another method like :
```
def _create_series_data(generator_matrix_constants: "GeneratorMatrixConstants",
    null_matrix: str,
    null_scenario_matrix: str,
    area_id: str,
    series_type: str,
) -> dict[str, Any]:    
return {
        "prepro": _create_prepro_data(generator_matrix_constants, null_matrix, area_id),
        "series": {f"{series_type}_{area_id}": null_scenario_matrix},
    }
```
where `series_type` is solar, load or wind.

and in the `build` method, you get for example
```
"load": _create_series_data(
                    generator_matrix_constants, null_matrix, null_scenario_matrix, area_id, "load"
                ),
```

as you wish

TheoPascoli (commented):
Looks like it works !
I added some suggestions, feel free to tell me what you think about them

antarest/study/dao/file/file_study_area_dao.py
MartinBelthle: The thing is I want to create in the next PR the method GET and SAVE for these matrices so they will disappear, we'll only have the prepro here as we don't want to have this in the DAO (these are deprecated data).

I actually added the DAO for these matrices so this comment is not relevant anymore

antarest/study/dao/file/file_study_area_dao.py
MartinBelthle: It works because I moved this logic inside the CreateArea command where I call several DAO methods now instead of just the `save_area`

antarest/study/dao/file/file_study_area_dao.py
MartinBelthle: Not sure to understand what's  the `save` method you're referencing.

And for the `.tree.delete()` I agree it's repetitive but IMO putting this in a loop will not make it easier to read.

antarest/study/dao/memory/in_memory_study_dao.py
TheoPascoli, line 1: When deleting an area, do you need to delete all the matrices as well ? 
You do 
```
self._area_names.remove(area_id)
self._area_ui.pop(area_id, None)
```
Do you also need to do:
```
del self._load[area_id]
del self._solar[area_id]
...
```

antarest/study/dao/file/file_study_user_resources_dao.py
```diff
@@ -38,6 +38,26 @@ def get_file_study(self) -> FileStudy:
...
+        assert blob_service is not None
+
+        user_path = self.get_file_study().config.study_path / "user"
+        if user_path.exists():
+            all_files = user_path.rglob("*")
+            expansion_folder = user_path / "expansion"
+            for file in all_files:
+                if file.is_file() and not file.is_relative_to(expansion_folder):
+                    content = file.read_bytes()
+                    file.as_posix()
```
TheoPascoli: This is useless here

antarest/study/dao/file/file_study_user_resources_dao.py
```diff
@@ -38,6 +38,26 @@ def get_file_study(self) -> FileStudy:
...
+                    blob_id = blob_service.save(content)
+                    yield UserResourceDataCreation(
+                        path=PurePosixPath(str(file)),
```
TheoPascoli: This is an absolute path to the file, maybe you should use a relative path. I'm not sure that it can leads to errors in this case though

antarest/study/dao/memory/in_memory_study_dao.py
MartinBelthle: There are thousands of issues inside the in memory dao, (not just in thie method), we should remove the links ,the clusters etc. I don't want to handle this in this PR as it's unrelated

antarest/study/dao/file/file_study_user_resources_dao.py
MartinBelthle: Indeed

antarest/study/dao/file/file_study_user_resources_dao.py
MartinBelthle: Mmmh you're right. I'll change this to relative to user path and add a test

TheoPascoli (approved):

PR #2828, opened 2025-10-31
fix(tasks): task status update on restart must be done only once

sylvlecl:
Currently, each worker changes tasks status from RUNNING/PENDING to FAILED.

This cannot work well in a distributed environment, it must be done only once,
or some tasks could be erroneously set to FAILED.

This PR introduces a new "admin" operation which is performed on startup
after migration and locks cleanup.

### DIFF
--- a/antarest/tools/admin_lib.py
+++ b/antarest/tools/admin_lib.py
@@ -11,16 +11,27 @@
 # This file is part of the Antares project.
 
 import logging
+import time
+from datetime import datetime, timezone
 from pathlib import Path
 
+from sqlalchemy import Engine, create_engine, update
+
 from antarest.core.config import Config
+from antarest.core.tasks.model import TaskJob, TaskStatus
 from antarest.core.utils.utils import get_local_path
 from antarest.launcher.adapters.slurm_launcher.slurm_launcher import WORKSPACE_LOCK_FILE_NAME
 
 logger = logging.getLogger(__name__)
 
 
-def get_config(config_path: Path) -> Config:
+def _create_engine(config: Config) -> Engine:
+    url = config.db.db_admin_url
+    assert url is not None, "db_admin_url can not be None"
+    return create_engine(url, echo=False)
+
+
+def _create_config(config_path: Path) -> Config:
     res = get_local_path() / "resources"
     config_obj = Config.from_yaml_file(res=res, file=config_path)
     return config_obj
@@ -39,7 +50,7 @@ def clean_locks_from_config(config: Config) -> None:
 
 def clean_locks(config: Path) -> None:
     """Clean app locks"""
-    config_obj = get_config(config)
+    config_obj = _create_config(config)
     clean_locks_from_config(config_obj)
 
 
@@ -47,7 +58,7 @@ def reindex_table(config: Path) -> None:
     import sqlalchemy
     from sqlalchemy import text
 
-    config_obj = get_config(config)
+    config_obj = _create_config(config)
     assert config_obj.db.db_admin_url is not None, "db_admin_url can not be None"
 
     engine = sqlalchemy.create_engine(str(config_obj.db.db_admin_url), echo=False)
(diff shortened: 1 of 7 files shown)


scripts/pre-start.sh
```diff
@@ -11,3 +15,4 @@ cd -
 
 export PYTHONPATH=$BASE_DIR
 python3 "$BASE_DIR/antarest/tools/admin.py" clean-locks -c "$ANTAREST_CONF"
+python3 "$BASE_DIR/antarest/tools/admin.py" fix-tasks-status -c "$ANTAREST_CONF"
```
sylvlecl, line 18: The operation is now done only once here.

antarest/tools/admin_lib.py
```diff
@@ -61,3 +72,34 @@ def reindex_table(config: Path) -> None:
         connection.execute(text("REINDEX INDEX study_pkey"))
         connection.execute(text("VACUUM ANALYSE rawstudy"))
         connection.execute(text("REINDEX INDEX rawstudy_pkey"))
+
+
+def fix_interrupted_tasks_status(engine_or_config_path: Engine | Path) -> None:
```
MartinBelthle: Why accept a Path or a engine here ? The real code always gives a Path so to me we should only accept this

tests/core/tasks/test_fix_tasks_status.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+    make_session = sessionmaker(bind=db_engine)
+    with make_session() as session:
+        session.add(task_job)
+        session.commit()
+    fix_interrupted_tasks_status(db_engine)
+    with make_session() as session:
+        now = datetime.now(timezone.utc).replace(tzinfo=None)
+        if status in [TaskStatus.RUNNING.value, TaskStatus.PENDING.value]:
+            update_tasks_count = (
```
MartinBelthle: Not sure if this is necessary as you only have 1 task in your test and you're checking it's set to failed in the following lines

antarest/tools/admin_lib.py
sylvlecl: Indeed it was just for testing but I will separate in 2 methods

tests/core/tasks/test_fix_tasks_status.py
sylvlecl: Indeed, I will simplify the test

sylvlecl: I changed the test, it's more verbose but I hope more straight forward to read

antarest/tools/admin_lib.py
```diff
@@ -61,3 +72,34 @@ def reindex_table(config: Path) -> None:
...
+        TaskJob.result_msg: "Task was interrupted due to server restart",
+        TaskJob.completion_date: datetime.now(timezone.utc).replace(tzinfo=None),
+    }
+    orphan_status = [TaskStatus.RUNNING.value, TaskStatus.PENDING.value]
+    logger.info("Setting status of interrupted tasks to FAILED.")
+    t0 = time.time()
+    with engine.connect() as connection:
+        stmt = update(TaskJob).where(TaskJob.status.in_(orphan_status)).values(updated_values)
+        connection.execute(stmt)
+        connection.commit()
```
TheoPascoli, line 90: `connection` has no `commit()` method. Only `Session` has it. 
```
 with engine.begin() as connection: 
      stmt = update(TaskJob).where(TaskJob.status.in_(orphan_status)).values(updated_values)
      connection.execute(stmt)
      # commit is automatic at the end of the 'with'
```

TheoPascoli (changes requested):

tests/core/tasks/test_fix_tasks_status.py
```diff
@@ -0,0 +1,115 @@
...
+
+    with make_session() as session:
+        tasks_in_progress = (
+            session.query(TaskJob)
+            .filter(TaskJob.status.in_([TaskStatus.RUNNING.value, TaskStatus.PENDING.value]))
+            .count()
+        )
+        assert tasks_in_progress == 0
+
+        fixed_pending = session.query(TaskJob).get("pending")
```
TheoPascoli: This is deprecated in SQLAlchemy 2.0. Use `session.get(TaskJob, "pending")`. There are a few more later on the code.

antarest/tools/admin_lib.py
sylvlecl: Here we use `connect()`, not `begin()`, so it requires an explicit `commit()`.

antarest/tools/admin_lib.py
```diff
@@ -61,3 +72,34 @@ def reindex_table(config: Path) -> None:
...
+    logger.info(f"Tasks status correctly updated in {time.time() - t0}s.")
+
+
+def fix_interrupted_tasks_status(config_file: Path) -> None:
```
TheoPascoli, line 94: I suggest error handling in this method in case of timeout, connexion error ...

sylvlecl: I think the global behaviour is OK, so I am not sure we should make things more complex:
if there is an exception here, the command line tool will return an invalid exit code, which will make the startup script fail, and will cause either a retry (the service will try to restart), or a human analysis if the problem persists.

antarest/tools/admin_lib.py
TheoPascoli: Yes this is indeed equivalent, my bad

TheoPascoli (approved):
Good work, nice catch on this weird behavior

PR #2829, opened 2025-10-31
fix(layer): bug update layer

TheoPascoli:
## Bug Fixes

- `POST /layers` now returns the id instead of the name. Fixing a bug where you would update a layer right after creating it. It needed the ID but was using the name that was previously returned.
- `PUT /areas/{area_id}/ui`: Frontend sent extra fields rejected by backend

### DIFF
--- a/webapp/src/services/api/studydata.ts
+++ b/webapp/src/services/api/studydata.ts
@@ -17,7 +17,7 @@ import {
   bindingConstraintModelAdapter,
 } from "../../components/App/Singlestudy/explore/Modelization/BindingConstraints/BindingConstView/utils";
 import type { StudyMapNode } from "../../redux/ducks/studyMaps";
-import type { UpdateAreaUi } from "../../types/types";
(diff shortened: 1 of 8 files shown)


webapp/src/services/api/studydata.ts
```diff
@@ -34,7 +34,12 @@ export const updateAreaUI = async (
   layerId: string,
   areaUi: UpdateAreaUi,
 ): Promise<string> => {
-  const res = await client.put(`/v1/studies/${uuid}/areas/${areaId}/ui?layer=${layerId}`, areaUi);
+  const payload: AreaUIUpdatePayload = {
```
sylvlecl, line 37: Shouldn't we also change the model upstream, to use this model instead of `UpdateAreaUi` everywhere ? 
Maybe @skamril has an opinion ?

Anyway we could merge it as a quick fix for release

sylvlecl (approved):
Approved for the quick fix

PR #2832, opened 2025-11-03
feat(build): add a change log generator

smailio:
Unfortunately this is not 100% iso with the process release. 

To generate the list of change I rely in the commits. Our standard so far is to use pull requets titles. Most of the the time it's identical but not always. I'm not sure how important it is to use the pull request title, maybe the commit message is enough ? 

You can still use this script, you'll just need to update the sscript by hard coding the list of change instead of retrieving them dynamically. 

To do so rewrite `get_commits_for_release()`so it returns the lines you got after drafting a new release.


TheoPascoli:
I wonder if we can just use a llm and ask him to collect all the commits from last release, and format it. We should try it

PR #2833, opened 2025-11-03
feat(output): add first variables view endpoint

MartinBelthle:
This PR doesn't tackle the storage issue


antarest/study/web/output_blueprint.py
```diff
@@ -530,4 +534,37 @@ def get_output_variables_list(uuid: str, output_id: str) -> OutputVariablesList:
...
+        summary="Fetches the variables view for a given output and a given configuration",
+    )
+    def get_output_variables_view(
+        uuid: str,
+        output_id: str,
+        type: OutputVariablesType,
+        variable_name: str,
+        frequency: MatrixFrequency,
+        area_id: str | None = None,
+        area_from_id: str | None = None,
```
sylvlecl, line 548: I notice that we are not completely consistent between use of `area1` and  `area_from`, but in the API it seems that we use more `area_from` which is consistent with this

antarest/study/storage/output_service.py
```diff
@@ -604,3 +615,63 @@ def get_output_variables_information(self, study_id: str, output_id: str) -> Out
...
+        download_name = f"aggregated_output_{study_id}_{output_id}{export_format.suffix}"
+        download_log = f"Exporting aggregated output data for study '{study_id}' as {export_format} file"
+        download_id = self.aggregate_output_data(
+            study_id,
+            output_id,
+            query_file,
+            frequency,
+            TableExportFormat.PARQUET,
+            [variable_name],
+            [id_to_consider],
```
sylvlecl: In the case where we ask for a variable of cluster "my_cluster" of area "my_area", it seems to me that we have lost the cluster information here, isn'it ? It seems that the data for all clusters will be returned ?

antarest/study/business/output/variables_management.py
```diff
@@ -168,3 +169,142 @@ def extract_variables_list(output_path: Path) -> OutputVariablesList:
...
+    variable_type: OutputVariablesType,
+    variable_name: str,
+    available_variables: OutputVariablesList,
+    area_id: str | None = None,
+    area_from_id: str | None = None,
+    area_to_id: str | None = None,
+    thermal_id: str | None = None,
+    renewable_id: str | None = None,
+    st_storage_id: str | None = None,
+) -> tuple[str, QueryFileType]:
```
sylvlecl: I think we'd better have a better return type here with a class hierarchy to have more explicit constraints about the identifiers that are expected for different types:
```python
class OutputIdentifier(ABC):
    @property
    @abstractmethod
    def variables_type(self) -> OutputVariablesType: ...

class ThermalClusterOutputIdentifier(OutputIdentifier):
    variables_type: Final[OutputVariablesType] = OutputVariablesType.THERMAL

    area_id: str
    cluster_id: str

class LinkOutputIdentifier(OutputIdentifier):
    variables_type: Final[OutputVariablesType] = OutputVariablesType.LINK

    area_from_id: str
    area_to_id: str

class AreaOutputIdentifier(OutputIdentifier):
    variables_type: Final[OutputVariablesType] = OutputVariablesType.AREA

    area_id: str
```
(we could remove the type if not useful)

antarest/study/business/output/variables_management.py
```diff
@@ -168,3 +169,142 @@ def extract_variables_list(output_path: Path) -> OutputVariablesList:
...
+    _checks_variables_view_arguments_coherence(
+        variable_type, output_id, area_id, area_from_id, area_to_id, thermal_id, renewable_id, st_storage_id
+    )
+
+    if variable_type == OutputVariablesType.LINK:
+        assert area_from_id is not None
+        assert area_to_id is not None
+        _checks_links_variables_view_coherence(output_id, available_variables, variable_name, area_from_id, area_to_id)
+        return f"{area_from_id} - {area_to_id}", MCIndLinksQueryFile.VALUES
```
sylvlecl: See comment above, would prefer to have a structure return type

tests/integration/raw_studies_blueprint/test_output_variables.py
```diff
@@ -150,3 +150,55 @@ def test_get_output_variables_imagrid_endpoint(client: TestClient, user_access_t
...
+
+def test_get_output_variables_view(client: TestClient, user_access_token: str, internal_study_id: str):
+    client.headers = {"Authorization": f"Bearer {user_access_token}"}
+    output_id = "20201014-1425eco-goodbye"
+    # Areas
+    body = {"type": "area", "variable_name": "OP. COST", "frequency": "weekly", "area_id": "de"}
+    res = client.get(f"/v1/studies/{internal_study_id}/output/{output_id}/variables-views/data", params=body)
+    assert res.json() == {"data": [[46452000.0, 46452000.0], [46452000.0, 46452000.0]], "columns": [1, 2]}
+    # Thermal clusters
+    body = {
```
sylvlecl: it's a little misleading to name it body, it will not be the body of the request but the query params

tests/integration/raw_studies_blueprint/test_output_variables.py
```diff
@@ -150,3 +150,55 @@ def test_get_output_variables_imagrid_endpoint(client: TestClient, user_access_t
...
+        "type": "thermal",
+        "variable_name": "NODU",
+        "frequency": "weekly",
+        "area_id": "de",
+        "thermal_id": "01_solar",
+    }
+    res = client.get(f"/v1/studies/{internal_study_id}/output/{output_id}/variables-views/data", params=body)
+    assert res.json() == {
+        "data": [
+            [167.0, 167.0],
```
sylvlecl: Actually here, we get the data for all clusters, not only 01_solar:
if you look at the output file, it has only 2 weeks of data, but 9 different clusters, hence the 18 rows here!

We also get data for `02_wind_on`, `03_wind_off`, etc

antarest/study/business/output/variables_management.py
```diff
@@ -168,3 +169,142 @@ def extract_variables_list(output_path: Path) -> OutputVariablesList:
...
+    variable_type: OutputVariablesType,
+    area_id: str,
+    thermal_id: str | None = None,
+    renewable_id: str | None = None,
+    st_storage_id: str | None = None,
+) -> QueryFileType:
+    error_msg = f"The variable '{variable_name}' does not exist for area '{area_id}' and type '{variable_type.value}'"
+    area_variables = available_variables.mc_ind.areas
+    for area_variable in area_variables:
+        if area_variable.name == area_id:
```
sylvlecl, line 310: I did not pay attention unfortunately but it seems that in the variables list we have stored the ID in "name" ?

We should rename those fields, otherwise it will be hard to maintain.
I think we can afford the breaking change in the API, but we need to be able to read the previous ones from the database or we will break things :(

antarest/study/business/output/variables_management.py
```diff
@@ -168,3 +169,142 @@ def extract_variables_list(output_path: Path) -> OutputVariablesList:
...
+def _checks_variables_view_arguments_coherence(
+    variable_type: OutputVariablesType,
+    output_id: str,
+    area_id: str | None = None,
+    area_from_id: str | None = None,
+    area_to_id: str | None = None,
+    thermal_id: str | None = None,
+    renewable_id: str | None = None,
+    st_storage_id: str | None = None,
+) -> None:
```
sylvlecl: Here you could return a valid "structure identifier" as proposed above, since we validate the consistency here

sylvlecl (changes requested):
- proposition to have a more structured representation for identifiers
- the handling of area items like thermal clusters needs to be fixed (returns all the clusters of the area, not only the one selected)
- there is an ambiguity between name and id in variables list, we need to fix this too
  It 'd better go in another PR though

antarest/study/web/output_blueprint.py
MartinBelthle: Indeed. I'll keep area_from as you said

antarest/study/business/output/variables_management.py
MartinBelthle: Indeed. Didn't touch this part in this PR as it's a bit unrelated; Should go in a next PR I think

antarest/study/business/output/variables_management.py
```diff
@@ -168,3 +172,213 @@ def extract_variables_list(output_path: Path) -> OutputVariablesList:
...
+    variable_name: str,
+    available_variables: OutputVariablesList,
+    area_id: str | None = None,
+    area_from_id: str | None = None,
+    area_to_id: str | None = None,
+    thermal_id: str | None = None,
+    renewable_id: str | None = None,
+    st_storage_id: str | None = None,
+) -> OutputIdentifier:
+    _checks_variables_view_arguments_coherence(
```
sylvlecl: Code would be simpler by returning here the identifier object, and provide it to the following methods, instead of passing again all the arguments

antarest/study/business/output/variables_management.py
```diff
@@ -168,3 +172,213 @@ def extract_variables_list(output_path: Path) -> OutputVariablesList:
...
+    area_id: str,
+    thermal_id: str | None = None,
+    renewable_id: str | None = None,
+    st_storage_id: str | None = None,
+) -> OutputIdentifier:
+    error_msg = f"The variable '{variable_name}' does not exist for area '{area_id}' and type '{variable_type.value}'"
+    area_variables = available_variables.mc_ind.areas
+    for area_variable in area_variables:
+        if area_variable.name == area_id:
+            if variable_type == OutputVariablesType.THERMAL:
```
sylvlecl: Here, if we already have the identifier object, we can use a match and use the right field:

```python
match output_id:
    case ThermalClusterOutputIdentifier():
         # use output_id.cluster_id etc
```

antarest/study/business/output/variables_management.py
sylvlecl: Still think it's better to return it here, and use it later in the code, to benefit as soon as possible from the structure

antarest/study/business/output/variables_management.py
```diff
@@ -168,3 +172,213 @@ def extract_variables_list(output_path: Path) -> OutputVariablesList:
...
+    query_file: Final[QueryFileType] = MCIndAreasQueryFile.DETAILS
+
+
+@dataclass(frozen=True)
+class RenewableClusterOutputIdentifier(ClusterOutputIdentifier):
+    query_file: Final[QueryFileType] = MCIndAreasQueryFile.DETAILS_RES
+
+
+@dataclass(frozen=True)
+class ShortTermStorageOutputIdentifier(ClusterOutputIdentifier):
```
sylvlecl: We'd better not name it cluster_id here, it's not a cluster.
It's better to a have a few more lines of code but better naming

sylvlecl (commented):
more code clarity suggestions

antarest/study/business/output/variables_management.py
```diff
@@ -168,3 +172,209 @@ def extract_variables_list(output_path: Path) -> OutputVariablesList:
...
+    area_variables = available_variables.mc_ind.areas
+    for area_variable in area_variables:
+        if area_variable.name == area_id:
+            match output_identifier:
+                case AreaOutputIdentifier():
+                    if variable_name in area_variable.variables:
+                        return
+                    raise OutputVariablesViewError(output_id, error_msg)
+                case ThermalClusterOutputIdentifier():
+                    attr = "thermal_clusters"
```
sylvlecl: why not just getting `area_variable.thermal_clusters` and assigning it to a list here?
It makes the typing very weal and things a little difficult to understand

antarest/study/business/output/variables_management.py
```diff
@@ -168,3 +172,209 @@ def extract_variables_list(output_path: Path) -> OutputVariablesList:
...
+                case RenewableClusterOutputIdentifier():
+                    attr = "renewable_clusters"
+                case ShortTermStorageOutputIdentifier():
+                    attr = "short_term_storages"
+                case _:
+                    raise OutputVariablesViewError(output_id, error_msg)
+            sub_id = output_identifier.get_sub_id_for_aggregation()
```
sylvlecl, line 324: we could name `variable` something like `cluster_variables`, in consistency with the class ?
it makes things a little difficult to follow to name it variable

sylvlecl (dismissed):

PR #2837, opened 2025-11-06
feat(launcher): add versions in the /launchers endpoint

TheoPascoli:
The front-end client want to be able to call only one endpoint to get both de launchers and the associated versions. Instead of calling `GET /launchers` then `GET /versions`
`GET /versions` is now deprecated

### DIFF
--- a/antarest/launcher/model.py
+++ b/antarest/launcher/model.py
@@ -306,7 +306,7 @@ class LauncherInfoDTO(AntaresBaseModel):
     name: str
     nb_cores: LauncherResourceRangeDTO
     time_limit: LauncherResourceRangeDTO
-    versions: list[str]
+    versions: list[SolverVersionStr]
 
 
 class LauncherListDTO(AntaresBaseModel):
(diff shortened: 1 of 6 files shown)


antarest/launcher/model.py
```diff
@@ -306,6 +306,7 @@ class LauncherInfoDTO(AntaresBaseModel):
     name: str
     nb_cores: LauncherResourceRangeDTO
     time_limit: LauncherResourceRangeDTO
+    versions: list[str]
```
sylvlecl: Should be a `SolverVersionStr` instead of a plain string

tests/integration/launcher_blueprint/test_launcher_local.py
```diff
@@ -44,6 +44,7 @@ def test_get_launchers(
                     "max": local_config.time_limit.max,
                     "default": local_config.time_limit.default,
                 },
+                "versions": ["700", "880"],
```
sylvlecl: Since it's a new field, we'd better switch to the correct form which is 7.0 or 8.8

sylvlecl (changes requested):
Minor: we'd better use the opportunity to have better typing and the "correct" format for those versions

sylvlecl (dismissed):
Nice thanks !

sylvlecl (approved):

PR #2838, opened 2025-11-06
refactor(study): remove study additional data

mehdiwahada:
What it modifies ? : putting the editor / author / horizon field into the Study class instead of the StudyAdditionalData class and adapting the methods using or modifying the class in question to bring those changes to Study instead

Why ? : StudyAdditionalData brought another level of complexity that was deemed unnecessary

How to test it ? : You can test the same way as the ticket #2700


antarest/study/service.py
```diff
@@ -911,13 +906,12 @@ def create_study(
...
             workspace=DEFAULT_WORKSPACE_NAME,
             path=str(study_path),
+            author=author,
+            editor=author,
             created_at=now_utc,
             updated_at=now_utc,
             version=f"{version or NEW_DEFAULT_STUDY_VERSION:ddd}",
-            additional_data=StudyAdditionalData(author=author, editor=author),
             directory_id=directory_id,
-            owner=owner,
```
sylvlecl, line 919: Here we don't fill in the owner anymore, this is one of the reasons some tests fail:
for example when a user list studies, it cannot see studies that are not owned by himself, so if the owner is not filled when he creates a study, he will not be able to see it.

alembic/versions/9du891hc3051_remove_study_additional_data.py
```diff
@@ -0,0 +1,32 @@
...
+depends_on = None
+
+def upgrade():
+    with op.batch_alter_table("study") as batch_op:
+        batch_op.add_column(sa.Column('editor', sa.String(length=255)))
+        batch_op.add_column(sa.Column('horizon', sa.String(), nullable=True))
+
+    bind = op.get_bind()
+    content = bind.execute(text("SELECT author, editor, horizon FROM study_additional_data"))
+    for row in content:
```
sylvlecl: This will make one database request for each study (more than 10 000 in production env), we should instead try to perfom the copy with only one SQL statement.
With something like (copy pasted from AI, to be tested):
```sql
UPDATE study s
INNER JOIN study_additional_data sad ON s.id = sad.id
SET s.editor = sad.editor;
```

This will be MUCH more efficient

antarest/study/model.py
```diff
@@ -309,7 +284,9 @@ class Study(Base):
     name: Mapped[Optional[str]] = mapped_column(String(255), nullable=True, index=True)
     type: Mapped[str] = mapped_column(String(50), index=True)
     version: Mapped[str] = mapped_column(String(255), index=True)
-    author: Mapped[Optional[str]] = mapped_column(String(255), nullable=True)
+    author: Mapped[str] = mapped_column(String(255), nullable=True)
```
sylvlecl: If it's nullable, it must be `Optional`, same for editor

antarest/study/model.py
```diff
@@ -309,7 +284,9 @@ class Study(Base):
...
+    editor: Mapped[str] = mapped_column(String(255), nullable=True)
+    horizon: Mapped[Optional[str]] = mapped_column(String)
```
sylvlecl: should be marked nullable since it's optional

antarest/study/model.py
```diff
@@ -328,12 +305,6 @@ class Study(Base):
     tags: Mapped[List[Tag]] = relationship(Tag, secondary=StudyTag.__table__, back_populates="studies")
     owner = relationship(Identity, uselist=False)
     groups = relationship(Group, secondary=StudyGroup.__table__, cascade="")
-    directory = relationship("Directory", uselist=False)
```
sylvlecl, line 331: Why is it removed ?

Seems it does not break any test though, so I guess it's safe to remove it

antarest/study/repository.py
```diff
@@ -190,9 +190,6 @@ def save(
         if metadata.owner:
             metadata.owner = session.merge(metadata.owner)
 
-        if metadata.additional_data:
```
sylvlecl, line 193: We should perform this for new fields, to keep consistent

antarest/study/storage/abstract_storage_service.py
```diff
@@ -91,12 +90,25 @@ def get_study_information(
...
             groups=[GroupDTO(id=group.id, name=group.name) for group in study.groups],
             public_mode=study.public_mode or PublicMode.NONE,
-            horizon=additional_data.horizon,
+            horizon=study.horizon,
             folder=folder_path or study.folder,
             tags=[tag.label for tag in study.tags],
             directory_id=study.directory_id,
         )
 
+    def _update_study_data_from_files(self, file_study: FileStudy, metadata: Study) -> None:
```
sylvlecl, line 100: Nice, the name is more clear now

antarest/study/storage/utils.py
```diff
@@ -84,9 +84,12 @@ def update_antares_info(metadata: Study, study_tree: FileStudyTree, update_autho
 
     author = "Unknown"
     editor = "Unknown"
-    if metadata.additional_data:
-        author = metadata.additional_data.author
-        editor = metadata.additional_data.editor
+
+    if metadata.editor != "Unknown":
```
sylvlecl: We can remove those if conditions, since anyway author and editor are initialized to that value

alembic/versions/9du891hc3051_remove_study_additional_data.py
```diff
@@ -0,0 +1,32 @@
...
+from sqlalchemy import text
+
+revision = "9du891hc3051"
+down_revision = "0b2141573d50"
+branch_labels = None
+depends_on = None
+
+def upgrade():
+    with op.batch_alter_table("study") as batch_op:
+        batch_op.add_column(sa.Column('editor', sa.String(length=255)))
```
sylvlecl: Should be nullable, the editor will not be defined for some existing studies (not all studies have additional data, and not all additional data have a non null editor)

sylvlecl (changes requested):
Looks good overall, a few requests to polish the PR

sylvlecl (dismissed):

PR #2846, opened 2025-11-10
refactor(output): do not use download mechanism inside variables endpoint

MartinBelthle:


antarest/study/storage/output_service.py
```diff
@@ -649,31 +681,32 @@ def get_output_variables_view(
...
+        self._task_service.await_task(task_id)
 
         # Transform the dataframe to have the expected format
         if cluster_id := output_identifier.get_sub_id_for_aggregation():
-            dataframe = pd.read_parquet(Path(download.path), columns=[MCYEAR_COL, variable_name, CLUSTER_ID_COL])
+            dataframe = pd.read_parquet(tmp_path, columns=[MCYEAR_COL, variable_name, CLUSTER_ID_COL])
             dataframe = dataframe[dataframe[CLUSTER_ID_COL] == cluster_id].drop(columns=[CLUSTER_ID_COL])
         else:
-            dataframe = pd.read_parquet(Path(download.path), columns=[MCYEAR_COL, variable_name])
+            dataframe = pd.read_parquet(tmp_path, columns=[MCYEAR_COL, variable_name])
```
sylvlecl: we need to make sure to remove the temporary file once we don't need it any more.

antarest/study/storage/output_service.py
```diff
@@ -485,7 +486,7 @@ def archive_output_task(notifier: ITaskNotifier) -> TaskResult:
 
         return task_id
 
-    def aggregate_output_data(
+    def aggregate_output_data_with_download(
```
sylvlecl: We should document that it returns the download ID.
We could also have a more explicit name like `create_aggregated_output_data_download`

antarest/study/storage/output_service.py
```diff
@@ -498,6 +499,46 @@ def aggregate_output_data(
...
+            columns_names,
+            ids_to_consider,
+            file_download_path,
+            download_id,
+            mc_years,
+        )
+
+        return download_id
+
+    def aggregate_output_data(
```
sylvlecl: Since it starts a task, we could rename it `start_aggregate_output_data` for example

sylvlecl (changes requested):
Looks good overall but we need to remove the temp file when it's not used anymore

sylvlecl (approved):

PR #2848, opened 2025-11-12
chore(web): remove rate limiter

sylvlecl:
Removes the marginal rate limiting feature that exists now. As is, it's:
- inactive in https
- broken with python 3.11 (see https://github.com/AntaresSimulatorTeam/AntaREST/pull/2809 by @insatomcat )
- not configurable (config is hard-coded, hidden)

If need arises, we will restore it with more care.


smailio (approved):
LGTM. I was able to run the app after uninstalling the ratelimit dependecy, checked there were no remaining usage in code.

TheoPascoli (approved):

PR #2849, opened 2025-11-12
refactor(ui-launcher): auto modification of field values ​​related to launcher selection

skamril:


webapp/src/components/App/shared/studies/dialogs/StudyLaunchDialog/Fields.tsx
```diff
@@ -57,8 +57,8 @@ function Fields() {
   );
 
   const configurationOptions = useMemo(
-    () => getConfigurationsOptionsForVersion?.(version) || [],
-    [getConfigurationsOptionsForVersion, version],
+    () => getConfigurationOptionsForVersion?.(version) || [],
```
smailio, line 60: I think the ? here is not relevant, getConfigurationOptionsForVersion is never undefined.

smailio (changes requested):
Nice job, I didn't see the  tooltip in previous PR, it's a good idea. 

As you're fixing thing related to the last PR I have a few suggestions. 

in `SelectFE.tsx`  here `const { emptyValueLabel = t("global.none") } = rest;` you don't remove emptyValueLavel from rest, so it's passed to the TextField component which causes an error that is logger in the web console. 

Also in Fields.txt this line 
```  
const [configuration, isXpansionEnabled, isSensitivityModeEnabled, launcherId, version] = watch([
    "configuration",
    "xpansion",
    "sensitivityMode",
    "launcher",
    "version",
  ]);
``` 
returns a configuration that is stale, which you would then try to set as the active option for a select, so a warning is logged in the web console.

webapp/src/components/App/shared/studies/dialogs/StudyLaunchDialog/Fields.tsx
skamril: `defaultValues` is obtained from an asynchronous function, so `getConfigurationOptionsForVersion` is undefined at startup.
The lib type didn't handle this.

skamril:
> Nice job, I didn't see the tooltip in previous PR, it's a good idea.
> 
> As you're fixing thing related to the last PR I have a few suggestions.
> 
> in `SelectFE.tsx` here `const { emptyValueLabel = t("global.none") } = rest;` you don't remove emptyValueLavel from rest, so it's passed to the TextField component which causes an error that is logger in the web console.
> 
> Also in Fields.txt this line
> 
> ```
> const [configuration, isXpansionEnabled, isSensitivityModeEnabled, launcherId, version] = watch([
>     "configuration",
>     "xpansion",
>     "sensitivityMode",
>     "launcher",
>     "version",
>   ]);
> ```
> 
> returns a configuration that is stale, which you would then try to set as the active option for a select, so a warning is logged in the web console.

I fixed the issue with the `emptyValueLavel` prop, thanks.

But I don't understand the issue with `configuration`.

smailio (dismissed):

smailio (dismissed):

smailio (approved):

PR #2861, opened 2025-11-20
fix(db,metrics): always explicitly close transactions on session exit

sylvlecl:
Currently, we don't explicitly rollback transactions before closing the session.

This does not make the behaviour very clear, and in particular it causes some
transactions to be incorrectly seen as "unclosed", when an exception occurs at
closing time.
This results, in prod env, in wrong metrics for "current transactions".

The new behaviour is to always explicitly either commit or rollback the current
transaction before closing the session. slqalchemy seems to hand errors at
rollback time better than errors at closing time.

### DIFF
--- a/antarest/core/filetransfer/repository.py
+++ b/antarest/core/filetransfer/repository.py
@@ -24,8 +24,9 @@ class FileDownloadRepository:
         db.session.commit()
 
     def get(self, download_id: str) -> Optional[FileDownload]:
-        file_download: Optional[FileDownload] = db.session.get(FileDownload, download_id)
-        return file_download
+        download = db.session.get(FileDownload, download_id)
+        db.session.refresh(download)
+        return download
 
     def save(self, download: FileDownload) -> None:
         db.session.merge(download)
--- a/tests/core/test_metrics.py
+++ b/tests/core/test_metrics.py
@@ -28,6 +28,7 @@ from antarest.core.metrics import (
     _add_metrics_middleware,
 )
 from antarest.core.tasks.model import TaskStatus, TaskType
+from antarest.core.utils.fastapi_sqlalchemy import db
 from antarest.main import add_exception_handlers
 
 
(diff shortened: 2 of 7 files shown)


antarest/core/filetransfer/repository.py
```diff
@@ -24,8 +24,9 @@ def add(self, download: FileDownload) -> None:
         db.session.commit()
 
     def get(self, download_id: str) -> Optional[FileDownload]:
-        file_download: Optional[FileDownload] = db.session.get(FileDownload, download_id)
-        return file_download
+        download = db.session.get(FileDownload, download_id)
+        db.session.refresh(download)
```
MartinBelthle, line 28: What does this achieve ?

tests/core/test_metrics.py
```diff
@@ -290,6 +291,29 @@ def test_db_connection_metrics_preping():
...
+
+    registry = CollectorRegistry()
+    _add_db_session_metrics(registry, session_factory)
+
+    try:
+        with db(session_factory):
+            db.session.execute(text("DROP TABLE IF EXISTS test"))
+            db.session.execute(text("CREATE TABLE test (id INTEGER PRIMARY KEY)"))
+
+            # Put a breakpoint here and restart postgres server before continuing
```
MartinBelthle, line 308: Not sure to understand

tests/core/test_metrics.py
```diff
@@ -290,6 +291,29 @@ def test_db_connection_metrics_preping():
...
+            assert _get_value(registry, "db_transactions_current") == 1
+    except Exception:
+        pass
```
MartinBelthle, line 311: Why do we need this

antarest/core/filetransfer/repository.py
sylvlecl: It forces to fetch the data from the database. This is the right solution compared to the previous one which was opening a new session just to get that data.

tests/core/test_metrics.py
sylvlecl: As discussed: 
those tests are skipped in CI because they need an actual server that we kill in the middle of the test.

I used that to check and demonstrate the behaviour, indeed it's not a great solution, but did not find a simple way to simulate that behaviour.

MartinBelthle (approved):

PR #2863, opened 2025-11-24
fix(db): remove nested sessions

sylvlecl:
Fixes CI on dev, which was broken by the change in sessions closing, which
now always rollbacks the connection.

It has revealed an issue in recent evolutions about output variables, which were
using nested sessions, and re-used objects from session to session.


MartinBelthle (approved):
Indeed I put `with db(): ` everywhere :innocent:

PR #2864, opened 2025-11-24
fix(user): always override data inside user folder

MartinBelthle:
[ANT-4142]


antarest/study/web/raw_studies_blueprint.py
```diff
@@ -290,10 +282,7 @@ def replace_study_file(
         uuid: str,
         path: PATH_TYPE = "/",
         file: bytes = File(default=None),
-        create_missing: bool = Query(
-            False,
-            description="Create file or parent directories if missing.",
-        ),  # type: ignore
+        create_missing: bool = True,  # Deprecated, always considered as True
```
sylvlecl: We can annotate it with deprecated

antarest/study/service.py
```diff
@@ -1498,36 +1491,28 @@ def _edit_study_using_command(
...
 
         file_relpath = PurePosixPath(url.strip().strip("/"))
-        file_path = study_service.get_study_path(study).joinpath(file_relpath)
-        create_missing &= not file_path.exists()
-        if create_missing:
-            context = self.storage_service.variant_study_service.command_factory.command_context
+        try:
+            tree_node = file_study.tree.get_node(list(file_relpath.parts))
+            commands.append(self._create_edit_study_command(tree_node, url, data, version))
+        except ChildNotFoundError:  # The file doesn't exist, we have to create it.
```
sylvlecl: It's a little weird that we have a specific "user" command for creation, but not for the edition of an existing one.
Wouldn'it be more consistent to add an additional case in the `_create_edit_study_command` which will create a `CreateUserCommand` resource as soon as `user` starts the path ?
This way, all editions of user resoruces will use that command, instead of having sometimes an `UpdateRawFile` command, depending on the parent state..

sylvlecl: Also, I see that `UpdateRawFile` does not use a blob_id, we should implement this to be consistent with `CreateUserResource` (but not related to that PR)

sylvlecl (commented):
See comment, I wonder if we should not change a little more the logic to use consistently the same command for all user data editions (in particular, this should still be supported for database studies, whereas generic `UpdateRawFile` shoud not)

antarest/study/web/raw_studies_blueprint.py
smailio: We can trigger an exception / return a 400, if it's set to Flase, otherwise using this endpoint could be confusing.

MartinBelthle: I'm gonna explicitely say it's deprecated but I don't want to raise in case it's set to False as it would be a breaking change for someone wanting to update his file.

antarest/study/service.py
MartinBelthle: As discussed I renamed the command `ReplaceUserResource` and I use it instead of `UpdateRawFile` when modifying an existing user resource

tests/variantstudy/test_command_factory.py
```diff
@@ -802,13 +802,13 @@
     ),
     pytest.param(
         CommandDTO(
-            action=CommandName.CREATE_USER_RESOURCE.value,
+            action=CommandName.REPLACE_USER_RESOURCE.value,
```
sylvlecl, line 805: Shouldn't we keep also the existing test in addition to that one ?

tests/variantstudy/model/command/test_create_user_resource.py
sylvlecl, line 1: We should rename the test file

PR #2865, opened 2025-11-24
perf(output): use Response instead of Pydantic object inside variables endpoint

MartinBelthle:


sylvlecl (approved):
Much faster! from 10s to 1.5s

PR #2868, opened 2025-11-25
refactor(output)!: use aggregation code inside `/download` imagrid endpoint

MartinBelthle:
### Breaking changes

They all concern the same `/download` endpoint
- The keyword `use_task` will always be considered False.
- Inside the endpoint body, `synthesis` will always be considered False
- Inside the endpoint body, `filterIn` and `filterOut` will be ignored

### Fix

When an area didn't exist in the input, it could not be requsted inside the outputs

### Benchmark

Performed on my local computer on an output of 1000 MC year, for 3 areas and hourly frequency

- Dev branch, takes around 175s. Memory ramps up continually to reach at the end around 60% of my RAM.
- This branch, takes around 190s. Memory stays really low until the end where it also reaches 60% of my RAM. Note: Now that we use polars inside aggregation it, seems that it's around 3% faster than `dev` branch


antarest/study/output/output_service.py
```diff
@@ -396,106 +406,111 @@ def export_task(notifier: ITaskNotifier) -> TaskResult:
...
 
-                return FileResponse(tmp_export_file, headers=headers, media_type=filetype)
+        finally:
+            for file_path in file_paths:
+                file_path.unlink(missing_ok=True)
 
-            else:
-                json_response = matrix.model_dump_json()
-                return Response(content=json_response, media_type="application/json")
+        return FileResponse(final_path, headers={"Content-Disposition": "inline"}, media_type="application/json")
```
TheoPascoli: `final_path` is never deleted after the response is sent. You can use a FastAPI BackgroundTask to clean it up. See `FileTransferManager.request_tmp_file` for a similar pattern.

TheoPascoli (changes requested):

antarest/study/output/output_service.py
MartinBelthle: Indeed, I now use `FileTransferManager.request_tmp_file` as you suggested and it cleans the tmp file after each request, thanks :)

antarest/core/serde/np_array.py
```diff
@@ -25,3 +25,6 @@ def _np_to_list(array: npt.NDArray[np.float64]) -> list[float]:
 
 
 NpArray: TypeAlias = Annotated[npt.NDArray[np.float64], PlainSerializer(_np_to_list), BeforeValidator(_list_to_np)]
+
+# The `BeforeValidator` uses a lot of memory so we won't use it inside the Imagrid `/download` endpoint
+NpArrayLowMemory: TypeAlias = Annotated[npt.NDArray[np.float64], PlainSerializer(_np_to_list)]
```
sylvlecl: In the PR about pydantic warnings, we have change the before validator so that it does not do anything when the input is an NP array.
I think this should solve that issue.

antarest/study/model.py
```diff
@@ -588,59 +589,27 @@ def inc_date(self, date: datetime) -> datetime:
...
     type: StudyDownloadType
-    years: Optional[List[int]]
+    years: list[int] = []
     level: StudyDownloadLevelDTO
-    filterIn: Optional[str]
-    filterOut: Optional[str]
-    filter: Optional[List[str]]
-    columns: Optional[List[str]]
-    synthesis: bool = False
+    filterIn: Optional[str]  # deprecated, we don't consider it
```
sylvlecl: We should mark them as deprecated so that it appears like so in the API doc

antarest/study/output/output_service.py
```diff
@@ -393,80 +404,106 @@ def export_task(notifier: ITaskNotifier) -> TaskResult:
...
+                    element_name = object_name
+                    if data.type == StudyDownloadType.LINK:
+                        element_name = "^".join(element_name.split(" - "))
+
+                    for year, year_group in object_group.groupby(MCYEAR_COL):
+                        variables_list = []
+                        for col in df_columns:
+                            # We're using arrays to avoid flooding memory
+                            output_data = year_group[col].to_numpy()
+                            splitted_col = col.split(" % ")
```
sylvlecl: We must externalize the concatenating and splitting of column headers in dedicated utility functions so that this implementation detail is not hidden that much.
It might also be re-used elsewhere in the future.

antarest/study/output/output_service.py
```diff
@@ -393,80 +404,106 @@ def export_task(notifier: ITaskNotifier) -> TaskResult:
...
+                            name, unit = splitted_col[0], splitted_col[1]
+                            variables_list.append({"name": name, "unit": unit or " ", "data": output_data})
+
+                        intermediary_dict.setdefault(element_name, {}).setdefault(str(year), []).extend(variables_list)
+
+            response = MatrixAggregationResultDTO.model_validate(
```
sylvlecl, line 483: Not sure why we go through dictionaries and `model_validate` instead of just using the pydantic models from the start ?
But not a real issue though

antarest/study/output/output_service.py
```diff
@@ -393,80 +404,106 @@ def export_task(notifier: ITaskNotifier) -> TaskResult:
...
-            export_name,
-            task_type=TaskType.EXPORT,
-            ref_id=study_id,
-            progress=None,
-            custom_event_messages=None,
-        )
-
-        return FileDownloadTaskDTO(file=export_file_download.to_dto(), task=task_id)
+                # Wait for the aggregation to end
+                self._task_service.await_task(task_id)
```
sylvlecl, line 454: Now we can just use the aggregation method of the `_storage`, which is synchronous and does not go through a task.

antarest/study/web/output_blueprint.py
```diff
@@ -141,50 +143,20 @@ def get_output_time_index(
...
-        response_model=None,  # only pydantic models are supported as response model
-    )
+    @bp.post("/studies/{study_id}/outputs/{output_id}/download", summary="Get outputs data")
     def output_download(
         study_id: str,
         output_id: str,
         data: StudyDownloadDTO,
-        request: Request,
-        use_task: bool = False,
+        request: Request,  # deprecated, we always return a JSON response
```
sylvlecl: This can be safely removed if we don't use it, it's not a breaking change (it's just injected by fastapi)

antarest/study/web/output_blueprint.py
```diff
@@ -141,50 +143,20 @@ def get_output_time_index(
...
+        use_task: bool = False,  # deprecated, we always consider it's False
```
sylvlecl: We need to mark it as deprecated so that it appears like so in the api.
It's also a breaking change, since we don't return the same thing as before in that case
--> we need to mark the PR as breaking change.

sylvlecl (changes requested):
A few change requests for code quality + using last merged refactos

tests/integration/raw_studies_blueprint/assets/output_downloads_list/res1.json
sylvlecl, line 1: That's a LOT of lines !! 
Can't we have either a smaller test, or maybe using digests like in previous tests ?

MartinBelthle: I wanted to have a test for hourly just because it's ultimately what Imagrid will use. I could divide the size by 4 to only check 1 area (currently we have 4). Not sure to understand what you mean by "using digests".

antarest/study/output/output_service.py
MartinBelthle: The issue with directly using this method is that i will have dataframes with possible columns mismatch as it doesn't go through the usual process. So I'll have to handle this myself. I could use the /variables endpoint to know in advance what will be the expected columns but that's a total new process

antarest/core/serde/np_array.py
MartinBelthle: I tested it and indeed it seems that it takes the same memory as `NpArrayLowMemory` so I removed the created class

antarest/study/output/output_service.py
MartinBelthle: Doesn't make a difference to me. The only thing is that I only import the base class and not the inner classes that are implementation details so perhaps it's a bit better this way, don't know. I'll keep it as it is unless it bothers you

antarest/study/output/output_service.py
MartinBelthle: I introduced such functions inside the `utils.py` file. Let me know if that's what you had in mind

antarest/study/output/output_service.py
sylvlecl: Ok indeed

sylvlecl (approved):

PR #2872, opened 2025-11-27
test(alembic): add a test for alembic migrations

MartinBelthle:


tests/core/test_alembic_migration.py
```diff
@@ -0,0 +1,32 @@
...
+        """db:
+  url: "sqlite:///memory"
+"""
+    )
+    try:
+        # This will fail if there's an issue inside our alembic migrations
+        upgrade_db(config_file)
+    finally:
+        # Remove the created DB
+        db_path = Path(__file__).parent.parent.parent / "memory"
```
sylvlecl: it's weird: why is there a file if we use an in-memory database ?

sylvlecl: I reverted to using a file database in order to add a minimal check after the migration.

PR #2877, opened 2025-11-28
CR18 Antares user club - Forced and planned outage counting for Thermal time series 

dusanparipovic:
# Work in progress
# Summary – Forced and Planned Outage Counting for Thermal Time Series

[Functional Specification ](https://rteinternational.sharepoint.com/:w:/r/sites/AntaresUser%27sClub-Interne/_layouts/15/Doc.aspx?sourcedoc=%7B26F8D920-C300-40A2-9242-9347CFE234D9%7D&file=CR18.docx&nav=eyJjIjo5OTkxOTcyNjV9&action=default&mobileredirect=true)
## To-Do

- [x] **Add user option for file generation**  
      Implement a **Yes/No toggle** that allows users to choose whether the CR18 output files should be generated in the user directory.

- [ ] **Confirm implementation approach with RTE**  
      RTE colleagues need to decide which implementation approach for CR18 should be adopted, based on the proposals described in the functional specification.


antarest/study/dao/api/study_dao.py
```diff
@@ -150,9 +116,8 @@ def get_file_study(self) -> FileStudy:
         """
         raise NotImplementedError()
 
-    @abstractmethod
-    def save_comments(self, comments: str) -> None:
-        raise NotImplementedError()
+    def get_study_path(self) -> Path:
```
MartinBelthle: You shouldn't have to introduce this

antarest/study/storage/rawstudy/model/filesystem/matrix/ts_generator_matrix.py
```diff
@@ -10,3 +10,9 @@
 #
 # This file is part of the Antares project.
 
+from antarest.study.storage.rawstudy.model.filesystem.matrix.input_series_matrix import InputSeriesMatrix
+
+
+class TsGeneratorMatrix(InputSeriesMatrix):
```
MartinBelthle: We discussed with the RTE team and we would like the `ts-generator` folder to be created inside the `user` folder.
Because the `input` folder should be used only for the Simulator needs which isn't the case of these data.

This will also simplify your PR as you don't have to introduce all the `build()` as the `user` folder automatically shows everything it contains inside the Debug View

antarest/study/web/study_data_blueprint.py
```diff
@@ -1195,47 +1370,51 @@ def set_advanced_parameters(uuid: str, field_values: AdvancedParametersUpdate) -
...
         return study_service.advanced_parameters_manager.update_advanced_parameters(study_interface, field_values)
 
+    # skok
     @bp.put(
         "/studies/{uuid}/timeseries/generate",
+        tags=[APITag.study_data],
         summary="Generate timeseries",
     )
-    def generate_timeseries(uuid: str) -> str:
+    def generate_timeseries(uuid: str, data: GenerateTimeSeriesRequestOutages) -> str:
```
MartinBelthle: IMo this should only be a flag instead of a class. The name could be `with_outage_details`

tests/core/test_tasks.py
```diff
@@ -523,6 +557,10 @@ def test_ts_generation_task(
     tasks = study_service.task_service.list_tasks(TaskListFilter())
     assert len(tasks) == 1
     task = tasks[0]
+    print("Task status:", task.status)
```
MartinBelthle: Remove these prints

tests/integration/study_data_blueprint/test_generate_thermal_cluster_timeseries.py
```diff
@@ -25,7 +25,11 @@
 class TestGenerateThermalClusterTimeseries:
     @staticmethod
     def _generate_timeseries(client: TestClient, user_access_token: str, study_id: str) -> TaskDTO:
-        res = client.put(f"/v1/studies/{study_id}/timeseries/generate")
+        res = client.put(
```
MartinBelthle: We should ensure that we didn't introduce a breaking change and that calling the code the old way still works so don't modify this code

tests/integration/study_data_blueprint/test_generate_thermal_cluster_timeseries.py
```diff
@@ -73,6 +77,26 @@ def test_lifecycle_nominal(self, client: TestClient, user_access_token: str) ->
         task = self._generate_timeseries(client, user_access_token, study_id)
         assert task.status == TaskStatus.COMPLETED
 
+        """
+        /input/thermal/clusters/area1_id/list.ini -> probably for adding one more test case to check correctness of outages counting 
+        """
+        # check outage generated files
```
MartinBelthle: Also check that we didn't generate anything inside the user folder when the flag is set to false

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -43,11 +47,20 @@ class GenerateThermalClusterTimeSeries(ICommand):
     Command used to generate thermal cluster timeseries for an entire study
     """
 
-    command_name: CommandName = CommandName.GENERATE_THERMAL_CLUSTER_TIMESERIES
+    generate_outage_files_thermal: bool = Field(
+        default=False
```
MartinBelthle: Check other commands such as `create_link`. You should introduce `    _SERIALIZATION_VERSION: Final[int] = 2` anda model_validate before method to parse old commands to fit with our code.

Also you should add a test inside `test_command_factory.py` to test this

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -118,10 +144,25 @@ def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener]
                     return command_failed(f"Area {area_id}, cluster {thermal.id.lower()}: " + e.args[0])
 
         # 11- Once we've written all matrices inside the matrix-store, modify the input folder.
+
+        if self.generate_outage_files:
```
MartinBelthle: What's the difference between `self.generate_outage_files_thermal` and ` self.generate_outage_files:` ?
 I'm confused.

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -118,10 +144,25 @@ def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener]
...
+            study_dir = study_data.get_study_path()
+            outage_dir = study_dir / "ts-generator" / "thermal"
```
MartinBelthle: You should do this, commands are supposed to be agnostic of the DAO implem. So you should use the `study_data.save_user_resource` method instead.

Also it means you shouldn't use the matrixstore but the blobstore lines 132 and 133

MartinBelthle (changes requested):

dusanparipovic:
Hi @flomnes @tbittar,

This approach for implementing CR18 works correctly on the web app side, but when running a simulation, the solver throws the following error:

`'outage-details-thermal': Unknown property`

The issue occurs because this datum is currently being written into the generaldata.ini file.

Our suggestion is to update the solver so it recognizes this property. It won’t be used during the simulation itself—its purpose is only for generating outage files within the web interface. Adding support for reading this parameter in the solver should not break any existing logic, and no additional updates would be required elsewhere.

If you have any alternative suggestions or see potential issues with this approach, please let us know. At the moment, we are writing this datum in generaldata.ini because it aligns with the existing time-series generation logic and avoids breaking current workflows.

dusanparipovic:
Hi @MartinBelthle,

Could you please recheck the new implementation approach when you have a free moment? It has been updated to align with your previous comments.
Here is the frontend code, so I think @skamril could also review this approach.

flomnes:
Yes we can ignore this field in antares-solver 9.4

MartinBelthle:
Not sure why we should store this data in the generaldata, seems to be a wrong practice as it has nothing to do with the Simulator itself. Why just setting the flag to True or False inside the API call for thermal generation is not enough ?

antarest/study/business/model/config/timeseries_config_model.py
```diff
@@ -15,6 +15,7 @@
 
 class TimeSeriesType(AntaresBaseModel, extra="forbid", validate_assignment=True, populate_by_name=True):
     number: int = 1
+    outage_details_thermal: bool = False
```
MartinBelthle, line 18: To me we shouldn't change this

antarest/study/storage/variantstudy/model/command/generate_thermal_cluster_timeseries.py
```diff
@@ -117,11 +137,27 @@ def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener]
...
                     return command_failed(f"Area {area_id}, cluster {thermal.id.lower()}: " + e.args[0])
 
-        # 11- Once we've written all matrices inside the matrix-store, modify the input folder.
+        # 11- Once we've written all matrices inside the matrix-store, modify the input and user folders.
+        outage_dir: Path | None = None
+        if outage_details_thermal:
+            file_study = study_data.get_file_study()
+            study_dir = file_study.tree.config.path
+            outage_dir = study_dir / "user" / "ts-generator" / "thermal"
+            outage_dir.mkdir(parents=True, exist_ok=True)
```
MartinBelthle, line 146: This is still wrong. You're assuming your study is in a FS format. But the whole point of what we're trying to do recently is to hide this inside the filestudy implementation and we're building a DB implem to store future studies. So commands shouldn't know which implem are given and just use DAO method, such as `save_user_resource` as I recommended to you in my last review

antarest/study/storage/variantstudy/model/command/remove_area.py
```diff
@@ -35,9 +43,211 @@ class RemoveArea(ICommand):
...
+            for key in list(ruleset):
+                symbol, *parts = key.split(",")
+                if (symbol in area_keys and parts[0] == self.id) or (
+                    symbol in link_keys and (parts[0] == self.id or parts[1] == self.id)
+                ):
+                    del ruleset[key]
+
+        study_data.tree.save(rulesets, ["settings", "scenariobuilder"])
+
+    def _remove_area_timeseries(self, study_data: FileStudy) -> None:
```
MartinBelthle, line 165: You're quite a bit late compared to `dev` branch. This should go in the FS implem not here

antarest/study/web/study_data_blueprint.py
```diff
@@ -1195,8 +1369,10 @@ def set_advanced_parameters(uuid: str, field_values: AdvancedParametersUpdate) -
         study_interface = study_service.get_study_interface(study)
         return study_service.advanced_parameters_manager.update_advanced_parameters(study_interface, field_values)
 
+    # skok
```
MartinBelthle, line 1372: ?

MartinBelthle (changes requested):
You have to write several tests otherwise your work will never be merged

dusanparipovic:
Hi @MartinBelthle,

Thank you for the quick response. I will add the required tests. I already have them locally, but I first wanted to check your thoughts on the implementation approach (hence the **Work in progress** status on this PR).

Before proceeding, I want to make sure I fully understand your position regarding the approach. From your last comments, it seems you prefer **not** introducing an additional parameter, in order to avoid breaking the current API blueprint. That means we should avoid sending any new parameter in the PUT request body.  
(Reference: [additional parameter inside PUT request body](https://github.com/rte-i/Anta_REST/blob/2d7ad391430c6517e8f5715ad328869ab6935312/tests/integration/study_data_blueprint/test_generate_thermal_cluster_timeseries.py))

If we decide to introduce a new parameter, it would require updating the API call (breaking the current contract) and propagating the parameter through the entire chain:  
`API endpoint → service method → task constructor (ThermalClusterTimeSeriesGeneratorTask) → task method → command constructor → command`.

I would also like to point out that the command responsible for generating the time series reads the number of thermal time series from the `generaldata.ini` file. This is the reason our current approach is structured as it is. 

That’s why I shared the [functional specification](https://rteinternational.sharepoint.com/:w:/r/sites/AntaresUser%27sClub-Interne/_layouts/15/Doc.aspx?sourcedoc=%7B26F8D920-C300-40A2-9242-9347CFE234D9%7D&file=CR18.docx&nav=eyJjIjo5OTkxOTcyNjV9&action=default&mobileredirect=true), to confirm whether this implementation aligns with your expectations, or if you would prefer us to adjust the approach.

Let me know your thoughts when you have a moment.

MartinBelthle:
Okay for the test part I understand.
For the `generaldata.ini`, it was a mistake at the time it got introduced and even if @flomnes says he can authorize it in a next version we shouldn't allow it. RTE decided that we should avoid if possible to put some data in the `input` folder that aren't real input for the Simulator.
Concerning the breaking change, I don't see why it would be one. You can add a flag inside the PUT `/timeseries/generate` endpoint like `generate_outage` that would be False by default and could be True for your use-case. It will go through the whole app and I'm okay with it. And we don't have to modify the GET or SET `timeseries/config` endpoints, the front-end will always show it to False by default and when the user put the value to True, the front end call the `/generate` endpoint with the right value

PR #2878, opened 2025-12-01
feat(celery): implement maintenance container

TheoPascoli:


antarest/maintenance/tasks/gc_matrix.py
```diff
@@ -0,0 +1,146 @@
...
+    start_time = time.time()
+    deleted_count = 0
+
+    logger.info("Beginning matrix GC process")
+    logger.info(f"Configuration: dry_run={dry_run}, retention_time={retention_time}s")
+
+    try:
+        with db():
+            logger.info(f"Attempting to acquire advisory lock {MATRIX_GC_LOCK_ID}")
+            result = db.session.execute(text("SELECT pg_try_advisory_lock(:lock_id)"), {"lock_id": MATRIX_GC_LOCK_ID})
```
sylvlecl: We should encapsulate the lock implementation in its own class, ideally a context manager, for example named `PostgresqlLock`. 
We should probably check that the sqlalchemy dialect is postgres, and raise an exception if not, at some point.

tests/maintenance/tasks/test_gc_matrix.py
```diff
@@ -0,0 +1,228 @@
...
+            mock_db.return_value.__exit__ = Mock(return_value=None)
+            mock_db.session = mock_session
+            yield mock_db
+
+    def test_returns_skipped_when_lock_not_acquired(self, mock_context, mock_db):
+        """Test that task returns 'skipped' status when advisory lock cannot be acquired."""
+        # Simulate lock not acquired
+        mock_result = Mock()
+        mock_result.scalar.return_value = False
+        mock_db.session.execute.return_value = mock_result
```
sylvlecl: The mocking gets complicated and very linked to implementation details here:
we'd better create a DistirbutedLock interface so that we can provide a mock or a simple implementation (for ex. just a threadlock based impl) which does not rely on such implementation details

requirements.txt
```diff
@@ -43,8 +43,13 @@ py7zr~=0.20.6
...
 SQLAlchemy~=2.0.41
 tables==3.9.2
 typing_extensions~=4.15.0
-xlsxwriter~=3.2.0
\ No newline at end of file
+xlsxwriter~=3.2.0
+
+# Celery for maintenance tasks scheduling
+celery[redis]~=5.5.0
+celery-types~=0.23.0
```
sylvlecl: Should go into requirements-dev

resources/application.yaml
```diff
@@ -60,3 +59,11 @@ debug: False
 #redis:
 #  host: localhost
 #  port: 6379
+
+# Celery configuration for maintenance tasks (optional, defaults to env vars)
+# celery:
+#   broker_url: "redis://localhost:6379/1"
```
sylvlecl: This could be discussed, but I really think we'd better re-use the information from the redis config above. In particular, we'll also need a password.

The DB name can be hard coded in the application.

I think it's better to have an opinionated setup where we decide that we use only one instance of redis, and we decide the name of the databases, instead of needing to configure every detail in multiple places.

antarest/celery/app.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+# Create Celery app instance
+celery_app = Celery("antarest-maintenance")
+
+# Configure from environment variables (fallback/default)
+# This allows Beat to work with just env vars
+celery_app.conf.update(
+    # Broker and backend
+    broker_url=os.getenv("CELERY_BROKER_URL", "redis://localhost:6379/1"),
```
sylvlecl: Not sure it's a good idea to introduce the possibility of overriding configuration with env variables this way:
if we do it, we should do it for the whole configuration, probably in an automated way in order to avoid that boiler plate code

antarest/celery/app.py
```diff
@@ -0,0 +1,154 @@
...
+    if not config_path.exists():
+        logger.error(f"Config file not found: {config_path}")
+        return
+
+    # Load config from YAML (only once, then pass to MaintenanceContext)
+    logger.info(f"Loading worker config from {config_path}")
+    res = get_local_path() / "resources"
+    config = Config.from_yaml_file(res=res, file=config_path)
+
+    # Override Celery config from YAML if present (YAML > env vars)
```
sylvlecl: It's often the opposite that is implemented (for ex. spring in java world), do we really want this ?

antarest/celery/context.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Provides access to all services needed by maintenance tasks:
+    - MatrixService
+    - BlobService
+    - ...
+
+    The context is initialized once per worker process in the worker_init signal.
+    """
+
+    _instance: Optional["MaintenanceContext"] = None
```
sylvlecl: we could make them capital letter, so that it's clear that they're class level

antarest/celery/context.py
```diff
@@ -0,0 +1,116 @@
...
+            return
+
+        with self._lock:
+            if self._initialized:
+                return
+
+            logger.info(f"Initializing MaintenanceContext from {config_path}")
+
+            # Import here to avoid circular dependencies
+            from typing import Dict, cast
```
sylvlecl: this should not be imported here

antarest/celery/context.py
```diff
@@ -0,0 +1,116 @@
...
+            logger.debug("MaintenanceContext already initialized")
+            return
+
+        with self._lock:
+            if self._initialized:
+                return
+
+            logger.info(f"Initializing MaintenanceContext from {config_path}")
+
+            # Import here to avoid circular dependencies
```
sylvlecl: Are you sure it's really needed ?
It does not seem to me that the application services will need to import celery-related stuff ? There should not be a circular dependency here

antarest/core/config.py
```diff
@@ -635,6 +635,37 @@ def from_dict(cls, data: JSON) -> "MetricsConfig":
...
+        broker_url: URL of the message broker (Redis)
+        result_backend: URL of the result backend (Redis or DB)
+        timezone: Timezone for scheduled tasks
+        result_expires: Time in seconds before task results expire
+        matrix_gc_interval: Interval in seconds between matrix GC runs
+    """
+
+    broker_url: str = "redis://localhost:6379/1"
+    result_backend: str = "redis://localhost:6379/1"
+    timezone: str = "UTC"
```
sylvlecl: Not sure why we would have a specific timezone for celery.
If we really need to provide some, it'd be better defined at the application level.

antarest/core/config.py
```diff
@@ -635,6 +635,37 @@ def from_dict(cls, data: JSON) -> "MetricsConfig":
...
+    result_expires: int = 86400  # 24 hours
+    matrix_gc_interval: int = 3600  # 1 hour
```
sylvlecl: Not sure we should move that feature specific config into the technical celery config

antarest/celery/__init__.py
sylvlecl, line 1: I feel it would be better to move it into the maintenance package, side to side with the tasks definitions, which are themselves celery-related.

tests/celery/test_context.py
```diff
@@ -0,0 +1,177 @@
...
+        ctx._initialized = True
+
+        mock_config = Mock()
+        # This should not raise any error and should return early
+        ctx.initialize(mock_config, Path("/fake/path"))
+
+        # Config should still be None since initialize returned early
+        assert ctx.config is None
+
+    @patch("antarest.service_creator.init_db_engine")
```
sylvlecl: Patching methods this way will make tests very hard to maintain, and refactoring very painful. As soon as we move or rename something, this will break the tests. We should think of unit tests that are less based on such implementation details

antarest/maintenance/tasks/gc_matrix.py
```diff
@@ -0,0 +1,147 @@
...
+    logger.info("Deleting unused saved matrices:")
+    for unused_matrix_id in unused_matrices:
+        logger.info(f"Matrix {unused_matrix_id} is set to be deleted")
+        if not dry_run:
+            logger.info(f"Deleting {unused_matrix_id}")
+            matrix_service.delete(unused_matrix_id)
+
+
+@celery_app.task(name="antarest.maintenance.tasks.clean_matrices_task")
+def clean_matrices_task() -> Dict[str, Any]:
```
sylvlecl: We'd better return a dataclass, not an opaque dict here.
Note that celery seems to support pydantic models as return types

tests/maintenance/tasks/test_gc_matrix.py
```diff
@@ -0,0 +1,228 @@
...
+
+            mock_matrix_service = Mock()
+            mock_ctx.matrix_service = mock_matrix_service
+
+            yield mock_ctx
+
+    @pytest.fixture
+    def mock_db(self):
+        """Create a mock db context manager."""
+        with patch("antarest.maintenance.tasks.gc_matrix.db") as mock_db:
```
sylvlecl: We must not patch/mock such internal details, again this will make future refactorings very very hard

antarest/maintenance/tasks/gc_matrix.py
sylvlecl: To improve testability and readability, the main logic should be implemented in a fucntion that explicitly states its arguments (dependency injection principle): matrix service etc.

Then, the method which is annotated as a celery task should do the work of getting the necessary stuff out of the worker context to pass it to the "main" method.

tests/maintenance/tasks/test_gc_matrix.py
```diff
@@ -0,0 +1,228 @@
...
+            unused_matrices=set(),
+            dry_run=False,
+        )
+
+        mock_matrix_service.delete.assert_not_called()
+
+
+class TestCleanMatricesTask:
+    @pytest.fixture
+    def mock_context(self):
```
sylvlecl: In methods that just test the logic of the GC, we should not need to instantiate celery related context, we must separate those 2 concerns.
The "logic" method should explicitly receives arguments it needs (matrix service etc), and not rely on an external singleton linked to the fact that we use it in a celery worker.

Then, the wiring of this method as a celery task can be another unit test, which just checks that the arguments are correctly extracted from the worker context.

tests/maintenance/tasks/test_gc_matrix.py
```diff
@@ -0,0 +1,228 @@
...
+        result = clean_matrices_task()
+
+        assert result["status"] == "skipped"
+        assert result["reason"] == "lock_not_acquired"
+        assert result["deleted_count"] == 0
+        assert "duration_seconds" in result
+
+    def test_returns_success_with_no_matrices_to_delete(self, mock_context, mock_db):
+        """Test successful execution when there are no matrices to delete."""
+        # Simulate lock acquired
```
sylvlecl: See other comments, this should be replaced by a real object in all tests

antarest/celery/app.py
```diff
@@ -0,0 +1,154 @@
...
+    This function is called automatically when Beat scheduler starts.
+    It registers all periodic maintenance tasks.
+    """
+    logger.info("Setting up periodic tasks")
+
+    # Import here to avoid circular imports and get the task signature
+    from antarest.maintenance.tasks.gc_matrix import clean_matrices_task
+
+    # Matrix Garbage Collector interval (default: 3600 seconds = 1 hour)
+    matrix_gc_interval = int(os.getenv("CELERY_MATRIX_GC_INTERVAL", "3600"))
```
sylvlecl: We'd better get it from configuration

antarest/maintenance/app.py
```diff
@@ -0,0 +1,154 @@
...
+    "Celery app created",
+    extra={
+        "broker": _mask_url_credentials(celery_app.conf.broker_url),
+        "backend": _mask_url_credentials(celery_app.conf.result_backend),
+    },
+)
+
+
+@worker_init.connect
+def init_worker(**kwargs: object) -> None:
```
sylvlecl: Note that this will be called only once for the worker, which makes the synchronization of the `MaintenanceContext` a little overkill (will not be useful in practice).

sylvlecl (commented):
Nice work! 

I added somme comments, I think in some places we should separate concerns more clearly in order to improve testability and furture possible refactorings. In particular, the tests are a LOT based on patching and mocking implementation details, which will make future refactorings very heard.

We should separate more:
- the task logic implementation from its wiring as a celery task
- the locking logic from the task logic

For the configuration, I think we'd better avoid to introduce env variables this way, because it will make unclear what can/should be configured by env var and what cannot.
If we want to do this (overriding config with en vars), it should be done at the whole config level, not only partially.

Also, we'd better re-use the existing redis config, even if we use a separate DB for celery.

scripts/start.sh
```diff
@@ -71,5 +71,29 @@ elif [ "$use_uvicorn" = true ]; then
   done
 else
   export PYTHONPATH=$BASE_DIR
-  python3 $BASE_DIR/antarest/main.py -c $ANTAREST_CONF --module "$1"
+
+  # Check if module is celery-related
+  case "$1" in
+    celery-beat)
+      echo "Starting Celery Beat scheduler..."
+      exec celery -A antarest.celery.app:celery_app beat \
```
sylvlecl: needs to be updated with last changes

antarest/maintenance/tasks/gc_matrix.py
```diff
@@ -0,0 +1,172 @@
...
+
+    return GCTaskResult(
+        status=TaskStatus.SUCCESS,
+        deleted_count=deleted_count,
+        duration_seconds=duration,
+        dry_run=dry_run,
+    )
+
+
+@celery_app.task(name="antarest.maintenance.tasks.clean_matrices_task")
```
sylvlecl: We need to add `pydantic=True` here, else we get an error

antarest/service_creator.py
```diff
@@ -297,10 +285,6 @@ def create_services(config: Config, app_ctxt: Optional[AppBuildContext], create_
     watcher = create_watcher(config=config, app_ctxt=app_ctxt, study_service=core_services.study_service)
     explorer_service = create_explorer(config=config, app_ctxt=app_ctxt)
 
-    matrix_garbage_collector = None
-    if config.server.services and Module.MATRIX_GC.value in config.server.services or create_all:
```
sylvlecl, line 342: Thinking that actually, we will still need this for the desktop version:
I think we need to restore it unfortunately, should not be too much of a problem since the cleanup method is celery-agnostic.
It will indeed use the `ThreadLock` you created :)

antarest/maintenance/app.py
```diff
@@ -0,0 +1,149 @@
...
+    accept_content=["json"],
+    enable_utc=True,
+    # Worker settings
+    worker_prefetch_multiplier=1,  # Don't prefetch (tasks can be long)
+    worker_max_tasks_per_child=100,  # Restart worker after 100 tasks
+    task_acks_late=True,  # Acknowledge after task completion
+    task_reject_on_worker_lost=True,  # Reject if worker crashes
+    # Timeouts
+    task_soft_time_limit=7000,  # 116 min soft limit (warning)
+    task_time_limit=7200,  # 2h hard limit (kill)
```
sylvlecl, line 100: If the task is killed after that time, it may be too short for some of the future tasks, like auto archiving, where archiving can take long.
(or we will need to ensure those long tasks are split ?)

antarest/core/utils/lock.py
```diff
@@ -0,0 +1,228 @@
...
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
+        pass
+
+
+class PostgresqlLock(DistributedLock):
```
sylvlecl, line 67: Nice :)

antarest/core/utils/lock.py
```diff
@@ -0,0 +1,228 @@
...
+    @override
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
+        if self._lock_acquired:
+            logger.debug(f"Releasing advisory lock {self._lock_id}")
+            self._session.execute(text("SELECT pg_advisory_unlock(:lock_id)"), {"lock_id": self._lock_id})
```
sylvlecl, line 136: I wonder what happens when we lose the connection.
It can happen, in particular, because idle connections are closed after 1h, in our env.

I am worried that it keeps locked.

sylvlecl: It seems it will be unlocked when the session is closed so that should be mostly OK.
But we can maybe have the case where a task lasts more than 1h and continues its work while the lock is released.

sylvlecl (changes requested):
Looks good overall :) 
There are minor things to fix (pydantic usage + startup script)

Then the only thing I think we need to do is restore the "in-process" maintenance task for the desktop version. But that could be done in another PR.

antarest/maintenance/app.py
TheoPascoli: I suggest we let it like this, and we overload it for long tasks (like auto archiving)

antarest/core/utils/lock.py
TheoPascoli: Yes. It would not cause issue for the matrix gc. However, for auto archiving, it can lead to some problems. One solution is to send an heartbeat periodically (`select 1`) to keep the connexion active. 
we can handle this in another PR

antarest/service_creator.py
TheoPascoli: Its done. Im not sure that I need to implement the `lock` though !

antarest/singleton_services.py
```diff
@@ -51,10 +50,6 @@ def _init(config_file: Path, services_list: List[Module]) -> list[IService]:
         watcher = create_watcher(config=config, app_ctxt=None, study_service=core_services.study_service)
         services.append(watcher)
 
-    if Module.MATRIX_GC in services_list:
```
sylvlecl, line 54: I'd prefer restoring this for the moment too.
We can remove the singleton_services as a whole when we have completely validated the celery-based.

antarest/main.py
```diff
@@ -269,8 +269,6 @@ def get_config() -> JwtSettings:
     if services.watcher and Module.WATCHER in config.server.services:
         services.watcher.start()
 
-    if services.matrix_gc and Module.MATRIX_GC in config.server.services:
```
sylvlecl, line 272: We should restore it so that it's started in "standalone" mode

antarest/matrixstore/matrix_garbage_collector.py
```diff
@@ -24,6 +31,13 @@
...
 
 class MatrixGarbageCollector(IService):
+    """
+    Background service that periodically cleans unused matrices.
+
+    This is used in non-Celery environments (desktop version) where
+    we can't rely on Celery Beat for scheduling.
+    """
+
     def __init__(self, matrix_service: MatrixService, sleeping_time: float, dry_run: bool, retention_time: int):
```
sylvlecl, line 39: I think we can use the function you have moved in the maintenance folder here

antarest/maintenance/app.py
```diff
@@ -0,0 +1,160 @@
...
+
+    logger.info("Setting up periodic tasks")
+
+    # Import here to avoid circular imports and get the task signature
+    from antarest.maintenance.tasks.gc_matrix_task import clean_matrices_task
+
+    matrix_gc_interval = _config.storage.matrix_gc_sleeping_time if _config else 3600
+
+    sender.add_periodic_task(
+        matrix_gc_interval,
```
sylvlecl, line 194: With this, the first execution of the task will be after "matrix_gc_interval", which can be a little late, in particular with the default config for matrices.
We'll need either to program a first execution earlier, or to make the delay shorter.

antarest/matrixstore/matrix_garbage_collector.py
sylvlecl: Done

antarest/main.py
sylvlecl: Done

antarest/singleton_services.py
sylvlecl: Done

antarest/maintenance/app.py
TheoPascoli: Done

sylvlecl (dismissed):
Working well :)

PR #2879, opened 2025-12-01
feat(tasks): change task status type

mehdiwahada:
What it modifies ? : it replaces the status type (an integer) by a string (it's name) in the Swagger UI

Why ? : To increase the readability of the status and it's usability (a user would be better typing the status name "PENDING" than "1")

How do you test it ? : in integration tests, instead of integers for the task statuses, you type their name (1 -> PENDING, 3 -> COMPLETED) in the client.get and you check if it gets the tasks with the statuses


tests/integration/test_tasks.py
```diff
@@ -42,3 +42,20 @@ def test_list_tasks(client: TestClient, user_access_token: str, internal_study_i
 
     assert len(res_task_list) == 2
     assert actual_task_list == expected_task_list
+
+    # pour les status, faire des GET récupérant les COMPLETED et les RUNNING (confirmer qu'il n'y en a pas)
```
TheoPascoli: You can remove this comment. We dont want french in the code in general, and I dont think this comment is relevant

antarest/core/tasks/model.py
```diff
@@ -59,6 +59,34 @@ def is_final(self) -> bool:
             TaskStatus.TIMEOUT,
         ]
 
+    @classmethod
+    def parse(cls, other: object) -> "TaskStatus":
+        if isinstance(other, TaskStatus):
+            return other
+        if isinstance(other, str):
+            if other in list(map(lambda c: c.name, cls)):  # cls.__members__:
```
TheoPascoli: After verification `cls.__members__` is more performant than recreating a list every time.

antarest/core/tasks/model.py
```diff
@@ -59,6 +59,34 @@ def is_final(self) -> bool:
             TaskStatus.TIMEOUT,
         ]
 
+    @classmethod
+    def parse(cls, other: object) -> "TaskStatus":
```
TheoPascoli, line 63: You should add unit test for this method. It handles many different types and we want to be sure that it works. You can try to parse a `TaskStatus`, a `str`, an `int`, an invalid type (like a `float`), and an invalid `str`.

TheoPascoli (changes requested):

tests/core/tasks/test_tasks_service.py
```diff
@@ -704,3 +705,43 @@ def dummy_task(notifier: ITaskNotifier) -> TaskResult:
         # accessing an implementation detail, but no other way to check it
         assert task_id not in task_service.tasks
         assert task_service.status_task(task_id).status == TaskStatus.COMPLETED
+
+
+def test_task_status_parsing() -> None:
+    # Parsing multiple primitive types and TaskStatus
+
+    # Preparing invalid task status and error messages
```
sylvlecl: Minor: You should hard-code those values in the actual test lines at the end of the test, to make it more readable

tests/core/tasks/test_tasks_service.py
```diff
@@ -704,3 +705,43 @@ def dummy_task(notifier: ITaskNotifier) -> TaskResult:
...
+    invalid_task_status = "INVALID_STATUS"
+    invalid_float_task_status = 4.2
+    float_type = type(invalid_float_task_status)
+    error_str_message = "invalid literal for int() with base 10: 'INVALID_STATUS'"
```
sylvlecl: It's surprising to have this message here, we should simply have a message stating that it's not a valid value for TaskStatus (it's important to provide understandable messages for the user of the API :) )

antarest/core/tasks/model.py
```diff
@@ -59,6 +59,34 @@ def is_final(self) -> bool:
...
+        if isinstance(other, TaskStatus):
+            return other
+        if isinstance(other, str):
+            if other in cls.__members__:
+                return cls[other]
+            return cls(int(other))
```
sylvlecl: Here, if the value is not an int, it would be more clear to just raise with a `ValueError`

tests/core/tasks/test_tasks_service.py
```diff
@@ -704,3 +705,43 @@ def dummy_task(notifier: ITaskNotifier) -> TaskResult:
...
+    assert task_status == TaskStatus.RUNNING
+
+    # Putting an invalid type (float) in the parse method
+    with pytest.raises(
+        TypeError,
+        match=f"Invalid status type: {float_type}",
+    ):
+        TaskStatus.parse(4.2)
+
+    # Putting an invalid type (float) in the parse method
```
sylvlecl: This comment is misleading, you are not inputing a float here but a string.

As mentioned above, it would be far more clear to have the values here instead of defning them at the start of the test

sylvlecl (changes requested):
Nice work, I just added a few requests to polish the test and the error messages

sylvlecl (dismissed):

PR #2880, opened 2025-12-01
perf(matrix): use `polars` library

MartinBelthle:
Use `polars` library to read `csv` input matrices and to save `csv` matrices

**Benchmark:** 
- Reading a 8760*200 matrix: ~ 100ms less
- Importing a BP study:  From ~ 40s to ~ 30s
- Exporting a BP study: From ~ 280s to ~ 105s


tests/storage/integration/test_STA_mini.py
```diff
@@ -82,15 +82,10 @@ def assert_url_content(client: TestClient, url: str, expected_output: dict[str,
     assert_study(res.json(), expected_output)
 
 
-def assert_with_errors(
```
MartinBelthle, line 85: The only change in the file is that I switched the test with formatted=False from the storage service to the real client

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
```diff
@@ -17,7 +17,8 @@
 
 import numpy as np
 import pandas as pd
-from pandas.errors import EmptyDataError
+import polars as pl
+from polars._plr import ComputeError, NoDataError
```
sylvlecl: Aren't they exposed in a public package ?

sylvlecl (commented):
just a question about using private polar package

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
MartinBelthle: Indeed, I should be able to do : `from polars.exceptions import ComputeError, NoDataError`

PR #2881, opened 2025-12-01
feat(compression_seven_zip): added compression parameter to the GET request and the export_study method

mehdiwahada:
What it does / modifies ? : it adds another compression mode (.7z) in the swagger UI and the code to permit the user to choose between .zip and .7z

Why ? : .7z use the similarities between file to reduce the compressed file size

How to test it ? : The user must add the "compression=" parameter in the GET request in integration in order to use the .7z compression, otherwise they'll just use .zip, due to the fact it's a default parameter


antarest/study/web/studies_blueprint.py
```diff
@@ -382,11 +383,13 @@ def get_study_matrix_index(uuid: str, path: str = "") -> MatrixIndex:
         "/studies/{uuid}/export",
         summary="Export Study",
     )
-    def export_study(uuid: str, no_output: Optional[bool] = False) -> FileDownloadTaskDTO:
+    def export_study(
```
sylvlecl, line 386: In the method signature you don't need to make the compression optional, since it has a defaut value

tests/integration/studies_blueprint/test_study_export.py
```diff
@@ -0,0 +1,34 @@
...
+    # Doing exports with both .zip and .7z
+    def test_export_study_with_both_compression(
+        self, client: TestClient, internal_study_id: str, user_access_token: str
+    ) -> None:
+        client.headers = {"Authorization": f"Bearer {user_access_token}"}
+
+        # exporting without any parameters (thus exporting in .zip by default)
+        export_res = client.get(f"/v1/studies/{internal_study_id}/export")
+        export_json_res = export_res.json()
+        assert ".zip" in export_json_res["file"]["filename"]
```
sylvlecl: Nice to have added this test for backwards compat !

tests/integration/studies_blueprint/test_study_export.py
```diff
@@ -0,0 +1,34 @@
...
+
+        # exporting with the .7z compression
+        export_res = client.get(f"/v1/studies/{internal_study_id}/export?compression=.7z")
+        export_json_res = export_res.json()
+        assert ".7z" in export_json_res["file"]["filename"]
+
+        # exporting with an unauthorized compression
+        failed_export_res = client.get(f"/v1/studies/{internal_study_id}/export?compression=.rar")
```
sylvlecl: nice too to have added that error case

tests/integration/studies_blueprint/test_study_export.py
```diff
@@ -0,0 +1,34 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+from starlette.testclient import TestClient
+
+
+class TestStudyExport:
```
sylvlecl: A suggestion:
with pytest, having such classes to embed unit test is not very useful, we can have plain functions instead.

I know in many pplaces we have such classes, but it's not a good practice, it makes things harder to read with no added value.

tests/integration/studies_blueprint/test_study_export.py
```diff
@@ -0,0 +1,34 @@
...
+
+        # exporting without any parameters (thus exporting in .zip by default)
+        export_res = client.get(f"/v1/studies/{internal_study_id}/export")
+        export_json_res = export_res.json()
+        assert ".zip" in export_json_res["file"]["filename"]
+
+        # exporting with the .7z compression
+        export_res = client.get(f"/v1/studies/{internal_study_id}/export?compression=.7z")
+        export_json_res = export_res.json()
+        assert ".7z" in export_json_res["file"]["filename"]
```
sylvlecl: We must check that the content of the response is indeed a valid 7zip file

tests/integration/studies_blueprint/test_study_export.py
sylvlecl: We should also check here it's a valid zip file

antarest/study/service.py
```diff
@@ -1281,8 +1283,14 @@ def export_study(
...
         export_name = f"Study {study.name} ({uuid}) export"
+
+        archive_format = ArchiveFormat.ZIP
+
+        if compression in list(map(lambda arch_format: arch_format, ArchiveFormat)):
+            archive_format = compression
+
         export_file_download = self.file_transfer_manager.request_download(
-            f"{study.name}-{uuid}{ArchiveFormat.ZIP}", export_name
+            f"{study.name}-{uuid}{archive_format}", export_name
```
sylvlecl, line 1288: It seems that you change the file name here, but you did not actually create a 7z file instead of the zip file.

In order to do this you need to push the compression argument lower in the services: method `export_study` of the `IStudyStorage` etc.
Otherwise, you will get a file named ".7z", but with the same zip compression as today :)

sylvlecl (changes requested):
The API change looks good ! In particular backwards compat is managed, good thing.
But I think you need to push the compression argument lower so that it **actually** creates a 7z and not a zip

antarest/study/service.py
```diff
@@ -1281,16 +1283,24 @@ def export_study(
 
         logger.info("Exporting study %s", uuid)
         export_name = f"Study {study.name} ({uuid}) export"
+
+        compression_format = ArchiveFormat.ZIP
```
sylvlecl: Not sure why you need that variable, is it not exactly the same as the `archive_format` argument ?

antarest/study/storage/abstract_storage_service.py
```diff
@@ -300,14 +300,17 @@ def import_output(
         return output_full_name
 
     @override
-    def export_study(self, metadata: Study, target: Path, outputs: bool = True) -> Path:
+    def export_study(
+        self, metadata: Study, target: Path, outputs: bool = True, archive_format: Optional[ArchiveFormat] = None
```
sylvlecl: we'd better have it non-optional and default to zip like the endpoint.
Generally speaking, when we can avoid optional, it's better not to use it, it makes the behaviour less clear (what will be the behaviour for None ? less clear than if the default is ZIP)

antarest/study/storage/study_storage.py
```diff
@@ -148,14 +149,17 @@ def _check_study_exists(self, metadata: Study) -> None:
             raise StudyNotFoundError(f"Study with the uuid {metadata.id} does not exist.")
 
     @abstractmethod
-    def export_study(self, metadata: Study, target: Path, outputs: bool = True) -> Path:
+    def export_study(
+        self, metadata: Study, target: Path, outputs: bool = True, archive_format: Optional[ArchiveFormat] = None
```
sylvlecl: Same let's not use an optional here

tests/integration/studies_blueprint/test_study_export.py
```diff
@@ -0,0 +1,29 @@
...
+def test_export_study_with_both_compression(client: TestClient, internal_study_id: str, user_access_token: str) -> None:
+    # exporting without any parameters (thus exporting in .zip by default)
+    export_res = client.get(f"/v1/studies/{internal_study_id}/export", stream=True)
+    export_json_res = export_res.json()
+    assert ".zip" in export_json_res["file"]["filename"]
+
+    # exporting with the .7z compression
+    export_res = client.get(f"/v1/studies/{internal_study_id}/export?compression=.7z")
+    export_json_res = export_res.json()
+    assert ".7z" in export_json_res["file"]["filename"]
```
sylvlecl: We still need to check that it's a valid 7z: you can unarchive it in the test to check that it does not raise an exception, which will confirm that it's indeed a valid 7z.
Same for the zip above

sylvlecl (changes requested):
Good it's working :)  I still requested very minor code improvements, and tests improvement too: we need to check that what we download is actually a valid 7z or zip, in the tests.

antarest/study/service.py
sylvlecl: This comment still holds

sylvlecl (approved):

PR #2884, opened 2025-12-02
perf(output): use polar inside aggregation

MartinBelthle:
**Benchmark performed on my local env:**

- For a BP study, endpoint GET /variables:
  - Before: 5.5s
  - Now: 1s

Not directly related to polars but still nice. Will only be useful once by output cause after that we put the value inside the DB.

- Aggregation on 1000 years on 1 area, hourly
  -  Before: 61s
  -  Now: 45s

*NB*
- The keyword `use_pyarrow_extension_array` is a bit less performant
- Using `file_path` is a bit faster than creating an `io.StringIO` object
- Using `n_threads=1` is also a bit faster


antarest/study/business/output/aggregator_management.py
```diff
@@ -116,19 +119,28 @@ def __init__(
             if (isinstance(query_file, MCIndAreasQueryFile) or isinstance(query_file, MCIndLinksQueryFile))
             else MCRoot.MC_ALL
         )
+        self._output_first_column = get_start_column(self.frequency)
+
+    def _parse_output_file(self, file_path: Path, normalize_column_names: bool) -> pd.DataFrame:
+        content = file_path.read_text(encoding="utf-8")
+        output_headers = parse_headers(content, self._output_first_column)
+        try:
+            polars_df = pl.read_csv(file_path, skip_lines=7, separator="\t", has_header=False, null_values="N/A")
```
sylvlecl: I think we should use n_threads=1 here

sylvlecl (approved):
Looks good, I think the next step would be to expand the use of polars DF everywhere we don't actually need something else, meaning:
- when we want the conversion to the current JSON format
- when we compute the hash of the matrix ...

PR #2885, opened 2025-12-03
perf(aggregation): use /variables endpoint to avoid having columns mismatch

MartinBelthle:
We could use this `pl.scan_csv("data.csv").select(pl.col("^A.*$")).collect()` to fasten the reading performances when the user asks for specific columns (i.e. for the variables view endpoint) But it will not be very useful as we only perform the aggregation once, after this it's stored in DB.

TODO:
- Remove the columns mismatch related code
- Benchmark the performances
- Fix potential failing tests
- Add test for new behavior for wrong columns


MartinBelthle:
Not for now

PR #2887, opened 2025-12-04
feat(ui-results): introduce results variable per variable views

hdinia:
ANT-4205


MartinBelthle (changes requested):
I tested it on my local env and I have a few remarks:

- Not sure what `recordYears` is in the list of possible categories to choose from. And it seems buggy.
- When I click on the links view, it doesn't perform an API call so I cannot see the matrices anymore
- The thermal, renewable and st storages views are broken as you're not calling the backend properly. For them, you should provide a choice for which cluster they want to check and use its name inside the `thermal_id`, `renewable_id` back-end call (with the right `OutputVariablesType` also)

hdinia:
@MartinBelthle  Thanks for the review ! To follow up on the meeting about the API responses, here's a summary of my observations :


### Overview
Now that we've successfully implemented the variable-per-variable feature on the frontend, I've had a chance to review how it integrates with the GET /v1/studies/${studyId}/output/${outputId}/variables-views/data endpoint. Through this process, I've identified some opportunities to enhance the endpoint to better support all our use cases and ensure a seamless user experience.

---

### Current Problem
The endpoint currently returns a `404` status code for multiple distinct scenarios, making it impossible for the frontend to differentiate between different states and provide appropriate UX. 

---

### Different possible states ans suggested changes

#### 1. **State 1 - Not Materialized Yet** (Valid variable but content not available)
**Current behavior:**
- Returns `404` error

**Required behavior:**
- Return an appropriate HTTP status code that indicates the absence of data (e.g., `204 No Content` or a `200` with empty response body)

- Suggested response structure:
```json
{
  "status": "not_materialized",
  "message": "Data not yet materialized",
  "data": null
}
```

**Frontend impact:**
- Will display the Empty view placeholder with the "Process" button to trigger materialization

---

#### 2. **State 2 - Materialization in Progress**
**Current behavior:**
- Returns `404` error

**Required behavior:**
- Return a status code that reflects the "pending" state (e.g., `202 Accepted`)
- Include the `taskId` for the ongoing materialization in the response (metadata is optional)

- Suggested response structure:
```json
{
  "status": "materializing",
  "taskId": "abc-123-def-456",
  "message": "Materialization in progress",
  "data": null
}
```

**Frontend impact:**
- Can track the materialization progress using the returned `taskId`
- If the user switches to another area and returns to this one, the frontend can read the `taskId` and display the appropriate loading state instead of allowing a duplicate materialization to be triggered

---

#### 3. **State 3 - Data Available** (Materialization completed)
**Current behavior:**
- Returns `200` with data ✅
- May include `null` values in the matrix data ❌

**Required behavior:**
- Continue returning `200` status code ✅
- **Replace `null` values with `NaN` in the matrix data**

**Rationale:**
The Matrix component expects consistent data types (numbers or strings). Handling `null` values would require significant foundational changes to the Matrix component, which is used throughout the application. Using `NaN` for missing numeric values is the standard approach and is already handled by the component.

---

#### 4. **State 4 - Actual Errors**
**Current behavior:**
- Mixed with other scenarios (returns `404` for non-error cases)

**Required behavior:**
- Reserve `404`, `500`, and other error codes exclusively for genuine errors

**Frontend impact:**
- Can display appropriate error messages for genuine failures

---

### Additional Issue: Invalid Variable for Area/Link

**Current behavior:**
- Requesting the endpoint with a non-existing variable for the current area/link returns `404`

**Required behavior:**
- Return `422 Unprocessable Entity` to indicate that the request is well-formed but semantically incorrect (invalid variable name for the given area/link)
- Include a descriptive error message

- Suggested response structure:
```json
{
  "error": "Invalid variable",
  "message": "Variable 'xyz' does not exist for area 'area1'"
}
```

MartinBelthle (approved):

Issue #1030, opened 2022-08-03
Fix gunicorn bad worker load balancing


pl-buiquang:
Maybe due to the hybrid nature of the fastapi application (using a lot of sync thread instead of async ones) the load balancing of gunicorn is not good (running always on same top 4 workers)

pl-buiquang:
Solution could be to run multiple dockers with gunicorn using a single worker, and let the loadbalancing part to nginx

PR #2888, opened 2025-12-04
feat(ui): replace React Router by Tanstack Router

skamril:
implemention of Tanstack Router: add missing routes, refactor navigation buttons and links into semantic anchor links (\<a href\>), implementation of Tanstack Query...

Other change includes:

#### Features

- feat(ui): new component for the list of items with view
- feat(ui): add a background for favorite studies menu item
- feat(ui-router): add Not Found page
- feat(ui-dialog): create useDialog hook
- feat(ui-dialog): update confirmation dialog
- feat(ui-form): create useFormBlocker hook that block navigation when a form is dirty or saving using Tanstack Router
- feat(ui-thermals,ui-renewables): split thermal and renewable views into multiple routes, like storage views
- feat(ui-thermals,ui-renewables,ui-storages): add a select to change the current item
- feat(ui-storages): add optimistic update with Tanstack Query for additionnal constraints
- feat(ui-bc): allow to change the first area on binding constraint terms
- feat(ui-bc): add optimistic update with Tanstack Query
- feat(ui-bc): update binding constraint terms editor
- feat(scripts): update license checker to ignore file and folders

#### Bug Fixes

- fix(ui): update documentation links
- fix(ui-theme): prevent dark mode flickering when users refresh the page
- fix(ui-matrix): prevent empty view flickering when loading end
- fix(ui-matrix): error thrown when in certain case when data is empty
- fix(ui-xpansion): delete config button hidden
- fix(ui-debug): navigate outside a a dirty JSON file no displayed a confirmation message
- fix(ui-links): output filters not saved correctly
- fix(ui-bc): removal of a negative offset cannot be saved


webapp/src/components/buttons/BackButton.tsx
hdinia, line 1: Its not better to keep this one for router links only ? (as we should mainly use it now)

Also note that their is an existing `ButtonBack` that is redundant with this one

webapp/src/components/fieldEditors/SelectFE.tsx
```diff
@@ -138,6 +142,17 @@ function SelectFE<OptionValue extends AllowedValue = AllowedValue>({
     [options, startCaseLabel, t],
   );
 
+  const groupsOfOptions = useMemo(() => {
```
hdinia: `optionGroups` or `groupedOptions`

webapp/src/components/page/EmptyView.tsx
```diff
@@ -12,23 +12,23 @@
...
-import { Box } from "@mui/material";
 import type { SvgIconComponent } from "@mui/icons-material";
+import LiveHelpRoundedIcon from "@mui/icons-material/LiveHelpRounded";
+import { Box, Typography } from "@mui/material";
+import { useTranslation } from "react-i18next";
 
 export interface EmptyViewProps {
   title?: string;
   icon?: SvgIconComponent;
+  actions?: React.ReactNode;
```
hdinia, line 23: this prop was added to display a single action button under the text, now its redundant with `extraActions`, it should stay `action`

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/areas/route.tsx
hdinia: 📚 **Official TanStack Router Documentation Reference**

Based on the official documentation, here's what's happening in your code and why it's problematic:

### **Understanding `useParams({ strict: false })`**

According to the [official TanStack Router docs](https://tanstack.com/router/v1/docs/framework/react/api/router/useParamsHook):

> **`strict: false` is designed for specific scenarios:**
> 1. When rendering a component that is **shared across multiple routes**
> 2. When rendering a component that is **not within a route**

**Important note from the docs:**
> "This will never fail at runtime, and what you get back is still better typed than most other solutions, because the router knows about all existing routes and can compute an intersection of all possibly existing params."

However, there's an [open issue (#1030)](https://github.com/TanStack/router/issues/1030) about type safety being reduced when using `strict: false`.

---

## 🔄 **Parent-Child Route Params: The Official Pattern**

According to [TanStack Router's Routing Concepts](https://tanstack.com/router/v1/docs/framework/react/routing/routing-concepts):

> **"Once a path param has been parsed, it is available to all child routes."**

And from the [Type Safety guide](https://tanstack.com/router/v1/docs/framework/react/guide/type-safety):

> **"When you look at a child route, you can see that you get all params from the route hierarchy back when calling `Route.useParams()`."**

### **✅ The Correct Pattern**

**Parent routes pass params DOWN to children, not the other way around:**

```typescript
// Parent route: /_authenticated/studies/$studyId/explore/modeling/areas
const { studyId } = Route.useParams(); // ✅ Only access params defined in THIS route

// Child route: /_authenticated/studies/$studyId/explore/modeling/areas/$areaId
const { studyId, areaId } = Route.useParams(); // ✅ Access parent AND own params
```

---

## 🔴 **Why Your Current Code is Wrong**

In your `areas/route.tsx`:

```typescript
// ❌ WRONG: Trying to access CHILD route params from PARENT route
const { areaId, thermalId, storageId, renewableId } = useParams({ strict: false });
```

**Problems:**
1. **Anti-pattern**: Parent routes should NOT access child route params
2. **Violates data flow**: TanStack Router's design is parent → child, not child → parent
3. **Type safety loss**: `strict: false` loses TypeScript guarantees
4. **Not the intended use case**: According to docs, `strict: false` is for **shared components**, not for accessing child params

---

## ✅ **Correct Solutions (Backed by Docs)**

### **Option 1: Use Search Params (Recommended)**

From [Search Params documentation](https://tanstack.com/router/v1/docs/framework/react/guide/search-params):

> "Search parameters are ideal for filtering, sorting, pagination, and **UI state** that doesn't fundamentally change the resource being accessed."

```typescript
export const Route = createFileRoute(
  "/_authenticated/studies/$studyId/explore/modeling/areas"
)({
  validateSearch: (search) => ({
    selectedArea: search.selectedArea as string | undefined,
  }),
  component: AreasLayout,
});

function AreasLayout() {
  const { studyId } = Route.useParams(); // ✅ Only this route's params
  const { selectedArea } = Route.useSearch(); // ✅ UI state in search params
  // ...
}
```

### **Option 2: Use Context (From Official Docs)**

From [Context Inheritance in TanStack Router](https://tkdodo.eu/blog/context-inheritance-in-tan-stack-router):

> "Router context merges down with the hierarchy of routes, so that each route has access to the context of all of its parents."

You could pass selected state through context:

```typescript
// Parent sets context
export const Route = createFileRoute("/_authenticated/studies/$studyId/explore/modeling/areas")({
  beforeLoad: () => ({
    selectedAreaId: undefined as string | undefined,
  }),
  component: AreasLayout,
});

// Children access it via Route.useRouteContext()
```

### **Option 3: Use `useMatch` for Active Child Detection**

From [useMatch documentation](https://tanstack.com/router/v1/docs/framework/react/api/router/useMatchHook):

```typescript
const thermalMatch = useMatch({
  from: "/_authenticated/studies/$studyId/explore/modeling/areas/$areaId/thermals/$thermalId",
  shouldThrow: false,
});

// Now you can safely access: thermalMatch?.params.thermalId
```

---

## 📖 **Key Documentation Quotes**

From [GitHub Discussion #364](https://github.com/TanStack/router/discussions/364) - "How to access current route params from a parent layout":

The official recommendation is to use `useMatch` or pass information through context/search params rather than trying to access child params from parents.

From [TkDodo's blog on TanStack Router](https://tkdodo.eu/blog/the-beauty-of-tan-stack-router):

> "The explicit opt-in nature of `strict: false` helps make codebases more maintainable and easier to refactor with confidence."

webapp/src/components/ButtonBack.tsx
hdinia, line 1: this can be deleted no ?

webapp/src/hooks/useFormBlocker.ts
hdinia, line 1: Its better 👍🏼

webapp/src/redux/selectors.ts
```diff
@@ -251,7 +252,7 @@ export const getAreasById = createSelector(getStudySynthesis, (synthesis) => {
         acc[id] = { ...synthesis.areas[id], id };
         return acc;
       },
-      {} as Record<string, Area & { id: string }>,
+      {} as Record<string, AreaWithId>,
```
hdinia, line 255: `AreaWithId` this type is cleaner than the previous `Area & { id: string }` everywhere

webapp/src/components/Matrix/index.tsx
```diff
@@ -64,6 +65,7 @@ interface MatrixProps {
 }
 
 function Matrix({
+  studyId,
```
hdinia, line 69: Any reason to add `studyId` prop ? It creates a duplication everywhere we use `Matrix`, can you revert that change and consume studyId directly from the router context ?

webapp/src/app/PageNotFound.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const isRoot = match.routeId === "__root__";
+
+  // In the root route, there is no menu to go back
+  const actions = isRoot ? (
+    <RouterButton to={location.pathname.split("/").slice(0, -1).join("/")} variant="contained">
+      {t("global.home")}
+    </RouterButton>
+  ) : null;
+
+  return <EmptyView title={t("page.notFound")} icon={SearchOffIcon} actions={actions} />;
```
hdinia, line 33: `to={location.pathname.split("/").slice(0, -1).join("/")}`

not the same as `to="/"`?


also the `isRoot` condition seems weird to me, I dont see a case where you have a 404 page and you dont want to display an action redirecting to the home page, can you please explain more ?

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/areas/$areaId/-hooks/useArea.ts
hdinia, line 1: maybe you can use a loader + router context for this ?

webapp/src/routes/_authenticated/studies/$studyId/-hooks/useStudy.ts
hdinia, line 1: you could optimize a bit and avoid Redux dependency by using a loader + route context

webapp/src/app/App.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+//! Redux store must be imported before any redux ducks
+import store from "../redux/store";
```
hdinia, line 16: this comment might be unecessary

webapp/src/queries/bindingConstraints.ts
hdinia, line 1: you can create a folder and split into 2 files `queries.ts` and `mutations.ts` and use a factory file for queries keys `keys.ts`

ressources :
https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories

webapp/src/queries/bindingConstraints.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const bindingConstraintQueries = {
+  all: () => [...studyQueries.all(), "bindingConstraints"] as const,
+  list: (studyId: StudyMetadata["id"]) => {
+    return queryOptions({
+      queryKey: [...bindingConstraintQueries.all(), { studyId }],
+      queryFn: async () => queryList(await getBindingConstraints({ studyId })),
+      refetchOnWindowFocus: (query) => !query.state.data?.some((c) => c.isOptimistic),
+      refetchOnReconnect: (query) => !query.state.data?.some((c) => c.isOptimistic),
+      refetchOnMount: (query) => !query.state.data?.some((c) => c.isOptimistic),
```
hdinia: This custom refetch logic is not the recommended way i think, if you invalidate queries this is still needed ? If you can explain more please

The official pattern is:
1. `onMutate`: `cancelQueries()` cancels outgoing refetches
2. `onError`: Rollback with `setQueryData(previousValue)`
3. `onSettled`: `invalidateQueries()` always refetch after mutation

you can check :
https://tanstack.com/query/latest/docs/framework/react/guides/optimistic-updates#via-the-cache
and
https://tkdodo.eu/blog/concurrent-optimistic-updates-in-react-query#concurrent-optimistic-updates

webapp/src/queries/bindingConstraints.ts
hdinia: can you check the storages queries too please

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/areas/$areaId/storages/$storageId/additional-constraints/-hooks/useDeleteStorageConstraint.ts
```diff
@@ -0,0 +1,98 @@
...
+    },
+    onSettled: (data, error, variables, onMutateResult, context) => {
+      const mutationNb = context.client.isMutating({
+        mutationKey: storageMutations.constraintList(studyId, areaId, storageId),
+      });
+
+      if (mutationNb === 1) {
+        context.client.invalidateQueries({ queryKey: queryListKey });
+      }
+    },
```
hdinia: the docs recommend to always invalidate, any reason you needed to do this ?

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/binding-constraints/$bindingConstraintId/time-series.tsx
```diff
@@ -0,0 +1,128 @@
...
+  "/_authenticated/studies/$studyId/explore/modeling/binding-constraints/$bindingConstraintId/time-series",
+)({
+  component: TimeSeries,
+});
+
+function TimeSeries() {
+  const study = useStudy();
+  const constraint = useBindingConstraint();
+  const { t } = useTranslation();
+  const urlBase = `input/bindingconstraints/${constraint.id}`;
```
hdinia: `baseUrl`

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/binding-constraints/$bindingConstraintId/time-series.tsx
```diff
@@ -0,0 +1,128 @@
...
+      return getBindingConstraint({
+        studyId: study.id,
+        constraintId: constraint.id,
+      });
+    },
+    {
+      deps: [study.id, constraint.id],
+      disabled: isLegacy,
+    },
+  );
```
hdinia: you can use the router cache updated via the update mutation

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/binding-constraints/$bindingConstraintId/time-series.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  ////////////////////////////////////////////////////////////////
+  // JSX
+  ////////////////////////////////////////////////////////////////
+
+  if (constraint.isOptimistic) {
+    return <DataGridSkeleton />;
+  }
+
+  if (isLegacy) {
```
hdinia: the correct buisness term is Scenarized vs Non-Scenarized

if you could document it too it would be nice

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/links/route.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { useTranslation } from "react-i18next";
+
+export const Route = createFileRoute("/_authenticated/studies/$studyId/explore/modeling/links")({
+  component: LinksLayout,
+});
+
+function LinksLayout() {
+  const navigate = Route.useNavigate();
+  const { studyId } = Route.useParams();
+  const { linkId } = useParams({ strict: false });
```
hdinia, line 31: this parent route is trying to read `linkId` that belongs to a child route, which breaks the type-safety and hierarchy that provides the router. This goes against framework conventions Tstack Router explicitly discourages this pattern (except for some cases)

Try to use another solution, one way to do that I see is `useMatches`

```ts
const matches = useMatches();
const isOnChildRoute = matches.some(match => match.routeId.includes('$linkId'));
```

Note: I saw this on other files to not repeat please check this on other places if needed

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/links/$linkId/time-series/route.tsx
```diff
@@ -0,0 +1,59 @@
...
+)({
+  component: TimeSeriesLayout,
+});
+
+function TimeSeriesLayout() {
+  const params = Route.useParams();
+  const study = useStudy();
+  const { t } = useTranslation();
+
+  if (semver.lt(study.version, TABS_VIEW_MIN_VERSION)) {
```
hdinia, line 33: `TABS_VIEW_MIN_VERSION` is generic I suggest you rename it for something like `LINK_MIN_VERSION`

hdinia (changes requested):

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/links/route.tsx
skamril: It's a layout route, so it's herited by the child route. How did you read it goes against framework conventions?
It's more safe and simple than `matches.some(match => match.routeId.includes('$linkId'));`, you can have `$linkIdXXX`.

webapp/src/components/page/EmptyView.tsx
skamril: Why limiting it to one action?

webapp/src/redux/selectors.ts
skamril: But it's temporary, we need to to use adapter and add it in `Area`

webapp/src/components/Matrix/index.tsx
skamril: Because src/components folder should not import from src/router.
Otherwise Matrix folder need to me moved inside router.

webapp/src/app/PageNotFound.tsx
skamril: Because PageNotFound takes full size only on root route, in other case, it's visible under a layout route (so the main menu is visible).

webapp/src/app/App.tsx
skamril: No, if a redux duck is imported before, the app will not working

webapp/src/app/PageNotFound.tsx
skamril: For `to={location.pathname.split("/").slice(0, -1).join("/")}` I don't know why I write this...
`to="/"` is better.

webapp/src/queries/bindingConstraints.ts
skamril: It's already a key factories. If you talk about the use of `queryOptions` inside the factory => https://tkdodo.eu/blog/the-query-options-api#datatag

skamril: image

?

webapp/src/queries/bindingConstraints.ts
skamril: I already read this, it's needed in our case. You can talk about it on Teams.

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/areas/$areaId/storages/$storageId/additional-constraints/-hooks/useDeleteStorageConstraint.ts
skamril: https://tkdodo.eu/blog/concurrent-optimistic-updates-in-react-query#limiting-the-scope

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/links/$linkId/time-series/route.tsx
skamril: It's the min version to display tabs instead of one matrix

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/binding-constraints/$bindingConstraintId/time-series.tsx
skamril: You talked about the variable name `isLegacy `?

webapp/src/components/ButtonBack.tsx
skamril: Yes, I need to replace it in some views with the new one

webapp/src/components/page/EmptyView.tsx
hdinia: It's about intent not technical limitation. In an empty state you want one clear call-to-action not multiple options. The singular `action` name reinforces that pattern. Also, having both `actions` and `extraActions` is confusing what distinguishes them?

webapp/src/queries/bindingConstraints.ts
hdinia: yes thanks for the change

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/binding-constraints/$bindingConstraintId/time-series.tsx
hdinia: Yes it's a bit misleading, optional but it would be nice to document this a bit for future changes

webapp/src/components/page/EmptyView.tsx
skamril: `extra` for secondary actions (display on top right). I can rename it `primaryActions` maybe.

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/links/$linkId/time-series/route.tsx
hdinia: ok nervermind

webapp/src/app/App.tsx
skamril: Circular import issue with ducks and the store. I take time to find this. You can test.

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/links/route.tsx
hdinia: The current code might work but uses an anti-pattern with untyped params access

It's an architecture problem, you should not access data from child -> parent. The flow must remain from parent to child (Tstack Router and React core pattern)

Proper pattern (used by binding-constraints):
- **`index.tsx`** - Handles redirect to first item (no UI)
- **`route.tsx`** - Pure layout that renders the list (no child awareness)

This separates routing logic from UI rendering and maintains proper route encapsulation

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/binding-constraints/$bindingConstraintId/time-series.tsx
hdinia:
```ts
  const updateMutation = useMutation({
    ...bindingConstraintMutations.update(),
    onSuccess: (updatedConstraint) => {
      queryClient.setQueryData(
        bindingConstraintQueries.list(study.id).queryKey,
        (oldData) =>
          oldData?.map((c) => (c.id === updatedConstraint.id ? updatedConstraint : c)) ?? [],
      );
    },
  });
  
  
  
      return updateMutation.mutateAsync({

  ```
  
  ```ts
    update: () => {
    return mutationOptions({
      mutationKey: [...bindingConstraintQueries.all(), "updateBindingConstraint"],
      mutationFn: updateBindingConstraint,
    });
  },
  ```
  
  in timeseries 
  ```ts
  
  const { operator } = constraint;
  
  ```

File: webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/links/route.tsx (hunk 12 above)
Comment (hdinia) [reply webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/links/route.tsx:31]: you can add an `index.ts`
```ts
export const Route = createFileRoute("/_authenticated/studies/$studyId/explore/modeling/links/")({
  loader: async ({ params: { studyId } }) => {
    const links = await getLinks({ studyId });

    if (links.length > 0) {
      throw redirect({
        to: "/studies/$studyId/explore/modeling/links/$linkId",
        params: { studyId, linkId: links[0].id },
      });
    }
  },
});
```

hdinia (approved):

PR #2890, opened 2025-12-05
fix(matrix-gc): don't clean variant used matrices in a limit case

MartinBelthle:
Bug that existed and is fixed inside this PR (never happened but could):
1) Create a study with 1 area and 1 thermal at least
2) Create a variant and launch a time series generation
3) The matrix garbage collector does not see the generated matrix as used ones and remove them from the matrix-store.
4) If the snapshot is not cleaned, fetching this matrix will raise an error

Now, if a variant contains a `GenerateThermalTimeSeries` command, we'll scan its snapshot. This way we avoid the problem.

Also improves a bit the speed of the `RawStudyMatrixProvider` by:
- Don't scan archived studies
- Only scan `input` and `expansion` folder

### DIFF
--- a/tests/matrixstore/test_matrix_usage_providers.py
+++ b/tests/matrixstore/test_matrix_usage_providers.py
@@ -156,7 +156,6 @@ def test_command_matrix_usage_provider(
         study_version = "880"
         variant_study_repository.save(VariantStudy(id=study_id, version=study_version, path=tmp_path.as_posix()))
         matrices_id = "a68de4b5e96a60c8ceb3c7b7ef93461725bdbbff3516b136585a743b5c0ec664"
-        use_description = f"Used by command {matrices_id} from variant study {study_id}"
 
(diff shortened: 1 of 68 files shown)


tests/matrixstore/test_matrix_usage_providers.py
```diff
@@ -145,7 +156,6 @@ def test_command_matrix_usage_provider(
         study_version = "880"
         variant_study_repository.save(VariantStudy(id=study_id, version=study_version, path=tmp_path.as_posix()))
         matrices_id = "a68de4b5e96a60c8ceb3c7b7ef93461725bdbbff3516b136585a743b5c0ec664"
-        use_description = f"Used by command {matrices_id} from variant study {study_id}"
```
MartinBelthle, line 148: This was wrong, we didn't use the cmd id but the matrix_id in `use_description`

sylvlecl (approved):

PR #2891, opened 2025-12-08
test(db): initialize DB only once for all integration tests

sylvlecl:
In integration tests, we currently initialize the database schema once per test, which takes
some time (around 1s on my computer).

This PR initializes the schema only once for a test session, and copies that base file
at the start of each test.

Surprisingly, no significant gain on ubuntu CI, but around 3min gained on windows CI.


MartinBelthle (approved):
You should just rename your PR `test(` instead of `tests(`

PR #2893, opened 2025-12-08
refactor(output): change some logic inside the /variables-views endpoints

MartinBelthle:
- Use `NaN` instead of `null` inside the GET `/variables-views/data` endpoint
- Raise an 404 exception when asking for unexisting data inside the GET and POST `/variables-views` endpoints
- The GET endpoint now returns a 404 response with the following body {"status": xxx, "task_id": xxx} when the data is not in DB yet. status is NOT_FOUND and task_id is None when the materializing hasn't been done. status is `IN_PROGRESS` and task_id is the running task when the materializing is in progress

### DIFF
--- a/antarest/study/storage/output_model.py
+++ b/antarest/study/storage/output_model.py
@@ -145,3 +145,8 @@ class OutputVariablesViewsModel(Base):
 class OutputVariablesViewStatus(StrEnum):
     NOT_FOUND = "NOT_FOUND"
     IN_PROGRESS = "IN_PROGRESS"
+
+
+class OutputVariablesViewResponse(AntaresBaseModel, extra="forbid", alias_generator=to_camel, populate_by_name=True):
+    status: OutputVariablesViewStatus
+    task_id: str | None
--- a/antarest/study/storage/output_model.py
+++ b/antarest/study/storage/output_model.py
@@ -140,3 +140,8 @@ class OutputVariablesViewsModel(Base):
     st_storage_id: Mapped[Optional[str]] = mapped_column(String, nullable=True)
     matrix_id: Mapped[str] = mapped_column(String, ForeignKey("matrix.id"), nullable=False)
     last_read: Mapped[datetime] = mapped_column(DateTime)
+
+
+class OutputVariablesViewStatus(StrEnum):
+    NOT_FOUND = "NOT_FOUND"
+    IN_PROGRESS = "IN_PROGRESS"
--- a/antarest/study/storage/output_service.py
+++ b/antarest/study/storage/output_service.py
@@ -107,12 +107,6 @@ class OutputVariablesViewMaterializationTask:
 
     def _materialize_view(self) -> None:
         """Run the task"""
-        # Checks the asked couple `variable name` / `object_id` exists for the output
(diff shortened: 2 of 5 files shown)


antarest/study/web/output_blueprint.py
```diff
@@ -553,12 +555,12 @@ def get_output_variables_view(
     ) -> Response:
         """
         Fetches the variables view for a given output and a given configuration.
-        If the view does not exist in DB yet, raises an HTTP 404 error.
+        If the view does not exist in DB yet, returns an HTTP 404 response.
```
sylvlecl, line 563: Would be nice to describe those responses in the swagger doc, see:
https://fastapi.tiangolo.com/advanced/additional-responses/#additional-response-with-model

antarest/study/storage/output_model.py
```diff
@@ -140,3 +140,8 @@ class OutputVariablesViewsModel(Base):
     st_storage_id: Mapped[Optional[str]] = mapped_column(String, nullable=True)
     matrix_id: Mapped[str] = mapped_column(String, ForeignKey("matrix.id"), nullable=False)
     last_read: Mapped[datetime] = mapped_column(DateTime)
+
+
+class OutputVariablesViewStatus(StrEnum):
+    NOT_FOUND = "NOT_FOUND"
```
sylvlecl, line 146: Don't we want to distinguish between "not materialized" and "variable does not exist" ?

sylvlecl: Ah I understand the second case is handled by the exception, why not.

antarest/study/storage/output_service.py
```diff
@@ -788,7 +804,17 @@ def get_output_variables_view(
...
+        available_variables = self.get_output_variables_list(study_id, output_id)
+        check_output_variable_exists(output_id, variable_type, variable_name, available_variables, output_identifier)
+
+        # Return a 404 Response with a body specifying if the materialization is in progress or not.
+        task_id, _ = self._get_ongoing_variables_view_materialization_task(
+            output_identifier, study_id, output_id, frequency
+        )
+        status = OutputVariablesViewStatus.IN_PROGRESS if task_id else OutputVariablesViewStatus.NOT_FOUND
+        content = {"status": status, "task_id": task_id}
+        return Response(status_code=404, content=to_json(content), media_type="application/json")
```
sylvlecl: We'd better not have the concept of Response in the service, we'd better return a small pydantic model here, and let the blueprint convert it

antarest/study/web/output_blueprint.py
```diff
@@ -571,8 +573,10 @@ def get_output_variables_view(
             renewable_id,
             st_storage_id,
         )
-        content = dataframe.to_json(orient="split", index=False)
-        return Response(content=content, media_type="application/json")
+        if isinstance(view, pd.DataFrame):
+            content = view.to_dict(orient="split", index=False)
+            return Response(content=to_json(content), media_type="application/json")
+        return view
```
sylvlecl: The transformation to response should be implemented here, not in the service layer

sylvlecl (changes requested):
Minor stuff: improving the api doc and moving transformation to response to the web layer

sylvlecl (approved):

PR #2894, opened 2025-12-08
refactor(matrix): add in-memory predefined matrices

sylvlecl:
Changes the handling of so-called "constant" matrices:
today, they are actually written to storage and read from there.

This PR proposes to have them created by a factory function every time instead.
The benefits are:

1. faster startup, in particular in integration tests.
    **Allows to reduce windows CI by 2-3 minutes more.**
2. some performance gains when retrieving those matrices, since we don't have
    to go read them in the storage anymore


antarest/matrixstore/service.py
```diff
@@ -144,13 +155,22 @@ class SimpleMatrixService(ISimpleMatrixService):
     def __init__(self, matrix_content_repository: MatrixContentRepository):
         self.matrix_content_repository = matrix_content_repository
         self.usage_providers: List[IMatrixUsageProvider] = []
+        self._predefined_matrices: dict[str, Callable[[], pd.DataFrame]] = {}
+
+    @override
+    def add_predefined_matrix(self, matrix_factory: Callable[[], pd.DataFrame]) -> str:
```
MartinBelthle, line 161: You have the exact same code for the 3 implem, you could instead put this inside the ISimpleMatrixService

MartinBelthle (approved):

antarest/matrixstore/service.py
sylvlecl: I prefer to keep the interface definition pure, but indeed there's some duplication.
In particular we cannot assume that all implementations will rely on the same `compute_hash` method for computing the ID, so it would be a little fragile to implement it at the interface level.

MartinBelthle: Why not. But for info the interface is not pure today as there's already a method `get_matrix_id` implemented inside it

PR #2896, opened 2025-12-09
refactor(matrix): perform less DB calls when normalizing/denormalizing

MartinBelthle:
Currently we perform 1 DB call for each matrix normalization or denormalization.
For a BP study it means we're performing around 5000 DB calls !!!
With this PR we'll perform only 1 DB call for the whole study

We'll also see minor performance improvements inside the denormalization process.
Before if we had 100 files that referenced the same matrix in the matrix-store, we read its content 100 times.
Now we only read it once.


antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -526,3 +526,92 @@ def checks_antares_web_compatibility(study: Study) -> None:
...
+            study = self.get_raw(study)
+        matrix_nodes = cast(list[MatrixNode], study.tree.normalize())
+        if not matrix_nodes:
+            return
+
+        matrix_mapper = matrix_nodes[0].matrix_mapper
+        matrix_ids = matrix_mapper.save_matrices(matrix_nodes)
+
+        for k, node in enumerate(matrix_nodes):
+            node.matrix_mapper.save_matrix(node, matrix_ids[k])
```
smailio, line 492: I don't get it here why you need to call save_matrix individually after you called save_matrices for all matrix_nodes. Wuould you mind adding a comment to explain the intent ?

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -526,3 +526,92 @@ def checks_antares_web_compatibility(study: Study) -> None:
...
+            return
+
+        matrices_mapping: dict[str, list[MatrixNode]] = {}
+        for node in matrix_nodes:
+            link_content = node.matrix_mapper.get_link_content(node)
+            assert link_content is not None
+            matrices_mapping.setdefault(build_matrix_uri(link_content), []).append(node)
+
+        matrix_mapper = matrix_nodes[0].matrix_mapper
+        for matrix_id, dataframe in matrix_mapper.get_matrices(list(matrices_mapping.keys())):
```
smailio: It's worth to refactor this, a lot of stuff happening at once here

antarest/study/storage/rawstudy/raw_study_service.py
MartinBelthle: `save_matrices` save the matrices in the matrix_store. `save_matrix` save the uri inside the node. I agree it's perhaps not really clear

antarest/matrixstore/repository.py
```diff
@@ -126,12 +127,25 @@ def save(self, matrix: Matrix) -> Matrix:
...
 
+    def save_batch(self, matrices: list[Matrix]) -> None:
+        try:
+            self.session.add_all(matrices)
+            self.session.commit()
+        except IntegrityError:
+            # Can happen if one the matrices is already inside DB.
+            self.session.rollback()
+            for matrix in matrices:
+                self.save(matrix)
```
smailio, line 138: you should add a commit after this loop, no ?

antarest/matrixstore/service.py
```diff
@@ -102,6 +113,10 @@ def get(self, matrix_id: str) -> pd.DataFrame:
     def get_matrices(self) -> list[MatrixMetadataDTO]:
         raise NotImplementedError()
 
+    @abstractmethod
+    def yield_matrices(self, matrix_ids: Sequence[str]) -> Iterator[tuple[str, pd.DataFrame]]:
```
smailio: this should be renamed tield_matrices_content

antarest/matrixstore/service.py
```diff
@@ -157,6 +176,11 @@ def get(self, matrix_id: str) -> pd.DataFrame:
     def get_matrices(self) -> list[MatrixMetadataDTO]:
         raise NotImplementedError()
 
+    @override
+    def yield_matrices(self, matrix_ids: Sequence[str]) -> Iterator[tuple[str, pd.DataFrame]]:
```
smailio: is this implementation ever called ?

antarest/matrixstore/repository.py
MartinBelthle: The commit is inside the `save` method

antarest/matrixstore/service.py
MartinBelthle: This implem I don't know but I implemented it in every service in case we need it

antarest/matrixstore/service.py
MartinBelthle: Or yield_matrices_id_and_content as it also yields the matrix id

smailio: Honestly I'm fine with this kind of long name. I didn't dare suggest it because I though it would be too long for most people. Maybe it's worth it to create a type for this but I can't think of a nice name.

smailio:
### PR schema

image

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -526,3 +526,92 @@ def checks_antares_web_compatibility(study: Study) -> None:
...
                     raise NotImplementedError("AntaresWeb doesn't support the value 'hourly' for the flag 'hydro-pmax'")
+
+    def normalize_study(self, study: Study | FileStudy) -> None:
+        """
+        Method used to normalize a study.
+        It will put every matrix in the study in the matrix-store.
+        """
+        if isinstance(study, Study):
+            study = self.get_raw(study)
+        matrix_nodes = cast(list[MatrixNode], study.tree.normalize())
```
sylvlecl: We cannot keep that name (normalize) if we change the behaviour that much.
It does not perform any action anymore

antarest/study/main.py
```diff
@@ -97,7 +98,8 @@ def build_study_service(
 
     if not generator_matrix_constants:
         generator_matrix_constants = GeneratorMatrixConstants(matrix_service=matrix_service)
-        generator_matrix_constants.init_constant_matrices()
+        with db():  # Needed to create matrices in DB at the app start
```
sylvlecl: Why didn't we have an issue here earlier ?

antarest/matrixstore/repository.py
```diff
@@ -126,12 +127,25 @@ def save(self, matrix: Matrix) -> Matrix:
         self.session.commit()
         return merged_matrix
 
+    def save_batch(self, matrices: list[Matrix]) -> None:
+        try:
+            self.session.add_all(matrices)
+            self.session.commit()
+        except IntegrityError:
+            # Can happen if one the matrices is already inside DB.
```
sylvlecl, line 135: I think this will happen often, maybe we should first have a request to check which matrices already exist and then send the other ones ? Gets tricky though

antarest/study/storage/rawstudy/model/filesystem/inode.py
```diff
@@ -107,21 +107,17 @@ def save(self, data: S, url: Optional[List[str]] = None) -> None:
         """
         raise NotImplementedError()
 
-    def normalize(self) -> None:
+    def normalize(self) -> list[Self]:
```
sylvlecl: We need at least to rename it. Also it should just return `MatrixNode`s, if we keep this role.
My feeling is that it should return something really more focused on the task, like an iterator of dataframes, and maybe even with a callable to be used with the ID of the matrix once it has been created.

antarest/matrixstore/matrix_uri_mapper.py
```diff
@@ -84,11 +76,11 @@ def save_matrix(self, node: MatrixNode, matrix_uri: str) -> None:
         pass
 
     @abstractmethod
-    def normalize(self, node: MatrixNode) -> None:
+    def save_matrices(self, nodes: Sequence[MatrixNode]) -> list[str]:
```
sylvlecl: We cannot name it this way, it's too different from the other `save_matrix` method.
We need to add docstrings to that class to make sure we understand its role, it's getting very confusing !

antarest/matrixstore/matrix_uri_mapper.py
```diff
@@ -84,11 +76,11 @@ def save_matrix(self, node: MatrixNode, matrix_uri: str) -> None:
...
         pass
 
     @abstractmethod
-    def denormalize(self, node: MatrixNode) -> None:
+    def get_matrices(self, uris: Sequence[str]) -> Iterator[tuple[str, pd.DataFrame]]:
```
sylvlecl: We need at least to have some docstring to explain the return type, or a small dataclass with named fields

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -526,3 +526,92 @@ def checks_antares_web_compatibility(study: Study) -> None:
...
+        if not matrix_nodes:
+            return
+
+        matrix_mapper = matrix_nodes[0].matrix_mapper
```
sylvlecl: We cannot do it this way, it show that dependencies are ill defined.

In that evolution, you take out from the nodes the responsibility of performing themselves the normalization/denormalization process.
This should appear in the structure of objects: 
now that "mapper" must be here in the raw study service, if you say it's that service that is responsible for doing the normalization / denormalization?

Probably, we need to separate the current "mapper" in  2 things:
- one which stays in nodes and which is used only when we want to write a matrix to the study
- one which is here outside of the study and handles normalizing/denormalizing whole studies

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -526,3 +526,92 @@ def checks_antares_web_compatibility(study: Study) -> None:
...
+            link_content = node.matrix_mapper.get_link_content(node)
+            assert link_content is not None
+            matrices_mapping.setdefault(build_matrix_uri(link_content), []).append(node)
+
+        matrix_mapper = matrix_nodes[0].matrix_mapper
+        for matrix_id, dataframe in matrix_mapper.get_matrices(list(matrices_mapping.keys())):
+            for node in matrices_mapping[matrix_id]:
+                node.write_dataframe(dataframe)
+
+    def export_study_flat_utils(
```
sylvlecl: Why this postfix `utils` ? what does it mean?

sylvlecl (changes requested):
I think the behaviour is right !  but that we really need to clarify the architecture and the naming.
In particular the so called "URI mapper" may need to be split into 2 different things:
- one thing which is just responsible for reading/writing individual matrices, which in particular handles the difference between managed and external studies
- another thing which handles the full-study operations, and which actually should not depend on managed or not, since we know what we want to do (writing the ID or not)

I don't have more detailed proposition for now but we need to find a way to improve the understandability.

antarest/study/main.py
MartinBelthle: Because we had this `with db()` inside the matrix_service. But we didn't needed it everytime, just here. Makes me realize that now that you put the default matrices in memory we shouldn't need it anymore. I removed it and everything works

antarest/matrixstore/repository.py
MartinBelthle: Why would this happen often ? To me it only happens if another worker tries to do the same thing in parallel. Because here we only add matrices that didn't exist in the matrixstore

antarest/study/storage/rawstudy/raw_study_service.py
MartinBelthle: Nothing, it's just that I moved it from the utils.py file and a method with this name already existed in the raw_study_service so I changed the name. But it's poor naming I agree

antarest/matrixstore/repository.py
sylvlecl: Ok indeed I didn't get right the `_create` method behaviour

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -443,6 +440,34 @@ def get_study_path(self, metadata: Study) -> Path:
             return self.find_archive_path(metadata)
         return Path(metadata.path)
 
+    def get_output_path(self, study: Study, output_id: str) -> Path:
```
sylvlecl: minor: I think this is not needed anymore (after refacto in other PR)

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -95,10 +94,8 @@ def __init__(
             cache=cache,
         )
 
-        RawStudyMatrixUsageProvider(
-            StudyMetadataRepository(cache_service=cache),
-            matrix_service=self.study_factory._matrix_mapper_factory._matrix_service,
-        )
+        self._matrix_service = self.study_factory._matrix_mapper_factory._matrix_service
```
sylvlecl, line 95: I know it can be a pain but we'd better add the matrix service as an explicit dependency in the constructor, not something hidden there

antarest/study/storage/rawstudy/raw_study_service.py
MartinBelthle: Not sure what's not needed anymore, the `BadArchiveContent` exception msg ? Because it can still be raised

antarest/study/storage/variantstudy/variant_command_generator.py
```diff
@@ -40,97 +40,89 @@ def __call__(self, x: float) -> None:
         logger.info(f"Command {self.index}/{self.total_count} [{self.study_id}] applied in {x}s")
 
 
-class VariantCommandGenerator:
-    def __init__(self, study_factory: StudyFactory) -> None:
-        self.study_factory = study_factory
-
-    @staticmethod
-    def _generate(
```
MartinBelthle, line 48: I just put the methods as functions and remove the class as they were only static methods

antarest/study/storage/abstract_storage_service.py
```diff
@@ -32,22 +32,16 @@
...
 class AbstractStorageService(IStudyStorage, ABC):
-    def __init__(
-        self,
-        config: Config,
-        study_factory: StudyFactory,
-        cache: ICache,
-    ):
+    def __init__(self, config: Config, cache: ICache):
         self.config: Config = config
-        self.study_factory: StudyFactory = study_factory
```
sylvlecl, line 50: nice

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -354,9 +351,8 @@ def export_commands_matrices(self, study_id: str) -> FileDownloadTaskDTO:
             )
             or []
         }
-        return cast(MatrixService, self.command_factory.command_context.matrix_service).download_matrix_list(
-            list(matrices), f"{study.name}_{study.id}_matrices"
-        )
+        matrix_service = cast(MatrixService, self._matrix_service)
```
sylvlecl, line 354: we'd better require a MatrixService instead of casting

antarest/study/storage/variantstudy/variant_command_generator.py
sylvlecl: great

sylvlecl (approved):

PR #2902, opened 2025-12-12
perf(aggregation): faster aggregation for clusters

MartinBelthle:
Benchmark performed on my local env for hourly matrices.

The modified code snippet took around 0.1s per dataframe and now it takes around 0.04s


sylvlecl (approved):
Nice it looks far better this way

PR #2905, opened 2025-12-15
chore(pydantic): remove pydantic deprecated class-based config

sylvlecl:
Spams the logs in CI


antarest/core/serde/np_array.py
```diff
@@ -9,19 +9,23 @@
...
+    if isinstance(array, list):
+        return np.array(array, dtype=np.float64)
+    elif isinstance(array, np.ndarray):
+        return array
+    raise ValueError("Input should be either a list or a numpy array")
 
 
-def _np_to_list(array: npt.NDArray[np.float64]) -> list[float]:
-    return cast(list[float], array.tolist())
+def _np_to_list(array: npt.NDArray[np.float64]) -> Any:
```
TheoPascoli: Not a big fan of `any`, is it always a 1 dim list ? In this case, we can check the dim et return a `list[float]`

TheoPascoli:
I added another dead code change as well

antarest/core/serde/np_array.py
sylvlecl: Actually it's always either a 1 or 2 dimension array, you're right that it'll be a bit cleaner to check it

TheoPascoli (dismissed):

antarest/core/serde/np_array.py
```diff
@@ -9,19 +9,27 @@
...
+    elif isinstance(array, np.ndarray):
+        return array
+    raise ValueError("Input should be either a list or a numpy array")
 
 
-def _np_to_list(array: npt.NDArray[np.float64]) -> list[float]:
-    return cast(list[float], array.tolist())
+def _np_to_list(array: npt.NDArray[np.float64]) -> list[float] | list[list[float]]:
+    if array.ndim == 1:
+        return [float(x) for x in array]
```
sylvlecl: I think this will be much slower than the numpy `tolist` method, we'd better stick with it

antarest/core/serde/np_array.py
```diff
@@ -9,19 +9,27 @@
...
+from typing import Annotated, Any, TypeAlias
 
 import numpy as np
 import numpy.typing as npt
 from pydantic import BeforeValidator, PlainSerializer
 
 
-def _list_to_np(array: list[float]) -> npt.NDArray[np.float64]:
-    return np.array(array, dtype=np.float64)
+def _list_to_np(array: Any) -> npt.NDArray[np.float64]:
```
sylvlecl, line 19: We should check the dimension here, while we are improving that.

antarest/core/serde/np_array.py
TheoPascoli: I added the tolist() method. However, it returns any so I added a cast ... I dont like it but dont see a better way

sylvlecl: It's ok as long as we check the dimension, we know that the cast is valid (or we could also cast to float first to make sure it's indeed floats).

PR #2912, opened 2025-12-19
feat(output): add variables view export endpoint

sylvlecl:
Adds options `export_format`, `index` and `header` to the existing variables view getter.
We might instead move that feature to another endpoint though.


MartinBelthle:
As discussed with Hatim we prefer another endpoint to separate concerns. I also prefer this inside the back-end as it will be easier to understand

PR #2913, opened 2025-12-19
refactor(matrix): remove study download level dto

sylvlecl:
Removes duplicate enum

### DIFF
--- a/antarest/core/serde/matrix_export.py
+++ b/antarest/core/serde/matrix_export.py
@@ -10,7 +10,7 @@
 #
 # This file is part of the Antares project.
 from pathlib import Path
-from typing import Iterator, Protocol
+from typing import IO, Iterator, Protocol
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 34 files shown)


antarest/core/serde/matrix_export.py
```diff
@@ -152,7 +152,7 @@ def get_stream_writer(self) -> DataframeStreamWriter:
     def export_table(
         self,
         df: pd.DataFrame,
-        export_path: str | Path,
+        export_path: str | Path | IO[bytes],
```
MartinBelthle, line 155: What's the link between this and the rest of the dev ? Anyway it doesn't really matter as it works in both cases

MartinBelthle (approved):

antarest/core/serde/matrix_export.py
sylvlecl: Indeed I mixed it up with the feature branch from which I extracted it

PR #2914, opened 2025-12-20
feat(celery): implement tasks for blob gc and auto archiver

TheoPascoli:


antarest/blobstore/blob_garbage_collector.py
```diff
@@ -9,54 +9,43 @@
...
-from antarest.core.utils.fastapi_sqlalchemy import db
-from antarest.core.utils.utils import StopWatch
+from antarest.maintenance.tasks.gc_blob import clean_blobs
 
 logger = logging.getLogger(__name__)
 
 
 class BlobGarbageCollector(IService):
+    """
+    Background service that periodically cleans unused blobs.
```
MartinBelthle: Minor but seems your doc is quite redundant with the one you wrote at the start of the file, we should only keep one IMO

antarest/maintenance/tasks/auto_archive_task.py
```diff
@@ -0,0 +1,38 @@
...
+from celery import Task
+
+from antarest.core.jwt import DEFAULT_ADMIN_USER
+from antarest.login.utils import current_user_context
+from antarest.maintenance.app import celery_app
+from antarest.maintenance.context import MaintenanceContext
+from antarest.maintenance.tasks.auto_archive import AutoArchiveTaskResult, archive_old_studies
+
+
+@celery_app.task(bind=True, name="antarest.maintenance.tasks.auto_archive_task", pydantic=True)
```
MartinBelthle: Why naming the task like this, `auto_archived_task` or `auto_archiver` seems simpler but perhaps you have to use the file path as a name ? If so it will be a pain everytime we'll move this code

antarest/maintenance/tasks/auto_archive_task.py
```diff
@@ -0,0 +1,38 @@
...
+def auto_archive_task(self: Task) -> AutoArchiveTaskResult:  # type: ignore[type-arg]
+    """Celery wrapper that delegates to archive_old_studies() with admin context."""
+    ctx: MaintenanceContext | None = getattr(self.app.conf, "maintenance_ctx", None)
```
MartinBelthle: minor: `self.app.conf.get("maintenance_ctx")` seems a bit more readable

antarest/maintenance/tasks/common.py
```diff
@@ -0,0 +1,45 @@
...
+
+"""Shared types for maintenance tasks."""
+
+from enum import IntEnum, StrEnum
+from typing import Optional
+
+from pydantic import BaseModel
+
+
+class TaskStatus(StrEnum):
```
MartinBelthle: I know it's not done inside this PR but I'm not a fan of naming this TaskStatus as we already have one. To me it would be clearer if we specified that they are BackGroundTaskStatus just to differentiate

antarest/maintenance/tasks/common.py
```diff
@@ -0,0 +1,45 @@
...
+    SUCCESS = "success"
+    PARTIAL_SUCCESS = "partial_success"
+    SKIPPED = "skipped"
+    ERROR = "error"
+
+
+class GCTaskResult(BaseModel):
```
MartinBelthle: I prefer to use GarbageCollector instead of GC as it doesnt' seem obvious to me when I read it. Or at least keep the full name in the docstring as it used to be

antarest/maintenance/tasks/gc_blob.py
```diff
@@ -0,0 +1,91 @@
...
+
+from antarest.blobstore.service import BlobService
+from antarest.core.utils.fastapi_sqlalchemy import db
+from antarest.core.utils.lock import LockNotAcquired, create_lock
+from antarest.maintenance.tasks.common import GCTaskResult, LockId, TaskStatus
+
+logger = logging.getLogger(__name__)
+
+
+def _delete_blobs(blob_service: BlobService, blobs: Set[str], dry_run: bool) -> int:
```
MartinBelthle: i prefer blob_ids instead of blobs as it's clearer

antarest/maintenance/tasks/gc_blob.py
```diff
@@ -0,0 +1,91 @@
...
+    except Exception as e:
+        logger.error("Blob GC failed", exc_info=e)
+        return GCTaskResult(
+            status=TaskStatus.ERROR,
+            error=str(e),
+            deleted_count=0,
+            duration_seconds=time.time() - start_time,
+        )
+
+    duration = time.time() - start_time
```
MartinBelthle, line 81: You could put this before the 2 exceptions as you're doing it 3 times

antarest/maintenance/tasks/gc_blob.py
```diff
@@ -0,0 +1,91 @@
...
+
+    except LockNotAcquired:
+        logger.warning("Could not acquire lock, another GC is probably running")
+        return GCTaskResult(
+            status=TaskStatus.SKIPPED,
+            reason="lock_not_acquired",
+            deleted_count=0,
+            duration_seconds=time.time() - start_time,
+        )
+    except Exception as e:
```
MartinBelthle, line 72: A bit of duplicated code between all tasks, perhaps we could factorize it as we'll have a lot of duplication in the end

antarest/maintenance/app.py
```diff
@@ -43,20 +35,14 @@
 
 def _mask_url_credentials(url: str) -> str:
     """Mask password in URL for safe logging."""
-    # Matches ://user:password@ or ://:password@ and replaces password with ***
```
MartinBelthle, line 46: I liked the regex explanation

antarest/maintenance/app.py
```diff
@@ -11,17 +11,9 @@
...
-This module creates the Celery app that will be used by both Beat and Worker.
-
-Configuration is loaded lazily via Celery signals to avoid side effects at import time:
-- `celeryd_init`: Loads config from ANTAREST_CONF and applies Celery-specific settings
-- `worker_init`: Creates the MaintenanceContext with all services
-- `on_after_configure`: Registers periodic tasks (Beat only)
-
-This design allows the module to be imported without side effects, making it
-easier to test and avoiding issues with config loading during imports.
+Config is loaded lazily via signals to avoid side effects at import time.
```
MartinBelthle: I also liked the more verbose doc, any reason you remove it ?

tests/integration/maintenance_blueprint/conftest.py
```diff
@@ -26,3 +32,16 @@
 dataset_repo = dataset_repo_fixture
 content_repo = content_repo_fixture
 matrix_service = matrix_service_fixture
+
+
+@pytest.fixture(name="simple_blob_service")
+def simple_blob_service_fixture(tmp_path: Path) -> BlobService:
```
MartinBelthle, line 38: Why not using the existing simple_blob_service fixture ?

tests/maintenance/conftest.py
```diff
@@ -0,0 +1,25 @@
...
+import pytest
+
+from antarest.maintenance.app import celery_app
+
+
+@pytest.fixture
+def celery_ctx_backup():
+    """Fixture to backup and restore celery_app.conf.maintenance_ctx."""
+    original_ctx = getattr(celery_app.conf, "maintenance_ctx", None)
+    yield
```
MartinBelthle, line 30: If i understand correctly you're doing this because the celery_app is global so after putting the context to None, you rehabilitate it for other tests. But you don't use it in your test you just call the fixture at test initialization and that's all. I'm a bit confused and it seems a bit sketchy, could you explain ?

antarest/maintenance/tasks/auto_archive.py
```diff
@@ -0,0 +1,162 @@
...
+class AutoArchiveTaskResult(BaseModel):
+    status: TaskStatus
+    archived_studies: int
+    duration_seconds: float
+    dry_run: bool
+    errors: List[str] = []
+    reason: Optional[str] = None
+
+
+def _get_studies_to_archive(study_service: StudyService, threshold_days: int) -> list[tuple[str, bool]]:
```
MartinBelthle: Could be a bit better to return a dataclass instead of a tuple for readability

antarest/maintenance/tasks/auto_archive.py
```diff
@@ -0,0 +1,162 @@
...
+    except Exception as e:
+        logger.error(f"Failed to archive study {study_id}: {e}", exc_info=e)
+        return ArchiveStudyResult(0, str(e))
+
+
+def _clear_old_snapshots(study_service: StudyService, retention_days: int, dry_run: bool) -> None:
+    """Clear variant snapshots older than retention_days."""
+    logger.info(f"Clearing snapshots older than {retention_days} days" + (" [dry-run]" if dry_run else ""))
+    if not dry_run:
+        study_service.storage_service.variant_study_service.clear_all_snapshots(datetime.timedelta(days=retention_days))
```
MartinBelthle: You should wait for this task to end as you waited for every other task

MartinBelthle (changes requested):
Several minor remarks

antarest/maintenance/tasks/auto_archive_task.py
TheoPascoli: Yea I just realised the name is only for information purposes. It can be anything. So I made it simpler for this one and the two others tasks as well

antarest/maintenance/tasks/auto_archive_task.py
TheoPascoli: Ok why not

antarest/maintenance/tasks/gc_blob.py
TheoPascoli: I'm not sure. I want to calculate all the time, including the time needed to process the exception. I think refactoring this would add complexity for minimal gains

antarest/maintenance/tasks/gc_blob.py
TheoPascoli: the `auto_archive` task has a different pattern. I would say no need to add an abstraction only for the two tasks that are basically the same. Maybe later

tests/integration/maintenance_blueprint/conftest.py
TheoPascoli: The existing fixture in conftest_services.py is scope="session". it's shared across all tests. For GC tests, we need isolation because each test creates and deletes blobs. If we use the session-scoped one, tests would interfere with each other. That's why I created a function-scoped version here, each test gets a fresh empty blob store.

tests/maintenance/conftest.py
TheoPascoli: You're right, it was confusing. I refactored it: the fixture is now called with_no_maintenance_ctx and does the setup itself (sets ctx to None) instead of letting the test do it. I tried using mock.patch but Celery's Settings object doesn't work well with standard mocking, so I kept the direct assignment approach but made it more explicit.

antarest/maintenance/tasks/auto_archive.py
TheoPascoli: The tuple is only 2 elements, documented in the docstring, and unpacked directly in the loop. I think it's fine for this case. But if you prefer, I can use a NamedTuple as a lightweight compromise

MartinBelthle: Yes I think the NamedTuple is a good compromise

antarest/maintenance/tasks/common.py
MartinBelthle: What's your opinion on this ?

antarest/maintenance/tasks/common.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class TaskStatus(StrEnum):
+    SUCCESS = "success"
+    PARTIAL_SUCCESS = "partial_success"
+    SKIPPED = "skipped"
+    ERROR = "error"
+
+
+class GarbageCollectorTaskResult(BaseModel):
```
MartinBelthle: Also minor but everywhere in the code we use `AntaresBaseModel` instead of `BaseModel` directly

antarest/maintenance/tasks/gc_blob.py
MartinBelthle: Mmmh I see, okay

antarest/maintenance/tasks/gc_blob.py
MartinBelthle: I think we can do better but I agree we can do it in a next PR

antarest/maintenance/app.py
```diff
@@ -69,161 +57,95 @@ def _load_config() -> Optional[Config]:
...
 
-# Static configuration that doesn't depend on external config files
 celery_app.conf.update(
-    # Serialization
     task_serializer="json",
     result_serializer="json",
     accept_content=["json"],
     enable_utc=True,
-    # Worker settings
-    worker_prefetch_multiplier=1,  # Don't prefetch (tasks can be long)
```
MartinBelthle, line 94: I think we should keep the existing comments, it helps understanding. For me at least

antarest/maintenance/app.py
```diff
@@ -69,161 +57,95 @@ def _load_config() -> Optional[Config]:
...
-    matrix_gc_interval = config.storage.matrix_gc_sleeping_time if config else 3600
+    storage = config.storage if config else StorageConfig()
 
-    # Register periodic task
-    sender.add_periodic_task(
-        matrix_gc_interval,
-        clean_matrices_task.s(),
-        name="matrix-gc",
-    )
+    sender.add_periodic_task(storage.matrix_gc_sleeping_time, clean_matrices_task.s(), name="matrix-gc")
```
MartinBelthle: Minor but there's a difference between the naming here and the naming of your celery tasks, shouldn't we harmonize this ?

tests/integration/maintenance_blueprint/conftest.py
MartinBelthle: Okay I see

tests/maintenance/conftest.py
MartinBelthle: Okay now it's indeed better documented thanks. Too bad Mock doesn't work as we could have issues with tests run in parallel I think

MartinBelthle (changes requested):
Minor changes and questions. After this we're good to go

antarest/maintenance/tasks/common.py
TheoPascoli: I changed it

MartinBelthle (approved):

PR #2922, opened 2025-12-23
feat(output): introduce second output storage

sylvlecl:
**Status:**
 - Introduced output storage type to select between 2 implementations
   at the output service level
 - Started implementation of "parquet" (should probably be renamed, see below) output storage,
   which only supports a few methods for now.
 - The storage uses an "LFS" (large file storage) abstraction to store 
   archives, which can have later an S3 implementation.
 - Proposition of configuration to enable the "new storage type"
 - Some unit tests for each new component, first integration tests for the
   import.

**TODOs:**
- **The main difficulty is to add the correct interactions with the study service:**
  In the process, we must take care of exposing only what's necessary to the study
  service
   - deleting a study must call delete of output service
   - copy of a study must call a new "copy" method of the output service
   - export of studies must call a method of the output service that will write
     the file output in the exported study directory
   - get_synthesis must get output information from the study service, to complement
     the information read from the file study

- polishing naming and configuration:
  - probably remove "parquet" from the name of the new output storage implementation,
    name it something like "internal output storage"
- more unit tests
- integration tests
- implement variables list extraction and saving in DB.


antarest/study/output/output_storage.py
```diff
@@ -24,6 +25,11 @@
 logger = logging.getLogger(__name__)
 
 
+class OutputStorageType(StrEnum):
+    FILE_TREE = "FILE_TREE"
+    PARQUET = "PARQUET"
```
sylvlecl, line 30: Actually I don't like the naming at all.
In particular, parquet is only part of  the solution, for variables, but not the whole solution.
It's also kind of an "implementation detail", that maybe we should not display in the API.

Maybe we should use something like "in-study" (for files stored in antares study format, inside the study), and "internal" (data is stored on the server, the actual implementation is a "detail" that the user of the API does not need to know).

Or "in-study" and "v2" or "v1" and "v2".

antarest/study/output/filestudy/extract_metadata.py
```diff
@@ -0,0 +1,53 @@
...
+        general=parameters_dict["general"],
+        input=parameters_dict["input"],
+        output=parameters_dict["output"],
+        optimization=parameters_dict["optimization"],
+        otherPreferences=parameters_dict["other preferences"],
+        advancedParameters=parameters_dict["advanced parameters"],
+        seedsMersenneTwister=parameters_dict["seeds - Mersenne Twister"],
+        playlist=list((get_playlist(parameters_dict) or {}).keys()),
+    )
+    # TODO: maybe the short version "eco" is expected, instead of "Economy"
```
sylvlecl, line 43: Seems consistent with current behaviour after all, to be removed

antarest/study/output/filestudy/extract_metadata.py
```diff
@@ -0,0 +1,53 @@
...
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+from pathlib import Path
+
+from antarest.core.serde.ini_reader import IniReader
+from antarest.study.model import StudySimResultDTO, StudySimSettingsDTO
+from antarest.study.storage.rawstudy.model.filesystem.config.files import get_playlist
+
+DUPLICATE_KEYS = [
```
MartinBelthle, line 18: This already existed, we should put code in common

antarest/study/output/filestudy/extract_metadata.py
```diff
@@ -0,0 +1,53 @@
...
+        input=parameters_dict["input"],
+        output=parameters_dict["output"],
+        optimization=parameters_dict["optimization"],
+        otherPreferences=parameters_dict["other preferences"],
+        advancedParameters=parameters_dict["advanced parameters"],
+        seedsMersenneTwister=parameters_dict["seeds - Mersenne Twister"],
+        playlist=list((get_playlist(parameters_dict) or {}).keys()),
+    )
+    # TODO: maybe the short version "eco" is expected, instead of "Economy"
+    #       is it anyway useful ??
```
MartinBelthle, line 44: Good question. Perhaps by the R scripts. The front doesn't seem to use it and antares-craft neither.
If so it could be good to not store it in DB

antarest/study/output/lfs/dir_lfs.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, directory: Path):
+        self._directory = directory
+        self._directory.mkdir(parents=True, exist_ok=True)
+
+    def _get_path(self, blob_id: str) -> Path:
+        return self._directory / blob_id
+
+    @override
+    def read_file(self, blob_id: str, target_path: Path) -> None:
+        shutil.copy(self._get_path(blob_id), target_path)
```
MartinBelthle, line 36: Doesn't seem to read something, it just copies the file to a target_path. The naming is misleading

MartinBelthle (commented):
Minor comments, did not read everything at all, I'll do another review later

sylvlecl:
Closing, will be contained in https://github.com/AntaresSimulatorTeam/AntaREST/pull/2923

PR #2923, opened 2025-12-24
refactor(output): extract remaing outputs handling to output service

sylvlecl:
**Big picture**

The `OutputService` now relies on a list of `IOutputStorage` implementations.

All operations on simulation outputs now need to go through those implementations.
It means that the following functionalities needed to move to the `IOutputStorage` interface, and to the output service so that they work with any output storage implementation:

 - retrieval of output synthesis data
 - deletion of outputs
 - copy of outputs
 - export of outputs as files (antares-simulator format)
 - storage and retrieval of logs

**Important implementation details**

- all output related features are moved to top level `antarest.output` module, to separate it clearly from `antarest.study`

- `FileStudyTreeConfigDTO` is not exposed anymore by the API, and is replaced with the `StudySynthesis` class.
   The previous DTO exposed internal details that should never have been exposed in the API.
   It does not break the frontend, but it's a breaking change, strictly speaking.
   Output storages are responsible for returning the metadata necessary to build the output information in that synthesis.

- **separation of responsibilities between study service and output service**:

  - The `StudyService` remain responsible for aggregating studies input data and output data. Therefore it needs some access to a subset of the output service functions, materialized by the `IOutputsAccess` interface (open to renaming)

  - The `StudyService`, when aggregating input and output data, must rely on the output service for output data, and not any more on the `IStudyStorage` interface, which is only responsible for input data. Therefore, the features of the `IOutputStorage` such as "flat export" should ignore, in general, the "output" folder.

**Proposition for the review**

I have split the work in 2 PRs:
- this one that just adds the possibility to have a list of `IOutputStorage` but does not use it yet (no "v2" implementation)
- the second that adds the start of "v2" implementation, lfs and so on

The big piece is the first one.


TheoPascoli:
I wrongly merged `origin/dev` in this. I rollbacked it.

antarest/study/storage/utils.py
```diff
@@ -165,17 +164,20 @@ def is_output_archived(path_output: Path) -> bool:
...
     if archived:
-        temp_dir = tempfile.TemporaryDirectory()
-        s = StopWatch()
         with ZipFile(path_output, "r") as zip_obj:
-            zip_obj.extract("info.antares-output", temp_dir.name)
-            info_antares_output = ini_reader.read(Path(temp_dir.name) / "info.antares-output")
-        logger.info(f"info.antares_output has been read in {s}s")
-        temp_dir.cleanup()
-
+            content = zip_obj.read("info.antares-output")
```
sylvlecl, line 179: Just a simplification, we can just read the content in memory for that small file, no need to have a temp directory and a file written

antarest/core/config.py
```diff
@@ -144,6 +144,26 @@ def from_dict(cls, data: JSON) -> "DbConfig":
         )
 
 
+@dataclass(frozen=True)
+class OutputStorageConfig:
+    """
+    Configuration for "new style" internal output storage
+    """
+
+    enable: bool = False
```
sylvlecl: rename it "enable_v2 ?"

antarest/study/storage/study_storage.py
```diff
@@ -69,8 +68,6 @@ def copy(
         dest_study_name: str,
         groups: Sequence[str],
         destination_folder: PurePosixPath,
-        output_ids: List[str],
```
sylvlecl, line 72: Now the copy of outputs must be handle by the output service, therefore it's removed from here.
The implementation must NOT copy the outputs.

antarest/study/storage/study_storage.py
```diff
@@ -148,30 +143,11 @@ def _check_study_exists(self, metadata: Study) -> None:
         if not self.exists(metadata):
             raise StudyNotFoundError(f"Study with the uuid {metadata.id} does not exist.")
 
-    @abstractmethod
```
sylvlecl, line 151: This is removed from here, because the aggregation of input data and output data into one single zip must be done at the study service level, which has both access to input data and to the output service.

antarest/study/storage/study_storage.py
```diff
@@ -180,17 +156,11 @@ def export_study_flat(
         Args:
             metadata: study.
             dst_path: destination path.
-            outputs: list of outputs to keep.
-            output_list_filter: list of outputs to keep (None indicate all outputs).
             denormalize: denormalize the study (replace matrix links by real matrices).
         """
 
     @abstractmethod
-    def get_synthesis(self, metadata: Study) -> FileStudyTreeConfigDTO:
```
sylvlecl, line 189: Now exposes the input data synthesis, not the "config" which is an implementation detail of the INode hierarchy

antarest/launcher/adapters/abstractlauncher.py
```diff
@@ -23,8 +24,26 @@
...
+    err: Path | None
+
+    @classmethod
+    def no_logs(cls) -> "SimulationLogs":
+        return cls(None, None)
+
+
 class ImportCallBack(Protocol):
-    def __call__(self, job_id: str, output_path: Path, additional_logs: Dict[str, List[Path]]) -> Optional[str]:
+    def __call__(self, job_id: str, output_path: Path, additional_logs: SimulationLogs) -> Optional[str]:
```
sylvlecl, line 46: In practice, we never had anything else than a stdout path and a stderr path, this clarifies what's expected

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -296,37 +320,23 @@ def _delete_workspace_file(self, study_path: Path) -> None:
...
-        xpansion_mode: Optional[str] = None,
-        log_dir: Optional[str] = None,
+        xpansion_mode: str | None,
+        log_dir: str | None,
     ) -> Optional[str]:
         if xpansion_mode:
             self._import_xpansion_result(job_id, xpansion_mode)
 
-        launcher_logs: Dict[str, List[Path]] = {}
-        if log_dir is not None:
```
sylvlecl, line 306: Was unnecessarily complicated, we only have one out file and one err file

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -451,34 +461,6 @@ def _handle_success(self, study: StudyDTO) -> None:
...
-
-    @staticmethod
-    def _get_log_path_from_log_dir(log_dir: Path, log_type: LogType = LogType.STDOUT) -> Optional[Path]:
-        pattern = {
-            LogType.STDOUT: "antares-out-*",
-            LogType.STDERR: "antares-err-*",
-        }[log_type]
-        return next(iter(log_dir.glob(pattern)), None)
-
-    def _clean_local_workspace(self) -> None:
```
sylvlecl, line 472: Was never used

antarest/launcher/service.py
```diff
@@ -408,31 +408,27 @@ def sort_log(log: JobLog, logs: Dict[JobLogType, List[str]]) -> Dict[JobLogType,
...
-                )
-                return "\n".join(app_logs[JobLogType.BEFORE] + [launcher_logs] + app_logs[JobLogType.AFTER])
-            return launcher_logs
+        if not job_result:
+            raise JobNotFound()
 
-        raise JobNotFound()
+        launcher_logs: str
+        if job_result.output_id:
+            launcher_logs = self.output_service.get_logs(job_result.study_id, job_result.output_id, log_type)
```
sylvlecl, line 418: we now get it from the output service when the output exists

antarest/output/output_blueprint.py
```diff
@@ -121,17 +121,16 @@ def create_output_routes(
         status_code=HTTPStatus.ACCEPTED,
         summary="Import Output",
     )
-    def import_output(uuid: UuidStr, output: UploadFile) -> str | None:
+    def import_output(uuid: UuidStr, output: UploadFile, storage_type: OutputStorageType | None = None) -> str | None:
```
sylvlecl, line 124: This is the only new thing in the API, is not actually used yet

antarest/study/web/studies_blueprint.py
```diff
@@ -298,6 +298,30 @@ def upgrade_study(uuid: SanitizedStr, target_version: SanitizedStr = "") -> str:
         # returns the task ID
         return study_service.upgrade_study(uuid, target_version)
 
+    def _output_selection(with_outputs: bool | None, output_ids: list[str]) -> OutputSelection:
```
sylvlecl, line 301: That logic was duplicated in multiple places, we now use a less ambiguous type as soon as possible

tests/integration/variant_blueprint/test_variant_manager.py
```diff
@@ -108,7 +108,7 @@ def test_variant_manager(
 
         res = client.get(f"v1/studies/{variant_id}/synthesis", headers=admin_headers)
 
-        assert variant_id in res.json()["output_path"]
```
sylvlecl, line 111: Not exposed anymore, should never have been

tests/integration/test_integration.py
```diff
@@ -1328,399 +1331,6 @@ def test_import_with_editor(
     assert imported_antares_data["editor"] == "importer"
 
 
-def test_copy_with_editor_preservation(client: TestClient, admin_access_token: str) -> None:
```
sylvlecl, line 1331: Those copy tests have been moved to the more specific `studies_blueprint/test_study_copy.py` file

tests/integration/studies_blueprint/test_synthesis.py
```diff
@@ -25,7 +25,7 @@
 
 def _compare_resource_file(actual: dict[str, Any], res_path: Path) -> None:
     # note: private data are masked in the resource file
-    masked = dict.fromkeys(["study_path", "path", "output_path", "study_id"], "DUMMY_VALUE")
```
sylvlecl, line 28: The paths are not exposed anymore, no need to mask them any more

tests/integration/studies_blueprint/test_study_export.py
```diff
@@ -10,44 +10,101 @@
...
 from starlette.testclient import TestClient
 
 
+def _zip_namelist(zip_path: Path) -> set[str]:
+    with zipfile.ZipFile(zip_path) as zip_file:
+        return set(zip_file.namelist())
+
+
 # Doing exports with both .zip and .7z
-def test_export_study_with_both_compression(
```
sylvlecl, line 19: Just using `admin_client` and the `download_to_file` utility in those existing tets

tests/integration/studies_blueprint/test_study_export.py
```diff
@@ -10,44 +10,101 @@
...
-        assert py7zr.is_7zfile(base_path / "export.7z")
+    download_to_file(client, json_export_download_id, base_path / "export.zip")
 
     # exporting with an unauthorized compression
     failed_export_res = client.get(f"/v1/studies/{internal_study_id}/export?compression=.rar")
     assert failed_export_res.status_code == 422, failed_export_res
+
+
+# Doing exports with both .zip and .7z
+def test_export_study_outputs(admin_client: TestClient, tmp_path) -> None:
```
sylvlecl, line 52: This is moved from `test_integration`

tests/integration/output_blueprint/test_digest.py
```diff
@@ -41,12 +41,11 @@ def test_get_digest_endpoint(self, client: TestClient, user_access_token: str, i
         assert area_matrix["columns"][:3] == [[""], ["OV. COST", "Euro", "EXP"], ["OP. COST", "Euro", "EXP"]]
 
         # Asserts we have a 404 Exception when the output doesn't exist
-        fake_output = "fake_output"
-        res = client.get(f"/v1/private/studies/{internal_study_id}/outputs/{fake_output}/digest-ui")
+        res = client.get(f"/v1/private/studies/{internal_study_id}/outputs/fake_output/digest-ui")
         assert res.status_code == 404
         assert res.json() == {
-            "description": f"'{fake_output}' not a child of Output",
-            "exception": "ChildNotFoundError",
```
sylvlecl, line 49: We now get a more specific error

tests/integration/output_blueprint/test_output_storage_v2.py
```diff
@@ -0,0 +1,73 @@
...
+#
+# This file is part of the Antares project.
+import io
+from pathlib import Path
+
+from integration.assets import ASSETS_DIR
+from starlette.testclient import TestClient
+
+
+def test_import(admin_client: TestClient, internal_study_id: str, tmp_path: Path) -> None:
```
sylvlecl: Those are the only new integration tests

antarest/output/storage/output_storage.py
```diff
@@ -0,0 +1,212 @@
+# Copyright (c) 2026, RTE (https://www.rte-france.com)
```
sylvlecl, line 1: That file is actually moved from `antarest.study.output`, methods have been added.

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,504 @@
...
+
+
+def _add_logs(output_dir: Path, logs: SimulationLogs) -> None:
+    if logs.out:
+        _copy_file(logs.out, output_dir, PurePosixPath("antares-out.log"))
+    if logs.err:
+        _copy_file(logs.err, output_dir, PurePosixPath("antares-err.log"))
+
+
+class InStudyFileOutputStorage(IOutputStorage):
```
sylvlecl, line 196: This file is actually moved from `study.antarest.output`, but methods have been added: copy, get_logs, write_to_dir ...

antarest/launcher/service.py
```diff
@@ -509,87 +505,75 @@ def _import_output(
...
-                Optional[str],
-                getattr(
-                    job_launch_params,
-                    LAUNCHER_PARAM_NAME_SUFFIX,
-                    None,
-                ),
-            )
 
             self._save_solver_stats(job_result, output_true_path)
-            if additional_logs and not output_is_zipped:
```
sylvlecl, line 538: The import of logs is now handled inside the output storage import function

antarest/launcher/service.py
```diff
@@ -509,87 +505,75 @@ def _import_output(
...
-                    concat_files(
-                        log_paths,
-                        output_true_path / log_name,
-                    )
 
-        if study_id:
-            zip_path: Optional[Path] = None
+        zip_path: Path | None = None
+        # Optimized path for studies stored on external devices, that will then be unarchived there.
+        # TODO: that whole optimization path should be refactored to:
```
sylvlecl, line 533: I added that TODO, but changing cannot be part of that PR, it would be a dedicated refacto

antarest/output/output_service.py
```diff
@@ -297,30 +313,35 @@ def unarchive_output_task(notifier: ITaskNotifier) -> TaskResult:
...
 
     def import_output(
         self,
         uuid: str,
         output: BinaryIO | Path,
         output_name_suffix: Optional[str] = None,
         auto_unzip: bool = True,
+        storage_type: OutputStorageType | None = None,
+        logs: SimulationLogs = SimulationLogs.no_logs(),
     ) -> Optional[str]:
```
TheoPascoli, line 365: You should add the associated docstring

antarest/output/output_service.py
```diff
@@ -197,27 +202,39 @@ def run_task(self, notifier: ITaskNotifier) -> TaskResult:
 class OutputService:
     def __init__(
         self,
-        storage: IOutputStorage,
+        storage: IOutputStorage | list[IOutputStorage],
```
TheoPascoli: It looks like it's always a list

antarest/core/exceptions.py
```diff
@@ -397,6 +397,16 @@ def __init__(self, output_id: str) -> None:
         super().__init__(HTTPStatus.EXPECTATION_FAILED, message)
 
 
+class OutputAlreadyExists(HTTPException):
+    """
+    Exception raised when a user wants to unarchive an output which is already unarchived.
```
MartinBelthle: Minor: To me the exception name does not really match its description. Could be something like `OutputUnarchivedAlreadyExists`. Perhaps it's a too long name

antarest/launcher/model.py
```diff
@@ -219,6 +219,8 @@ class JobResult(Base):
     msg: Mapped[Optional[str]] = mapped_column(String())
     output_id: Mapped[Optional[str]] = mapped_column(String())
     exit_code: Mapped[Optional[int]] = mapped_column(Integer)
+
+    # TODO: not used, should be removed
```
MartinBelthle: You talk about the `solver_stats` ? If so, how do we know it's not used ? I recall using it at some point used it for Alexander. But we can perhaps find this information in other ways

antarest/study/storage/rawstudy/model/filesystem/config/model.py
```diff
@@ -95,6 +95,8 @@ class AreaConfig(AntaresBaseModel, extra="forbid"):
 class Simulation(AntaresBaseModel):
     """
     Object linked to /output/<simulation_name>/about-the-study/** information
+
+    # TODO: Used in "config" and therefore inside filestudy implementation, but is it used elsewhere ?
```
MartinBelthle: Seems it's only used for the filesystem implementation indeed

antarest/study/storage/rawstudy/model/filesystem/root/output/output.py
```diff
@@ -30,9 +29,6 @@ def build(self) -> TREE:
             )
             for i, s in self.config.outputs.items()
         }
-
-        if (self.config.path / "logs").exists():
```
MartinBelthle, line 34: Not sure of what I'm saying but this might be useful for Desktop app as perhaps the local launcher writes the logs here. To be confirmed. Otherwise indeed we can remove it

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,511 @@
...
+        """
+        study_outputs = self._outputs_provider.get_outputs(study_id)
+
+        logger.info(f"Exporting output {output_id} from study {study_id}")
+
+        path_output = study_outputs.outputs_path / output_id
+        path_output_zip = study_outputs.outputs_path / f"{output_id}{ArchiveFormat.ZIP}"
+
+        if path_output_zip.exists():
+            shutil.copyfile(path_output_zip, target)
```
TheoPascoli: You did not modify this but this condition is useless because you test it above

antarest/core/exceptions.py
```diff
@@ -397,6 +397,16 @@ def __init__(self, output_id: str) -> None:
...
+    """
+
```
TheoPascoli, line 404: You have the wrong docstring

tests/integration/maintenance_blueprint/test_gc_variable_view.py
```diff
@@ -69,6 +69,7 @@ def test_deletes_old_variable_views(self, matrix_service: MatrixService):
         with db():
             matrix_id = matrix_service.create(matrix_data)
             db.session.add(_create_study(study_id))
+            db.session.commit()
```
MartinBelthle: WHy do we need to add these `commit()` ?

tests/integration/utils.py
```diff
@@ -54,7 +53,7 @@ def duration_threshold(base_duration: float) -> float:
 
 def wait_task_completion(
     client: TestClient,
-    access_token: str,
+    access_token: str | None,
```
MartinBelthle, line 56: Minor but you added this because of `tes_study_copy.py` where you explicitely call this with None. Why did you wanted to not have an access token ?

TheoPascoli (changes requested):
Hello, first mini review. I'm not done yet !

tests/storage/integration/test_STA_mini.py
```diff
@@ -492,8 +492,15 @@ def test_sta_mini_import_output(tmp_path: Path, storage_service: StudyService, c
...
+
+    result = client.post(
+        "/v1/studies",
+        params={"name": "test"},
+    )
+    assert result.status_code == 201
+
     result = client.post(
-        f"/v1/studies/{UUID}/output",
+        f"/v1/studies/{result.json()}/output",
```
sylvlecl, line 503: Previously, this test tried to import an already existing import, this now fails.
We now import it in an empty study.

tests/storage/web/test_output_bp.py
```diff
@@ -51,37 +45,3 @@ def test_output_whole_download(tmp_path: Path) -> None:
         f"/v1/studies/{UUID}/outputs/{output_id}/export",
     )
     assert res.status_code == HTTPStatus.OK
-
-
-def test_sim_result() -> None:
```
sylvlecl, line 56: Did not bother to update that, it's covered by integration tests

antarest/output/output_blueprint.py
MartinBelthle: Minor but I think it's a bit weird to have by default a `None`, it's not explicit what this will do. I prefer when we have this explicit pattern: `storage_type: OutputStorageType | None = xxx`

antarest/output/output_blueprint.py
```diff
@@ -353,7 +342,6 @@ def aggregate_links_raw_data(
         )
 
         # Avoid vulnerabilities by sanitizing the `uuid` and `output_id` parameters
```
MartinBelthle: You can remove this comment

tests/storage/business/test_export.py
```diff
@@ -36,81 +32,29 @@
 from tests.conftest import empty_study_fixture
 from tests.db_statement_recorder import DBStatementRecorder
 from tests.helpers import create_raw_study, dirhash, with_db_context
+from tests.integration.test_helpers.outputs import create_minimal_output_dir_from_name
 
 
-def test_export(
-    empty_study_930: FileStudy, raw_study_service: RawStudyService, command_context: CommandContext
+def test_export_flat_export_all_files_except_output(
```
sylvlecl, line 38: Checks that the "study storage" does not export outputs.
It's now the output storage which is reponsible for this

antarest/output/storage/file_output_storage.py
sylvlecl: Indeed it was weird, I improved the logic

antarest/output/output_service.py
sylvlecl: Updated to accept only sequences

antarest/output/output_service.py
sylvlecl: done!

antarest/output/storage/output_storage.py
```diff
@@ -0,0 +1,217 @@
...
+    @abstractmethod
+    def delete_output(self, study_id: str, output_id: str) -> None:
+        """
+        Delete a simulation output
+        """
+
+    @abstractmethod
+    def write_output_to_dir(self, study_id: str, output_id: str, parent: Path) -> None:
+        """
+        Writes outputs in filestudy format into the specified parent directory.
```
MartinBelthle, line 146: Shouldn't this be storage type insensitive ?

antarest/output/output_service.py
```diff
@@ -197,27 +202,39 @@ def run_task(self, notifier: ITaskNotifier) -> TaskResult:
...
         self._matrix_service = matrix_service
         self._tmp_dir = tmp_dir
         self._studies_repository = studies_repository
-        self._cache = cache
 
         OutputVariablesMatrixUsageProvider(self._matrix_service)
 
+    def _find_output_storage(self, study_id: str, output_id: str) -> IOutputStorage:
+        for storage in self._storages:
+            if storage.output_exists(study_id, output_id):
```
MartinBelthle, line 240: With this logic if the output exist in the 2 storage form we'll "randomly" choose one. Perhaps it's not possible for an output to be stored in the 2 formats. If so, it should probably be mentioned just for clarity sake

antarest/output/output_service.py
```diff
@@ -197,27 +202,39 @@ def run_task(self, notifier: ITaskNotifier) -> TaskResult:
...
+                return storage
+        raise OutputNotFound(output_id)
+
+    def _get_storage(self, storage_type: OutputStorageType | None) -> IOutputStorage:
+        if not storage_type:
+            return self._storages[0]
```
MartinBelthle, line 249: Due to this, we sohuld probably mention when building the service that the tuple order matters (perhaps it's already documented)

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,510 @@
...
+        return False
+    zip_path = outputs_root / f"{output_id}{ArchiveFormat.ZIP}"
+    if zip_path.is_file():
+        return True
+    raise OutputNotFound(output_id)
+
+
+def _output_exists(outputs_root: Path, output_id: str) -> bool:
+    output_path = outputs_root / output_id
+    zip_path = outputs_root / f"{output_id}{ArchiveFormat.ZIP}"
```
MartinBelthle: We could check for `.7z` too even if I think this case cannot happen at the moment.
Same for `_is_output_archived` as `_find_archived_output` supports both

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,510 @@
...
+    study_outputs_path.mkdir(exist_ok=True)
+    shutil.copyfile(output_zip_path, final_path)
+
+    _add_logs(final_path, logs)
+
+    logger.info(f"Copied output for {study_id} in {t}s")
+    return output_full_name
+
+
+def _import_dir(output_dir: Path, study_outputs_path: Path, output_name_suffix: str | None) -> str:
```
MartinBelthle: Never used

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,510 @@
...
+    output_full_name = extract_output_name(output_dir, output_name_suffix)
+
+    # Moving temporary directory to actual directory
+    output_dir.rename(study_outputs_path / output_full_name)
+    return output_full_name
+
+
+def _copy_file(src: Path, dst_root: Path, dst: PurePosixPath) -> None:
+    """
+    Copies src to dst inside dst_root, dst_root being either a dict or a zip file.
```
MartinBelthle: `dst_root` is just typed as a Path not sure the doc is up to date

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,510 @@
...
+
+        # Optimized path to copy outputs on external devices. Will then be unarchived there.
+        if isinstance(output, Path) and is_zip(output):
+            output_name = _import_zip_as_archived(
+                study_id, output, study_outputs.outputs_path, output_name_suffix, logs
+            )
+            remove_from_cache(self._cache, study_id)
+            return output_name
+
+        # Output is first written to a temporary directory which is then moved to the actual output path
```
MartinBelthle, line 238: Nice to add some comments in this code as it's pretty hard to follow our own old logic :/

antarest/output/storage/file_output_storage.py
MartinBelthle: In fact after reading the rest of the code we should remove the support of .7z inside `_find_archived_output` as it's the only place we do

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,510 @@
...
+        else:
+            raise OutputNotFound(output_id)
+
+    @override
+    def export_output(self, study_id: str, output_id: str, target: Path) -> None:
+        """
+        Export and compresses study inside zip
+        Args:
+            output_id: output id
+            target: path of the file to export to
```
MartinBelthle, line 346: Minor but `study_id` is missing in the docstring

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,510 @@
...
+        # Specific logic for "external" workspaces: we ask for unarchiving to the remote executor ...
+        workspace = study_outputs.study_workspace
+        if workspace != DEFAULT_WORKSPACE_NAME:
+            self._remote_unarchive(output_id, study_outputs)
+            remove_from_cache(self._cache, study_id)
+        else:
+            study_outputs = self._outputs_provider.get_outputs(study_id)
+            outputs_path = study_outputs.outputs_path
+            try:
+                # TODO: should remove the zip ?
```
MartinBelthle: Indeed we should

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,510 @@
...
+                unzip(outputs_path / output_id, outputs_path / f"{output_id}{ArchiveFormat.ZIP}")
+                remove_from_cache(self._cache, study_id)
+            except Exception as e:
+                # TODO: we should probably raise here and remove partially unzipped files
```
MartinBelthle: Harder to do I think but yes

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,510 @@
...
+            columns_names,
+            transform_columns_headers,
+            mc_years,
+        )
+        return aggregator_manager.aggregate_output_data()
+
+    @override
+    def extract_variables_list(self, study_id: str, output_id: str) -> OutputVariablesList:
+        study_outputs = self._outputs_provider.get_outputs(study_id)
+        return extract_variables_list(study_outputs.outputs_path / output_id)
```
MartinBelthle: Makes me realize that for instance the `variables_management.py` is implicety linked to the current output implem, such as `aggregator_management.py`. Shouldn't we create a directory for file_output things and put them there ?

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,510 @@
...
+            LogType.STDERR: [
+                ["output", output_id, "antares-err"],
+            ],
+        }
+        empty_log = False
+        for log_location in log_locations[log_type]:
+            try:
+                # Assume UTF-8 but ignore errors, it's difficult to be sure of log encoding
+                # especially because of windows error messages
+                log = cast(
```
MartinBelthle: Really minor but you could replace with `cast(bytes, file_study.tree.get(log_location, depth=1))` to be one lined

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -245,8 +221,7 @@ def copy(
 
         shutil.copytree(src_path, dest_path, ignore=shutil.ignore_patterns("output"))
 
-        copy_output_folders(src_path / "output", dest_path / "output", with_outputs, output_ids)
-
+        # TODO: now we create the config too early without the outputs, maybe ?
```
MartinBelthle: What about this todo ? Indeed we'll have to  update the config after copying the outputs

antarest/study/web/studies_blueprint.py
```diff
@@ -298,6 +298,30 @@ def upgrade_study(uuid: SanitizedStr, target_version: SanitizedStr = "") -> str:
...
+            raise IncorrectArgumentsForCopy("output_ids can only be used with with_outputs=True")
+        if with_outputs:
+            if output_ids:
+                return output_ids
+            else:
+                return "all"
+        elif with_outputs is None:
+            return output_ids if output_ids else "none"
+        else:
+            return "none"
```
MartinBelthle: Wouldn't it be better for this to return a list of ids ? As here you introduce `all` and `none` with a match case later in the code. I think the list would be simpler

antarest/study/web/studies_blueprint.py
```diff
@@ -298,6 +298,30 @@ def upgrade_study(uuid: SanitizedStr, target_version: SanitizedStr = "") -> str:
...
+        """
+        Translates API input to OutputSelection type.
+
+        Output copy behavior:
+            - If `with_outputs` is True and `output_ids` are specified: only the specified outputs are copied.
+            - If `with_outputs` is True and `output_ids` is empty: all outputs are copied.
+            - If `with_outputs` is False and `output_ids` are specified: an error is raised (incoherent configuration).
+            - If `with_outputs` is False: no outputs are copied
+            - If `with_outputs` is None and `output_ids` are specified: outputs will be copied; behaves like `with_outputs=True`.
+            - If `with_outputs` is None and `output_ids` is empty: no outputs are copied.
```
MartinBelthle, line 311: Thanks for the docstring

antarest/service_creator.py
```diff
@@ -180,6 +182,13 @@ def build_favorite_service(
     return favorite_study_service, favorite_directory_service
 
 
+# TODO: move it elswhere ?
+def _delete_study_outputs(storage: IOutputStorage, study_id: str) -> None:
```
MartinBelthle: Never used

antarest/study/dtos.py
```diff
@@ -0,0 +1,82 @@
...
+from antarest.study.business.model.district_model import District
+from antarest.study.model import StudyVersionInt
+from antarest.study.storage.rawstudy.model.filesystem.config.model import (
+    AreaConfig,
+    EnrModelling,
+    FileStudyTreeConfig,
+)
+
+
+class StudyDataSynthesis(AntaresBaseModel):
```
MartinBelthle, line 27: Nice to add this class :). But we're still deeply dependant on filesystem ugly ass classes such as `AreaConfig` :\

MartinBelthle (changes requested):
Didn't read everything but a big chunk of it. Not that much interesting remarks on my side

antarest/output/storage/file_output_storage.py
sylvlecl: That's right, will change it

sylvlecl: Definition of output paths (archived or not, is now defined in a single function (wth only zip support for archives).

antarest/output/storage/file_output_storage.py
sylvlecl: it's a path either to a zip file or a directory (I fixed the docstring)

antarest/output/storage/file_output_storage.py
sylvlecl: Actually I did not changed it but updated the comment:
the fact that we don't remove the archive allows to speed up a second archival, for the moment I leave it this way.

antarest/output/storage/file_output_storage.py
sylvlecl: I added removal and raise

antarest/output/storage/file_output_storage.py
sylvlecl: Yes indeed, in the next PR for addition of a 2nd storage, I added a separate module to add those file processing, since they will be useful for both implementations

antarest/output/storage/file_output_storage.py
```diff
@@ -0,0 +1,510 @@
...
+        return extract_variables_list(study_outputs.outputs_path / output_id)
+
+    @override
+    def get_logs(self, study_id: str, output_id: str, log_type: LogType) -> str:
+        study_outputs = self._outputs_provider.get_outputs(study_id)
+        file_study = study_outputs.get_file_study()
+
+        log_locations = {
+            LogType.STDOUT: [
+                ["output", output_id, "antares-out"],
```
sylvlecl, line 498: Note that here I removed the possibility to have logs in an "output/logs" dir, which was never used and did not actually work

antarest/output/storage/output_storage.py
sylvlecl: Not sure what you mean, but basically this requires that any storage is able to provide the output in the "legacy" format, which is needed for export for example

antarest/output/output_service.py
sylvlecl: I added a docstring to the service which mentions this.

antarest/output/output_service.py
sylvlecl: Indeed this is something we should forbid, but it's not checked for now, I will add some checks.

sylvlecl: I added a chek at import time, which is not great for now because the ID is determined by the storage, so the best we can do is checking after the import that the ID does not conflict.

antarest/study/storage/rawstudy/model/filesystem/config/model.py
sylvlecl: Removed the TODO, it's already handled by the separation of the "config" class which is specific to the filestudy implementation, and the StudySynthesis class which will be independent from the study storage.
In synthesis, OutputDetail is used instead of Simulation

antarest/study/storage/rawstudy/model/filesystem/root/output/output.py
sylvlecl: I checked this:
this directory is currently not used.
It was introduced with an old dev about importing zipped output. I guess it required to store the log outside from the output itself.
The "save" actually did not work since like one year ago, because the path is considered to not exist when the directory does not exist, which prevents to save it in the first place.

antarest/study/storage/rawstudy/raw_study_service.py
sylvlecl: I removed the comment, it's already handled:
the output storage invalidates the cache every time an output is added, so this will be OK

antarest/study/web/studies_blueprint.py
sylvlecl: Actually I just copied it from elsewhere :)

antarest/study/web/studies_blueprint.py
sylvlecl: The resolution of "all" to a list of IDs cannot be done here (or we would need to add more logic here to get the outputs of the study).
Indeed we could remove the "none" though.

sylvlecl: Removed the none value

tests/integration/maintenance_blueprint/test_gc_variable_view.py
sylvlecl: Otherwise I have a foreign key violation at commit time later, which I don't understand yet, will dig a little more

tests/integration/utils.py
sylvlecl: The access token is already configured in the client, which makes this useless

tests/integration/maintenance_blueprint/test_gc_variable_view.py
sylvlecl: So it's interesting:
sqlalchemy ORM does not guarantee the execution order of requests.

In order to send requests to the database, we need to "flush" the session, which I added here.

An interesting thing is that we disable autoflushing in our application, which does not seem really relevant.

At first view, it would seem more relevant to enable it, and only disable it where performance matters.

https://docs.sqlalchemy.org/en/21/orm/session_basics.html#flushing

sylvlecl: About why it has changed: I guess that the execution order depends on implementation details of sqlalchemy.
I moved the output variable view model to a different directory, maybe that's why the ordering in sqlalchemy internals has changed.

MartinBelthle (dismissed):
To me we can merge this work as it is and if there are things we want to refactor we can do this afterwards as for now it's hard to tell as I'm not familiar with the code

PR #2925, opened 2026-01-07
feat(favorite): add backend favorite

mehdiwahada:
What it does ? : it adds persistance on the user favorite studies on the API and DB side

Why ? : Before, the favorites were only linked to the front, making it difficult when users switched PCs. 

How to test it ? : 

- API : 
- DB : put the main in debug mode and breakpoints on the favorite method, and check the datas are good


alembic/versions/7b406752d80_add_favorite_table.py
```diff
@@ -0,0 +1,31 @@
...
+branch_labels = None
+depends_on = None
+
+def upgrade():
+    op.create_table(
+        "favorite",
+        sa.Column("user_id", sa.String(), nullable=False),
+        sa.Column("study_id", sa.String(), sa.ForeignKey("study.id"), nullable=False),
+        sa.PrimaryKeyConstraint("user_id", "study_id", name="pk_favorite"),
+        sa.UniqueConstraint("user_id", "study_id", name="uid_favorites"),
```
TheoPascoli: You dont need this constraint because you specified`sa.PrimaryKeyConstraint("user_id", "study_id", name="pk_favorite")`. And a `primaryKey` must be unique. So the `Constraint` is redondant

antarest/favorite/main.py
```diff
@@ -0,0 +1,31 @@
...
+from typing import Optional
+
+from antarest.core.application import AppBuildContext
+from antarest.core.config import Config
+from antarest.favorite.repository import FavoriteRepository
+from antarest.favorite.service import FavoriteService
+from antarest.favorite.web import create_favorite_routes
+
+
+def build_favorite_service(
```
TheoPascoli: You can define this method directly in the `service_creator` and delete this file.

antarest/favorite/service.py
```diff
@@ -0,0 +1,32 @@
...
+from antarest.favorite.repository import FavoriteRepository
+from antarest.login.utils import get_user_id
+
+
+class FavoriteService:
+    def __init__(self, favorite_repository: FavoriteRepository):
+        self.favorite_repository = favorite_repository
+
+    def list_favorites(self) -> list[FavoriteDTO]:
+        favorites = self.favorite_repository.get_all()
```
TheoPascoli, line 23: Here you are returning all favorites. You must return only the user's favorite by filtering with the user id.

antarest/favorite/web.py
```diff
@@ -0,0 +1,40 @@
...
+def create_favorite_routes(favorite_service: FavoriteService, config: Config) -> APIRouter:
+    auth = Auth(config)
+    bp = APIRouter(prefix="/v1", tags=[APITag.favorite], dependencies=[auth.required()])
+
+    @bp.get("/favorite", summary="Listing favorites for current user")
+    def list_favorites() -> list[FavoriteDTO]:
+        logger.info("Listing favorites for current user")
+        return favorite_service.list_favorites()
+
+    @bp.post("/favorite/{uuid}", summary="Create new favorite")
```
TheoPascoli: Here you say `{uuid}` and right after `{study_uuid}`, you should be consistent. `uuid` is fine

alembic/versions/7b406752d80_add_favorite_table.py
```diff
@@ -0,0 +1,31 @@
...
+revision = "7b406752d80"
+down_revision = "0146b79f723c"
+branch_labels = None
+depends_on = None
+
+def upgrade():
+    op.create_table(
+        "favorite",
+        sa.Column("user_id", sa.String(), nullable=False),
+        sa.Column("study_id", sa.String(), sa.ForeignKey("study.id"), nullable=False),
```
TheoPascoli: In the model, you defined `ondelete="CASCADE"` but not here. You must include it here as well.

TheoPascoli (changes requested):
Hello Mehdi, good work overall, just some comments.
moreover, I'm really not a fan of the "favorite" word to define user's favorite studies. We may want to add favorite directories later on and it might cause some trouble. I suggest that you rename it with something like : "starred_study" (like the ihm star), bookmarked_study" ou just "starred" like github.
Any opinion on this @MartinBelthle ?

MartinBelthle:
> Hello Mehdi, good work overall, just some comments. moreover, I'm really not a fan of the "favorite" word to define user's favorite studies. We may want to add favorite directories later on and it might cause some trouble. I suggest that you rename it with something like : "starred_study" (like the ihm star), bookmarked_study" ou just "starred" like github. Any opinion on this @MartinBelthle ?

I agree with you, we'll probably add the directories later so we have to specify that it concerns studies. "favorite_study" seems fine. I personnaly think "starred" and "bookmark" are a bit harder to understand but I don't have a strong opinion

antarest/favorite/model.py
```diff
@@ -0,0 +1,54 @@
...
+
+from antarest.core.persistence import Base
+from antarest.core.serde import AntaresBaseModel
+from antarest.study.model import Study
+
+logger = logging.getLogger(__name__)
+
+
+class FavoriteStudyDTO(AntaresBaseModel, extra="forbid"):
+    id_study: str
```
sylvlecl: Better name it study_id

antarest/core/utils/web.py
```diff
@@ -12,6 +12,7 @@
 
 
 class APITag:
+    favorite = "Favorite"
```
sylvlecl: Let's follow the other tags wording "Manage favorites"

sylvlecl (commented):
minor comments

sylvlecl (approved):

PR #2926, opened 2026-01-07
perf(matrix)!: use polars instead of pandas in the back-end

MartinBelthle:
Breaking changes:
- We do not support reading data inside an archived study anymore. Before the GET /raw endpoint worked for input files but not output ones. And all other endpoints failed. 
- When reading a matrix, its headers were [0, 1, 2, ...]. Now it's ['0', '1', '2', ...]


tests/integration/test_matrix_blueprint.py
```diff
@@ -18,7 +18,7 @@ def test_get_matrices(client: TestClient, admin_access_token: str) -> None:
     client.headers = {"Authorization": f"Bearer {admin_access_token}"}
 
     matrix = [[1, 2], [1, 7]]
-    matrix_1 = [[3, 6], [5, 10, 28]]
+    matrix_1 = [[3, 6, 4], [5, 10, 28]]
```
MartinBelthle, line 21: Before pandas used to assume the missing value was a NaN. Didn't make sense as we're not supposed to have NaN inside inputs so I changed it as polars does not assume, it raises.

tests/matrixstore/test_repository.py
```diff
@@ -250,11 +252,12 @@ def test_concurrent_save(self, tmp_path: str) -> None:
         for more chances to generate problems.
         """
         trial_count = 10
+        matrix = pl.DataFrame(data=np.zeros(shape=(8760, 1)), schema=["0"])
```
MartinBelthle: I used a matrix in legacy format, otherwise the test never ends. I do not understand why but it's probably important. Seems the `df.to_pandas()` inside the `compute_hash` method never works but I can't figure out why

MartinBelthle: We could also change the `compute_hash` method for non legacy matrix as we only have the variables view ones inside the matrixstore at the moment I think. If the problem comes from here

antarest/study/output/utils.py
```diff
@@ -107,3 +112,38 @@ def split_concatenated_columns_from_dataframe(df: pd.DataFrame) -> Iterator[Time
...
+        )
+
+
+@dataclass(frozen=True)
+class OutputDataFrame:
+    data: pd.DataFrame
+    headers: list[list[str]]
+
+
+def parse_output_file(file_path: Path, first_column: int) -> OutputDataFrame:
```
MartinBelthle, line 142: To be used when reading output matrices in a future work as for now we still use pandas for this

tests/matrixstore/test_repository.py
sylvlecl: Weird, I think it's very important to make this work though, otherwise we'll probably encounter the issue in production some day.

We can use the following "hack" to make sure `to_pandas` is never executed concurrently:
```python
# module level
_LOCK = threading.Lock()

...
# in function
    with _LOCK:
        pandas_df = df.to_pandas()
```

It seems to solve the issue

MartinBelthle: I implemented this solution

sylvlecl (approved):
Looks good to me!

PR #2929, opened 2026-01-10
feat(celery): watcher task

TheoPascoli:


antarest/core/config.py
```diff
@@ -169,6 +169,8 @@ class StorageConfig:
     blobstore: Path = Path("./blobstore")
     blob_gc_sleeping_time: int = 86400
     blob_gc_dry_run: bool = False
+    watcher_scan_sleeping_time: int = 60
```
TheoPascoli, line 172: Right now it is 2 seconds. I wonder if it's not too agressive ?

smailio: what do you mean by right now it's 2 seconds ?

TheoPascoli: I mean in the actual code, not in the PR

antarest/maintenance/app.py
```diff
@@ -117,22 +117,26 @@ def _setup_periodic_tasks(sender: Celery, **_: object) -> None:
...
     from antarest.maintenance.tasks.gc_matrix_task import clean_matrices_task
+    from antarest.maintenance.tasks.watcher_scan_task import watcher_scan_task
 
     config: Optional[Config] = getattr(celery_app.conf, "antarest_config", None)
     storage = config.storage if config else StorageConfig()
 
     sender.add_periodic_task(storage.matrix_gc_sleeping_time, clean_matrices_task.s(), name="matrices_cleaner")
     sender.add_periodic_task(storage.blob_gc_sleeping_time, clean_blobs_task.s(), name="blobs_cleaner")
     sender.add_periodic_task(storage.auto_archive_sleeping_time, auto_archive_task.s(), name="auto_archiver")
+    sender.add_periodic_task(storage.watcher_scan_sleeping_time, watcher_scan_task.s(), name="watcher_scan")
```
smailio, line 128: Currently scans are sequential : scan => sleep(2s) => repeat , your implementation seems to trigger a scan every minute whether  the current scan finished or not. 

It shoudl'nt matter to you but just in case note that current implementation support concurrent scan : a scan is triggered by a web or api user.

Besides you're adding a periodic task even when it's missing in the config file.

TheoPascoli: If a previous scan is not finished, the new task is skipped because the lock is not available

antarest/core/config.py
MartinBelthle: I think it's 2 seconds because it scans the whole app which takes several hours I think. This value does not really matter IMO as long as it's not too big

antarest/maintenance/tasks/common.py
```diff
@@ -43,3 +43,15 @@ class LockId(IntEnum):
     MATRIX_GC = 1001
     BLOB_GC = 1002
     AUTO_ARCHIVE = 1003
+    WATCHER_SCAN = 1004
+
+
+class WatcherScanTaskResult(AntaresBaseModel):
```
MartinBelthle, line 49: Minor but seems you can create a base class with error, reason etc. and just fill the different values in the main classes

antarest/maintenance/app.py
```diff
@@ -117,22 +117,26 @@ def _setup_periodic_tasks(sender: Celery, **_: object) -> None:
...
 
     # Stagger first execution to avoid hitting everything at once
     clean_matrices_task.apply_async(countdown=60)
     clean_blobs_task.apply_async(countdown=90)
     auto_archive_task.apply_async(countdown=120)
+    watcher_scan_task.apply_async(countdown=30)  # First scan early
```
MartinBelthle: What does the comment mean, I'm not sure to understand

tests/maintenance/tasks/test_watcher_scan.py
```diff
@@ -0,0 +1,126 @@
...
+                ),
+            }
+        )
+
+        result = _collect_studies(config)
+
+        assert len(result) == 1
+        assert result[0].path == study_path
+
+    def test_skips_default_workspace(self, tmp_path):
```
MartinBelthle, line 71: You could create a study inside the default workspace to ensure it isn't scan. As for now even if the code doesn't work the test will still pass

antarest/maintenance/app.py
TheoPascoli: it starts the task early. It is relevant only for task that are started every hour for example. This run the task immediately at the start of the container instead of waiting one hour. I'll remove this one.

antarest/maintenance/tasks/common.py
TheoPascoli: With only 2 classes sharing this, it feels like unecessery coupling, maybe later

tests/maintenance/tasks/test_watcher_scan.py
TheoPascoli: done

MartinBelthle (approved):

PR #2932, opened 2026-01-15
feat(variable-view-gc): clean old materialized output variable view

smailio:
This PR implements a garbage collector for materialized output variable views. 

**How materialized_output_view are created ?**  
On user demand, by the OutputService. Also when we GET an OutputVariablesViewsModel view its last_read is updated, each time.  
 
  **How variable view gc  works ?**  
  - Deletes rows from OutputVariablesViewsModel where last_read < now - retention_days
  - Once a row is deleted,  OutputVariablesMatrixUsageProvider (who returns all ids in the table) no longer returns its matrix ID 
  - MatrixGarbage collector will see that the matrix is in the store but not returned by any provider so he'll delete it from the matrix store 
  
  **Note**  : if the matrice is for some reason referenced somewhere else, it won't be deleted as its id will be returned by another usage provider. 
  
  **Once a variable view is deleted, how long before its matrix is deleted ?** 
It depends, Matrix gc rely on creation date while variable view gc relies on last read date. With current default conf : 30 days retention for variable gc, one hour for matrix gc, the matrice will be deleted the next matrix gc run. If the matrix gc retention time was one year (this will never happen), then the matrix will be deleted only when it reaches both gc conditions : created one year ago, hasn't been read for a month.   
  
  **Additional changes:** 
  - Added MaintenanceContextNotFoundError exception in common.py to reduce code duplication
  
  
 #### How garbage collection works diagram
 The new component to review is VariableViewGC, bottow left of this diagram : 

image

### DIFF
--- a/antarest/core/config.py
+++ b/antarest/core/config.py
@@ -169,6 +169,9 @@ class StorageConfig:
     blobstore: Path = Path("./blobstore")
     blob_gc_sleeping_time: int = 86400
     blob_gc_dry_run: bool = False
+    variable_view_gc_sleeping_time: int = 3600
+    variable_view_gc_dry_run: bool = False
+    variable_view_gc_retention_days: int = 30
 
     @classmethod
     def from_dict(cls, data: JSON, desktop_mode: bool = False) -> "StorageConfig":
@@ -206,6 +209,13 @@ class StorageConfig:
             blobstore=Path(data["blobstore"]) if "blobstore" in data else defaults.blobstore,
             blob_gc_sleeping_time=data.get("blob_gc_sleeping_time", defaults.blob_gc_sleeping_time),
             blob_gc_dry_run=data.get("blob_gc_dry_run", defaults.blob_gc_dry_run),
+            variable_view_gc_sleeping_time=data.get(
+                "variable_view_gc_sleeping_time", defaults.variable_view_gc_sleeping_time
+            ),
+            variable_view_gc_dry_run=data.get("variable_view_gc_dry_run", defaults.variable_view_gc_dry_run),
+            variable_view_gc_retention_days=data.get(
+                "variable_view_gc_retention_days", defaults.variable_view_gc_retention_days
+            ),
         )
 
     @classmethod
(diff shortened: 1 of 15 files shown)


antarest/core/config.py
```diff
@@ -169,6 +169,9 @@ class StorageConfig:
     blobstore: Path = Path("./blobstore")
     blob_gc_sleeping_time: int = 86400
     blob_gc_dry_run: bool = False
+    variable_view_gc_sleeping_time: int = 3600
```
smailio, line 172: I went with "variable_view" but I can accept an alternative name. 

Option 1 : output_variable_view. It's explicit but long. 

Option 2: ovv. it's an abbrevation, it's maybe too short. But if we all agree on what it means then it wont make the code less readable imo.

antarest/maintenance/app.py
```diff
@@ -126,6 +127,9 @@ def _setup_periodic_tasks(sender: Celery, **_: object) -> None:
     sender.add_periodic_task(storage.blob_gc_sleeping_time, clean_blobs_task.s(), name="blobs_cleaner")
     sender.add_periodic_task(storage.auto_archive_sleeping_time, auto_archive_task.s(), name="auto_archiver")
     sender.add_periodic_task(storage.watcher_scan_sleeping_time, watcher_scan_task.s(), name="watcher_scan")
+    sender.add_periodic_task(
```
smailio, line 130: I setup a periodic task whether the service was defined or not in conf file. I'm not sure about this but that's how it was done for other services.

antarest/maintenance/tasks/gc_blob_task.py
```diff
@@ -25,6 +25,6 @@ def clean_blobs_task(self: Task) -> GarbageCollectorTaskResult:  # type: ignore[
     """Celery wrapper that delegates to clean_blobs()."""
     ctx: MaintenanceContext | None = self.app.conf.get("maintenance_ctx")
     if not ctx:
-        raise RuntimeError("MaintenanceContext not in app.conf - worker not initialized?")
+        raise MaintenanceContextNotFoundError()
```
sylvlecl, line 28: Nice

antarest/maintenance/app.py
sylvlecl: Yes let's keep it this way.
We can consider there are 2 different setups:
- either we don't use celery, we use the "usual" main, and therefore we need to select
  which services are started
- or we use celery, then we consider the app is dedicated to execute all those background services
  (simpler for now)

sylvlecl (approved):
looks good to me!

PR #2934, opened 2026-01-16
fix(matrix): fix unit test and revert lock around polars

sylvlecl:
The reason for the deadlock was the concurrent use of the same dataframe **object**,
which will not happen in real use cases.

I converted global variables to factory methods to ensure that we indeed use different
instances every time.

It simplifies logic where we had read-only instances that we had to copy and make
writeable for later use.

### DIFF
--- a/antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
+++ b/antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
@@ -16,12 +16,12 @@ from pathlib import Path
 from typing import Callable, Optional, TypeAlias
 
 import numpy as np
+import numpy.typing as npt
 import polars as pl
 from typing_extensions import override
 
 from antarest.core.exceptions import ChildNotFoundError
 from antarest.core.serde.matrix_export import write_dataframe_in_tsv_format
-from antarest.core.serde.np_array import NpArray
 from antarest.core.utils.archives import read_original_file_in_archive
 from antarest.core.utils.polars import create_polars_dataframe, read_input_dataframe
 from antarest.core.utils.utils import StopWatch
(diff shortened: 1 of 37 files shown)


antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
```diff
@@ -45,17 +48,12 @@ def __init__(
...
         should_exist: bool = True,
     ):
         super().__init__(matrix_mapper=matrix_mapper, config=config, freq=freq)
         self.nb_columns = nb_columns
-        if default_empty is None:
-            self.default_empty = None
-        else:
-            # Clone the template value and make it writable
-            self.default_empty = np.copy(default_empty)
-            self.default_empty.flags.writeable = True
```
sylvlecl, line 58: Note that the use of factory functions makes the logic a lot more clear here, we don't need anymore that fragil logic.

antarest/study/storage/rawstudy/model/filesystem/root/input/hydro/common/capacity/capacity.py
```diff
@@ -28,24 +28,30 @@ class MatrixInfo(TypedDict, total=False):
     name: str
     freq: MatrixFrequency
     start_version: StudyVersion
-    default_empty: NpArray
+    default_empty: Callable[[], NpArray]
```
MartinBelthle: I find a bit weird that we use `NpArray` which is a TypeAlias with validator and serializer as it seems we could just use `npt.NDArray[np.float64]`

tests/matrixstore/test_repository.py
```diff
@@ -252,13 +252,16 @@ def test_concurrent_save(self, tmp_path: Path) -> None:
         for more chances to generate problems.
         """
         trial_count = 10
-        matrix = pl.DataFrame(data=np.zeros(shape=(8760, 1)))
+        # note: important to have different instances, which will be the cas in real usage
```
MartinBelthle: typo: `cas` -> `case`

MartinBelthle (approved):

antarest/study/storage/rawstudy/model/filesystem/root/input/hydro/common/capacity/capacity.py
sylvlecl: I changed it here and in input_series_matrix.py

PR #2935, opened 2026-01-16
chore(matrix)!: various cleaning and improve some test performance

MartinBelthle:
Breaking change:
- The GET /raw endpoint for output matrices when called with `formatted=True` or `MatrixFormat.JSON` do no longer return the index of the matrix.


tests/storage/integration/test_STA_mini.py
```diff
@@ -580,36 +538,6 @@ def test_sta_mini_import_output(tmp_path: Path, storage_service: StudyService, c
...
-            {
-                "de": [1],
-                "fr": [1],
-                "it": [1],
-                "es": [1],
-            },
-        ),
-    ],
-)
-def test_sta_mini_filter(storage_service: StudyService, url: str, expected_output: Any) -> None:
```
MartinBelthle, line 605: These 2 tests did not really work on `dev` branch. Because the `get_node_and_remainder` method fails for these weird paths. The only reason they passed was because we tested the `get` method and not the real metohd used inside the GET /raw endpoint,  `get_raw_content`

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -97,8 +96,10 @@ def serialize_dataframe(self, dataframe: pl.DataFrame) -> Response:
 
         buffer = io.BytesIO()
         if self == MatrixFormat.JSON:
-            convert_polars_dataframe_to_pandas(dataframe).to_json(buffer, orient="split")
-            return Response(content=buffer.getvalue(), media_type="application/json")
+            # We have to do `to_json(DataFrame.to_dict())` to keep the NaNs for backward compatibility.
```
MartinBelthle, line 99: Actually, this is only used by the front-end (R script use MatrixFormat.PLAIN and Antares-Craft MatrixFormat.ARROW_COMPRESSED). So we could synchronize ourselves with them and make them adapt their code to receive nulls. Currently there is a function somewhere that convert NaNs as empty string or someting like this to render it

tests/storage/integration/conftest.py
```diff
@@ -69,8 +64,11 @@ def sta_mini_seven_zip_path(project_path: Path, sta_mini_zip_path: Path) -> Path
     return target
 
 
-@pytest.fixture
-def services(tmp_path: Path, project_path: Path, sta_mini_zip_path: Path) -> tuple[StudyService, OutputService, Config]:
+@pytest.fixture(scope="session")
```
MartinBelthle: I added the scope session to initialize only this once for tests in `test_STA_mini.py`. It makes them run really faster

sylvlecl: I don't think we should do that: if the tests are not read-only, it breaks tests isolation, which can cause issues in the future. For example outputs posted in one test will be visible from following tests, but not previous ones.

I had tried to unzip the study only once and copy it for each test, but it's actually longer.

antarest/study/web/raw_studies_blueprint.py
sylvlecl: indeed, would be a nice improvement. Let's track this in jira maybe

sylvlecl (changes requested):
Only, the change in tests dot not seem relevant to me.

PR #2938, opened 2026-01-20
fix(matrices): change Simulator default values

MartinBelthle:
[ANT-4404]

This PR does 3 things:
- Change the Simulator default values as we considered wrong ones
- During the normalization process, do not remplace empty matrices by their default ones, instead keep them empty
- Refactor the code concerning default matrices to explicitely put at one place the default ones to show in the front-end and on the other hand the Simulator default ones


antarest/study/storage/variantstudy/business/matrix_constants_generator.py
```diff
@@ -127,7 +131,7 @@ def init_constant_matrices(
 
         # Some short-term storage matrices use np.ones((8760, 1))
         self.hashes[ONES_SCENARIO_MATRIX] = self.matrix_service.add_predefined_matrix(
-            lambda: pl.DataFrame(matrix_constants.st_storage.series.pmax_injection())
+            lambda: pl.DataFrame(ones_scenario_matrix())
```
sylvlecl: we don't need to use a lambda which copies the result, we can just use the function as is

antarest/study/storage/rawstudy/model/filesystem/matrix/simulator_default.py
```diff
@@ -43,3 +44,39 @@ def default_6_fixed_hourly() -> np.ndarray:
...
+
+def default_water_values() -> np.ndarray:
+    return np.zeros((365, 101), dtype=np.float64)
+
+
+def default_energy() -> np.ndarray:
+    return np.zeros((12, 5), dtype=np.float64)
+
+
+def cost_level() -> np.ndarray:
```
sylvlecl: Indeed let's follow the convention of using `default_` prefix

sylvlecl (commented):
minor stuff

sylvlecl (approved):

PR #2943, opened 2026-01-21
perf(aggregation): use polars instead of pandas

MartinBelthle:
Benchmark:
- A task that took ~27s on `dev` now takes ~20.5s on this branch


antarest/study/output/aggregator_management.py
```diff
@@ -185,11 +186,15 @@ def _gather_all_files_to_consider(self) -> Sequence[Path]:
             raise MCRootNotHandled(f"Unknown Monte Carlo root: {self.mc_root}")
         return all_output_files
 
-    def columns_filtering(self, df: pd.DataFrame, is_details: bool) -> pd.DataFrame:
+    def _ensures_typing(self, headers: list[str] | list[list[str]]) -> list[str]:
+        # Method used to fix mypy typing inside `columns_filtering` method
+        return [col[0] for col in headers] if not self.transform_columns_headers else headers  # type: ignore
```
TheoPascoli: Its the opposite: when `transform_columns_headers` is `False`, headers are `list[list[str]`. I think you should remove the `not`

TheoPascoli (changes requested):

tests/integration/raw_studies_blueprint/assets/columns_mismatch.txt
```diff
@@ -5,4 +5,4 @@ FR	area	va	annual
 FR	annual	OV. COST	OP. COST	MRG. PRICE	BALANCE	ROW BAL.	PSP	MISC. NDG	LOAD	H. ROR	WIND	SOLAR	NUCLEAR	LIGNITE	COAL	GAS	OIL	MIX. FUEL	MISC. DTG	H. STOR	H. PUMP	H. LEV	H. INFL	H. OVFL	H. VAL	H. COST	UNSP. ENRG	SPIL. ENRG	LOLD	LOLP	AVL DTG	DTG MRG	MAX MRG	NP COST	NODU
 		Euro	Euro	Euro	MWh	MWh	MWh	MWh	MWh	MWh	MWh	MWh	MWh	MWh	MWh	MWh	MWh	MWh	MWh	MWh	MWh	%	MWh	%	Euro/MWh	Euro	MWh	MWh	Hours	%	MWh	MWh	MWh	Euro	 
 																																			
-	Annual	46452000	46452000	47.14	0	0	0	0	1402800	0	0	0	0	0	0	0	0	0	1402800	0	0	N/A	0	N/A	N/A	0	0	0	0	0	3024000	1621200	1621200	0	783
```
sylvlecl, line 8: Is this change useful ?

antarest/study/output/aggregator_management.py
```diff
@@ -118,13 +119,13 @@ def __init__(
         self._output_first_column = get_start_column(self.frequency)
         self.transform_columns_headers = transform_columns_headers
 
-    def _parse_output_file(self, file_path: Path, normalize_column_names: bool) -> pd.DataFrame:
+    def _parse_output_file(self, file_path: Path, normalize_column_names: bool) -> OutputDataFrame:
         output_data = parse_output_file(file_path, self._output_first_column)
 
         if normalize_column_names:
-            output_data.data.columns = pd.Index(normalize_df_column_names(self.mc_root, output_data.headers))
+            output_data.headers = normalize_df_column_names(self.mc_root, output_data.headers)  # type: ignore
```
sylvlecl: we'd better not have that type ignore:
either make the function accept both types of lists, or ensure we have a 2-levels list up front

antarest/study/output/aggregator_management.py
```diff
@@ -252,35 +264,44 @@ def _build_dataframes(self, files: Sequence[Path]) -> Iterator[pd.DataFrame]:
...
             # columns filtering
-            df = self.columns_filtering(df, is_details)
+            output_data = self.columns_filtering(output_data, is_details)
 
             if not self.transform_columns_headers:
-                concatenate_dataframe_multi_indexed_columns(df)
+                concatenate_dataframe_multi_indexed_columns(output_data)
+
+            # Starting from here, output_data.headers are just a list of strings.
+            # We can use them as colons for our dataframe.
```
sylvlecl: typo columns

antarest/study/output/utils.py
```diff
@@ -57,6 +59,12 @@ class MCAllLinksQueryFile(StrEnum):
 QueryFileType: TypeAlias = MCIndAreasQueryFile | MCAllAreasQueryFile | MCIndLinksQueryFile | MCAllLinksQueryFile
 
 
+@dataclass
+class OutputDataFrame:
```
sylvlecl, line 66: We can add a docstring to explain why we need this:
polars does not handle multi-headers columns

antarest/study/output/aggregator_management.py
```diff
@@ -198,10 +203,14 @@ def columns_filtering(self, df: pd.DataFrame, is_details: bool) -> pd.DataFrame:
                 filtered_columns = [c for c in df_columns if any(regex in c.lower() for regex in lower_case_columns)]
             else:
                 filtered_columns = [c for c in df_columns if c.lower() in lower_case_columns]
-            df = df.loc[:, filtered_columns]
-        return df
 
-    def _process_df(self, file_path: Path, is_details: bool) -> pd.DataFrame:
+            indices = [k for k, c in enumerate(df_columns) if c in filtered_columns]
+            data.data = data.data.select([data.data.columns[i] for i in indices])
+            data.headers = [data.headers[i] for i in indices]  # type: ignore
```
sylvlecl: the type ignore does not seem necessary here ?
(I guess it's because mypy cannot know that data.headers has the same type every time we call it, so it thinks we could mix some `str` with some `list[str]`, which is forbidden by the type).

tests/integration/output_blueprint/test_aggregate_raw_data.py
```diff
@@ -616,51 +597,39 @@ class TestRawDataAggregationMCAll:
...
         """
         Test the aggregation of areas data
         """
         client.headers = {"Authorization": f"Bearer {user_access_token}"}
 
-        output_id = params.pop("output_id")
-        res = client.get(f"/v1/studies/{internal_study_id}/areas/aggregate/mc-all/{output_id}", params=params)
-        assert res.status_code == 200, res.json()
-        download_id = res.json()
+        for params, expected_result_filename in AREAS_REQUESTS__ALL:
```
sylvlecl, line 606: Those changes are just for execution time ?

sylvlecl (commented):
Mostly comments about typing

tests/integration/output_blueprint/test_aggregate_raw_data.py
MartinBelthle: Yes and to fit with the other tests in the file that did the same thing

tests/integration/raw_studies_blueprint/assets/columns_mismatch.txt
MartinBelthle: Yes it's to have the first df with type integers and the 2nd with type float. This way before I introduced the `_adapt_polars_schema` function the test failed

antarest/study/output/aggregator_management.py
MartinBelthle: I just copied the code 3 lines below so I don't think it's wrong

antarest/study/output/aggregator_management.py
MartinBelthle: Yes it is. When I remove ot I have this message `Incompatible types in assignment (expression has type "list[list[str] | str]", variable has type "list[list[str]] | list[str]"`

MartinBelthle: I fixed issues with typing by using `cast`

antarest/core/serde/parquet_writer.py
```diff
@@ -66,22 +74,22 @@ def write_dataframes_in_parquet_format_by_column_sets(
...
-                    table = pa.Table.from_pandas(df)
+                    table = df.to_arrow()
                     current_schema = table.schema
                     current_writer = _parquet_writer(file_path, current_schema)
                 else:
-                    df = df.reindex(new_index, axis="columns")
-                    # We're specifying the schema to use to be able to append NaN values to existing values.
-                    table = pa.Table.from_pandas(df, schema=current_schema)
+                    if df.columns != new_index:
+                        expr = pl.lit(None, dtype=pl.Float64)
```
smailio, line 90: how come columns can be different from new_index and that the same time should_write_new_file is false ?

MartinBelthle: If the new dataframe has less column than `new_index`, we don't write a new file but we still have to fill the missing columns to have the same structure.

sylvlecl: It happens when the currently processed dataframe has fewer columns than the current set of columns.

For example:
 - file 1 had columns A and B --> new_index is now A,B
 - file 2 has columns A B and C --> new_index is now A,B,C
 - file 4 has columns A and B --> it differs from new_index, but we don't want to create a new file, we just adapt it by adding an empty C column to append to the current file. It avoids creating too many files

@MartinBelthle 
Note that we could remove that complex processing now that we have a method to get the list of variables up-front  (future possible work, of course)

antarest/study/output/utils.py
```diff
@@ -118,32 +133,30 @@ def _parse_output_dataframe(file_path: Path) -> pl.DataFrame:
     try:
         return pl.read_csv(file_path, skip_lines=7, separator="\t", has_header=False, null_values="N/A", n_threads=1)
     except ComputeError:
-        # Happens if polars wrongly inferred the schema. If so, we specify that he shouldn't try.
-        # This way the parsing does not fail, but it is significantly slower.
-        # This case does not seem to happen very often.
+        # Happens if polars wrongly inferred the schema.
+        # If so, we specify that it should read the entire file to be sure it doesn't infer a false schema.
+        # It's significantly slower but it does not fail.
+        # As no file is longer than 10.000 rows we use this value.
```
smailio, line 139: Suggestion for an alternative design : persist the df.schema alongside the data file in parquet_writer and use it here to read the df.

antarest/core/serde/parquet_writer.py
MartinBelthle: Indeed. As we discussed, we'll have to enrich the list of variables we persist in DB to do so but it could be done in future works.

antarest/study/output/utils.py
MartinBelthle: Not sure to fully understand. What would be do for parsing only one file ? As it's the case inside `output_series_matrix.py` or the first file here. And we cannot know in advance how many columns we have as it differs from file to file so we cannot use the schema

sylvlecl (approved):

PR #2949, opened 2026-01-22
fix(typing): fix more typing issues in tests

sylvlecl:
Fixes a few typing issues in tests, with help of LLMs


TheoPascoli:
Do you have an idea how many type issues are left?

PR #2953, opened 2026-01-22
fix(alembic): bug during migration

TheoPascoli:


MartinBelthle (approved):
How did it work before (see for instance recent migration `fb5ad6c72ad4_output_variables_views`) where we create enums as you did before we just used MAJs ?

sylvlecl (approved):

TheoPascoli:
> How did it work before (see for instance recent migration `fb5ad6c72ad4_output_variables_views`) where we create enums as you did before we just used MAJs ?

On a create table it works fine, not on a add_column

PR #2956, opened 2026-01-26
feat(district): implement database support

TheoPascoli:


alembic/versions/8f3d2a1b5c7e_add_district_tables.py
```diff
@@ -0,0 +1,65 @@
...
+# revision identifiers, used by Alembic.
+revision = "8f3d2a1b5c7e"
+down_revision = "2572145aff71"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # 1. Create district table
+    op.create_table(
```
MartinBelthle, line 22: As we said in the previous meeting, we should store all the model in one table by flattening the list of areas. And we also added another guideline which is to have the same schema in BD as the one that exists in the DAO.

alembic/versions/8f3d2a1b5c7e_add_district_tables.py
```diff
@@ -0,0 +1,65 @@
...
+        "district",
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("district_id", sa.String(length=255), nullable=False),
+        sa.Column("name", sa.String(length=255), nullable=False),
+        sa.Column("output", sa.Boolean(), nullable=False),
+        sa.Column("comments", sa.String(length=500), nullable=False),
+        sa.Column("apply_filter", sa.String(length=50), nullable=False),
```
MartinBelthle: This should be an Enum

antarest/study/dao/database/database_area_dao.py
```diff
@@ -304,6 +317,16 @@ def save_area_ui(self, area_id: str, layer: str, area_ui_data: AreaUI) -> None:
             )
             session.execute(stmt_insert)
 
+    @override
+    def get_invalid_areas_in_district(self, areas: list[str]) -> list[str]:
```
MartinBelthle: I know that you just moved the method but we could rename it as it has nothing to do with districts, it could just be `get_invalid_area_ids`

antarest/study/dao/file/file_study_district_dao.py
```diff
@@ -83,7 +83,7 @@ def save_district(self, district: District) -> None:
         If the district already exists, it will be overwritten.
         """
         study_data = self.get_file_study()
-        invalid_areas = self.get_invalid_areas_in_district(district.add_areas + district.subtract_areas)
+        invalid_areas = self.get_invalid_areas_in_district(district.add_areas + district.subtract_areas)  # type: ignore[attr-defined]
```
MartinBelthle: Should be self.get_impl() I think

MartinBelthle (changes requested):

antarest/study/dao/file/file_study_district_dao.py
TheoPascoli: it changes nothing but ok

alembic/versions/8f3d2a1b5c7e_add_district_tables.py
TheoPascoli: What do you mean by flattening ? Having only one table for the districts ?

MartinBelthle: yes we should have one 1 table for the districts and put the areas list inside a String inside one column of the table

antarest/study/dao/file/file_study_district_dao.py
MartinBelthle: It allows you to remove the type: ignore

alembic/versions/8f3d2a1b5c7e_add_district_tables.py
TheoPascoli: I like it with FK cascade, when when delete an area, it automatically deletes the associated district. With only one table, I would have to parse all the districts to find the deleted area, it seems a bit heavy. What do you think

MartinBelthle: Mmmmh  great remark. I'm discussing the matter with Sylvain

antarest/study/dao/file/file_study_district_dao.py
TheoPascoli: yes but I had to add a type checking above, its the same honestly its fine

alembic/versions/8f3d2a1b5c7e_add_district_tables.py
MartinBelthle: The 2 approaches have pros and cons.
Having one table will be :
- Easier to implement and faster for `get` and `creation` operation
- Faster for `update` operation
- Harder to implement and slower for `delete` operation

As the deletion is the operation users perform the less, we perfer the solution with 1 table. Feel free to give other arguments if you want :)

alembic/versions/8f3d2a1b5c7e_add_district_tables.py
```diff
@@ -0,0 +1,44 @@
...
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+from sqlalchemy import String
+
+# revision identifiers, used by Alembic.
+revision = "8f3d2a1b5c7e"
+down_revision = "2572145aff71"
```
MartinBelthle: I messed up in my alembic revision, so now the down_revision should be `4bbe4f165a05`

alembic/versions/8f3d2a1b5c7e_add_district_tables.py
```diff
@@ -0,0 +1,44 @@
...
+            name=op.f("fk_district_study_id_study"),
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint("study_id", "district_id", name=op.f("pk_district")),
+    )
+
+
+def downgrade() -> None:
+    op.drop_table("district")
+    sa.Enum(name="districtapplyfilter").drop(op.get_bind(), checkfirst=True)
```
MartinBelthle: Does this work on sqlite ? As in previous alembic migrations we checked if we were inside postgre

antarest/study/dao/database/models/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+#
+# See AUTHORS.txt
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
```
MartinBelthle, line 11: I see why you're introducing this. To me the folder should be named `models` and we should move the `models.py` inside this and renamed `area.py`

antarest/study/dao/database/model/district.py
```diff
@@ -0,0 +1,42 @@
...
+    Column("name", String(255), nullable=False),
+    Column("output", Boolean, nullable=False),
+    Column("comments", String(500), nullable=False),
+    Column("apply_filter", Enum(DistrictApplyFilter), nullable=False),
+    Column("add_areas", String, nullable=False),
+    Column("subtract_areas", String, nullable=False),
+    ForeignKeyConstraint(
+        ["study_id"],
+        ["study.id"],
+        ondelete="CASCADE",
```
MartinBelthle: Minor: remove this `,` to be onelined

antarest/study/dao/database/database_district_dao.py
```diff
@@ -0,0 +1,184 @@
...
+from antarest.study.dao.api.district_dao import DistrictDao
+from antarest.study.dao.database.model.district import DISTRICT_TABLE
+from antarest.study.dao.database.models import AREA_TABLE
+
+
+class DatabaseDistrictDao(DistrictDao):
+    """
+    Database implementation of DistrictDao.
+
+    Note: Write operations do NOT commit transactions. The caller (service layer)
```
MartinBelthle: We should remove this out of date doc

antarest/study/dao/database/database_district_dao.py
```diff
@@ -0,0 +1,184 @@
...
+                output=district.output,
+                comments=district.comments,
+                apply_filter=district.apply_filter,
+                add_areas=json.dumps(district.add_areas),
+                subtract_areas=json.dumps(district.subtract_areas),
+            )
+        )
+        session.commit()
+
+    def _get_invalid_areas(self, areas: list[str]) -> list[str]:
```
MartinBelthle: You should probably do this instead `self.get_impl().get_invalid_area_ids()` and remove this method

antarest/study/dao/database/database_district_dao.py
```diff
@@ -0,0 +1,184 @@
...
+        Returns the list of districts in this study.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        stmt = select(DISTRICT_TABLE).where(DISTRICT_TABLE.c.study_id == study_id)
+        district_rows = session.execute(stmt).fetchall()
+
+        return [
+            District(
```
MartinBelthle: Minor but you could create a function inside this file that does the conversion from BD to inside class, see `_convert_db_properties_to_model` for instance

antarest/study/dao/database/database_district_dao.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        session.execute(
+            delete(DISTRICT_TABLE).where(
+                (DISTRICT_TABLE.c.study_id == study_id) & (DISTRICT_TABLE.c.district_id == district_id)
+            )
+        )
+        session.commit()
```
MartinBelthle, line 129: Shouldn't we raise if there's nothing to delte saying that the district does not exist

antarest/study/dao/database/database_area_dao.py
```diff
@@ -233,9 +248,30 @@ def delete_area(self, area_id: str) -> None:
 
         validate_area_exists(session, study_id, area_id)
 
+        # Remove area from districts that reference it
+        stmt = select(DISTRICT_TABLE).where(DISTRICT_TABLE.c.study_id == study_id)
+        district_rows = session.execute(stmt).fetchall()
+
+        for row in district_rows:
+            add_areas: list[str] = json.loads(row.add_areas)
```
MartinBelthle: Minor: This code could be faster but not really important:
```
add_areas = set(json.loads(row.add_areas))
subtract_areas = set(json.loads(row.subtract_areas))

add_areas.discard(area_id)
subtract_areas.discard(area_id)

```

tests/study/dao/test_database_district_dao.py
```diff
@@ -0,0 +1,225 @@
...
+            db_session.add(study)
+            db_session.commit()
+        return study_id
+
+    @pytest.fixture
+    def dao(self, db_session: Session, study_id: str) -> DatabaseDistrictDao:
+        return DatabaseDistrictDao(study_id, db_session)
+
+    @pytest.fixture
+    def study_dao(self, db_session: Session, study_id: str) -> DatabaseStudyDao:
```
MartinBelthle: These fixtures are inside the conftest, you shouldn't need to redeclare them

tests/study/dao/test_database_district_dao.py
```diff
@@ -0,0 +1,225 @@
...
+        return DatabaseStudyDao(study_id, db_session)
+
+    def test_save_district_creates_district(self, db_session: Session, dao: DatabaseDistrictDao) -> None:
+        with db_session:
```
MartinBelthle: No need for `with db_session` and `db_session.commit()`, it will be lighter to read

MartinBelthle (changes requested):
Minor comments only

tests/study/dao/conftest.py
```diff
@@ -35,3 +36,9 @@ def study_id(db_session: Session) -> str:
 def dao(db_session: Session, study_id: str) -> DatabaseStudyDao:
     """Create a DatabaseStudyDao instance for testing."""
     return DatabaseStudyDao(study_id, db_session)
+
+
+@pytest.fixture
+def district_dao(db_session: Session, study_id: str) -> DatabaseDistrictDao:
```
MartinBelthle: Seems you could just use the `dao` fixture just above

antarest/study/dao/database/database_district_dao.py
```diff
@@ -0,0 +1,172 @@
...
+
+    @override
+    def district_exists(self, district_id: str) -> bool:
+        """
+        Returns whether a district with the given id exists in the study.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        stmt = select(DISTRICT_TABLE.c.district_id).where(
```
MartinBelthle, line 169: Really minor but you could put this in a utils method because it's used twice

MartinBelthle (commented):
2 minor comments and we're good to go

antarest/study/dao/database/database_district_dao.py
TheoPascoli: I like it being this explicit

MartinBelthle (approved):

PR #2958, opened 2026-01-26
fix(constraint): fix bug when deleting additionnal constraints

TheoPascoli:


antarest/study/dao/file/file_study_st_storage_dao.py
```diff
@@ -237,7 +238,8 @@ def delete_st_storage(self, area_id: str, storage: STStorage) -> None:
         # Deleting the short-term storage in the configuration must be done AFTER deleting the files and folders.
         study_data.config.areas[area_id].st_storages.remove(storage)
         if study_data.config.version >= STUDY_VERSION_9_2:
-            study_data.config.areas[area_id].st_storages_additional_constraints.pop(storage.id)
+            if storage_id in study_data.config.areas[area_id].st_storages_additional_constraints:
```
MartinBelthle: We should add a test. Also I think you could just use this instead, `.pop(storage.id, None)`. IMO it's easier to read

MartinBelthle (changes requested):

MartinBelthle (dismissed):

PR #2961, opened 2026-01-28
fix(celery): fix initialising issues

TheoPascoli:
Problem: Periodic maintenance tasks were registered dynamically in the `on_after_configure` hook using `add_periodic_task()`. Celery Beat requires the schedule to be defined statically at startup, so tasks weren't being picked up properly.                                                                                                                                                                                                      
                                                                                                                                                                                                                           Solution: Moved the `beat_schedule` configuration to a static definition in `celery_app.conf.update()`. Created a `TaskName` num to centralize task names and ensure consistency between task definitions and routing. The `on_after_configure` hook now only triggers initial staggered task execution instead of registering the schedule.


sylvlecl (approved):
Looks good to me.

PR #2965, opened 2026-01-28
feat(dao): support links inside DB

MartinBelthle:


antarest/study/dao/database/models/link.py
```diff
@@ -0,0 +1,53 @@
...
+from antarest.dbmodel import Base
+from antarest.study.business.model.link_model import AssetType, LinkStyle, TransmissionCapacity
+
+metadata = Base.metadata
+
+# Relation: One to many with `AREA_TABLE`: 1 area can have multiple links
+LINK_TABLE = Table(
+    "link",
+    metadata,
+    Column("study_id", String(length=36), nullable=False),
```
sylvlecl: The 3 columns should have primary key = True ?

tests/study/dao/test_database_link_dao.py
```diff
@@ -0,0 +1,141 @@
...
+    with pytest.raises(LinkNotFound):
+        dao.get_link("london", "paris")
+
+    _create_default_link(dao)
+
+    # Should not raise this time
+    dao.get_link("london", "paris")
+
+
+def test_get_all_links(dao: DatabaseStudyDao, db_session: Session) -> None:
```
sylvlecl, line 87: Can we already have a test which checks the behaviour when we delete an area ?

antarest/study/dao/database/database_link_dao.py
```diff
@@ -0,0 +1,173 @@
...
+        return self._db_session
+
+    @abstractmethod
+    def get_impl(self) -> "DatabaseStudyDao":
+        pass
+
+    @override
+    def save_link(self, link: Link) -> None:
+        if self.link_exists(link.area1, link.area2):
+            self._delete_link(link)
```
sylvlecl: maybe slightly better to have an update in that case ?

sylvlecl (commented):
A few questions / comments

antarest/study/dao/database/models/link.py
sylvlecl: Note sur if that's actually useful though

antarest/study/dao/database/database_link_dao.py
```diff
@@ -0,0 +1,173 @@
...
+        stmt_insert = insert(LINK_TABLE).values(
+            study_id=self.get_study_id(),
+            area1=link.area1,
+            area2=link.area2,
```
smailio: Minor : you don't check for area validity, which isn't a major issue as the DB will do the check anyway. Just the error message won't be as explicit as it used to be with the file study dao, instead we'll have something about foreign keys etc...

smailio (commented):
PR is neat. Seems to handle all cases handled by the file study dao except for the not implemented stuff. One minor comment

antarest/study/dao/database/models/link.py
MartinBelthle: Not sure either but we do this for other tables so I'll add it

tests/study/dao/test_database_link_dao.py
MartinBelthle: There's the test `test_delete_area` at the end of the file

antarest/study/dao/database/database_link_dao.py
MartinBelthle: Indeed you're right. We do this check inside `file_study_dao` so I should probably add it too. Because the command does not check and the manager neither. I'll add it

antarest/study/dao/database/database_link_dao.py
MartinBelthle: I used the same logic as Theo used inside districts but indeed it could be faster

antarest/study/dao/database/database_link_dao.py
MartinBelthle: I added it

sylvlecl (approved):

PR #2968, opened 2026-01-29
fix(favorite): replace user id in favorite queries

mehdiwahada:
What it does ? : 
- It links users of the identities table to the favorite_study table 
- Replace the get_user_id() method by the get_user_impersonator()

Why ? : 
- It makes the code in favorite easier to understand while resolving the token problem that could occur with the user_id


alembic/versions/y7x41d4a8yc_add_foreign_key_users_favorite_table.py
```diff
@@ -0,0 +1,22 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_foreign_key("fk_user_id_favorite_study", "identities", "favorite_study", ["user_id"], ["id"], ondelete="CASCADE")
+
+
+def downgrade():
+    op.drop_constraint("fk_user_id_favorite_study", "identities", type_="foreignkey")
```
MartinBelthle: FYI : SQLite doesn't support dropping foreign keys, so we need to ignore it here.
You should add this check before your downgrade: `if dialect_name == "postgresql":`

alembic/versions/y7x41d4a8yc_add_foreign_key_users_favorite_table.py
```diff
@@ -0,0 +1,22 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'y7x41d4a8yc'
+down_revision = '7b406752d80'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_foreign_key("fk_user_id_favorite_study", "identities", "favorite_study", ["user_id"], ["id"], ondelete="CASCADE")
```
MartinBelthle: I don't think this is the right syntax, for instance in other alembic migrations we wrote something like `batch_op.create_foreign_key(COMMAND_BLOCK_FK, "variantstudy", ["study_id"], ["id"])`

MartinBelthle (changes requested):
We should add a test concerning the user impersonator, to ensure it works well

alembic/versions/y7x41d4a8yc_add_foreign_key_users_favorite_table.py
sylvlecl: What we should do is change the type `user_id` in the table `favorite_study` to integer:
for this, the simplest solution (maybe the only one) is to drop and recreate the table.

There is no problem to do it, the previous table creation has not been released.

antarest/favorite/model.py
```diff
@@ -37,8 +37,9 @@ class FavoriteStudy(Base):
 
     __tablename__ = "favorite_study"
 
-    user_id: Mapped[str] = mapped_column(
-        String(255),
+    user_id: Mapped[int] = mapped_column(
```
sylvlecl, line 40: see comment on alembic migration, it's not enought to change it here, it must be consistent with the actual table schema that is created by alembic (so we need to change the alembic migration so that user_id has integer type in database).

alembic/versions/y7x41d4a8yc_add_foreign_key_users_favorite_table.py
```diff
@@ -0,0 +1,32 @@
...
+
+def upgrade():
+
+    op.drop_table('favorite_study')
+    op.create_table(
+        "favorite_study",
+        sa.Column("user_id", sa.Integer(), nullable=False),
+        sa.Column("study_id", sa.String(36), sa.ForeignKey("study.id", ondelete="CASCADE"), nullable=False),
+        sa.PrimaryKeyConstraint("user_id", "study_id", name="pk_favorite_study"),
+        sa.ForeignKeyConstraint(["user_id"], ["identities.id"], name="fk_user_id_favorite_study", ondelete="CASCADE"),
```
sylvlecl, line 27: Very very minor: it is suprising for the reader to define the 2 foreign keys in 2 different ways, once in the column definition, once outside.

tests/favorite_study/test_favorite_study.py
```diff
@@ -30,22 +30,24 @@ def test_add_favorite(favorite_service: FavoriteService, admin_user: JWTUser) ->
         actual_favorite_list = favorite_service.list_favorites()
 
     expected_favorite_list = [favorite_dto_1, favorite_dto_2]
+    print("Liste : ", actual_favorite_list)
```
sylvlecl: We need to delete those prints

sylvlecl (commented):
Very minor comments

sylvlecl (dismissed):

sylvlecl (approved):

PR #2970, opened 2026-01-29
feat(layer): implement database support

TheoPascoli:


antarest/study/dao/database/models/layer.py
```diff
@@ -0,0 +1,30 @@
...
+
+from antarest.dbmodel import Base
+
+metadata = Base.metadata
+
+LAYER_TABLE = Table(
+    "layer",
+    metadata,
+    Column("study_id", String(36), nullable=False, primary_key=True),
+    Column("layer_id", String(10), nullable=False, primary_key=True),
```
MartinBelthle, line 27: To me we should create a foreign key between this and `layer_id` inside AreaUi table. This would simplify the deletion and ensure consistance.

antarest/study/dao/database/database_layer_dao.py
```diff
@@ -0,0 +1,193 @@
...
+        The default layer (id="0") always contains all areas.
+        Other layers contain the areas that have UI entries for that layer.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        # Get all layer names from LAYER_TABLE
+        stmt_layers = select(LAYER_TABLE).where(LAYER_TABLE.c.study_id == study_id)
+        layer_rows = session.execute(stmt_layers).fetchall()
+        layer_names = {row.layer_id: row.name for row in layer_rows}
```
MartinBelthle: Could be renamed `layers_id_to_name_mapping` to be clearer

antarest/study/dao/database/database_layer_dao.py
```diff
@@ -0,0 +1,193 @@
...
+        for row in area_ui_rows:
+            all_areas.add(row.area_id)
+            areas_by_layer[row.layer_id].append(row.area_id)
+
+        # Build result - ensure default layer is first and contains all areas
+        layers: list[Layer] = []
+
+        # Default layer (always exists, always contains all areas)
+        default_name = layer_names.get(DEFAULT_LAYER_ID, "All")
+        layers.append(Layer(id=DEFAULT_LAYER_ID, name=default_name, areas=sorted(all_areas)))
```
MartinBelthle: Minor but you could declare: `layers = ...` here instead of line 96

antarest/study/dao/database/database_layer_dao.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # Other layers
+        for layer_id in sorted(layer_names.keys()):
```
MartinBelthle: Minor: no need for `.keys()`

antarest/study/dao/database/database_layer_dao.py
```diff
@@ -0,0 +1,193 @@
...
+            if layer_id == DEFAULT_LAYER_ID:
+                continue
+            name = layer_names.get(layer_id)
```
MartinBelthle: Seems you could just use `[]` instead of `.get()` I find it easier to read.
Also next line, could `layer_id` not be inside `areas_by_layer` ? I think you could alos just use `[]`

antarest/study/dao/database/database_layer_dao.py
```diff
@@ -0,0 +1,193 @@
...
+                    layer_id=layer.id,
+                    name=layer.name or "",
+                )
+            )
+
+        # Update area associations if areas are provided
+        if layer.areas is not None:
+            self.get_impl().save_layer_areas(layer.id, layer.areas)
+
+        session.commit()
```
MartinBelthle: Minor but you need this commit only if you did not go through `save_layer_areas` otherwise you'll have 2 `.commit()` Perhaps it does not affect performance I don't know

tests/study/dao/test_database_layer_dao.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def test_delete_layer_removes_area_associations(self, dao: DatabaseStudyDao) -> None:
+        dao.save_area("Paris")
+        dao.save_area("London")
+
+        dao.save_layer(Layer(id="1", name="Layer 1", areas=["paris", "london"]))
+        dao.delete_layer(Layer(id="1"))
+
+        layers = dao.get_layers()
+        assert len(layers) == 1
```
MartinBelthle, line 129: Seems you're not checking that the Area_ui table was cleaned

tests/study/dao/test_database_layer_dao.py
```diff
@@ -0,0 +1,163 @@
...
+        dao.save_layer(Layer(id="1", name="Layer 1", areas=["paris", "london"]))
+
+        dao.delete_area("paris")
+
+        layers = dao.get_layers()
+        layer_1 = next(layer for layer in layers if layer.id == "1")
+        assert layer_1.areas == ["london"]
+
+    def test_get_layers_order(self, dao: DatabaseStudyDao) -> None:
+        dao.save_layer(Layer(id="2", name="Layer 2"))
```
MartinBelthle: Does the order really matters ? If not we could remove the sorting inside the src code

MartinBelthle (changes requested):
Not a big fan of the Layer code but it's too late as it's already in the DAO

MartinBelthle:
I think we should also add a test on area `save_ui` with a wrong layer to ensure we raise

antarest/study/dao/database/models/layer.py
TheoPascoli: Good idea. But currently, layer "0" is treated as always existing implicitly (layer_exists() returns True without checking DB). Adding this FK means we need to always insert layer "0" explicitly when creating a study. I can do it, but it changes the laayer creation logic

antarest/study/dao/database/database_layer_dao.py
TheoPascoli: I renamed it `layer_id_to_name`

antarest/study/dao/database/database_layer_dao.py
TheoPascoli: for the next line, I need `.get()` with default because a layer can exist in `LAYER_TABLE` without having any entries in `AREA_UI_TABLE` (empty layer).

tests/study/dao/test_database_layer_dao.py
TheoPascoli: yea it was for fun ...

antarest/study/service.py
```diff
@@ -958,6 +959,10 @@ def create_study(
 
         self._save_study(raw)
 
+        if storage_mode == StorageMode.DATABASE:
```
MartinBelthle: I discussed this with sylvain and he says that it would be better to do one of the 2 folowing things:
1- Create a factory system to do this, with 2 parts one BD the other FS
2- Create a method `initialize_study` inside the DAO where the DB implem would do the `save_layer`.

Next, I'll have to use your work for the study settings as we also want to create settings at study creation

alembic/versions/9c4e7f2a3b1d_add_layer_table.py
```diff
@@ -0,0 +1,48 @@
...
+            op.f("fk_area_ui_study_id_layer_id_layer"),
+            "layer",
+            ["study_id", "layer_id"],
+            ["study_id", "layer_id"],
+            ondelete="CASCADE",
+        )
+
+
+def downgrade() -> None:
+    with op.batch_alter_table("area_ui") as batch_op:
```
MartinBelthle: Seems that in other migrations we did this before dropping the constraint:
`# SQLite doesn't support dropping foreign keys, so we need to ignore it here
    if dialect_name == "postgresql":`

antarest/core/exceptions.py
```diff
@@ -541,11 +541,15 @@ def __str__(self) -> str:
 
 
 class LayerNotFound(HTTPException):
-    def __init__(self) -> None:
-        super().__init__(
-            HTTPStatus.NOT_FOUND,
-            "Layer not found",
-        )
+    def __init__(self, *layer_ids: str) -> None:
```
MartinBelthle: Minor but seems you always instantiate it with only one layer, we could say it's the only thing this exception accepts

antarest/study/dao/database/database_layer_dao.py
```diff
@@ -0,0 +1,189 @@
...
+        layer_id_to_name = {row.layer_id: row.name for row in layer_rows}
+
+        # Get all area-layer associations from AREA_UI_TABLE
+        stmt_area_ui = select(AREA_UI_TABLE.c.layer_id, AREA_UI_TABLE.c.area_id).where(
+            AREA_UI_TABLE.c.study_id == study_id
+        )
+        area_ui_rows = session.execute(stmt_area_ui).fetchall()
+
+        # Build areas per layer
+        areas_by_layer: defaultdict[str, list[str]] = defaultdict(list)
```
MartinBelthle, line 89: Seems to me that with what you done with the default layer at study creation, the only thing you need to do is: 
`areas_by_layer[DEFAULT_LAYER_ID] = list(all_areas)` and then the rest of the method does not need to do anything differently for this layer or other ones

MartinBelthle (changes requested):

MartinBelthle (approved):

PR #2976, opened 2026-01-30
fix(celery): fix setup

TheoPascoli:
The worker was not getting the redis credentials.
Added a cronjob for auto-archiver. It is launched only at night (with new retry mecanism if it fails).
Added a retry mecanism for the blob (because launched only once per day).


antarest/core/config.py
```diff
@@ -180,6 +180,7 @@ class StorageConfig:
     auto_archive_threshold_days: int = 60
     auto_archive_dry_run: bool = False
     auto_archive_sleeping_time: int = 3600
+    auto_archive_cron: str = "0 20-23,0-7 * * * "
```
sylvlecl, line 183: Nice addition :+1: 

Should it not be exclusive with the sleeping time ?
(like raising an error if both are defined ? and if none is defined, keeping only the sleeping time which was the default behaviour before)

antarest/maintenance/app.py
```diff
@@ -33,14 +36,23 @@
 
 logger = logging.getLogger(__name__)
 
+MAINTENANCE_QUEUE = "maintenance"
```
sylvlecl, line 40: Good to make it a constant here ! :+1:

antarest/maintenance/app.py
```diff
@@ -33,14 +36,23 @@
...
+
+TASK_NAMES = [
```
sylvlecl: Should we not have them as constants to (or enum) so that we don't have to repeat them in the tasks definition?

antarest/maintenance/tasks/auto_archive_task.py
```diff
@@ -37,3 +73,26 @@ def auto_archive_task(self: Task) -> AutoArchiveTaskResult:  # type: ignore[type
...
+    """
+    Setup auto-archive task with either cron schedule or sleeping time.
+
+    Uses cron if configured, otherwise falls back to sleeping_time.
+
+    Args:
+        sender: Celery app instance
+        storage: StorageConfig instance
+    """
+    if storage.auto_archive_cron:
```
sylvlecl, line 88: We'd better raise if both are defined (or handle it in config parsing)

antarest/maintenance/tasks/watcher_scan_task.py
```diff
@@ -20,9 +20,17 @@
 from antarest.maintenance.tasks.watcher_scan import scan_workspaces
 
 
-@celery_app.task(bind=True, name="watcher_scan", pydantic=True)
+@celery_app.task(
+    bind=True,
+    name="watcher_scan",
+    pydantic=True,
+    soft_time_limit=1500,
+    time_limit=1800,
```
sylvlecl: Not sure it's a good idea to have that limit:
on large disks the scan can be really long. If we kill it, we may never actually import new studies.

TheoPascoli: alright

antarest/maintenance/app.py
TheoPascoli: Yes, I changed it

antarest/core/config.py
TheoPascoli: I added a check at the config level. If both are specified in the config, it raises a `ValueError`

sylvlecl (approved):

PR #2977, opened 2026-02-02
feat(dao): support study settings inside DB

MartinBelthle:


antarest/study/dao/database/database_study_factory_dao.py
```diff
@@ -42,4 +49,16 @@ def session(self) -> Session:
...
+        dao.save_timeseries_config(TimeSeriesConfiguration())
+        dao.save_advanced_parameters(AdvancedParameters())
+        dao.save_optimization_preferences(OptimizationPreferences())
+        # todo: uncomment this when method is implemented
+        # dao.save_thematic_trimming(ThematicTrimming())
+        study_version = StudyVersion.parse(study.version)
+        if study_version > STUDY_VERSION_8_3:
+            dao.save_adequacy_patch_parameters(AdequacyPatchParameters())
+        if study_version > STUDY_VERSION_9_2:
+            dao.save_compatibility_parameters(CompatibilityParameters())
```
TheoPascoli: Can you extract this logic inside a method ? To keep create_study_dao "clean"

antarest/study/dao/database/database_study_settings_dao.py
```diff
@@ -0,0 +1,337 @@
...
+            building_mode=config.building_mode,
+            selection_mode=config.selection_mode,
+            year_by_year=config.year_by_year,
+            simulation_synthesis=config.simulation_synthesis,
+            mc_scenario=config.mc_scenario,
+            filtering=config.filtering,
+            geographic_trimming=config.geographic_trimming,
+            thematic_trimming=config.thematic_trimming,
+        )
+        session = self.get_session()
```
TheoPascoli: You can replace this by 
```suggestion
values = dict(
    study_id=self.get_study_id(),
    **config.model_dump(),
)
```

And in all the other methods I guess

tests/study/dao/test_database_settings_dao.py
```diff
@@ -0,0 +1,84 @@
...
+from antarest.study.business.model.config.timeseries_config_model import TimeSeriesConfiguration, TimeSeriesType
+from antarest.study.dao.database.database_study_dao import DatabaseStudyDao
+from antarest.study.dao.database.database_study_factory_dao import DatabaseStudyDaoFactory
+from antarest.study.model import StorageMode
+from tests.helpers import create_study
+
+
+def test_nominal_case(dao: DatabaseStudyDao) -> None:
+    # General Config
+    new_general_config = GeneralConfig(first_month=Month.OCTOBER, geographic_trimming=True, mode=Mode.ADEQUACY)
```
sylvlecl: Not fun to do but we should cover all the values of the settings, here we cover just a few.
AI may help ! :)

antarest/study/dao/database/database_study_settings_dao.py
```diff
@@ -0,0 +1,337 @@
...
+            accurate_shave_peaks_include_short_term_storage=row.accurate_shave_peaks_include_short_term_storage,
+        )
+
+    @override
+    def get_compatibility_parameters(self) -> CompatibilityParameters:
+        study_id = self._study_id
+        stmt = select(COMPATIBILITY_PARAMETERS_TABLE).where((COMPATIBILITY_PARAMETERS_TABLE.c.study_id == study_id))
+        row = self.get_session().execute(stmt).fetchone()
+        if not row:
+            raise StudyNotFoundError(study_id)
```
sylvlecl, line 157: Not related specifically to that PR, but for that kind of case we should probably raise another error, something like "inconsistent study state", so that it's more clear in the logs when that happens.

We should do that also in other places, like when some expected matrices are absent (in areas, thermal clusters etc).

sylvlecl (changes requested):
Looks very good, I think we should just harden a little the tests to cover all values.

antarest/study/dao/database/database_study_settings_dao.py
```diff
@@ -0,0 +1,337 @@
...
+            building_mode=row.building_mode,
+            selection_mode=row.selection_mode,
+            year_by_year=row.year_by_year,
+            simulation_synthesis=row.simulation_synthesis,
+            mc_scenario=row.mc_scenario,
+            filtering=row.filtering,
+            geographic_trimming=row.geographic_trimming,
+            thematic_trimming=row.thematic_trimming,
+        )
+
```
TheoPascoli: You can use row._mapping that returns a dict. But you must exlude the study_id from it before doing GeneralConfig(**row._mapping). Same for the other methods
I Suggest : 
```suggestion
        data = {k: v for k, v in row._mapping.items() if k != "study_id"}
        return GeneralConfig(**data)
```

antarest/study/dao/database/database_study_settings_dao.py
```diff
@@ -0,0 +1,337 @@
...
+        if isinstance(config.export_mps, bool):
+            mps = str(config.export_mps)
+        else:
+            mps = config.export_mps
+        values["export_mps"] = mps
+
+        session = self.get_session()
+        upsert_one(session, OPTIMIZATION_PREFERENCES_TABLE, values)
+        session.commit()
+
```
TheoPascoli, line 109: Those 3 lines are duplicated multiple times. Worth to extract it to a method ?

TheoPascoli (changes requested):
Hello Martin, good work its clean :) just minor comments

antarest/study/dao/database/database_study_settings_dao.py
MartinBelthle: I don't really like that `_mapping` is private but it' so much concise that I'll use it

TheoPascoli: sqlalchemy does this in case you have a column named "mapping" to avoid conflicts but it is a public property

antarest/study/dao/database/database_study_settings_dao.py
MartinBelthle: We could but I actually prefer the current code as now it's only 4 lines per method and it's clear as is. We could change this later, I don't really mind

antarest/study/dao/database/database_study_settings_dao.py
MartinBelthle: I agree but I'll do this in a next dedicated PR to separate concerns

sylvlecl (dismissed):

sylvlecl:
We need to change the down revision in migration file

PR #2978, opened 2026-02-02
feat(dao): thermal clusters DAO

sylvlecl:
Implements thermal clusters part of the database DAO, following decided conventions.


alembic/versions/0e628a645214_add_thermal_cluster_tables.py
```diff
@@ -0,0 +1,165 @@
...
+        sa.Column("volatility_forced", sa.Float(), nullable=False),
+        sa.Column("volatility_planned", sa.Float(), nullable=False),
+        sa.Column("law_forced", sa.Enum("uniform", "geometric", name="lawoption"), nullable=False),
+        sa.Column("law_planned", sa.Enum("uniform", "geometric", name="lawoption"), nullable=False),
+        sa.Column("marginal_cost", sa.Float(), nullable=False),
+        sa.Column("spread_cost", sa.Float(), nullable=False),
+        sa.Column("fixed_cost", sa.Float(), nullable=False),
+        sa.Column("startup_cost", sa.Float(), nullable=False),
+        sa.Column("market_bid_cost", sa.Float(), nullable=False),
+        sa.Column("co2", sa.Float(), nullable=False),
```
sylvlecl, line 57: we'd probably better have all those emissions in a serialized field

sylvlecl: Discussion convinced me it was a better choice to keep it as is:
cost of null columns is low in postgres, JSON has its own overhead of course (serde), and having explicit columns is closer to what we have done for other fields.

Another alternative would be to have a separate table for those emission properties, but it does not really make sense from the application point of view, since we will always request all the props together.

alembic/versions/0e628a645214_add_thermal_cluster_tables.py
```diff
@@ -0,0 +1,165 @@
...
+            nullable=False,
+        ),
+        sa.Column("min_stable_power", sa.Float(), nullable=False),
+        sa.Column("min_up_time", sa.Integer(), nullable=False),
+        sa.Column("min_down_time", sa.Integer(), nullable=False),
+        sa.Column("must_run", sa.Boolean(), nullable=False),
+        sa.Column("spinning", sa.Float(), nullable=False),
+        sa.Column("volatility_forced", sa.Float(), nullable=False),
+        sa.Column("volatility_planned", sa.Float(), nullable=False),
+        sa.Column("law_forced", sa.Enum("uniform", "geometric", name="lawoption"), nullable=False),
```
sylvlecl: Shouldn't we define that enum only once ?

tests/study/dao/test_database_thermal_dao.py
```diff
@@ -0,0 +1,135 @@
...
+    THERMAL_CLUSTER_TABLE,
+    THERMAL_CO2_COST_TABLE,
+    THERMAL_FUEL_COST_TABLE,
+    THERMAL_MODULATION_TABLE,
+    THERMAL_PREPRO_TABLE,
+    THERMAL_SERIES_TABLE,
+)
+
+
+def test_save_thermal_creates_cluster(dao: DatabaseStudyDao) -> None:
```
sylvlecl, line 41: What's the expected behaviour for matrices at creation ?

tests/study/dao/test_database_area_dao.py
```diff
@@ -105,8 +105,8 @@ def test_get_all_areas_info_returns_areas(dao: DatabaseStudyDao) -> None:
     # Check original names are preserved
     area_names = {a.name for a in areas}
     assert area_names == {"Paris", "London", "Berlin"}
-    # All areas should have thermals=None for now
-    assert all(a.thermals is None for a in areas)
+    # All areas should have empty thermals by default
+    assert all(a.thermals == [] for a in areas)
```
sylvlecl: We should have a test where we have created some thermal clusters

tests/study/dao/test_database_thermal_dao.py
sylvlecl: File study DAO does not create any. The creation command ensures this, OK.

tests/study/dao/test_database_area_dao.py
sylvlecl: Done

alembic/versions/0e628a645214_add_thermal_cluster_tables.py
MartinBelthle: yes it would be a bit better but minor

alembic/versions/0e628a645214_add_thermal_cluster_tables.py
```diff
@@ -0,0 +1,165 @@
...
+            ["area.study_id", "area.area_id"],
+            name=op.f("fk_thermal_cluster_area"),
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint("study_id", "area_id", "thermal_id", name=op.f("pk_thermal_cluster")),
+    )
+
+    op.create_table(
+        "thermal_prepro",
+        sa.Column("study_id", sa.String(length=36), nullable=False),
```
MartinBelthle: Minor but for readability we could do this:
```python
study_id_col = sa.Column("study_id", sa.String(length=36), nullable=False)
area_id_col = sa.Column("area_id", sa.String(length=255), nullable=False)
matrix_id_col = sa.Column("matrix_id", sa.String(length=64), nullable=False)
```
and reuse them inside tables as i did inside area_matrices migration

alembic/versions/0e628a645214_add_thermal_cluster_tables.py
```diff
@@ -0,0 +1,165 @@
...
+        ),
+        sa.PrimaryKeyConstraint("study_id", "area_id", "thermal_id", name=op.f("pk_thermal_series")),
+    )
+
+    op.create_table(
+        "thermal_fuel_cost",
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("area_id", sa.String(length=255), nullable=False),
+        sa.Column("thermal_id", sa.String(length=255), nullable=False),
+        sa.Column("matrix_id", sa.String(length=64), nullable=False),
```
MartinBelthle: How do we handle the fact that these matrices only exist since v8.7. Prior to v8.7 we don't put a line in this table and we ensure we don't call the DAO methods ?

tests/study/dao/test_database_thermal_dao.py
```diff
@@ -0,0 +1,135 @@
...
+    dao.save_thermal("paris", ThermalCluster(id="gas", name="Gas"))
+    dao.save_thermal("london", ThermalCluster(id="coal", name="Coal"))
+
+    all_thermals = dao.get_all_thermals()
+    assert set(all_thermals.keys()) == {"paris", "london"}
+    assert set(all_thermals["paris"].keys()) == {"gas"}
+    assert set(all_thermals["london"].keys()) == {"coal"}
+
+
+def test_delete_thermal(dao: DatabaseStudyDao) -> None:
```
MartinBelthle, line 190: We should ensure the deletion failed for a wrong thermal id or a wrong area id

tests/study/dao/test_database_thermal_dao.py
```diff
@@ -0,0 +1,135 @@
...
+
+    assert dao.thermal_exists("paris", "gas")
+    dao.delete_thermal("paris", thermal)
+    assert not dao.thermal_exists("paris", "gas")
+
+    with pytest.raises(ThermalClusterNotFound):
+        dao.get_thermal("paris", "gas")
+
+
+def test_thermal_matrices_lifecycle(db_session: Session, dao: DatabaseStudyDao) -> None:
```
MartinBelthle, line 217: We should add test for garbage collection on the thermal matrices

tests/study/dao/test_database_thermal_dao.py
```diff
@@ -0,0 +1,135 @@
...
+    dao.save_area("Paris")
+
+    dao.save_thermal("paris", ThermalCluster(id="gas", name="Gas", nominal_capacity=100.0))
+    dao.save_thermal("paris", ThermalCluster(id="gas", name="Gas", nominal_capacity=200.0))
+
+    result = dao.get_thermal("paris", "gas")
+    assert result.nominal_capacity == 200.0
+
+
+def test_get_all_thermals(dao: DatabaseStudyDao) -> None:
```
MartinBelthle, line 174: We should ensure that saving with a wrong area id raises a clear exception

antarest/study/dao/database/database_thermal_dao.py
```diff
@@ -0,0 +1,362 @@
...
+if TYPE_CHECKING:
+    from antarest.study.dao.database.database_study_dao import DatabaseStudyDao
+
+
+def _normalize_thermal_id(thermal_id: str) -> str:
+    return thermal_id.lower()
+
+
+def _normalize_group(group: Any) -> Any:
+    return group.value if hasattr(group, "value") else group
```
MartinBelthle: `group` is a string inside the thermal model so I don't think we need this function

antarest/study/dao/database/database_thermal_dao.py
```diff
@@ -0,0 +1,362 @@
...
+    @override
+    def save_thermal(self, area_id: str, thermal: ThermalCluster) -> None:
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        validate_area_exists(session, study_id, area_id)
+
+        cluster = thermal.model_copy()
+        version = self.get_impl().get_version()
+        validate_thermal_cluster_against_version(version, cluster)
```
MartinBelthle: We shouldn't need to do this here

antarest/study/dao/database/database_thermal_dao.py
```diff
@@ -0,0 +1,362 @@
...
+                )
+                .values(**values)
+            )
+            session.execute(stmt_update)
+        session.commit()
+
+    @override
+    def save_thermals(self, area_id: str, thermals: Sequence[ThermalCluster]) -> None:
+        for thermal in thermals:
+            self.save_thermal(area_id, thermal)
```
MartinBelthle: This is not optimized at all, this will perform N+1 queries. Also we didn't test it

antarest/study/dao/database/database_thermal_dao.py
```diff
@@ -0,0 +1,362 @@
...
+        )
+        row = session.execute(stmt).fetchone()
+        if not row:
+            raise ThermalClusterNotFound(area_id, thermal_id)
+
+        return self._convert_db_row_to_thermal(row)
+
+    @override
+    def thermal_exists(self, area_id: str, thermal_id: str) -> bool:
+        study_id = self.get_study_id()
```
MartinBelthle: We could put code in common with `get_thermal` as it is really similar

MartinBelthle (changes requested):

alembic/versions/0e628a645214_add_thermal_cluster_tables.py
MartinBelthle: Seems it's what you do

antarest/study/dao/database/database_thermal_dao.py
sylvlecl: you're right

antarest/study/dao/database/database_thermal_dao.py
```diff
@@ -0,0 +1,362 @@
...
+    THERMAL_MODULATION_TABLE,
+    THERMAL_PREPRO_TABLE,
+    THERMAL_SERIES_TABLE,
+)
+
+if TYPE_CHECKING:
+    from antarest.study.dao.database.database_study_dao import DatabaseStudyDao
+
+
+def _normalize_thermal_id(thermal_id: str) -> str:
```
sylvlecl: This is likely an issue: the save to database will transform the ID of the cluster.

The problem is that so far we have not forced thermal cluster IDs to be lower case, but when we search for items in the file implementation, we search with case-insensitivity.

Maybe we'll need to add a column to store the "legacy" ID, which can be upper case, and the lower case ID which will be used for searching.

antarest/study/dao/database/database_thermal_dao.py
sylvlecl: Note that we'll probably have to divide between update and insert lists, so possibly 2 queries

sylvlecl: Its now replaced by an "upsert" which updates or insert everything in one request, benefiting from specific dialect when possible (postgresql, sqlite)

antarest/study/dao/database/database_thermal_dao.py
sylvlecl: Done

tests/study/dao/test_database_thermal_dao.py
sylvlecl: Done, I also fixed the test, its logic was wrong because all objects referred to the same matrix, so it was enough that only one was working to prevent garbage collection and make the test pass

tests/study/dao/test_database_thermal_dao.py
sylvlecl: Done.
I added missing tests about the single `get_thermal` method too

TheoPascoli (dismissed):

alembic/versions/0e628a645214_add_thermal_cluster_tables.py
```diff
@@ -0,0 +1,172 @@
...
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0e628a645214"
+down_revision = "124274d80f2e"
+branch_labels = None
+depends_on = None
```
TheoPascoli, line 17: the revision in the comment is not the same as in `down_revision`

antarest/study/dao/database/sql_utils.py
```diff
@@ -0,0 +1,121 @@
...
+def _update_columns(table: Table) -> list[Column[Any]]:
+    return [column for column in table.columns.values() if not column.primary_key]
+
+
+def upsert_one(
+    session: Session,
+    table: Table,
+    values: dict[str, Any],
+) -> None:
+    return upsert_multiple(session, table, [values])
```
TheoPascoli: You still have the None and return

antarest/study/dao/database/database_area_dao.py
```diff
@@ -102,7 +102,8 @@ def get_all_areas_info(self) -> List[AreaInfo]:
 
         areas_info = []
         for row in result:
-            areas_info.append(AreaInfo(id=row.area_id, name=row.area_name, thermals=None))
+            thermals = list(self.get_impl().get_all_thermals_for_area(row.area_id))
```
TheoPascoli: Maybe this introduce a n+1 query ? You could do only one `get_all_thermals` and do a lookup in the dict I think

alembic/versions/0e628a645214_add_thermal_cluster_tables.py
sylvlecl: Indeed ! we probably have that in other places, I hope the comments are not actually used by the tool

antarest/study/dao/database/database_area_dao.py
sylvlecl: Indeed, well spotted

sylvlecl: Fixed

sylvlecl:
Thanks @TheoPascoli for the license check in pre-commit config

PR #2980, opened 2026-02-03
feat(launcher): add export mps option

TheoPascoli:
See antares-cd repo for the other PR


MartinBelthle (approved):
We should just say in the PR name that we're exporting the `named` mps

PR #2981, opened 2026-02-03
refactor(dao): introduce a factory to initialize studies

MartinBelthle:


antarest/study/dao/factory/factory.py
```diff
@@ -0,0 +1,79 @@
...
+from antarest.study.dao.database.database_study_dao import DatabaseStudyDao
+from antarest.study.dao.file.file_study_dao import FileStudyTreeDao
+from antarest.study.model import RawStudy, StorageMode, StudyContentStatus
+from antarest.study.repository import StudyMetadataRepository
+from antarest.study.storage.rawstudy.model.filesystem.factory import StudyFactory
+from antarest.study.storage.utils import create_new_empty_study, is_managed, update_antares_info
+from antarest.study.storage.variantstudy.model.command_context import CommandContext
+
+
+class DaoFactory:
```
sylvlecl: We'd better keep the Study prefix too, Dao is too generic.

This should instead be an abstract class in api packahe, with 2 implementations

antarest/study/service.py
```diff
@@ -574,6 +575,9 @@ def __init__(
         self.config = config
         self.on_deletion_callbacks: List[Callable[[str], None]] = []
         StudyDatabaseMatrixUsageProvider(command_context.matrix_service)
+        self._dao_factory = DaoFactory(
```
sylvlecl: here we'd better have a dict storage mode -> factory

antarest/study/dao/factory/factory.py
```diff
@@ -0,0 +1,79 @@
...
+        self._session = session
+
+    @property
+    def session(self) -> Session:
+        """Get the SqlAlchemy session or create a new one on the fly if not available in the current thread."""
+        if self._session is None:
+            return db.session
+        return self._session
+
+    def create_study_dao(self, study: RawStudy) -> tuple[StudyDao, RawStudy]:
```
sylvlecl: should be the abstract method of this abstract class.
The "if/else" part will happen in the study service where we will have a mapping StorageMode -> StudyDaoFactory.

sylvlecl (changes requested):
almost there but the idea was more to have an ABC with 2 implementations, to separate well the file study creation and the database creation, and leave them in the implementations packages

antarest/study/dao/file/file_study_factory_dao.py
```diff
@@ -0,0 +1,54 @@
...
+        self, command_context: CommandContext, study_repository: StudyMetadataRepository, study_factory: StudyFactory
+    ) -> None:
+        self._command_context = command_context
+        self._study_repository = study_repository
+        self._study_factory = study_factory
+
+    @override
+    def create_study_dao(self, study: RawStudy) -> tuple[FileStudyTreeDao, RawStudy]:
+        study.content_status = StudyContentStatus.VALID
+        self._study_repository.save(study)
```
sylvlecl: Kind of a detail but I think the responsibility of the DAO should not include the management of that study metadata object, only of the study actual data.

It would even be better that we don't pass in an sqlalchemy object (`RawStudy`), but only the data needed, so that it's clear in the API that the DAO does not mess with the study table.

But the body of the function uses it quite heavily so it may require changes too large for that PR

sylvlecl (commented):
Nice, looks good, I left a comment about separation of concerns between the DAO and the study metadata, but we may postpone that topic, what do you think ?

antarest/study/dao/api/study_factory_dao.py
```diff
@@ -0,0 +1,22 @@
...
+
+from abc import ABC, abstractmethod
+
+from antarest.study.dao.api.study_dao import StudyDao
+from antarest.study.model import RawStudy
+
+
+class StudyFactoryDao(ABC):
+    @abstractmethod
+    def create_study_dao(self, study: RawStudy) -> tuple[StudyDao, RawStudy]:
```
TheoPascoli: I know this tuple is needed for the test, but returning two different values here is questionable. you may want to separate responsibilities between methods, and actually the `_, raw` reassignment is useless since raw is mutable and already modified in place in the `create_study_dao` method . So `raw` before and after the call points to the same object, so there’s no need to return it.

antarest/study/dao/file/file_study_factory_dao.py
MartinBelthle: I agree with you. But as you said the file study implem needs lots of stuff to create the files (the path, if it's managed to determine whether matrices should be normalized or not, the autor and editor to save it inside the study.antares file ...) so i don't really know how I could do it in another way

TheoPascoli (changes requested):

MartinBelthle:
@sylvlecl , @TheoPascoli what I could do is move back this code 
```python
study.content_status = StudyContentStatus.VALID
self._study_repository.save(study)
```
inside the study service as we do this no matter the DAO. This would remove the dependency inside the factory to `StudyMetadataRepository`

And then the `create_study_dao` method could return only a DAO as Theo said we modify the study object inplace (inside the FS implem) and take as an argument only a `Study` object instead of `RawStudy`. What do you think about this ?

TheoPascoli:
> @sylvlecl , @TheoPascoli what I could do is move back this code
> 
> ```python
> study.content_status = StudyContentStatus.VALID
> self._study_repository.save(study)
> ```
> 
> inside the study service as we do this no matter the DAO. This would remove the dependency inside the factory to `StudyMetadataRepository`
> 
> And then the `create_study_dao` method could return only a DAO as Theo said we modify the study object inplace (inside the FS implem) and take as an argument only a `Study` object instead of `RawStudy`. What do you think about this ?

I agree with moving back the code to remove duplucation and the `StudyMetadataRepository`. Its cleaner.
You could also documente in the docstring of `create_study_dao` that the study is modified as a side effect of calling it to make sure the caller know what he is doing

antarest/study/dao/database/database_study_factory_dao.py
```diff
@@ -0,0 +1,45 @@
...
+from antarest.core.utils.fastapi_sqlalchemy import db
+from antarest.matrixstore.service import ISimpleMatrixService
+from antarest.study.business.model.area_model import DEFAULT_LAYER_ID, DEFAULT_LAYER_NAME
+from antarest.study.business.model.layer_model import Layer
+from antarest.study.dao.api.study_factory_dao import StudyFactoryDao
+from antarest.study.dao.database.database_study_dao import DatabaseStudyDao
+from antarest.study.model import Study
+
+
+class DataBaseStudyDaoFactory(StudyFactoryDao):
```
TheoPascoli: Minor: We usually write `Database` with lowercase `b` like `DatabaseStudyDao`

antarest/study/dao/database/database_study_factory_dao.py
```diff
@@ -0,0 +1,45 @@
...
+    def session(self) -> Session:
+        """Get the SqlAlchemy session or create a new one on the fly if not available in the current thread."""
+        if self._session is None:
+            return db.session
+        return self._session
+
+    @override
+    def create_study_dao(self, study: Study) -> DatabaseStudyDao:
+        dao = DatabaseStudyDao(study.id, self.session, self._matrix_service)
+        dao.save_layer(Layer(id=DEFAULT_LAYER_ID, name=DEFAULT_LAYER_NAME))
```
TheoPascoli, line 44: What happens if `save_layer` fails ? We get a study created but not initialized. Maybe we should try/except it, or something else, I honestly have no idea right now

antarest/study/service.py
```diff
@@ -954,13 +962,9 @@ def create_study(
...
         )
 
-        raw = self.storage_service.raw_study_service.create(raw)
-
-        self._save_study(raw)
-
-        dao = self._create_study_dao(raw)
-
-        dao.initialize_study()
+        raw.content_status = StudyContentStatus.VALID
```
TheoPascoli, line 965: Minor: you set it valid before the creation, maybe move it after

TheoPascoli (changes requested):

antarest/study/service.py
```diff
@@ -954,13 +962,9 @@ def create_study(
...
+        self.repository.save(raw)
+        self._study_dao_factories[raw.storage_mode].create_study_dao(raw)
```
TheoPascoli, line 967: You save the study here, but in create_study_dao, you modify the path but dont save it again. So the new field is ignored.
However, is the create_study_dao, you do:
```
path_study = Path(study.path)
...
study.path = str(path_study)
```
which look like a no-op
I think you can remove this logic in create_study_dao

TheoPascoli (changes requested):

antarest/study/service.py
MartinBelthle: We already did this i think as we have to save the data in DB.
And we have to create the study in DB before everything as otherwise the `save_layer` will fail for foreginkey issues so I think we'll have to keep it as is

antarest/study/service.py
MartinBelthle: Indeed and now I can remove the msg saying I'm moditying the `study` object as I'm not

antarest/study/dao/database/database_study_factory_dao.py
MartinBelthle: I don't see what could cause the `save_layer` to fail so I don't think we need to handle this case

antarest/study/service.py
MartinBelthle: In fact it seems we never use `content_status`, we could remove it in a future work

antarest/study/dao/database/database_study_factory_dao.py
sylvlecl: This could very rarely for example if we get some network error. In that case the error will propagate, it could be handled outside but not here.

antarest/study/service.py
```diff
@@ -954,13 +962,9 @@ def create_study(
...
 
-        raw = self.storage_service.raw_study_service.create(raw)
-
-        self._save_study(raw)
-
-        dao = self._create_study_dao(raw)
-
-        dao.initialize_study()
+        raw.content_status = StudyContentStatus.VALID
+        self.repository.save(raw)
```
sylvlecl, line 966: Indeed here we commit the study before it is initialized, maybe we could move it to after initialization, to avoid having incorrectly initialized studies ?

sylvlecl: But that would go against what we have decided, that DAO methods should handle the commits themselves.

So we'd better leave it this way. The best we could do is if creation fails, remove the metadata from the repo. That's really and edge case anyway

sylvlecl (approved):
Pending topic about the edge case where DAO creation fails, but it's not a blocking issue

TheoPascoli (approved):

PR #2983, opened 2026-02-04
refactor(dao): add a utility upsert method

sylvlecl:
This will ensure less queries and stronger consistency guarantees when
updating or inserting data.


antarest/study/dao/database/database_area_dao.py
```diff
@@ -462,21 +452,16 @@ def _get_matrix_row(self, area_id: str, table: Table) -> Row[Any] | None:
...
-            session.execute(stmt_insert)
-        else:
-            stmt_update = (
-                update(table)
-                .where((table.c.study_id == study_id) & (table.c.area_id == area_id))
-                .values(matrix_id=matrix_id)
-            )
-            session.execute(stmt_update)
+
+        validate_area_exists(session, study_id, area_id)
```
sylvlecl, line 458: Note that indeed here, in the case of an udpate, that validation is not actually needed... but code is simpler !
The alternative would be to check for `IntegrityError`, but it's not easy to check that it's indeed the foreign key constraint which is violated. The check would be DB-specific here too.

tests/study/dao/test_sql_utils.py
```diff
@@ -0,0 +1,72 @@
...
+
+    insertions = [
+        {"id": "1", "sub_id": "2", "str_value": "val1", "int_value": 12},
+        {"id": "1", "sub_id": "3", "str_value": "val2", "int_value": 52},
+        {"id": "2", "sub_id": "3", "str_value": "val3", "int_value": 66},
+    ]
+
+    # generic_upsert(session, TEST_TABLE, values=insertions, key_columns=[TEST_TABLE.c.id, TEST_TABLE.c.sub_id],
+    # update_columns=["str_value", "int_value"])
+    upsert_method(session, TEST_TABLE, values=insertions)
```
MartinBelthle: Could we add a test ensuring it does not perform N+1 queries ?

antarest/study/dao/database/database_layer_dao.py
```diff
@@ -114,27 +115,8 @@ def save_layer(self, layer: Layer) -> None:
...
-            session.execute(
-                insert(LAYER_TABLE).values(
-                    study_id=study_id,
-                    layer_id=layer.id,
-                    name=layer.name or "",
-                )
-            )
-
+        values = {"study_id": study_id, "layer_id": layer.id, "name": layer.name or ""}
+        upsert_one(session, LAYER_TABLE, values)
```
MartinBelthle, line 118: There's a difference between this and the old logic.
If someone updated an existing layer without giving it a name we did nothing. Now we're replacing the existing name by `""`

antarest/study/dao/database/sql_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    if not values:
+        return
+
+    dialect = session.get_bind().dialect.name
+    if dialect in {"postgresql", "sqlite"}:
+        key_columns = _key_columns(table)
+        update_columns = _update_columns(table)
+
+        # Note for the ignorant: the postgres and sqlite syntax requires to explicitly
```
MartinBelthle: Seems a bit harsh :), could just be `Note: `

antarest/study/dao/database/sql_utils.py
```diff
@@ -0,0 +1,121 @@
...
+def _update_columns(table: Table) -> list[Column[Any]]:
+    return [column for column in table.columns.values() if not column.primary_key]
+
+
+def upsert_one(
+    session: Session,
+    table: Table,
+    values: dict[str, Any],
+) -> None:
+    return upsert_multiple(session, table, [values])
```
MartinBelthle: You could also add a check: `if not values` as otherwise you'll enter `upsert_multiple` with `[{}]` and you'll pass the check

TheoPascoli: You have `return` with `-> None` in the signature

antarest/study/dao/database/sql_utils.py
```diff
@@ -0,0 +1,121 @@
...
+        else:
+            rows_to_insert.append(row)
+
+    if rows_to_insert:
+        session.execute(insert(table).values(rows_to_insert))
+
+    # 1 query per update unfortunately. Might be possible to optimize, but not worth it for us now.
+    for row in rows_to_update:
+        filters = [column == row[column.name] for column in key_columns]
+        stmt_update = update(table).where(and_(*filters)).values({k: row[k.name] for k in update_columns})
```
TheoPascoli: Could you rename `k` with `col` for better readability ?

antarest/study/dao/database/database_layer_dao.py
```diff
@@ -114,27 +115,8 @@ def save_layer(self, layer: Layer) -> None:
         session = self.get_session()
 
         # Check if layer already exists
```
TheoPascoli: This comment has no point anymore

antarest/study/dao/database/database_layer_dao.py
```diff
@@ -114,27 +115,8 @@ def save_layer(self, layer: Layer) -> None:
...
-        else:
-            session.execute(
-                insert(LAYER_TABLE).values(
-                    study_id=study_id,
-                    layer_id=layer.id,
-                    name=layer.name or "",
-                )
-            )
-
+        values = {"study_id": study_id, "layer_id": layer.id, "name": layer.name or ""}
```
TheoPascoli, line 117: The `layer.name` is optional in the model. If the user only wants to update the areas of a layer without specifying the name, you will update the name with "" as well
edit: Martin already noted this

TheoPascoli (changes requested):

antarest/study/dao/database/sql_utils.py
sylvlecl: All the methods here assume that the dict is consistent with the table definition, they don't check for this for perf reason.
I prefer to not check it for the "upsert_one" version since we don't check it for the "update_multiple" version.

sylvlecl: (I am adding that in the docstring)

antarest/study/dao/database/database_layer_dao.py
sylvlecl: I don't think it was the expected behaviour: here we have a layer, not a layer update.

The UpdateLayer command actually implements the update logic, which does not do anything if name is None

antarest/study/dao/database/database_layer_dao.py
sylvlecl: See other comment, I don't think it's the expected behaviour, here we overwrite the Layer, we don't update it.
The UpdateLayer command takes care of the update logic.

antarest/study/dao/database/sql_utils.py
sylvlecl: Of course it mainly included myself :)

tests/study/dao/test_sql_utils.py
sylvlecl: yes, done

antarest/study/dao/database/database_layer_dao.py
sylvlecl: Removed

antarest/study/dao/database/sql_utils.py
sylvlecl: Done

antarest/study/dao/database/sql_utils.py
sylvlecl: Done

sylvlecl:
About the optional layer name, I propose to change it in another PR for clarity sake

antarest/study/dao/database/sql_utils.py
sylvlecl: I also added a test which shows that the method raises anyway when we have absent primary keys

TheoPascoli (approved):

PR #2984, opened 2026-02-04
feat(favorite directory): adding favorite directory class

mehdiwahada:
It does the same thing as the favorite studies (#2968 )


alembic/versions/y7x41d4a8yc_add_foreign_key_users_favorite_table.py
```diff
@@ -5,29 +5,28 @@
...
-import sqlalchemy as sa
-
 from alembic import op
+import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
-revision = "y7x41d4a8yc"
-down_revision = "10318e5320f6"
+revision = 'y7x41d4a8yc'
+down_revision = '6a6d36e3c6ed'
```
sylvlecl: we should not change this after it has been merged on dev, otherwise it will change the history for all other dévelopers

alembic/versions/9v8a86r0g4xy_merge_link_and_favorite.py
```diff
@@ -0,0 +1,24 @@
+"""merge_link_and_favorite
```
sylvlecl: This file here is not needed anymore, you just need to ensure the migration you are adding has the previous head (the one on dev branch) as its down_revision

alembic/versions/u9t13x9i6kt_add_favorite_directory_table.py
```diff
@@ -0,0 +1,30 @@
...
+Revises: 6a6d36e3c6ed
+Create Date: 2026-01-26 10:45:36.111886
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = 'u9t13x9i6kt'
+down_revision = 'y7x41d4a8yc'
```
sylvlecl: you will need to change it so that it corresponds to the head on dev branch

alembic/versions/y7x41d4a8yc_add_foreign_key_users_favorite_table.py
```diff
@@ -5,29 +5,28 @@
...
+
+    op.drop_table('favorite_study')
     op.create_table(
         "favorite_study",
         sa.Column("user_id", sa.Integer(), nullable=False),
         sa.Column("study_id", sa.String(36), nullable=False),
         sa.PrimaryKeyConstraint("user_id", "study_id", name="pk_favorite_study"),
         sa.ForeignKeyConstraint(["user_id"], ["identities.id"], name="fk_user_id_favorite_study", ondelete="CASCADE"),
-        sa.ForeignKeyConstraint(["study_id"], ["study.id"], name="fk_study_id_favorite_study", ondelete="CASCADE"),
+        sa.ForeignKeyConstraint(["study_id"], ["study.id"], name="fk_study_id", ondelete="CASCADE")
```
sylvlecl: Same thing: you must not change this after it has been merged on dev branch

antarest/favorite/web.py
```diff
@@ -14,28 +14,45 @@
...
 
     @bp.get("/favorites/studies/", summary="Listing favorites for current user")
-    def list_favorites() -> list[FavoriteStudyDTO]:
+    def list_favorite_studies() -> list[FavoriteStudyDTO]:
         logger.info("Listing favorites for current user")
         return favorite_service.list_favorites()
 
-    @bp.post("/favorites/studies/{uuid}", summary="Create new favorite study")
-    def add_favorite(uuid: str) -> FavoriteStudyDTO:
+    @bp.post("/favorites/studies/{uuid}", summary="Create a new favorite study")
```
sylvlecl: Note that it would be more clear for the user to say "Add a study in the list of favorite studies", same thing for the directories below

sylvlecl (changes requested):
A few comments, I will review more in depth later

TheoPascoli:
Hello Mehdi, I haven't watch a lot of the code, but you must have only one migration file, and you must not modify an existing alembic migration fail because it would break the "chain" of the migration. 
I'll check the code a bit later

tests/integration/favorite_blueprint/test_favorite_directory.py
```diff
@@ -0,0 +1,51 @@
...
+
+    directory_test_1 = client.post("/v1/directories", json={"name": "directory_test_1"}).json()
+    directory_test_2 = client.post("/v1/directories", json={"name": "directory_test_2"}).json()
+    dt_1_id = directory_test_1["id"]
+    dt_2_id = directory_test_2["id"]
+    # adding favorite directories and checking the API returns the good amount of favorites
+
+    resp = client.post(f"/v1/favorite/directories/{dt_1_id}")
+    assert resp.status_code == 200
+    directory_dto_1 = resp.json()
```
sylvlecl, line 32: We should check explicitly the content of the JSON.

Your test only checks that the GET which retrieves the list is consistent with the GET which retrieves the favorites one by one, but it does not guarantee that they have the right content.

tests/integration/favorite_blueprint/test_favorite_directory.py
```diff
@@ -0,0 +1,51 @@
...
+    resp = client.post(f"/v1/favorite/directories/{dt_2_id}")
+    assert resp.status_code == 200
+    directory_dto_2 = resp.json()
+
+    actual_favorite_directories_list = client.get("/v1/favorite/directories").json()
+    expected_favorite_directories_list = [directory_dto_1, directory_dto_2]
```
sylvlecl: we'd better hard code the expected value, it will make test more straightforward (of course you cannot hard code the IDs, but the name yes)

tests/integration/favorite_blueprint/test_favorite_directory.py
```diff
@@ -0,0 +1,51 @@
...
+
+    assert actual_favorite_directories_list == expected_favorite_directories_list
+
+    resp = client.delete(f"/v1/favorite/directories/{dt_1_id}")
+    assert resp.status_code == 200
+    expected_favorite_directories_list.remove(directory_dto_1)
```
sylvlecl, line 49: we'd better remove that logic from the test and hard code the expected value

tests/integration/favorite_blueprint/test_favorite_directory.py
```diff
@@ -0,0 +1,51 @@
...
+    actual_favorite_directories_list = client.get("/v1/favorite/directories").json()
+
+    assert actual_favorite_directories_list == expected_favorite_directories_list
+
+    resp = client.delete(f"/v1/favorite/directories/{dt_2_id}")
+    assert resp.status_code == 200
+    expected_favorite_directories_list.remove(directory_dto_2)
+    actual_favorite_directories_list = client.get("/v1/favorite/directories").json()
+
+    assert actual_favorite_directories_list == expected_favorite_directories_list
```
sylvlecl: Same, you can just check it's equal to `[]`, much simpler to read

tests/integration/favorite_blueprint/test_favorite_directory.py
```diff
@@ -0,0 +1,51 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+
+from starlette.testclient import TestClient
+
+
+def test_favorite_directory(client: TestClient, admin_access_token: str) -> None:
```
sylvlecl, line 17: We should also check that when we delete a directory, it's also deleted from the favorites list.

Also, we need to add error case tests, in particular:
- when adding a directory that does not exist in favorite, we should get a 404 response
- same when deleting a directory that does not exist.

We need to add the same tests in the favorite study tests

antarest/service_creator.py
```diff
@@ -155,20 +155,32 @@ class CoreServices:
...
 
 def build_favorite_service(
-    config: Config, app_ctxt: Optional[AppBuildContext] = None, service: Optional[FavoriteService] = None
-) -> FavoriteService:
-    if service is None:
-        favorite_repository = FavoriteRepository()
-        service = FavoriteService(favorite_repository=favorite_repository)
+    config: Config,
+    app_ctxt: Optional[AppBuildContext] = None,
+    favorite_study_service: Optional[FavoriteStudyService] = None,
```
sylvlecl: Id does not seem useful to have those extra arguments, we never use them

antarest/service_creator.py
```diff
@@ -336,7 +350,7 @@ class Services:
     event_bus: IEventBus
     study: StudyService
     matrix: MatrixService
-    favorite: FavoriteService
+    favorite: FavoriteStudyService
```
sylvlecl: we'd better add the other, new service here too for consistency

antarest/launcher/service.py
```diff
@@ -106,7 +106,7 @@ def __init__(
         self,
         config: Config,
         study_service: StudyService,
-        favorite_service: FavoriteService,
+        favorite_service: FavoriteStudyService,
```
sylvlecl: Should be removed, it is not used by the launcher service, which is only used to launch computations

antarest/launcher/main.py
```diff
@@ -31,7 +31,7 @@ def build_launcher(
     app_ctxt: Optional[AppBuildContext],
     config: Config,
     study_service: StudyService,
-    favorite_service: FavoriteService,
```
sylvlecl, line 34: to be removed

tests/integration/favorite_blueprint/test_favorite.py
```diff
@@ -13,7 +13,7 @@
 from starlette.testclient import TestClient
 
 
-def test_favorite(client: TestClient, admin_access_token: str) -> None:
```
sylvlecl, line 16: As for the new directory tests, we should check that when we delete a study, the study is also removed from the user favorites

sylvlecl (changes requested):
A few more improvements requested, mainly in tests to cover more cases and improve readability

antarest/favorite/web.py
```diff
@@ -14,28 +14,45 @@
...
         return favorite_service.add_favorite(uuid)
 
-    @bp.delete("/favorites/studies/{uuid}", summary="Delete a favorite study")
-    def delete_favorite(uuid: str) -> None:
+    @bp.delete("/favorites/studies/{uuid}", summary="Delete a study from the ")
+    def delete_favorite_study(uuid: str) -> None:
         logger.info("Deleting favorite study for current user")
         favorite_service.delete_favorite(uuid)
 
+    @bp.get("/favorite/directories", summary="Listing favorite directories for current user")
```
sylvlecl: We need to use "favorite**s**", like for studies. Same in the other endpoints

sylvlecl (approved):

PR #2985, opened 2026-02-04
feat(hydro): dao db hydro

smailio:


antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,637 @@
...
+
+        stmt = select(HYDRO_ALLOCATION_TABLE).where(
+            (HYDRO_ALLOCATION_TABLE.c.study_id == study_id) & (HYDRO_ALLOCATION_TABLE.c.source_area_id == area_id)
+        )
+        rows = session.execute(stmt).fetchall()
+
+        allocation_areas = [
+            HydroAllocationArea(area_id=row.target_area_id, coefficient=row.coefficient) for row in rows
+        ]
+        return HydroAllocation(allocation=allocation_areas)
```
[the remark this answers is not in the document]
smailio: In that case HydroAllocation will fail to initialise and raises a ValueError.

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,639 @@
...
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        validate_area_exists(session, study_id, area_id)
+
+        stmt = select(HYDRO_ALLOCATION_TABLE).where(
+            (HYDRO_ALLOCATION_TABLE.c.study_id == study_id) & (HYDRO_ALLOCATION_TABLE.c.source_area_id == area_id)
+        )
+        rows = session.execute(stmt).fetchall()
```
smailio, line 267: I don't check for empty rows as it is never empty, except in case of a bug or data base corruption.  HydroAllocation will raise a value error is rows is empty during model validation. I though adding a check here would be redundant. Let me know if you think I should add one.

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,639 @@
...
+            hydro_management: The HydroManagement configuration to save.
+            area_id: The area identifier.
+
+        Raises:
+            AreaNotFound: If the area does not exist.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        validate_area_exists(session, study_id, area_id)
```
smailio: I'm not doing the validation for fields exclusive to 9.2 I think this is not the concern fo the DAO.

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,637 @@
...
+        session.execute(stmt_delete)
+
+        # Insert new correlations (only upper triangle: area_from < area_to)
+        # Diagonal is implicit, not stored
+        insert_values = []
+        for i in range(len(area_ids)):
+            # not saved: values from the diagonal are always == 1.0
+            for j in range(i + 1, len(area_ids)):
+                coefficient = current_correlation_matrix.data[i][j]
+                if not coefficient:
```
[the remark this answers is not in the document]
smailio: I want to ignore zeros as well.  Matrices are sparses, we only store values other than zero, later we'll interpret missing values as zero,  to save space.

alembic/versions/ae838d5fd166_add_hydro_tables.py
```diff
@@ -0,0 +1,109 @@
...
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("source_area_id", sa.String(length=255), nullable=False),
+        sa.Column("target_area_id", sa.String(length=255), nullable=False),
+        sa.Column("coefficient", sa.Float, nullable=False),
+        sa.ForeignKeyConstraint(
+            ["study_id", "source_area_id"],
+            ["area.study_id", "area.area_id"],
+            name=op.f("fk_hydro_allocation_study_id_source_area_id_area"),
+            ondelete="CASCADE",
+        ),
```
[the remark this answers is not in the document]
smailio: done

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,639 @@
...
+            Dictionary mapping source area_id to HydroAllocation.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        stmt = select(HYDRO_ALLOCATION_TABLE).where(HYDRO_ALLOCATION_TABLE.c.study_id == study_id)
+        rows = session.execute(stmt).fetchall()
+
+        if not rows:
+            raise HydroAllocationNotFound("No hydro allocation data found for study")
```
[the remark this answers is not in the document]
smailio: If rows is empty the DB is in an inconsistent state so I raise. Unlike the other function a ValueError won't be raised here so I do it manually.

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,609 @@
...
+        session = self.get_session()
+
+        # Validate source area exists
+        validate_area_exists(session, study_id, area_id)
+
+        # Validate all target areas exist
+        target_area_ids = {a.area_id for a in allocation.allocation}
+        validate_areas_exists(session, study_id, target_area_ids)
+
+        # Delete existing allocations for this source area
```
smailio, line 324: I don't use upsert multiple here because I would have to run a delete to remove allocation that are missing in `allocation` anyway. It's more straight forward to do like this 1) delete all related allocation 2) insert allocation sent by the user, rather than 1) upsert allocation sent by user 2) check if some allocaiton are in db but not in user payload 3) delete thise.

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,639 @@
...
+        insert_values = [
+            {
+                "study_id": study_id,
+                "source_area_id": area_id,
+                "target_area_id": alloc_area.area_id,
+                "coefficient": alloc_area.coefficient,
+            }
+            for alloc_area in allocation.allocation
+        ]
+
```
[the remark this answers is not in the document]
smailio: Can't be empty the allocation model validator prevent it.

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,637 @@
...
+                    # null values are not saved
+                    continue
+                area_from = area_ids[i]
+                area_to = area_ids[j]
+                insert_values.append(
+                    {
+                        "study_id": study_id,
+                        "area_from": area_from,
+                        "area_to": area_to,
+                        "coefficient": coefficient,
```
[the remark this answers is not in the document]
smailio: conversion is done when initializint hydro matrix object

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,637 @@
...
+                    }
+                )
```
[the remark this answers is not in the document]
smailio: fixed

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,639 @@
...
+
```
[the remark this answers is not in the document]
smailio: Updating the correlation for one are is as updating correlation for the whole study because of this matrix system.

antarest/study/dao/database/models/hydro.py
```diff
@@ -0,0 +1,112 @@
...
+        ondelete="CASCADE",
+    ),
+)
+
+HYDRO_CORRELATION_TABLE = Table(
+    "hydro_correlation",
+    metadata,
+    Column("study_id", String(36), nullable=False, primary_key=True),
+    Column("area_from", String(255), nullable=False, primary_key=True),
+    Column("area_to", String(255), nullable=False, primary_key=True),
```
smailio, line 97: Idelally I'd be able to add a constraint so that it's impossible to insert a Area2 => Area1 row if Area1 => Area2 already exists. But I can't think of a way to do it. Let me know if you can think of a solution.

alembic/versions/ae838d5fd166_add_hydro_tables.py
```diff
@@ -0,0 +1,115 @@
...
+            name=op.f("fk_hydro_inflow_structure_study_id_area_id_area"),
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint("study_id", "area_id", name=op.f("pk_hydro_inflow_structure")),
+    )
+
+    # hydro_allocation table
+    op.create_table(
+        "hydro_allocation",
+        sa.Column("study_id", sa.String(length=36), nullable=False),
```
smailio, line 69: I saw in other PR that the definition of study_id was centralized in a study_id_col variable. But when I do so here I get a weird warning 

Table 'hydro_correlation' specifies columns 'study_id' as primary_key=True, not matching locally specified columns 'study_id', 'area_from', 'area_to'; setting the current primary key columns to 'study_id', 'area_from', 'area_to'. This warning may become an exception in a future release

MartinBelthle (changes requested):
To me we should directly look inside the hydro table instead of using the `validate_area_exists` as currently we perform 2 requests and we could just do one and if it returns None, return an AreaNotFoundError

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,609 @@
...
+            HydroManagement configuration for the area.
+
+        Raises:
+            AreaNotFound: If the area does not exist.
+            HydroConfigNotFound: If hydro management config does not exist for the area.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        validate_area_exists(session, study_id, area_id)
```
MartinBelthle: Same remark

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,609 @@
...
+
+    @staticmethod
+    def _convert_row_to_inflow_structure(row: Row[Any]) -> InflowStructure:
+        """Convert a database row to InflowStructure model."""
+        return InflowStructure(inter_monthly_correlation=row.inter_monthly_correlation)
+
+    @staticmethod
+    def _hydro_management_to_dict(hydro_management: HydroManagement) -> dict[str, Any]:
+        """Convert HydroManagement model to dict for database insert/update."""
+        return {
```
MartinBelthle: It makes me realize that we could just do a `model_dump()` as it's a Pydantic model. We didn't do this in other DAOs but we could, it would simplify the code IMO

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,609 @@
...
+        Raises:
+            HydroConfigNotFound: If any area is missing hydro management configuration.
+            HydroInflowStructureNotFound: If any area is missing inflow structure configuration.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        # Get ALL areas in the study
+        stmt_areas = select(AREA_TABLE.c.area_id).where(AREA_TABLE.c.study_id == study_id)
+        all_area_ids = [row.area_id for row in session.execute(stmt_areas)]
```
MartinBelthle: To me we could just directly check inside the hydro tables to perform only 2 requests instead of 3. That's what's done inside the thermal dao for instance

antarest/study/dao/database/database_hydro_dao.py
MartinBelthle: You're saying that this method performs a replace of all existing allocation by the new ones that are given if I understand correctly. If so I understand but we should specify it inside the method documentation

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,609 @@
...
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        # Validate area exists
+        validate_area_exists(session, study_id, area_id)
+
+        # Validate all correlated areas exist
+        for corr_area in correlation.correlation:
+            if corr_area.area_id != area_id:
+                validate_area_exists(session, study_id, corr_area.area_id)
```
MartinBelthle: This will perform N+1 queries

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,609 @@
...
+            j = area_ids.index(row.area_to)
+            array[i][j] = row.coefficient
+            array[j][i] = row.coefficient
+
+        return HydroCorrelationMatrix(index=area_ids, columns=area_ids, data=array)
+
+    @override
+    def save_hydro_correlation(self, area_id: str, correlation: HydroCorrelation) -> None:
+        """
+        Save hydro correlation for a specific area.
```
MartinBelthle, line 406: Same comment we should document that this does a 'replace'

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,609 @@
...
+
+        # Validate area exists
+        validate_area_exists(session, study_id, area_id)
+
+        # Validate all correlated areas exist
+        for corr_area in correlation.correlation:
+            if corr_area.area_id != area_id:
+                validate_area_exists(session, study_id, corr_area.area_id)
+
+        # Get current matrix and apply changes (validates diagonal = 1, symmetry, etc.)
```
MartinBelthle: All of this seems complicated and unecessary. To me we should: 
1- Remove where study_id=studyid and area_from_id=area_id or area_to_id = area_id
2- Insert the new values

That's it

antarest/study/dao/database/common.py
```diff
@@ -29,6 +29,29 @@ def area_exists(session: Session, study_id: str, area_id: str) -> bool:
     return session.execute(stmt).fetchone() is not None
 
 
+def validate_areas_exists(session: Session, study_id: str, area_ids: set[str]) -> None:
```
MartinBelthle: As we have foreign key constraints, we could just do this instead (example in the link dao that I copied here)
```python
try:
    upsert_one(session, LINK_TABLE, values)
except IntegrityError:
    # Happens if the link's areas did not existed -> ForeignKey constraint fails
    session.rollback()
    raise AreaNotFound(*[link.area1, link.area2])

session.commit()

```
It will be more performant in 99% of the cases where we don't have any issue with wrong areas.
I think we don't use the foreign key as we should as the only example we currently have is this one. 
What are your thoughts on this @sylvlecl , @TheoPascoli ?

tests/study/dao/test_database_hydro_dao.py
```diff
@@ -0,0 +1,484 @@
...
+    HYDRO_CORRELATION_TABLE,
+    HYDRO_INFLOW_STRUCTURE_TABLE,
+    HYDRO_MANAGEMENT_TABLE,
+)
+
+
+class TestHydroManagement:
+    """Tests for hydro management CRUD operations."""
+
+    def test_get_hydro_management_raises_error_for_area_without_hydro_config(self, dao: DatabaseStudyDao) -> None:
```
MartinBelthle, line 37: To me this test is not really useful

MartinBelthle (changes requested):

antarest/study/dao/database/database_hydro_dao.py
smailio: It's the same implementation as the file dao. It's to handle the Area1 => Area2 also updates Area 2 => Area 1 rule. That logic is already implemented in set_correlation, if I do as you say I'll have to duplicate that logic in this DAO.

antarest/study/dao/database/database_hydro_dao.py
smailio: You made me realize that maybe we need to exclude None values.

antarest/study/dao/database/database_hydro_dao.py
smailio: A SELECT can't raise an integrity error. So either we check the areas like this or we decide that returning a HydroConfigNotFound is enough. I'd rather keep the check. What do youu think  ?

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,596 @@
...
+        return self._db_session
+
+    @abstractmethod
+    def get_impl(self) -> "DatabaseStudyDao":
+        pass
+
+    @staticmethod
+    def _convert_row_to_hydro_management(row: Row[Any]) -> HydroManagement:
+        """Convert a database row to HydroManagement model."""
+        return HydroManagement(
```
MartinBelthle: You could simply do 
```python
data = row._asdict()
del data["study_id"]
return HydroManagement(**data)
```

antarest/study/dao/database/database_hydro_dao.py
MartinBelthle: As we said we shouldn't verify the area exists as in 99% of the case it will so it only degrades our perfs.
But in the case you don't have any row, you should check this to raise a proper exception.

Also, you can remove the exceptions you've introduced and simply raise a ValueError for now. In a future PR we'll create a dedicated exceptions for these cases that should not happen (an area but not its related hydr, etc.)

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,596 @@
...
+
+        values = hydro_management.model_dump()
+        values["study_id"] = study_id
+        values["area_id"] = area_id
+        try:
+            upsert_one(session, HYDRO_MANAGEMENT_TABLE, values)
+            session.commit()
+        except IntegrityError as e:
+            session.rollback()
+            invalid = self.get_impl().get_invalid_area_ids([area_id])
```
MartinBelthle: You should remove this line and just raise with the given `area_id`

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,596 @@
...
+            InflowStructure configuration for the area.
+
+        Raises:
+            AreaNotFound: If the area does not exist.
+            HydroInflowStructureNotFound: If inflow structure does not exist for the area.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        validate_area_exists(session, study_id, area_id)
```
MartinBelthle: Samre remark as the first one

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,596 @@
...
+            "study_id": study_id,
+            "area_id": area_id,
+            "inter_monthly_correlation": inflow_structure.inter_monthly_correlation,
+        }
+        try:
+            upsert_one(session, HYDRO_INFLOW_STRUCTURE_TABLE, values)
+            session.commit()
+        except IntegrityError as e:
+            session.rollback()
+            invalid = self.get_impl().get_invalid_area_ids([area_id])
```
MartinBelthle: Same

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,596 @@
...
+
+        Raises:
+            HydroConfigNotFound: If any area is missing hydro management configuration.
+            HydroInflowStructureNotFound: If any area is missing inflow structure configuration.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        # Get ALL areas in the study
+        all_area_ids = self.get_impl().get_all_area_ids()
```
MartinBelthle: This is pretty inefficient. I discussed the matter with Sylvain and what he proposes is to use 1 Join

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,596 @@
...
+            HydroAllocation containing the allocation coefficients.
+            Returns empty allocation if no allocation data exists.
+
+        Raises:
+            AreaNotFound: If the area does not exist.
+        """
+        study_id = self.get_study_id()
+        session = self.get_session()
+
+        validate_area_exists(session, study_id, area_id)
```
MartinBelthle: Same

antarest/study/dao/database/database_hydro_dao.py
MartinBelthle: Isn't there a way with DBs to do a "replace" in one query instead of this delete all and then insert ?

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,596 @@
...
+            area_id: The area identifier.
+
+        Returns:
+            HydroCorrelation containing the correlation coefficients for all areas.
+            Self-correlation is always 100%, missing correlations default to 0%.
+
+        Raises:
+            AreaNotFound: If the area does not exist.
+        """
+        validate_area_exists(self.get_session(), self.get_study_id(), area_id)
```
MartinBelthle: This is not really performant, couldn't you just do a select where area_from = area_id or area_to = area_id ?

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,596 @@
...
+        for row in rows:
+            i = area_ids.index(row.area_from)
+            j = area_ids.index(row.area_to)
+            array[i][j] = row.coefficient
+            array[j][i] = row.coefficient
+
+        return HydroCorrelationMatrix(index=area_ids, columns=area_ids, data=array)
+
+    @override
+    def save_hydro_correlation(self, area_id: str, correlation: HydroCorrelation) -> None:
```
MartinBelthle, line 404: This method is really really unefficient. Why can you not do a delete on the related area_id and just insert afterwards ?

MartinBelthle (changes requested):

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,596 @@
...
+        insert_values = [
+            {
+                "study_id": study_id,
+                "source_area_id": area_id,
+                "target_area_id": alloc_area.area_id,
+                "coefficient": alloc_area.coefficient,
+            }
+            for alloc_area in allocation.allocation
+        ]
+
```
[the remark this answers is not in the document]
smailio: can't be empty because allocation model validator prevent that

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,596 @@
...
+        # Get stored correlations and fill the matrix
+        stmt = select(HYDRO_CORRELATION_TABLE).where(HYDRO_CORRELATION_TABLE.c.study_id == study_id)
+        rows = session.execute(stmt).fetchall()
+
+        for row in rows:
+            i = area_ids.index(row.area_from)
+            j = area_ids.index(row.area_to)
+            array[i][j] = row.coefficient
+            array[j][i] = row.coefficient
```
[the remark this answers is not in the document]
smailio: done

antarest/study/dao/database/database_hydro_dao.py
smailio: Not that I'm aware of

antarest/study/dao/database/database_hydro_dao.py
smailio: Now I need this function in hydro properties, the row it receives there contains fields that aren't in hydro management (thos from inflow, because now I retrieve data with a join) so I need to filter them out.

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,564 @@
...
+                HYDRO_MANAGEMENT_TABLE,
+                (AREA_TABLE.c.study_id == HYDRO_MANAGEMENT_TABLE.c.study_id)
+                & (AREA_TABLE.c.area_id == HYDRO_MANAGEMENT_TABLE.c.area_id),
+            )
+            .join(
+                HYDRO_INFLOW_STRUCTURE_TABLE,
+                (AREA_TABLE.c.study_id == HYDRO_INFLOW_STRUCTURE_TABLE.c.study_id)
+                & (AREA_TABLE.c.area_id == HYDRO_INFLOW_STRUCTURE_TABLE.c.area_id),
+            )
+            .where(AREA_TABLE.c.study_id == study_id)
```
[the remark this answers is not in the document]
smailio: If db state isn't corrupt each area has a management and an inflow structure.

alembic/versions/9e96cc5d95de_add_hydro_tables.py
```diff
@@ -0,0 +1,115 @@
...
+
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "9e96cc5d95de"
+down_revision = "0e628a645214"
```
MartinBelthle: You have to bump your down revision

antarest/study/dao/database/database_study_dao.py
```diff
@@ -355,6 +361,38 @@ def get_st_storage_additional_constraints(
     ) -> list[STStorageAdditionalConstraint]:
         raise NotImplementedError("This method is not yet implemented for database storage mode")
 
+    @override
+    def save_general_config(self, config: GeneralConfig) -> None:
```
MartinBelthle: You should remove this, I think you merged `dev` a bit too quickly

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,564 @@
...
+        return self._db_session
+
+    @abstractmethod
+    def get_impl(self) -> "DatabaseStudyDao":
+        pass
+
+    @staticmethod
+    def _convert_row_to_hydro_management(row: Row[Any]) -> HydroManagement:
+        """Convert a database row to HydroManagement model."""
+        return HydroManagement(**{c.name: row._mapping[c.name] for c in _MANAGEMENT_COLS})
```
MartinBelthle: Really minor but you should use the new method `get_row_representation_as_dict` to fit with other parts of the code

antarest/study/dao/database/database_study_dao.py
```diff
@@ -372,7 +410,7 @@ def save_hydro_correlation(self, area_id: str, correlation: HydroCorrelation) ->
         raise NotImplementedError("This method is not yet implemented for database storage mode")
 
     @override
-    def get_all_hydro_properties(self) -> Dict[str, HydroProperties]:
+    def get_all_hydro_properties(self) -> dict[str, HydroProperties]:
```
MartinBelthle: You should remove the `hydro` related methods from this file

tests/study/dao/test_database_hydro_dao.py
```diff
@@ -0,0 +1,536 @@
...
+            HydroAllocation(
+                allocation=[
+                    HydroAllocationArea(area_id="paris", coefficient=0.6),
+                    HydroAllocationArea(area_id="london", coefficient=0.4),
+                ]
+            ),
+        )
+
+        result = dao.get_hydro_allocation("paris")
+        assert len(result.allocation) == 2
```
MartinBelthle, line 219: We should check the result content

antarest/study/dao/database/database_hydro_dao.py
```diff
@@ -0,0 +1,564 @@
...
+        stmt_delete = delete(HYDRO_CORRELATION_TABLE).where(
+            (HYDRO_CORRELATION_TABLE.c.study_id == study_id)
+            & ((HYDRO_CORRELATION_TABLE.c.area_from == area_id) | (HYDRO_CORRELATION_TABLE.c.area_to == area_id))
+        )
+        session.execute(stmt_delete)
+
+        # Insert new correlations
+        insert_values = []
+        for corr_area in correlation.correlation:
+            if corr_area.area_id == area_id:
```
MartinBelthle: Minor but I think it could be simplified for readability:
```python
if corr_area.area_id == area_id or corr_area.coefficient == 0:
   continue
coefficient = corr_area.coefficient / 100
insert_values.append(
{
    "study_id": study_id,
    "area_from": area_from,
    "area_to": area_to,
    "coefficient": coefficient,
}
)
```
So both if at the same place and no need for sorting

tests/study/dao/test_database_hydro_dao.py
```diff
@@ -0,0 +1,536 @@
...
+                correlation=[
+                    HydroCorrelationArea(area_id="paris", coefficient=100.0),
+                    HydroCorrelationArea(area_id="london", coefficient=60.0),
+                ]
+            ),
+        )
+
+        result = dao.get_hydro_correlation_matrix()
+
+        assert result.data[0][0] == 1.0
```
MartinBelthle: Minor but yo ucould just do an array equality, this way it would be clearer and we won't need the `matrix_sanity_checks` as we see it visually that it's okay

MartinBelthle (changes requested):
Last round of reviews

alembic/versions/9e96cc5d95de_add_hydro_tables.py
smailio: Omg it's the third or four time with this branch 🥲

MartinBelthle (approved):

PR #2986, opened 2026-02-05
feat(ui-studies): add managed studies tree explorer

hdinia:


webapp/public/locales/en/main.json
```diff
@@ -848,6 +848,13 @@
   "studies.tree.error.failToFetchWorkspace": "Failed to load workspaces",
   "studies.tree.error.failToFetchFolder": "Failed to load subfolders for {{path}}",
   "studies.tree.unscannedStudyFolder": "Unscanned study",
+  "studies.tree.addRootDirectory": "Add root directory",
+  "studies.tree.addSubFolder": "Add subdirectory",
+  "studies.tree.noDirectories": "No directories found",
+  "studies.tree.deleteDirectory.title": "Delete directory",
+  "studies.createFolder.error": "Failed to create directory. Please try again.",
+  "studies.updateFolder.error": "Failed to update directory. Please try again.",
+  "studies.deleteFolder.error": "Failed to delete directory. Please try again.",
```
hdinia: folder -> directory

webapp/src/queries/directories/keys.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Key hierarchy:
+ * - ['directories'] - All directory-related queries
+ * - ['directories', 'list'] - All list queries
+ * - ['directories', 'list', filters] - List with specific filters
+ * - ['directories', 'detail'] - All detail queries
+ * - ['directories', 'detail', id] - Specific directory detail
+ */
+export const directoryKeys = {
+  all: ["directories"] as const,
```
skamril: It's more simple to have only functions `all: () => ["directories"],`.

webapp/src/queries/directories/keys.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import type { DirectoryFilters } from "@/services/api/directories/types";
+
+/**
+ * Query keys factory for directories
+ * Provides centralized query key management for consistent cache handling
+ *
+ * Key hierarchy:
+ * - ['directories'] - All directory-related queries
```
skamril: Not sure documenting the keys in comments is worth it. It adds maintenance overhead and can easily become outdated. Relying on the actual implementation might be safer.

webapp/src/queries/directories/mutations.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+import { mutationOptions } from "@tanstack/react-query";
+import { directoriesApi } from "@/services/api/directories";
+import type { CreateDirectoryInput, UpdateDirectoryInput } from "@/services/api/directories/types";
+
+export const directoryMutations = {
+  create: () => {
+    return mutationOptions({
+      mutationKey: ["createDirectory"],
```
skamril: We should keep using shared/global keys for directories. This makes it easier to group and find all directory-related mutations.

```typescript
export const directoryKeys = {
  // Use it for `mutationKey`
  create: () => [...directoryKeys.all(), "createDirectory"] as const, 
}
```

This enables patterns like:

```typescript
const isMutatingDirectories = useIsMutating({ mutationKey: ['directories'] })
```

webapp/src/queries/directories/queries.ts
```diff
@@ -0,0 +1,64 @@
...
+import type { DirectoryFilters } from "@/services/api/directories/types";
+import { directoryKeys } from "./keys";
+
+export const directoryQueries = {
+  /**
+   * Base key for all directory queries
+   *
+   * @returns Array of directory keys
+   */
+  all: () => directoryKeys.all,
```
skamril: This file should focus on queries only. For getting keys only, using `directoryKeys` directly should be sufficient.

webapp/src/queries/directories/queries.ts
```diff
@@ -0,0 +1,64 @@
...
+   * Note: Filters from getAll() since no GET by ID endpoint exists
+   *
+   * @param directoryId - ID of the directory to fetch
+   * @returns Query options object
+   */
+  detail: (directoryId: string) => {
+    return queryOptions({
+      queryKey: directoryKeys.detail(directoryId),
+      queryFn: async () => {
+        const directories = await directoriesApi.getAll();
```
skamril: This bypasses the TanStack Query cache by calling `directoriesApi.getAll()` directly here.
Instead, we could create a dedicated hook (e.g. `useDirectory`) that calls `useSuspenseQuery` and derives the directory using the `select` option.

skamril (changes requested):

webapp/src/queries/explorer/keys.ts
```diff
@@ -0,0 +1,31 @@
...
+ * - ['explorer'] - All explorer-related queries
+ * - ['explorer', 'workspaces'] - All workspace queries
+ * - ['explorer', 'folders'] - All folder queries
+ * - ['explorer', 'folders', workspace, path] - Folders for specific workspace and path
+ */
+export const explorerKeys = {
+  all: ["explorer"] as const,
+  workspaces: () => [...explorerKeys.all, "workspaces"] as const,
+  folders: () => [...explorerKeys.all, "folders"] as const,
+  foldersByPath: (workspace: string, path: string) =>
```
skamril:
```ts
folderDetail: (workspace: string, path: string) => [...explorerKeys.folders(), "detail", { workspace, path }] as const;
```

It might be easier to work with an object here, as TanStack Query hashes query keys deterministically: https://tanstack.com/query/v5/docs/framework/react/guides/query-keys#query-keys-are-hashed-deterministically

We could also rename `foldersByPath` to `folderDetail` and add the "detail" segment to keep the factory naming consistent with the other key factories.

skamril (changes requested):

webapp/src/queries/directories/queries.ts
hdinia: Maybe it was there before refactoring i will check it thanks

webapp/src/queries/directories/keys.ts
hdinia: Thanks for the remark, but their is a reason for this: the other keys are functions because they compose/extend the base key (creating new arrays via spread). The root doesn't need that, so using a plain value both signal this intent and avoids unnecessary allocations (each call to the function creates a new array instance)
https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories

skamril: Thanks for the clarification. The allocation cost is negligible in practice, while not using a function for the root key makes the distinction between keys less clear. Keeping everything as functions ensures a more consistent and easier-to-understand API.

webapp/src/services/api/directories/types.ts
```diff
@@ -0,0 +1,30 @@
...
+import type {
+  createDirectoryInputSchema,
+  directoriesListResponseSchema,
+  directoryFiltersSchema,
+  directorySchema,
+  updateDirectoryInputSchema,
+  updateDirectoryResponseSchema,
+} from "./schemas";
+
+export type Directory = z.infer<typeof directorySchema>;
```
skamril, line 22: Keeps internal convention for api files please:
- 'DTO' suffix for orignal API response.
- 'Params' suffix instead if 'Input'.
- 'Response' not needed.

webapp/src/queries/directories/hooks.ts
```diff
@@ -0,0 +1,31 @@
+/**
```
skamril: Wrong place, but can be removed it's not used.

webapp/src/routes/_authenticated/studies/$studyId/route.tsx
```diff
@@ -34,6 +35,9 @@ import NavHeader from "./-components/NavHeader";
 
 export const Route = createFileRoute("/_authenticated/studies/$studyId")({
   component: StudyHomeLayout,
+  loader: ({ context }) => {
+    return context.queryClient.ensureQueryData(directoryQueries.list());
```
skamril: Please do not return the `context.queryClient.ensureQueryData()` call, as it can cause issues (preventing the use of `useLoaderData()`). Instead, use `await`. Returning it also makes the purpose of the call less clear.

webapp/src/routes/_authenticated/studies/-components/StudyTree/ManagedTree/hooks/useCreateDirectory.ts
```diff
@@ -0,0 +1,95 @@
+/**
```
skamril, line 1:
```typescript
interface UseCreateDirectoryOptions {
  onSuccess?: (directory: Directory) => void;
}

export function useCreateDirectory(options?: UseCreateDirectoryOptions) {
  const queryClient = useQueryClient();

  const { queryKey: listQueryKey } = directoryQueries.list();

  return useMutation({
    ...directoryMutations.create(),
    onMutate: async (newDirValues) => {
      const tempId = crypto.randomUUID();

      await queryClient.cancelQueries({ queryKey: listQueryKey });

      queryClient.setQueryData(listQueryKey, (old = []) => {
        return [
          ...old,
          {
            ...newDirValues,
            id: tempId,
          },
        ];
      });

      return { tempId };
    },
    onError: (error, _variables, context) => {
      queryClient.setQueryData(listQueryKey, (old = []) => {
        return old.filter((dir) => dir.id !== context?.tempId);
      });

      enqueueErrorSnackbar(t("studies.createDirectory.error"), toError(error));
    },
    onSuccess: (newDirectory, _variables, context) => {
      queryClient.setQueryData(listQueryKey, (old) => {
        if (!old) {
          return newDirectory ? [newDirectory] : [];
        }
        return old.map((dir) => (dir.id === context.tempId ? newDirectory : dir));
      });

      options?.onSuccess?.(newDirectory);
    },
  });
}
```

You don't use `directoryKeys.detail(data.id)` for queries.

webapp/src/routes/_authenticated/studies/$studyId/route.tsx
hdinia: The return is required here

TanStack Router automatically handles promises returned from loaders it will wait for the promise to resolve before rendering the component regardless of whether you use `await`

WHen you want `await`
1. If you need try/catch error handling within the loader
2. If you need to do something with the result before returning it
3. If you have multiple sequential async calls that depend on each other

Since your loader just returns the promise directly without any additional processing, the current code is more concise and idiomatic.

The loader currently returns the result of `ensureQueryData`:

```ts
export const Route = createFileRoute("/_authenticated/studies/")({
  component: Studies,
  loader: ({ context }) => {
    return context.queryClient.ensureQueryData(directoryQueries.list());
  },
});
```

And the returned data is actively consumed via `useLoaderData()` in at least two places:

`useBreadcrumbs.ts` and `Breadcrumb/index.tsx`

webapp/src/queries/directories/hooks.ts
hdinia: Next time be more specific on the location cause you did ask to add this hook in a prior comment.. But it was indeed useless so I removed it

webapp/src/services/api/directories/types.ts
hdinia: DTO -> not to be used everywhere, only when data model transformations are needed. In the UI layer we should not see any type named with "DTO" suffix, an error largely made across all the codebase

webapp/src/routes/_authenticated/studies/$studyId/route.tsx
skamril: `useLoaderData()` will not subscribe to query cache changes, unlike `useSuspenseQuery()`.
This is why I said that returning the data in the loader when using `queryClient.ensureQueryData()` can cause issues.

```typescript
export const Route = createFileRoute("/_authenticated/studies/")({
  component: Studies,
  loader: async ({ context }) => {
    await context.queryClient.ensureQueryData(directoryQueries.list());
  },
});
```

With `await` we prevent the using of `useLoaderData()`.

webapp/src/queries/directories/keys.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+export const directoryKeys = {
+  all: ["directories"] as const,
+  lists: () => [...directoryKeys.all, "list"] as const,
+  details: () => [...directoryKeys.all, "detail"] as const,
```
skamril: `details` and `detail` can be removed, they are not used.

skamril (approved):

PR #2987, opened 2026-02-05
refactor(archive): improve perfs

TheoPascoli:
## Performance comparison (8 GB study)

Archiving / unarchiving:

Local:
- Archive: current 26 min → with improvements 5 min

Integration server:
- Archive: current 41 min → with improvements 15 min
- Unarchive: current 54 s → with improvements 36 s
- Import: identical for both methods

The archive size is identical with both methods.


antarest/core/utils/archives.py
```diff
@@ -45,8 +46,20 @@ def archive_dir(
             f"Non matching archive format {archive_format} and target archive suffix {target_archive_path.suffix}"
         )
     if target_archive_path.suffix == ArchiveFormat.SEVEN_ZIP:
-        with py7zr.SevenZipFile(target_archive_path, mode="w") as szf:
-            szf.writeall(src_dir_path, arcname="")
+        # if 7z is available on the machine, uses it
+        if shutil.which("7z") is not None:
```
sylvlecl: We should try to do that detection only once, not every time the functions are called.
We could have a module level flag which is initialized on first call

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -368,16 +368,33 @@
...
             study: The study to be unarchived.
 
         Raises:
             BadArchiveContent: If the archive is corrupted or in an unknown format.
         """
-        with open(self.find_archive_path(study), mode="rb") as fh:
-            self.import_study(study, fh)
+        archive_path = self.find_archive_path(study)
+        study_path = Path(study.path)
+        study_path.mkdir()
```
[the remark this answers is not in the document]
TheoPascoli: :rage:

Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM python:3.11-slim-bullseye
 
 # RUN apt update && apt install -y procps gdb
+RUN dnf install -y p7zip p7zip-plugins && dnf clean all
```
sylvlecl: For that Dockerfile, we need to use `apt`, `dnf` will not work.
We should keep dnf for our internal Dockerfile

antarest/core/utils/archives.py
```diff
@@ -104,22 +165,20 @@ def extract_archive(stream: BinaryIO, target_dir: Path) -> None:
...
-        except py7zr.exceptions.Bad7zFile as error:
-            raise BadArchiveContent("Unsupported 7z format") from error
-
+        suffix = ArchiveFormat.SEVEN_ZIP
     else:
         raise BadArchiveContent
 
+    with tempfile.NamedTemporaryFile(suffix=suffix, delete=False) as tmp:
+        tmp_path = Path(tmp.name)
+        shutil.copyfileobj(stream, tmp)
```
sylvlecl: Not sure we want to write it to a temp directory, this adds a copy. For ex. when using the "import output" API, the file will likely already be on disk elsewhere, it's sad to copy it again before unzipping it.

The best would probably be to either:
- never receive a stream in the first place but a path to the actual patfileh, but it may not be possible
- provide the stream to the underlying process in its standard output, if 7z can handle that, but it starts to get more difficult

If all that is no possible, why not keep the copy, but is it really useful to do it also for zips ?

antarest/core/utils/archives.py
```diff
@@ -60,9 +73,54 @@ def archive_dir(
...
+                raise BadArchiveContent("Unsupported ZIP format") from e
+
+
 def unzip(dir_path: Path, zip_path: Path) -> None:
-    with zipfile.ZipFile(zip_path, mode="r") as zipf:
-        zipf.extractall(dir_path)
+    if shutil.which("7z") is not None:
+        logger.info("Using 7z CLI to unzip %s", zip_path)
+        try:
+            run(["7z", "x", str(zip_path), f"-o{dir_path}", "-y"], check=True)
```
sylvlecl: Do you know if we have also performance improvements for zips ?

antarest/core/utils/archives.py
```diff
@@ -104,22 +165,20 @@ def extract_archive(stream: BinaryIO, target_dir: Path) -> None:
...
-
+        suffix = ArchiveFormat.SEVEN_ZIP
     else:
         raise BadArchiveContent
 
+    with tempfile.NamedTemporaryFile(suffix=suffix, delete=False) as tmp:
+        tmp_path = Path(tmp.name)
+        shutil.copyfileobj(stream, tmp)
+    try:
+        extract_archive_from_path(tmp_path, target_dir)
```
sylvlecl: Can this work ? isn't `tmp_path` removed when exiting the `with` statement ?

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -368,16 +368,36 @@ def archive(self, study: RawStudy) -> Path:
...
+        try:
+            extract_archive_from_path(archive_path, study_path)
+            fix_study_root(study_path)
+            self.update_from_raw_meta(study, study_path=study_path)
+        except Exception:
+            shutil.rmtree(study_path)
+            raise
+
+        try:
+            self.checks_antares_web_compatibility(study)
```
sylvlecl: It's quite a lot of logic duplication with `import_study` we should try to keep the code shared, otherwise it will be hard to maintain

sylvlecl (changes requested):
A few comments, great perf improvement !

antarest/core/utils/archives.py
TheoPascoli: No improvements for zips

antarest/core/utils/archives.py
TheoPascoli: no because I set it with the flag `delete=False` in the initialisation

antarest/core/utils/archives.py
TheoPascoli: Now I handle directly the stream without copy

antarest/study/storage/rawstudy/raw_study_service.py
TheoPascoli: I extracted a method with a test on the file type: Path or BinaryIO

antarest/core/utils/archives.py
TheoPascoli: Introduced a `@functools.cache`

antarest/core/utils/archives.py
```diff
@@ -25,6 +27,11 @@
 logger = logging.getLogger(__name__)
 
 
+@functools.cache
```
sylvlecl, line 30: Nice :)

antarest/core/utils/archives.py
```diff
@@ -105,18 +166,18 @@ def extract_archive(stream: BinaryIO, target_dir: Path) -> None:
...
-            raise BadArchiveContent("Unsupported ZIP format") from error
-
+            with zipfile.ZipFile(stream, mode="r") as zf:
+                zf.extractall(target_dir)
+        except zipfile.BadZipFile as e:
+            raise BadArchiveContent("Unsupported ZIP format") from e
     elif file_format[:2] == b"7z":
+        with tempfile.NamedTemporaryFile(suffix=ArchiveFormat.SEVEN_ZIP, delete=False) as tmp:
+            tmp_path = Path(tmp.name)
+            shutil.copyfileobj(stream, tmp)
```
sylvlecl, line 176: We still perform an additional copy in that case.
It's probably worth it, since the extraction is much faster.
However, we should probably use the temp directory which is configured at the application level instead of a default tmpr dir, for consistency and because archives can be quite large (although compressed :) )

antarest/core/utils/archives.py
sylvlecl, line 30: Nice :)

Dockerfile
sylvlecl: we still need to change it

antarest/core/utils/archives.py
TheoPascoli: Done

antarest/study/output/file_output_storage.py
```diff
@@ -117,7 +117,7 @@ def import_output(
                     path_output = Path(str(path_output) + f"{ArchiveFormat.ZIP}")
                     shutil.copyfile(output, path_output)
             else:
-                extract_archive(output, path_output)
+                extract_archive_from_stream(output, path_output)
```
sylvlecl: We need to use the tmp_dir from storage config here, otherwise for large outputs we'll fill up `/tmp`.
We should add the tmp_dir as a constructor argument of the service, since it's now a dependency of the service

sylvlecl (changes requested):
Last request: we should use the configured tmp_dir also when extracting outputs

sylvlecl (approved):

PR #2988, opened 2026-02-06
feat(ui-output): add export variable view button

smailio:
ANT-4299

Add this export button 

image


webapp/src/components/buttons/DownloadMatrixButton.tsx
```diff
@@ -17,16 +17,12 @@ import { downloadFile } from "@/utils/fileUtils";
...
-import type { Options } from "./SplitButton";
-
-type ExportFormat =
-  | "csv"
-  | "csv (semicolon)"
-  | "csv (header)"
-  | "csv (semicolon header)"
-  | "xlsx"
-  | "raw";
+import {
```
smailio, line 20: This is refactoring. I centralized the button options code in utils/buttonOption.ts

webapp/src/components/buttons/DownloadMatrixButton.tsx
```diff
@@ -74,7 +44,7 @@ function DownloadMatrixButton(props: DownloadMatrixButtonProps) {
   // Event Handlers
   ////////////////////////////////////////////////////////////////
 
-  const handleDownload = async (format: ExportFormat) => {
+  const handleDownload = async (format: ExportFormat | "raw") => {
```
smailio, line 47: All the format in ExportFormat are common to both button, but matrix export also handle the raw format (which isn't supported by the variable view export as there's no raw file, the views are computed by the AW backend, not outputed by the simulator). Anyway, I though it'd be too much to add a second type and there's too much common option to make two separates types. Let me know if you disagree.

webapp/src/components/buttons/SplitButton.tsx
```diff
@@ -25,19 +25,9 @@ import {
...
-interface Option<T extends string = string> {
-  value: T;
-  label?: string | ((t: TFunction) => string);
-  disabled?: boolean;
-}
-
-export type Options<T extends string = string> =
-  | Array<T | Option<T>>
-  | ReadonlyArray<T | Option<T>>;
+import type { Options } from "../utils/buttonOptions";
```
smailio: Other button used to import this type from this component. There's no reason for SplitButton to centralize all typing code related to options, so I moved this in utils/buttonOptions.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-components/ResultMatrixViewer/index.tsx
```diff
@@ -131,17 +133,22 @@ function ResultMatrixViewer({
...
             onToggleFilter={onToggleFilter}
             variablesMetadata={variablesMetadata}
             itemType={itemType}
             selectedItemId={selectedItemId}
+            selectedItem={selectedItem}
             selectedVariable={selectedVariable}
             onVariableSelect={onVariableSelect}
             selectedClusterId={selectedClusterId}
             onClusterSelect={onClusterSelect}
+            canExportVariableView={
```
smailio, line 149: We don't show the variable view button if we didn't finish retrieving variable view data. If the data is empty it means that the user didn't request yet for the computing of the  variable view, or that the compute didn't complete yet or that it failed, either way we don't want to enable the button untill we know there's something to be exported.

webapp/src/components/buttons/SplitButton.tsx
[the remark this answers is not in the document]
smailio: No it doesn't

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-components/ResultFilters/index.tsx
```diff
@@ -399,8 +408,24 @@ function ResultFilters({
                 <FilterListIcon />
               </IconButton>
             </Tooltip>
-            {/* TODO: Export functionality for variable per variable mode will be implemented later */}
-            {!isVariablePerVariable && <DownloadMatrixButton studyId={studyId} path={path} />}
+            {isVariablePerVariable && selectedItem ? (
```
smailio, line 411: Look a few line bellow, buildVariableViewParams doesn't accept an undefined selectedItem, so I need to check selectedItem. 

I feel like I just added this check to made the code work and I'm not sure this is the best design. Let me know.

webapp/src/components/buttons/SplitButton.tsx
skamril: These components rely on `<SplitButton />` as the base, so `Options` is part of its public API. I’d prefer keeping the type next to SplitButton (source of truth close to the contract) rather than moving it to utils/*.

skamril (changes requested):

skamril (approved):

PR #2990, opened 2026-02-10
fix(ci): fix broken deps

TheoPascoli:
## Why it broke

- `checksumdir` depends on `pkg_resources`
- `pkg_resources` was removed from `setuptools >= 78`
- CI installs `setuptools 82` → `ModuleNotFoundError`

## Fix

- Removed `checksumdir`
- Replaced with a small `dirhash` helper using only stdlib (`hashlib`, `os`)

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,5 +1,4 @@
 -r requirements.txt
-checksumdir~=1.2.0
 pytest~=8.3.0
 pytest-xdist~=3.6.0
 pytest-cov~=4.0.0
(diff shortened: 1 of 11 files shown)


smailio (approved):
LGTM 

There's no actively maintained and popular alternative to checksumdir. Plus, it wouldn't be worth adding a dependency that would eventually break just for a 20-line function, so it makes sense to implement it here. 

No additional tests are required for dirhash as it's already tested by existing code.

PR #2995, opened 2026-02-11
fix(ui-study): error displayed when opening the study home directly

skamril:


webapp/src/routes/_authenticated/studies/$studyId/-components/NavHeader/index.tsx
```diff
@@ -12,37 +12,69 @@
...
+    const parentStudy = parents.length > 0 ? parents[0] : undefined;
+
+    const root = parents.length > 0 ? parents[parents.length - 1] : study;
+    const variantTree = await getVariantTree(root.id);
+
+    const tree = findNodeInTree(study.id, variantTree);
+    const variantNb = tree ? countDescendants(tree) : 0;
+
+    return { parentStudy, variantNb };
+  }, [study]);
```
hdinia, line 45: `NavHeader` and `StudyHome` both fetch variant metadata independently 
-> 2+ duplicate API calls per page load
-> Fetching logic duplicated 

Suggested fix: add Tstack query cache (like `directoryQueries`):

Files to create:
- `queries/variants/keys.ts` - Query key factory for variants
- `queries/variants/queries.ts` - `variantQueries.metadata(studyId)` that fetches parents + tree + computes variant count

Files to modify:
- `NavHeader/index.tsx` - Replace `usePromise` with the query
- `index.tsx` - Replace `usePromise` with same query
- `route.tsx` - No changes needed (don't prefetch in loader to avoid Redux issues)

-> Single API call, shared cache, proper invalidation on WebSocket events.

webapp/src/routes/_authenticated/studies/$studyId/route.tsx
```diff
@@ -79,67 +54,28 @@ function StudyHomeLayout() {
...
   ////////////////////////////////////////////////////////////////
 
+  // Status must be "succeeded" to ensure that the study is loaded in the Redux store
+  if (studiesStatus === "idle" || studiesStatus === "loading") {
+    return <SimpleLoader />;
+  }
+
+  if (studiesStatus === "failed") {
+    throw new Error(t("studies.error.loadStudies"));
+  }
```
hdinia, line 72: When an error occurs the boundary shows it, but their is no snackbar and no action possible on the error screen, is that wanted ?

hdinia: Image

hdinia (changes requested):

webapp/src/routes/_authenticated/studies/$studyId/-components/NavHeader/index.tsx
skamril: I agree with the principle: sharing the query through TanStack Query caching would indeed avoid duplicated API calls and improve cache management.

However, this goes beyond the scope of the current ticket. The current behavior relies on already existing code, and the proposed change fits more into a broader effort to improve our API querying layer and progressively adopt TanStack Query.

Additionally, introducing TanStack Query here would also require handling proper cache invalidation when variants are created or deleted. These operations are currently managed through Redux, so integrating TanStack Query would require additional work to ensure proper synchronization between Redux updates and the query cache.

I suggest addressing this improvement in a dedicated ticket so it can be implemented properly and consistently with our overall TanStack Query strategy.

hdinia (approved):
Approving with reservation: while the implementation is correct, the duplicate API calls between NavHeader and StudyHome remain unaddressed.
c.f. https://github.com/AntaresSimulatorTeam/AntaREST/pull/2995#discussion_r2798811626

PR #2998, opened 2026-02-12
feat(ui-studies): add studies batch delete

hdinia:


antarest/study/web/studies_blueprint.py
```diff
@@ -424,6 +424,20 @@ def delete_study(uuid: str, children: bool = False) -> None:
...
+        summary="Delete Multiple Studies",
+    )
+    def delete_studies(
+        study_ids: list[str] = Body(description="List of study UUIDs to delete"),
+        children: bool = False,
+    ) -> None:
+        logger.info(f"Deleting multiple studies: {study_ids}")
+        sanitized_ids = [sanitize_uuid(sid) for sid in study_ids]
+        for uuid in sanitized_ids:
+            study_service.delete_study(uuid, children)
```
smailio: if one delete fail you stop everything, this is what you want ? 

Also through this endpoint an admin may delete the whole base. Maybe we can add a safeguard ? Like a limit ?

webapp/src/services/api/study.ts
```diff
@@ -192,8 +192,15 @@ export const upgradeStudy = async (studyId: string, targetVersion: string): Prom
...
 };
 
-export const deleteStudy = async (sid: string, deleteAllChildren?: boolean): Promise<void> => {
-  const res = await client.delete(`/v1/studies/${sid}?children=${deleteAllChildren || false}`);
+export const deleteStudy = async (studyId: string, deleteChildren = false): Promise<void> => {
+  const res = await client.delete(`/v1/studies/${studyId}?children=${deleteChildren}`);
+  return res.data;
+};
+
+export const deleteStudies = async (studyIds: string[], deleteChildren = false): Promise<void> => {
```
smailio: deleteChildren is always true, would it even make sense to pass False here someday ?

antarest/study/web/studies_blueprint.py
hdinia: yep I was initially thinking of stopping at the first error, but we could provide more granular feedback by collecting all validation failures and returning them as a batch. open to suggestions on this

the security concerns are valid yes @MartinBelthle , what do you think ? the permissions should mitigate this no ?

MartinBelthle: We should first ensure that the user has permissions on all studies before deleting them. For that we should introduce a method `delete_studies` in the study service that will first check all permissions and then remove them in one DB query (for that we should also add a method in the study repository)

MartinBelthle (changes requested):

antarest/study/web/studies_blueprint.py
```diff
@@ -424,6 +424,20 @@ def delete_study(uuid: str, children: bool = False) -> None:
 
         study_service.delete_study(uuid_sanitized, children)
 
+    @bp.delete(
+        "/studies",
+        status_code=HTTPStatus.OK,
+        summary="Delete Multiple Studies",
+    )
+    def delete_studies(
+        study_ids: list[str] = Body(description="List of study UUIDs to delete"),
```
sylvlecl: I would advise do have a nested model here, in case we want to add other arguments later, something like:
```
{
    "study_ids": ["id1", "id2"]
}
```

This will allow to delete on criteria other than the ID later if we want to.

antarest/study/service.py
```diff
@@ -1433,6 +1434,114 @@ def delete_study(self, uuid: str, children: bool) -> None:
...
+            StudyDeletionNotAllowed: If a study has variants and with_variants=False
+            HTTPException(BAD_REQUEST): If study_ids is empty or exceeds MAX_BATCH_DELETE_SIZE
+        """
+        if not study_ids:
+            raise HTTPException(
+                status_code=http.HTTPStatus.BAD_REQUEST,
+                detail="study_ids list cannot be empty",
+            )
+
+        if len(study_ids) > MAX_BATCH_DELETE_SIZE:
```
hdinia, line 1453: Mainly added as a safeguard against large accidental/malicious deletions. Do you think this limit is actually necessary? @sylvlecl @MartinBelthle 

also since Variant expansion could potentially push the resolved study_ids beyond the batch limit this might behave unexpectedly.

would a rate limiter (X delete requests/hour) make more sense or be complementary? Considering we have DB backups + permissions, I’m wondering whether this extra restriction is needed

antarest/study/model.py
```diff
@@ -553,6 +553,13 @@ def _normalize_tags(cls, v: List[str]) -> List[str]:
         return tags
 
 
+class DeleteManyStudies(AntaresBaseModel):
+    model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True)
+
+    study_ids: List[str] = Field(..., description="List of study UUIDs to delete")
+    with_variants: bool = Field(default=False, description="Whether to delete variant studies as well")
```
hdinia, line 560: as mentioned by @smailio , the frontend always sends true for this flag, and the UI doesn’t allow users to control it. In practice, deleting a study also deletes its variants.

given that, is this flag still meaningful in the API contract? Should we consider removing it and always enforcing true on the backend instead?

antarest/study/service.py
```diff
@@ -1433,6 +1434,114 @@ def delete_study(self, uuid: str, children: bool) -> None:
...
+
+        studies_to_delete: Dict[str, Study] = {}
+
+        for study in studies_to_check:
+            if study.id in studies_to_delete:
+                continue
+
+            if variant_service.has_children(study):
+                if not with_variants:
+                    raise StudyDeletionNotAllowed(study.id, "Study has variant children")
```
hdinia: If a user has permission to delete study A, should they automatically be allowed to delete all its variants, or should each variant require separate permission checks?

antarest/study/web/studies_blueprint.py
hdinia: it's done

webapp/src/services/api/study.ts
hdinia: yep, https://github.com/AntaresSimulatorTeam/AntaREST/pull/2998#discussion_r2817331883

antarest/study/service.py
sylvlecl: Indeed variants may have different permissions than their parent, so we need to check all children.

antarest/study/service.py
sylvlecl: No need for a rate limiter indeed.

Not sure about the usefulness of the max delete either to be honest, but it does not hurt. Indeed when deleting from the frontend 100 should be enough

webapp/public/locales/en/main.json
```diff
@@ -786,6 +786,7 @@
   "study.area.storage.additionalConstraint.notFound": "No additional constraint found with ID: {{id}}",
   "studies.question.delete": "Are you sure you want to delete the study: {{studyName}}?",
   "studies.question.deleteWithVariants": "Are you sure you want to delete the study '{{studyName}}' and its {{variantNb}} sub-variant(s)?",
+  "studies.question.deleteMany": "Are you sure you want to delete {{count}} study(ies)?",
```
skamril:
```
"studies.question.deleteMany_one": "Are you sure you want to delete the selected study?",
"studies.question.deleteMany_other": "Are you sure you want to delete the selected {{count}} studies?",
```

Then i18n will use the appropriate translation based on `count`.

```ts
t("studies.question.deleteMany", { count: studyIds.length }):
```

antarest/study/service.py
```diff
@@ -1433,6 +1434,114 @@ def delete_study(self, uuid: str, children: bool) -> None:
...
+                    payload=study_info,
+                    permissions=PermissionInfo.from_study(study),
+                )
+            )
+
+            self._on_study_delete(uuid=study.id)
+
+        # Commit the transaction after all filesystem deletions and events succeed.
+        # `delete_many` only flushes changes we commit here to finalize the transaction.
+        self.repository.session.commit()
```
sylvlecl: Not sure it's a good idea to perform the commit only at the end:
if any exception occurs during the deletion of actual study data, the studies will be seen as still existing but the data will still have been removed. The likeliness of such thing happening is increased by the fact that we are removing multiple studies, which could be a little long.

--> would probably be best to keep the behaviour of the single delete for each study (that implies to start with leafs, though)

On the long run, it will probably be a better implementation to just flag the studies as deleted, and having a background job delete the actual data, but that would be another work.

webapp/public/locales/en/main.json
hdinia: todo: update keys to use pluralization

webapp/src/routes/_authenticated/studies/-components/StudiesList/Header/BatchActions.tsx
```diff
@@ -40,6 +40,12 @@ function BatchActions({ selectedCount, onLaunch, onDeselectAll }: BatchActionsPr
           {t("global.launch")} ({selectedCount})
         </Button>
       </Tooltip>
+      <Tooltip title={t("global.delete")}>
+        <Button onClick={onDelete} color="error">
+          <DeleteIcon />
```
skamril: You have `startIcon` for the button icon.

I think we should keep the selected count only for the checkbox button:

Image

```jsx
    <>
      <Tooltip title={t("studies.batchMode")}>
        <Button onClick={onLaunch} color="primary" startIcon={<BoltIcon />}>
          {t("global.launch")}
        </Button>
      </Tooltip>
      <Tooltip title={t("global.delete")}>
        <Button onClick={onDelete} color="error" startIcon={<DeleteIcon />}>
          {t("global.delete")}
        </Button>
      </Tooltip>
      <Tooltip title={t("studies.deselectAll")}>
        <Stack>
          <IconButton color="primary" onClick={onDeselectAll}>
            <CheckBoxIcon />
          </IconButton>
          <Typography variant="body2">({selectedCount})</Typography>
        </Stack>
      </Tooltip>
    </>
```

webapp/public/locales/en/main.json
sylvlecl: From what I understand, this will also delete variants:
it could be misleading to the user.
We should probably at least mention that this will also delete variants, in addition to that count of studies.

webapp/src/routes/_authenticated/studies/-components/StudiesList/Header/hooks/useDeleteStudies.ts
```diff
@@ -0,0 +1,38 @@
+/**
```
skamril, line 1: No hooks folder?

antarest/study/service.py
```diff
@@ -1433,6 +1434,114 @@ def delete_study(self, uuid: str, children: bool) -> None:
...
+            if isinstance(study, RawStudy):
+                study_info = study.to_enhanced_json_summary()
+
+            if self.assert_study_unarchived(study=study, raise_exception=False):
+                self.storage_service.get_storage(study).delete(study)
+            else:
+                if isinstance(study, RawStudy):
+                    os.unlink(self.storage_service.raw_study_service.find_archive_path(study))
+
+            self.event_bus.push(
```
sylvlecl: we should probably not send the event before the change is committed, otherwise there will be a contradiction: the study will still be seen as existing while we have said it has been removed

antarest/study/service.py
```diff
@@ -1433,6 +1434,114 @@ def delete_study(self, uuid: str, children: bool) -> None:
...
+
+        for task_id in generation_tasks:
+            self.task_service.await_task(task_id, 600)
+
+        study_ids_to_delete = list(studies_to_delete.keys())
+
+        self.repository.delete_many(study_ids_to_delete)
+
+        for study in studies_to_delete.values():
+            study_info = study.to_json_summary()
```
sylvlecl: There is quite a lot of logic duplication between here and single delete, is it possible to share some ?
Possibly just having the single delete call the multiple delete ?

antarest/study/web/studies_blueprint.py
```diff
@@ -424,6 +425,16 @@ def delete_study(uuid: str, children: bool = False) -> None:
 
         study_service.delete_study(uuid_sanitized, children)
 
+    @bp.delete(
```
sylvlecl, line 428: We need at least some integration tests to cover that endpoint.

In particular we need to cover the error cases:

- no permission on one of the studies
- no permission on one of the variants
- one of the ID is not valid
- some IDs are duplicate, or one of the ID is a variant of another ID
- ...

antarest/study/model.py
sylvlecl: I think it's a good idead to keep it, because it will force the user of the API to think about what he wants to do.

I we just document it through a string in the API doc, it will be really not obvious that all children are deleted.

antarest/study/repository.py
```diff
@@ -394,6 +394,12 @@ def delete(self, id_: str, *ids: str) -> None:
         session.execute(stmt)
         session.commit()
 
+    def delete_many(self, ids: List[str]) -> None:
```
sylvlecl: We kind of break the implicit convention that the repository is responsible for commits here, I think I understand that you want to commit all the changes of the request at once, but I don't think it's worth it to change that convention just for that use case. Also related to the comment about committing after having deleted actual data.

Overall it would probably be simpler to just extend the behaviour of single commit:
- delete row and commit it
- delete data
- go on with next study ...

sylvlecl (changes requested):
Looks pretty good:
for the backend part we need to add some tests though.
Also, I think we'd better align the synchronization of data deletion and commits with the single delete implementation

webapp/src/routes/_authenticated/studies/-components/StudiesList/Header/DeleteStudiesDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+import type { StudyMetadata } from "@/types/types";
+import { useDeleteStudies } from "./useDeleteStudies";
+
+interface Props {
+  studyIds: Array<StudyMetadata["id"]>;
+  open: boolean;
+  onClose: VoidFunction;
+}
+
+function DeleteStudiesDialog({ studyIds, open, onClose }: Props) {
```
skamril, line 28: `useDialog()` hook can be used in Header component instead:

```ts
  const { confirm } = useDialog();

  const handleDeleteStudies = async () => {
    const isConfirmed = await confirm({
      content: t("studies.question.deleteMany", { count: selectedStudyIds.length }),
      alert: "error",
      titleIcon: DeleteIcon,
    });

    if (isConfirmed) {
      setSelectedStudyIds([]); // Reset selection
      deleteStudies.mutate({ studyIds: selectedStudyIds });
    }
  };
```

webapp/src/routes/_authenticated/studies/-components/StudiesList/Header/useDeleteStudies.ts
```diff
@@ -0,0 +1,38 @@
...
+export function useDeleteStudies(options?: UseDeleteStudiesOptions) {
+  const enqueueErrorSnackbar = useEnqueueErrorSnackbar();
+  const { t } = useTranslation();
+
+  return useMutation({
+    ...studyMutations.deleteMany(),
+    onError: (error) => {
+      enqueueErrorSnackbar(t("studies.error.deleteStudies"), toError(error));
+    },
+    onSuccess: () => {
```
skamril: Please update the Redux store here as well; relying only on the WebSocket update is not reliable.
The WebSocket event may be delayed, lost, or not received in some cases.

skamril (changes requested):

webapp/public/locales/en/main.json
hdinia: c.f. https://github.com/AntaresSimulatorTeam/AntaREST/pull/2998#discussion_r2817742334

hdinia: yes I plan to do it at the same time as https://github.com/AntaresSimulatorTeam/AntaREST/pull/2998#discussion_r2817742334

antarest/study/service.py
hdinia: surely we can avoid some duplications, I forgot to refactor this thanks for the feedback

antarest/study/web/studies_blueprint.py
hdinia: not really familiar with the back tests setup but I'll try my best

antarest/study/model.py
hdinia: Alright I see. Would it make sense to let users control this via an option in the UI then?

webapp/src/routes/_authenticated/studies/-components/StudiesList/Header/useDeleteStudies.ts
hdinia: Not sure about that but I will check it thanks for the feedback

antarest/study/service.py
hdinia: My intent was to defer the commit to avoid the "DB committed but filesystem deletion fails" scenario
It also seems to me that this failure mode is better whereas deferring the commit creates the opposite and worse problem: filesystem data gone but studies still visible in DB
Will align with the single delete behavior thanks for catching this

hdinia: >On the long run, it will probably be a better implementation to just flag the studies as deleted, and having a background job delete the actual data, but that would be another work.

yeah that would be best

webapp/src/routes/_authenticated/studies/-components/StudiesList/Header/useDeleteStudies.ts
hdinia: Good catch it was indeed needed, see : https://github.com/AntaresSimulatorTeam/AntaREST/pull/2998/changes/9485d1b9cbd07f3c75e510e107c8484435b8181e

skamril (dismissed):

antarest/study/service.py
```diff
@@ -1409,29 +1445,86 @@ def delete_study(self, uuid: str, children: bool) -> None:
...
+                studies_to_delete[study.id] = study
+
+        self._await_generation_tasks(studies_to_delete.values())
+        self.repository.delete(*studies_to_delete.keys())
+
+        for study in studies_to_delete.values():
+            self._delete_study_from_filesystem(study)
+
+        for study in studies_to_delete.values():
+            self._notify_study_deleted(study, study_infos[study.id])
```
sylvlecl, line 1504: would make more sens to have it int the loop above, notifying for each study as soon as it's actually deleted

antarest/study/service.py
```diff
@@ -1384,22 +1430,12 @@ def delete_study(self, uuid: str, children: bool) -> None:
...
-        study_info = study.to_json_summary()
-
-        # this prefetch the workspace because it is lazy loaded and the object is deleted
-        # before using workspace attribute in raw study deletion
-        # see https://github.com/AntaresSimulatorTeam/AntaREST/issues/606
-        if isinstance(study, RawStudy):
-            _ = study.workspace
-            study_info = study.to_enhanced_json_summary()
-
+        study_info = self._prefetch_study_info(study)
```
sylvlecl: Why not just call the multi-delete ? Is the behaviour different?

antarest/study/service.py
```diff
@@ -1409,29 +1445,86 @@ def delete_study(self, uuid: str, children: bool) -> None:
...
+        if len(study_ids) > MAX_BATCH_DELETE_SIZE:
+            raise HTTPException(
+                status_code=http.HTTPStatus.BAD_REQUEST,
+                detail=f"Cannot delete more than {MAX_BATCH_DELETE_SIZE} studies at once. Got {len(study_ids)} studies.",
+            )
+
+        logger.warning(
+            "Batch delete requested for %d studies by user %s (with_variants=%s)",
+            len(study_ids),
+            get_user_id(),
```
sylvlecl: You don't need to include the user ID in the log, it's already logged as a metadata associated to the log

tests/storage/test_service.py
```diff
@@ -1216,6 +1221,466 @@ def walk_children(
...
+    forbidden_variant.to_json_summary = Mock(return_value={"id": "forbidden-variant"})
+
+    def get_study(study_id: str) -> Study:
+        return {"parent": parent, "forbidden-variant": forbidden_variant}[study_id]
+
+    repository.get.side_effect = get_study
+    variant_study_service = Mock(spec=VariantStudyService)
+    variant_study_service.has_children.side_effect = lambda s: s.id == "parent"
+
+    def walk_children(
```
sylvlecl, line 1297: The test really relies on ta LOT of mocks of implementation details (mocking multipple methodds on underlying objects): it will make it kind of a headache to maintain and future evolutions difficult (we already feel it a lot :) ).

Also we don't test the web layer here
--> that's why I recommended to have integration tests (see `tests/integration` folder) which would not rely on that many mocks.

However coerage is very good here, so I propose as a middle ground to keep them, but we should at least add an integration test to cover the whole stack; for example in `tests/integration/studies_blueprint`. Single deletion is tested in `tests/integration/test_integration` for now but that file is bloated, we must not add more there.

tests/storage/test_service.py
```diff
@@ -1216,6 +1221,466 @@ def walk_children(
...
+        directory_service=Mock(spec=DirectoryService),
+        repository=repository,
+        config=Config(),
+        variant_study_service=variant_study_service,
+    )
+
+    # Admin user has permission, so this should succeed
+    service.delete_studies(["parent"], with_variants=True)
+    deleted_ids = set(repository.delete.call_args.args)
+    assert "forbidden-variant" in deleted_ids
```
sylvlecl: I think you only test that admin can delete anything, not that other users cannot delete others users studies ?
So the test does not really show that permissions are checked, however this is checked with tests below so coverage is good, but the naming here is misleading

skamril (dismissed):

tests/storage/test_service.py
```diff
@@ -1216,6 +1221,466 @@ def walk_children(
...
+        path=str(tmp_path / "parent"),
+        archived=False,
+        workspace=DEFAULT_WORKSPACE_NAME,
+        owner=None,
+        groups=[],
+    )
+    parent.to_json_summary = Mock(return_value={"id": "parent"})
+    parent.to_enhanced_json_summary = Mock(return_value={"id": "parent", "workspace": DEFAULT_WORKSPACE_NAME})
+
+    forbidden_variant = create_variant_study(
```
sylvlecl: Not sure why it's called forbidden variant, it seems misleading

tests/storage/test_service.py
```diff
@@ -1216,6 +1221,466 @@ def walk_children(
...
+    )
+
+    service.delete_studies(["ev-study"], with_variants=False)
+    assert call_order == ["delete", "event"]
+
+
+@with_admin_user
+def test_delete_studies_empty_list_raises() -> None:
+    """Passing an empty list should raise an HTTPException."""
+    from fastapi import HTTPException
```
sylvlecl: the import should be at module level, not repeated in each test

sylvlecl (changes requested):
Thx for the additions! a few minor comments on implementation + we should still add at least one integration test to cover the new endpoint

tests/storage/test_service.py
hdinia: the scenario sets up a variant that would normally be inaccessible to regular users (hence "forbidden-variant"), but since it's the admin performing the deletion it is indeed misleading I'll rename it

antarest/study/service.py
hdinia: forgot this refactor 😅

sylvlecl (approved):

MartinBelthle (approved):

PR #2999, opened 2026-02-12
fix(celery): fix logging setup

sylvlecl:
Restores the `@setup_logging` signal: currently, our logging configuration is not correctly taken into account.

Clarifies more the initialization:
an exception is now raised at beat or worker startup if the ANTAREST_CONF en variable is not defined
(or the config file does not exist).

This loading from env var is now defined in another module, so that it can be mocked in unit tests.


antarest/maintenance/app.py
```diff
@@ -97,40 +84,28 @@ def _apply_celery_config(app: Celery, celery_config: CeleryConfig) -> None:
...
-        return
 
-    if config.celery:
-        _apply_celery_config(celery_app, config.celery)
-        logger.info(f"Broker: {_mask_url_credentials(celery_app.conf.broker_url or '')}")
-
-    configure_logger(config)
-    celery_app.conf.antarest_config = config
-    logger.info("Celery configured")
+@setup_logging.connect
```
sylvlecl, line 92: This is crucial to keep, otherwise log config is not taken into account.

antarest/maintenance/app.py
```diff
@@ -97,40 +84,28 @@ def _apply_celery_config(app: Celery, celery_config: CeleryConfig) -> None:
 
 celery_app.autodiscover_tasks(["antarest.maintenance.tasks"])
 
-_startup_config = _load_config()
-if _startup_config and _startup_config.celery:
-    _apply_celery_config(celery_app, _startup_config.celery)
-
+load_config()
+_apply_celery_config(celery_app, get_config().celery)
```
sylvlecl, line 89: We now raise here (in get_config implementation) if the configuration is not defined.

We need to be sure that the config is defined when we start celery.

antarest/maintenance/app.py
```diff
@@ -170,3 +138,10 @@ def _log_critical_task_failure(sender: Any, task_id: str, exception: Exception,
             f"CRITICAL: Auto-archive task [{task_id}] failed permanently after all retries. Exception: {exception}. ",
             exc_info=exception,
         )
+
+
+def get_maintenance_context(task: Task[Any, Any]) -> MaintenanceContext:
```
sylvlecl: Just a helper method for tasks

antarest/maintenance/config.py
```diff
@@ -0,0 +1,50 @@
...
+def load_config() -> None:
+    """
+    Load config from ANTAREST_CONF env var, else raises.
+
+    In practice, this will abort celery app startup, since it's called at the app module level.
+    """
+    config_path_str = os.environ.get("ANTAREST_CONF")
+
+    if not config_path_str or not Path(config_path_str).exists():
+        raise ConfigurationError(
```
sylvlecl, line 37: We now raise here if config is not OK.
This way, the user gets an error message as soon as he starts the beat or the worker.

tests/maintenance/test_app.py
```diff
@@ -64,95 +89,52 @@ def test_worker_settings(self):
...
+    def test_load_config_from_file(self, tmp_path: Path):
+        Path(tmp_path / "config.yml").write_text("storage: {matrix_gc_sleeping_time: 5432}")
+        with env_var("ANTAREST_CONF", str(tmp_path / "config.yml")):
+            load_config()
+        assert isinstance(get_config(), Config)
+        assert get_config().storage.matrix_gc_sleeping_time == 5432
 
 
 class TestSetupPeriodicTasks:
-    def test_uses_defaults_without_config(self, celery_app_config_backup):
```
sylvlecl, line 116: Removed this: the config cannot be None any more

tests/maintenance/test_app.py
```diff
@@ -64,95 +89,52 @@ def test_worker_settings(self):
...
-        _configure_celery(sender="test", conf=celery_app.conf)
-        assert celery_app.conf.antarest_config is None
 
+class TestLoadConfig:
+    def test_load_config_without_env_var_raises(self, tmp_path: Path):
+        with pytest.raises(ConfigurationError):
+            load_config()
 
-class TestInitWorker:
-    def test_no_context_without_config(self, celery_app_config_backup, monkeypatch):
```
sylvlecl, line 97: Replaced those test by tests on the new behaviour : ANTAREST_CONF is required, and config is loaded from it

tests/maintenance/test_app.py
```diff
@@ -12,17 +12,42 @@
...
+def get_config_mock() -> Config:
+    return Config()
+
+
+# Explanation:
+#   The Celery app setup requires to initialize it at the module level.
+#   Here in tests, we override the initialization functions used in prod, otherwise they will raise because
+#   they require the env var ANTARES_CONF to be filled with a path to a working configuration.
+#
+with (
```
sylvlecl: Important part here to enable the unit tests in the dir to work, otherwise the import of the app module would fail, because ANTAREST_CONF is not defined

antarest/maintenance/tasks/auto_archive_task.py
```diff
@@ -61,10 +60,7 @@ def auto_archive_task(self: Task) -> AutoArchiveTaskResult:  # type: ignore[type
     if self.request.retries > 0:
         logger.warning(f"Auto-archive retry attempt {self.request.retries}/5")
 
-    ctx: MaintenanceContext | None = self.app.conf.get("maintenance_ctx")
-    if not ctx:
-        raise MaintenanceContextNotFoundError()
-
+    ctx = self.context
```
MartinBelthle, line 63: Way cleaner now that you centralized the ugly code :)

MartinBelthle (commented):
Thanks for the comments :), but I'm not really at ease with the celery code, I'll let @TheoPascoli review this as he knows better

antarest/maintenance/tasks/auto_archive_task.py
```diff
@@ -36,6 +34,7 @@
 
 
 @celery_app.task(
+    base=MaintenanceTask,
```
TheoPascoli, line 37: Nice

TheoPascoli (dismissed):
Cleaner this way :)

PR #3003, opened 2026-02-16
feat(ui-bc,ui-storages): add enabled status icon to constraints lists

skamril:
ANT-4511

### DIFF
--- a/webapp/src/components/icons/EnabledIcon.tsx
+++ b/webapp/src/components/icons/EnabledIcon.tsx
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) 2026, RTE (https://www.rte-france.com)
+ *
+ * See AUTHORS.txt
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+import CircleIcon from "@mui/icons-material/Circle";
+
+interface EnabledIconProps {
+  enabled?: boolean;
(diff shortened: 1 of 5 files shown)


hdinia (dismissed):
"Enabled" is domain specific, for a global component that is not ideal. you can make it more generic like : `StatusIndicator` or `StatusDot` to make it reusable across different contexts (task status, cluster status, etc.)

Suggested refactor:

```tsx
  interface StatusIndicatorProps {
    status: "success" | "error" | "warning" | "info" |
  "neutral";
    size?: "small" | "medium" | "large";
  }

  function StatusIndicator({ status, size = "small" }:
  StatusIndicatorProps) {
    const sizeMap = {
      ...
    };

    return <CircleIcon color={status} sx={{ fontSize:
  sizeMap[size] }} />;
  }
```

  Usage :

```tsx
  // Current use case - binding constraints
  <StatusIndicator status={constraint.enabled ? "success" :
  "error"} />
```

webapp/src/components/icons/EnabledIcon.tsx
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+import CircleIcon from "@mui/icons-material/Circle";
+
+interface EnabledIconProps {
+  enabled?: boolean;
```
hdinia: `enabled` should not be optional. If `constraint.enabled` is accidentally undefined due to a data issue, the component silently shows red instead of surfacing the problem. Can you explain this choice ? 
I think its better to handle a possible undefined value on the caller instead, or handle it directly in the component e.g :
```tsx
 if (enabled === undefined) {
   return <CircleIcon color="disabled" sx={{ fontSize: "0.5rem" }} />; // Gray color to indicate status in between 
}
```

hdinia (approved):

PR #3005, opened 2026-02-16
refactor(ui): rework variants tree with responsive layout

hdinia:
### Fix: Make variants tree fill available width

The variants tree now dynamically adjusts to fill the available horizontal space in its container:

- Added `ResizeObserver` to track container width changes
- Computed SVG `viewBox` width scales with container size while maintaining minimum dimensions
- Text label column stretches to absorb any extra horizontal space
- Tree scales uniformly with a zoom-out factor for better overview at wider viewports

Before:
image

After:
image


skamril:
Please ensure that the Git history is preserved when moving or renaming files.

skamril (approved):

PR #3008, opened 2026-02-17
feat(st-storage): implement database support

TheoPascoli:
- Added a method to get additional constraint matrix that was missing from the original file study dao


alembic/versions/7e3fd21b1c65_add_st_storage_table.py
```diff
@@ -0,0 +1,142 @@
...
+    "cost_injection",
+    "cost_withdrawal",
+    "cost_level",
+    "cost_variation_injection",
+    "cost_variation_withdrawal",
+]
+
+
+def upgrade():
+    group_enum = sa.Enum(
```
MartinBelthle: Since v9.3 st storage groups are free so you should not used this enum and simply declare the group as a string

antarest/study/dao/api/st_storage_dao.py
```diff
@@ -93,6 +93,12 @@ def get_st_storage_additional_constraints(
     ) -> list[STStorageAdditionalConstraint]:
         raise NotImplementedError()
 
+    @abstractmethod
+    def get_st_storage_additional_constraints_matrix(
```
MartinBelthle: Minor, I think we should remove the `s` at  the end of `constraints` as it's only for one matrix

antarest/study/dao/memory/in_memory_study_dao.py
```diff
@@ -155,6 +155,7 @@ def __init__(self, version: StudyVersion, matrix_service: ISimpleMatrixService)
         self._storage_cost_variation_withdrawal: Dict[ClusterKey, str] = {}
         # Short-term storages additional constraints
         self._st_storages_constraints: STStorageAdditionalConstraintsMap = {}
+        self._st_storages_constraints_matrix: Dict[AdditionalConstraintKey, str] = {}
```
MartinBelthle: Minor as we don't use this but i think we should add the storage id inside `AdditionalConstraintKey`

antarest/study/dao/database/database_st_storage_dao.py
```diff
@@ -0,0 +1,434 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+
+"""
+Database implementation of StStorageDao.
```
MartinBelthle, line 14: We should add all the matrices inside the matrix_provider and add tests for them

antarest/study/dao/database/database_st_storage_dao.py
```diff
@@ -0,0 +1,434 @@
...
+        if exc:
+            raise STStorageNotFound(area_id, storage_id) from exc
+        raise STStorageNotFound(area_id, storage_id)
+
+    def _raise_the_right_constraint_exception(
+        self, area_id: str, storage_id: str, constraint_id: str, exc: IntegrityError | None = None
+    ) -> NoReturn:
+        validate_area_exists(self._db_session, self._study_id, area_id)
+        if not self.st_storage_exists(area_id, storage_id):
+            raise STStorageNotFound(area_id, storage_id) from exc
```
MartinBelthle, line 121: As `exc` is optional we should also handle the case where it's not defined

antarest/study/dao/database/database_st_storage_dao.py
```diff
@@ -0,0 +1,434 @@
...
+        values = [self._convert_st_storage_to_row(area_id, st_storage) for st_storage in storages]
+
+        try:
+            upsert_multiple(session, ST_STORAGE_TABLE, values)
+        except IntegrityError as e:
+            validate_area_exists(session, self._study_id, area_id)
+            existing_storage_ids = {th.id for th in self.get_all_st_storages_for_area(area_id)}
+            for storage in storages:
+                if storage.id not in existing_storage_ids:
+                    self._raise_the_right_storage_exception(area_id, storage.id, e)
```
MartinBelthle: Should just raise StStorageNotFound here as otherwise we'll double check that the area exists

antarest/study/dao/database/database_st_storage_dao.py
```diff
@@ -0,0 +1,434 @@
...
+    def get_all_st_storages(self) -> dict[str, dict[str, STStorage]]:
+        session, study_id = self._db_session, self._study_id
+
+        stmt = select(ST_STORAGE_TABLE).where(ST_STORAGE_TABLE.c.study_id == study_id)
+        rows = session.execute(stmt).fetchall()
+
+        st_storages_by_areas: dict[str, dict[str, STStorage]] = {}
+        for row in rows:
+            st_storage = self._convert_db_row_to_st_storage(row)
+            st_storages_by_areas.setdefault(row.area_id, {})[st_storage.id.lower()] = st_storage
```
MartinBelthle: Don't think we need to do `.lower()`, we had to do this for thermal and renewables as ids could be in uppercase but that's not the case for st_storages

tests/study/dao/test_database_st_storage_dao.py
```diff
@@ -0,0 +1,347 @@
...
+    ST_STORAGE_ADDITIONAL_CONSTRAINT_TABLE,
+    ST_STORAGE_TABLE,
+    UPPER_RULE_CURVE_TABLE,
+)
+
+
+def test_save_st_storage(dao: DatabaseStudyDao) -> None:
+    dao.save_area("area_1")
+
+    dao.save_st_storage("area_1", STStorage(id="st_storage_id", name="st-storage", efficiency=0.8))
```
MartinBelthle: We should check all values not just `efficiecny` even if it's boring :/

tests/study/dao/test_database_st_storage_dao.py
```diff
@@ -0,0 +1,347 @@
...
+
+def test_st_storage_matrices_lifecycle(db_session: Session, dao: DatabaseStudyDao) -> None:
+    dao.save_area("area_1")
+    dao.save_st_storage("area_1", STStorage(id="battery", name="Battery"))
+
+    matrix_service = dao._matrix_service
+    dataframe = pl.DataFrame(data=[[1, 2.5], [3, 4.7]], orient="row")
+    series_id = matrix_service.create(dataframe)
+
+    dao.save_st_storage_pmax_injection("area_1", "battery", series_id)
```
MartinBelthle: Just to ensure we're not mixing the tables with the matrices it could be good to use different matrices for each type

MartinBelthle (changes requested):
We should also use the new method you've introduced inside the `study_converter.py`

antarest/study/dao/database/database_st_storage_dao.py
TheoPascoli: `from None` is fine. It will just raise without a cause. In the thermal `raise_the_right_exception` method, you dont have to test that `exc` is not None

antarest/study/dao/study_conversion/study_converter.py
```diff
@@ -272,8 +272,17 @@ def _convert_short_term_storages(
             self._new_dao.save_st_storage_inflows(area_id, sts_id, inflows_id)
 
             if self._study_version >= STUDY_VERSION_9_2:
-                if sts_id in constraints:
-                    self._new_dao.save_st_storage_additional_constraints(area_id, sts_id, constraints[sts_id])
+                storage_constraints = constraints.get(sts_id)
+                if storage_constraints is None:
```
MartinBelthle: Don't understand this logic, seems you're doing twice the same thing ?

antarest/study/dao/database/database_st_storage_dao.py
MartinBelthle: Ah okay, you could do the change inside thermal and renewable DAO then :)

tests/study/dao/test_database_st_storage_dao.py
```diff
@@ -0,0 +1,392 @@
...
+    ST_STORAGE_ADDITIONAL_CONSTRAINT_TABLE,
+    ST_STORAGE_TABLE,
+    UPPER_RULE_CURVE_TABLE,
+)
+
+
+def test_save_st_storage(dao: DatabaseStudyDao) -> None:
+    dao.save_area("area_1")
+
+    dao.save_st_storage(
```
MartinBelthle, line 46: Could be good to also check that we're able to save values that are set to None in this test

MartinBelthle (changes requested):
We should enrich the test_matrix_provider with sts matrices

antarest/study/dao/study_conversion/study_converter.py
TheoPascoli: yes you're right

tests/matrixstore/test_matrix_usage_providers.py
```diff
@@ -164,9 +165,12 @@ def test_command_matrix_usage_provider(
     with db():
         study_id = "study_id"
 
-        study_version = "880"
+        study_version = "930"
```
MartinBelthle: Nice of you to add this test but it is for the `CommandMatrixUsageProvider` not related to this PR. I meant to add a test inside `test_garbage_collector.py` which is in the same directory as others dao tests to test the actual provider we added with the DAO `StudyDatabaseMatrixUsageProvider`

MartinBelthle (changes requested):
Last comment :)

MartinBelthle (approved):

PR #3010, opened 2026-02-17
feat(matrix-store): add an admin endpoint to solve app issues

MartinBelthle:
[ANT-4543]

### DIFF
--- a/antarest/core/core_blueprint.py
+++ b/antarest/core/core_blueprint.py
@@ -14,9 +14,11 @@
 from fastapi import APIRouter
 
 from antarest.core.config import Config
+from antarest.core.requests import UserHasNotPermissionError
 from antarest.core.serde import AntaresBaseModel
 from antarest.core.utils.web import APITag
 from antarest.core.version_info import VersionInfoDTO, get_commit_id, get_dependencies
+from antarest.login.utils import require_current_user
 from antarest.matrixstore.model import MatrixMismatchDTO
 from antarest.matrixstore.service import ISimpleMatrixService
 
@@ -88,6 +90,10 @@ def create_utils_routes(config: Config, matrix_service: ISimpleMatrixService) ->
         - Deletes lines from `Matrix` table in DB for matrices that do not exist on the filesystem.
         - Insert lines inside `Matrix` table in DB for matrices that exist on the filesystem but not in the DB.
         """
+        user = require_current_user()
+        if not user.is_site_admin():
+            raise UserHasNotPermissionError()
+
         return matrix_service.synchronize_matrix_store(dry_run=dry_run)
 
     return bp
--- a/antarest/core/core_blueprint.py
+++ b/antarest/core/core_blueprint.py
@@ -17,6 +17,8 @@ from antarest.core.config import Config
 from antarest.core.serde import AntaresBaseModel
 from antarest.core.utils.web import APITag
 from antarest.core.version_info import VersionInfoDTO, get_commit_id, get_dependencies
+from antarest.matrixstore.model import MatrixMismatchDTO
+from antarest.matrixstore.service import ISimpleMatrixService
 
 
 class StatusDTO(AntaresBaseModel):
@@ -27,7 +29,7 @@ class DesktopModeDTO(AntaresBaseModel):
     desktop_mode: bool
 
 
-def create_utils_routes(config: Config) -> APIRouter:
+def create_utils_routes(config: Config, matrix_service: ISimpleMatrixService) -> APIRouter:
     """
     Utility endpoints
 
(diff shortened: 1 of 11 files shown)


antarest/core/core_blueprint.py
```diff
@@ -75,4 +79,21 @@ def desktop_mode() -> DesktopModeDTO:
 
         return DesktopModeDTO(desktop_mode=config.desktop_mode)
 
+    @bp.post(
```
sylvlecl: I think we'd better have it in the /matrixstore endpoint

antarest/matrixstore/service.py
```diff
@@ -689,3 +698,56 @@ def get_matrices_references(self, disk_usage: bool) -> dict[str, MatrixReference
                 references_dto[matrix_id].refs.append(ref_dto)
 
         return references_dto
+
+    @override
+    def synchronize_matrix_store(self, dry_run: bool) -> dict[str, MatrixMismatchDTO]:
+        db_matrices = {m.id for m in self.repo.get_matrices()}
+        fs_matrices = {
+            f.stem for f in self.matrix_content_repository.bucket_dir.iterdir() if not f.name.endswith(".tsv.lock")
```
sylvlecl: we'd better not expose repos internals here, would be better to have a method in the repo to retrieve the list of matrices.

antarest/matrixstore/service.py
```diff
@@ -689,3 +698,56 @@ def get_matrices_references(self, disk_usage: bool) -> dict[str, MatrixReference
...
+
+        new_matrices = []
+        current_date = current_time()
+        for matrix in only_fs_matrices:
+            result[matrix] = MatrixMismatchDTO(database=False, filesystem=True)
+            if not dry_run:
+                # We have to create a fake entry for the matrix in DB to fit with the filesystem.
+                # But for that we have to find what's the matrix version as it's used for parsing.
+                version = self._infer_matrix_version(matrix)
+                obj = Matrix(id=matrix, width=10, height=10, created_at=current_date, version=version)
```
sylvlecl: Why hard code the width and height, we should get it from the matrix content instead, I think ?

antarest/matrixstore/repository.py
```diff
@@ -363,12 +363,12 @@ def delete(self, matrix_hash: str) -> None:
...
 
     def get_matrix_disk_usage(self, matrix_hash: str) -> int:
-        matrix_path = self._get_matrix_path_n_format(matrix_hash)[0]
+        matrix_path = self.get_matrix_path_n_format(matrix_hash)[0]
         if matrix_path:
             return os.stat(matrix_path).st_size
         raise FileNotFoundError(str(self.bucket_dir.joinpath(matrix_hash)))
 
-    def _get_matrix_path_n_format(self, matrix_hash: str) -> tuple[Optional[Path], InternalMatrixFormat]:
+    def get_matrix_path_n_format(self, matrix_hash: str) -> tuple[Optional[Path], InternalMatrixFormat]:
```
sylvlecl: we'd better not expose a `Path` in the public API of the repo, it's an implementation detail that should not be accessible from outside

antarest/matrixstore/service.py
```diff
@@ -689,3 +698,56 @@ def get_matrices_references(self, disk_usage: bool) -> dict[str, MatrixReference
...
+                new_matrices.append(obj)
+
+        if new_matrices:
+            self.repo.save_batch(new_matrices)
+
+        return result
+
+    def _infer_matrix_version(self, matrix_id: str) -> int:
```
sylvlecl: see comment about exposing path:
it would probably be better to implement that method in the content repo instead to avoid leaking the path outside from it

antarest/matrixstore/service.py
```diff
@@ -689,3 +698,56 @@ def get_matrices_references(self, disk_usage: bool) -> dict[str, MatrixReference
...
+            if not dry_run:
+                # We remove the line from DB as it has no match on the filesystem
+                self.repo.delete(matrix)
+
+        new_matrices = []
+        current_date = current_time()
+        for matrix in only_fs_matrices:
+            result[matrix] = MatrixMismatchDTO(database=False, filesystem=True)
+            if not dry_run:
+                # We have to create a fake entry for the matrix in DB to fit with the filesystem.
```
sylvlecl: Let's add a log which explains what is done

antarest/matrixstore/service.py
```diff
@@ -689,3 +698,56 @@ def get_matrices_references(self, disk_usage: bool) -> dict[str, MatrixReference
...
+            f.stem for f in self.matrix_content_repository.bucket_dir.iterdir() if not f.name.endswith(".tsv.lock")
+        }
+        only_fs_matrices = fs_matrices - db_matrices
+        only_db_matrices = db_matrices - fs_matrices
+
+        result = {}
+        for matrix in only_db_matrices:
+            result[matrix] = MatrixMismatchDTO(database=True, filesystem=False)
+            if not dry_run:
+                # We remove the line from DB as it has no match on the filesystem
```
sylvlecl: Let's add a log here too

sylvlecl (changes requested):
The behaviour looks good, I propose some changes about:
 - where we add things
 - adding some logs so that we can check easily what is done

antarest/matrixstore/service.py
MartinBelthle: Just because it's faster (for matrices other than tsv) than reading the content and we don't really care about this info. But I can change this

sylvlecl: Yes it's better to have a "right" metadata. It's really not an issue to spend more time for this, performance is not a matter at all here.
If it's really useless we could remove it but that's another topic

MartinBelthle: Done in the 2 last commits

sylvlecl (approved):

PR #3012, opened 2026-02-19
feat(gc_tasks_task): add purge tasks table task by creation date

mehdiwahada:
What is it supposed to do ? : It cleans the tasks table by deleting the tasks under a certain creation date

Why ? : There's actually more than 400.000 tasks in the database, and that history contains less value, hence why we create a daily cleaning tasks task

How to test it ? : We must create some blank tasks, add them to the repository with a creation date and add the retention duration in int


antarest/core/config.py
```diff
@@ -191,6 +191,9 @@ class StorageConfig:
     variable_view_gc_retention_days: int = 30
     watcher_scan_sleeping_time: int = 900
     watcher_scan_dry_run: bool = False
+    tasks_gc_retention_duration: int = 3600
```
sylvlecl: we'd better have a shorter default duration, like 30 days.
We'd better name it with `days` too, like for the variable views just above

antarest/core/config.py
```diff
@@ -191,6 +191,9 @@ class StorageConfig:
...
+    tasks_gc_sleeping_time: int = 3600
```
sylvlecl: Let's have a longer default delay, like once a day (24 * 3600), it will not be very useful to execute it more often since retention duration is anyway in days

sylvlecl (changes requested):
Looks very good overall, we can just improve the parameter name and the default values and we're good :)

sylvlecl (approved):

PR #3014, opened 2026-02-20
feat(ui): improve study tree UX with collapsible sections, consistent icons, and stability fixes

hdinia:
ANT-4546

### DIFF
--- a/webapp/src/components/TreeItemEnhanced.tsx
+++ b/webapp/src/components/TreeItemEnhanced.tsx
@@ -12,64 +12,95 @@
  * This file is part of the Antares project.
  */
 
-import { mergeSxProp } from "@/utils/muiUtils";
-import { Box, CircularProgress, Tooltip } from "@mui/material";
+import { Box, CircularProgress, type SxProps, type Theme, Tooltip } from "@mui/material";
 import { TreeItem, type TreeItemProps } from "@mui/x-tree-view/TreeItem";
-import * as R from "ramda";
 import { Children } from "react";
+import { mergeSxProp } from "@/utils/muiUtils";
 
 export interface TreeItemEnhancedProps extends TreeItemProps {
   loading?: boolean;
+  disableTooltip?: boolean;
 }
+
+////////////////////////////////////////////////////////////////
+// Utils
+////////////////////////////////////////////////////////////////
+
+function getStyles(canExpand: boolean): SxProps<Theme> {
+  return {
+    "& > .MuiTreeItem-content": {
+      p: 0,
+      alignItems: "normal",
+      // Expand/collapse icon
+      "& > .MuiTreeItem-iconContainer": {
+        alignItems: "center",
+        borderTopLeftRadius: "inherit",
+        borderBottomLeftRadius: "inherit",
+        "&:hover": {
+          background: canExpand ? "inherit" : "none",
+        },
+      },
+      "& > .MuiTreeItem-label": {
+        py: 0.5,
+        whiteSpace: "nowrap",
+        textOverflow: "ellipsis",
+        overflow: "hidden",
+      },
+    },
+  };
+}
+
+function withTooltip(label: TreeItemEnhancedProps["label"], disableTooltip: boolean) {
+  if (disableTooltip || typeof label !== "string") {
+    return label;
+  }
+
+  return (
+    <Tooltip title={label}>
+      <span>{label}</span>
+    </Tooltip>
+  );
+}
+
+function withLoading(label: TreeItemEnhancedProps["label"], loading: boolean) {
+  if (!loading) {
+    return label;
+  }
+
+  return (
+    <Box sx={{ display: "flex", alignItems: "center", gap: 1 }}>
+      {/* Wrapped with a <Box> to prevent display issue when item is expanded */}
+      <Box>
+        <CircularProgress size={12} color="secondary" />
+      </Box>
+      {label}
+    </Box>
+  );
+}
+
+////////////////////////////////////////////////////////////////
+// Component
+////////////////////////////////////////////////////////////////
 
 function TreeItemEnhanced({
   onClick,
   sx,
-  label: labelFromProps,
-  loading,
+  label,
+  loading = false,
+  disableTooltip = false,
   ...rest
 }: TreeItemEnhancedProps) {
   const canExpand = Children.toArray(rest.children).length > 0;
+  const enhancedLabel = withLoading(withTooltip(label, disableTooltip), loading);
+  const styles = getStyles(canExpand);
 
-  ////////////////////////////////////////////////////////////////
-  // Label
-  ////////////////////////////////////////////////////////////////
-
-  const addLabelTooltip = (label: TreeItemEnhancedProps["label"]) => {
-    return typeof label === "string" ? (
-      <Tooltip title={label}>
-        <span>{label}</span>
-      </Tooltip>
-    ) : (
-      label
-    );
-  };
-
-  const addLabelLoading = (label: TreeItemEnhancedProps["label"]) => {
-    return loading ? (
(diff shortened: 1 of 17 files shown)


webapp/src/components/TreeItemEnhanced.tsx
```diff
@@ -12,64 +12,95 @@
...
 
-  const enhanceLabel = R.compose(addLabelLoading, addLabelTooltip);
+  return (
+    <Tooltip title={label}>
+      <span>{label}</span>
+    </Tooltip>
+  );
+}
 
-  ////////////////////////////////////////////////////////////////
```
skamril, line 64: Keep the "Event Handlers" comment block

webapp/src/components/TreeItemEnhanced.tsx
```diff
@@ -12,64 +12,95 @@
...
 
-function TreeItemEnhanced({
-  onClick,
-  sx,
-  label: labelFromProps,
-  loading,
-  ...rest
-}: TreeItemEnhancedProps) {
-  const canExpand = Children.toArray(rest.children).length > 0;
+////////////////////////////////////////////////////////////////
```
skamril, line 25: Using comment blocks both inside and outside the component makes it unclear whether we are currently inside or outside the component when reading the code.

In my opinion, this refactor wasn’t necessary. Adding the `disableTooltip` condition would have been sufficient.
The refactor does not meaningfully improve the component, which is already simple and clear, and appears to be driven more by personal preference than by a functional or structural need.

webapp/src/routes/_authenticated/studies/-components/StudyTree/ExternalTree/ExternalTreeNode.tsx
```diff
@@ -60,10 +44,8 @@ function ExternalTreeNode({
   const isWorkspace = isWorkspacePath(path);
 
   const sortedChildren = useMemo(() => {
-    const nodesToDisplay = filterScannedStudies(children);
-    const sortedByName = sortByName(nodesToDisplay);
-    return name === ROOT_NODE_NAME ? sortDefaultFirst(sortedByName) : sortedByName;
-  }, [children, name]);
+    return sortByName(filterScannedStudies(children));
```
skamril, line 47: 💡 There is a `sortByName` in "service/utils"

webapp/src/routes/_authenticated/studies/-components/StudyTree/ManagedTree/EditableTreeItem.tsx
```diff
@@ -53,6 +53,12 @@ function EditableTreeItem({
   ////////////////////////////////////////////////////////////////
 
   const handleSave = () => {
+    if (hasSavedRef.current) {
+      return;
+    }
+
+    // Prevent double-save caused by Firefox triggering blur event when Enter is pressed
+    hasSavedRef.current = true;
```
skamril, line 63: You don't reset it to false?

💡You can use `useThrottleFn` for `handleSave`

webapp/src/components/TreeItemEnhanced.tsx
hdinia: > Using comment blocks both inside and outside the component makes it unclear whether we are currently inside or outside the component when reading the code.

It's a subjective point of view, but I think that when defining utilities that return JSX, it's important to differentiate them from the main component. And I don't want to create separate files for that.

> In my opinion, this refactor wasn’t necessary. Adding the disableTooltip condition would have been sufficient.
The refactor does not meaningfully improve the component, which is already simple and clear, and appears to be driven more by personal preference than by a functional or structural need.

Your opinion is not facts it's subjective, I agree the refactoring was not strictly necessary however it's objectively improving this component 

Here's a detailed breakdown of why:

---

## 1. Helper Functions Defined Outside the Component

**V1** defines `addLabelTooltip`, `addLabelLoading`, and `enhanceLabel` *inside* the component body, which means they are re-created on every render.

**V2** defines `withTooltip`, `withLoading`, and `getStyles` *outside* the component at module scope. They are created once, are pure, referentially stable and are unit-testable in isolation.

---

## 2. Removes an Unnecessary Dependency

**V1** imports all of ramda (`import * as R from "ramda"`) just to do a two-function `R.compose`. This adds  an external dependency for something that can be expressed more readably without it:

```ts
const enhancedLabel = withLoading(withTooltip(label, disableTooltip), loading);
```

This is equally expressive and explicit about the execution order (inner-to-outer), with zero cost.

---

## 3. More Precise TypeScript Typing

**V1:**
```ts
const handleClick: TreeItemEnhancedProps["onClick"] = (event) => {
```
This types `handleClick` as `MouseEventHandler | undefined`, which is technically incorrect since the function can never be `undefined`.

**V2:**
```ts
const handleClick: NonNullable<TreeItemEnhancedProps["onClick"]> = (event) => {
```
This correctly narrows the type to exclude `undefined`.

---

## 4. Explicit Default Values

**V1** relies on implicit falsy checks for `loading` throughout the code. **V2** declares defaults directly in destructuring:

```ts
  loading = false,
  disableTooltip = false,
```

This makes the API contract clear and removes the need for defensive checks inside helper functions.

---

## 5. Early Returns vs Ternaries

**V1** uses nested ternaries inside the helper functions **V2** uses early returns:

```ts
function withTooltip(label: TreeItemEnhancedProps["label"], disableTooltip: boolean) {
  if (disableTooltip || typeof label !== "string") {
    return label;
  }
  ...
}
```

Early returns reduce cognitive load and a well known best practice

---

7. Cleaner Code Organization 

No styles bloat in the JSX, and the utils are clearly separated from the main operational component

webapp/src/components/TreeItemEnhanced.tsx
hdinia: Seriously ? there is only one function..

webapp/src/routes/_authenticated/studies/-components/StudyTree/ExternalTree/ExternalTreeNode.tsx
hdinia: updated thanks

webapp/src/routes/_authenticated/studies/-components/StudyTree/ManagedTree/EditableTreeItem.tsx
hdinia: Thanks for the remark but reset is not needed each new editing session gets a fresh component instance with `hasSavedRef.current` initialized to `false`.

About `useThrottleFn` it's not what's I am trying to do, `useThrottleFn` is designed for rate-limiting repeatable actions. Using a throttle would just mean "you can't save twice within 500ms", which is a weaker and less accurate contract than "you can only save once, period."

webapp/src/components/TreeItemEnhanced.tsx
skamril: 1. They are lightweight functions. Re-creation on render is negligible here and does not justify the added structural change.
2. Ramda is already a dependency of the application. We are not introducing anything new, so there is no real cost or impact here. This is a stylistic preference, not an optimization.
3. Both typings are valid and safe in practice. The change is technically more precise, but it does not solve a real issue.
4. OK
5. This is purely stylistic. Nested ternaries are readable in this case and do not justify refactoring.

This is not a blocker for validation. However, it introduces additional review and refactoring overhead without delivering real practical benefits. It’s small in this case, but we should avoid making this a habit.

webapp/src/components/TreeItemEnhanced.tsx
skamril: The JSX block is still present here.

skamril (dismissed):

skamril (approved):

PR #3016, opened 2026-02-20
feat(ui-output): add districts list with areas

skamril:
- LIstView: allow rendering content without TanStack routing
- use ListView
- refactor of some code
- disable search for synthesis list

ANT-4472


webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-components/OutputMatrixViewer/index.tsx
```diff
@@ -0,0 +1,315 @@
+/**
```
skamril, line 1: In my local commit, this is not being recognized as a new file...

webapp/src/components/page/ListView.tsx
```diff
@@ -41,27 +42,119 @@ interface BaseListItem {
...
+  content?: never;
+  data?: never;
 }
 
-interface ListViewProps {
-  list: RouteListItem[];
-  splitId: string;
+export interface ContentListItem<TData = unknown> extends BaseListItem {
+  linkOptions?: never;
+  // It will be used in priority over `renderItemContent` if provided.
```
hdinia: tip: to document interfaces properties you can use `/** */` to make the doc visible on hover (like a JSdoc)

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-components/OutputMatrixViewer/VariableMatrix/index.tsx
```diff
@@ -88,10 +78,7 @@ function VariableMatrix({
   const { t } = useTranslation();
 
   // TODO: !variablesMetadata check may be unnecessary
```
hdinia: we can safely remove this todo

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-utils.ts
```diff
@@ -22,26 +24,39 @@ import type {
...
+export type ListType = "areas" | "links" | "synthesis";
+export type DataType = "values" | "details" | "details-res" | "id" | "details-STstorage";
+export type Frequency = "hourly" | "daily" | "weekly" | "monthly" | "annual";
+export type MonteCarloMode = "mc-ind" | "mc-all" | "variable-per-variable";
 
-export type OutputItemType = (typeof OUTPUT_ITEM_TYPES)[number];
-export type DataType = (typeof DATA_TYPES)[number];
-export type Frequency = (typeof FREQUENCIES)[number];
-export type MonteCarloMode = (typeof MONTE_CARLO_MODES)[number];
+export type Item = AreaWithId | StudyMapDistrict | LinkElement;
```
hdinia, line 32: since it's not tied to only the Map view `StudyMapDistrict` is confusing

webapp/src/components/page/ListView.tsx
```diff
@@ -41,27 +42,119 @@ interface BaseListItem {
...
-function ListView({ list, splitId, emptyListContent, onAdd, actions }: ListViewProps) {
+function isRouteListItem(item: RouteListItem | ContentListItem): item is RouteListItem {
+  return !!item?.linkOptions;
+}
+
+function isRouteList(list: RouteListItem[] | ContentListItem[]): list is RouteListItem[] {
+  return list.length > 0 && isRouteListItem(list[0]);
+}
+
+function ListView<TItemData = unknown>({
```
hdinia: `ListView` mixes two unrelated modes

This component tries to handle two fundamentally different use cases in one place:

- **Route list** (7/8 usages): items have `linkOptions`, the right panel renders `<Outlet />`, no active-item state is needed.
- **Content list** (1/8 usages — `Output`): items carry data, the right panel is driven by local `activeContentItem` state.

Because the mode is inferred at runtime from the list data, the component is full of logic that only applies to one mode but always runs:

```tsx
// Only meaningful for content lists, but lives in the shared component
const [activeContentItem, setActiveContentItem] = useState(...);
const currentRouteId = useCurrentRouteId(); // Only meaningful for route lists

useEffect(() => {
  // Syncs activeContentItem — irrelevant for route lists
}, [list, hasRouteList, activeContentItem]);
```

And the mode detection itself is fragile — it breaks for empty route lists:

```ts
function isRouteList(list): list is RouteListItem[] {
  return list.length > 0 && isRouteListItem(list[0]); // false when list is empty
}
```

---

### Proposed refactor

Split into two focused components sharing a private `ListPanel` for the left column (search, add button, actions, item list):

- **`RouteListView`** — no selection state, renders `<Outlet />`, uses `useCurrentRouteId`.
- **`ContentListView`** — owns `activeContentItem` state and the sync effect, exposes `renderItemContent`.

```tsx
// RouteListView — dead simple, no local state for item selection
function RouteListView({ list, splitId, emptyListContent, onAdd, actions, splitMinSize }) {
  return (
    <SplitView splitId={splitId} minSize={splitMinSize}>
      <ListPanel list={list} onAdd={onAdd} actions={actions} ... />
      <Box>{list.length === 0 ? emptyListContent : <Outlet />}</Box>
    </SplitView>
  );
}
```

```tsx
// ContentListView — owns selection state, no routing concerns
function ContentListView<TData>({ list, renderItemContent, ... }) {
  const [activeItem, setActiveItem] = useState(() => list[0]);

  useEffect(() => {
    if (!activeItem || !list.find((item) => item.id === activeItem.id)) {
      setActiveItem(list[0]);
    }
  }, [list]); // no longer depends on activeItem

  return (
    <SplitView ...>
      <ListPanel list={list} ... />
      <Box>{activeItem && (activeItem.content ?? renderItemContent?.({ id: activeItem.id, data: activeItem.data }))}</Box>
    </SplitView>
  );
}
```

Each component only carries the props and hooks it actually needs, the fragile runtime type detection disappears, and the `as ContentListItem<TData>` cast in the effect goes away too.


## why this design could be problematic :

- Hard to evolve independently : Any new requirement for one mode affects the other, so new use-cases end up as props on a shared interface the other mode ignores. Over time the component accumulates dead weight.

- Hard to read and reason about : When reading a usage like `<ListView list={areas} ... />`, you can't tell which "mode" is active without tracing the type of `list` back through the call chain. The component's name and signature don't reveal its behavior. 

- Every future change touches unrelated code: If you need to adjust the content list's selection behavior, you're editing a component that 7 out of 8 callers use purely for routing. Any regression risk applies to all of them.

- No real benefit from the coupling : The shared value here is the left panel (search + list + add button). That's a layout concern, not a reason to unify state management and routing logic. There's nothing preventing that shared panel from being a private `ListPanel` used by two clearly scoped components.

- Testing is harder : You can't test the two behaviors in isolation , any test has to account for the shared surface, the mode inference logic, and both logics.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-components/OutputMatrixViewer/index.tsx
```diff
@@ -0,0 +1,315 @@
...
+    }
+  }, [mcMode, year]);
+
+  // Reset variable and cluster selections when dataType changes in variable-per-variable mode
+  useEffect(() => {
+    if (isVariablePerVariable) {
+      setSelectedVariable("");
+      setSelectedClusterId("");
+    }
+  }, [dataType, isVariablePerVariable, setSelectedVariable]);
```
hdinia: It's legacy code but I think `dataType` can be removed

webapp/src/components/page/ListView.tsx
skamril: I didn’t know this worked with types, thanks.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-utils.ts
skamril: True, we need to rename the Map types in a dedicated PR.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-components/OutputMatrixViewer/index.tsx
skamril: Are you certain? I noticed this comment:
“Reset variable and cluster selections when dataType changes in variable-per-variable mode.”

webapp/src/components/page/ListView.tsx
skamril: In a your proposition `ListPanel` needs a Route version also, because click behavior are not the same.

I preferred not to break the component into multiple files, given that it’s relatively simple.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-components/OutputMatrixViewer/index.tsx
hdinia: Yes, it seems safe to me to remove that dependency

The intent is to reset `selectedVariable` and `selectedClusterId` when `dataType` changes while in variable-per-variable mode. However, `dataType` is already passed to `useVariablePerVariable` (line 86) so when `dataType` changes, that hook re-runs anyway and `selectedVariable`/`selectedClusterId` resetting here would be redundant because the hook will re-derive its state.

The comment can be removed or updated `// Reset variable and cluster selections when switching into variable-per-variable mode`

webapp/src/components/page/ListView.tsx
hdinia: > In a your proposition ListPanel needs a Route version also, because click behavior are not the same.

No was implicit but I should perhaps have specified `ListPanel` owns the item wrapper (tooltip, loading spinner, padding) but delegates the button itself via a `renderItemButton` prop. This keeps the click behavior "navigate to a route vs update local state" out of `ListPanel` entirely, so it stays a pure layout shell with no knowledge of either mode.
edit: see examples below

> I preferred not to break the component into multiple files, given that it’s relatively simple.

 Did you read **why this design could be problematic :** ? the component isn't simple, it just looks small.

hdinia:
```tsx
function ListPanel({ list, onAdd, actions, renderItemButton }) {
  return (
    <>
      {/* search, add button, actions... */}
      <List>
        {list
          .filter((item) => isSearchMatching(search, item.label))
          .map((item) => (
            <Tooltip key={item.id} title={item.label}>
              <ListItem>{renderItemButton(item)}</ListItem>
            </Tooltip>
          ))}
      </List>
    </>
  );
}
```

```tsx
// RouteListView
<ListPanel
  renderItemButton={(item) => (
    <RouterListItemButton key={currentRouteId} {...item.linkOptions}>
      ...
    </RouterListItemButton>
  )}
/>

// ContentListView
<ListPanel
  renderItemButton={(item) => (
    <ListItemButton selected={activeItem?.id === item.id} onClick={() => setActiveItem(item)}>
      ...
    </ListItemButton>
  )}
/>

```

does that look better to you ?

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-utils.ts
hdinia: Districts are not tied to the Map (they could be in their own view) layers too, they are global Areas features. Please rename the type `District(s)` or add a TODO

webapp/src/components/page/ListView.tsx
skamril: It’s not a matter of knowing how to do it 🙂
I just didn’t see a clear benefit at this stage, since there’s no bug or incorrect behavior.
That said, I’m happy to split it if it improves readability or makes things clearer for others 👍

hdinia: Alright, I thought my initial comment was not detailed enough. That said, thanks for the changes and you got to the main point, it’s a matter of simplicity (both in reading and using this component) and avoiding future maintenance + evolution problems that we can easily prevent now.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-utils.ts
skamril: I’d prefer not to expand the scope of this PR. I didn’t introduce the original type name, and in this PR I only added District to the union type. Since these types are used across multiple files, a broader rename would be better handled in a dedicated PR.

hdinia (approved):

PR #3017, opened 2026-02-20
feat(ui-studies): scope study search recursively to selected managed directory

hdinia:
ANT-4605

### DIFF
--- a/webapp/src/redux/ducks/studies.ts
+++ b/webapp/src/redux/ducks/studies.ts
@@ -44,6 +44,8 @@ export interface StudyFilters {
   activeTree: "managed" | "external";
   managed: {
     directoryId: string | null;
+    // All directory IDs in scope (selected dir + descendants). null = root (all managed studies).
+    directoryIds: string[] | null;
   };
   external: {
     path: string;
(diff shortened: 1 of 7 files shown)


webapp/src/redux/ducks/studies.ts
```diff
@@ -44,6 +44,8 @@ export interface StudyFilters {
   activeTree: "managed" | "external";
   managed: {
     directoryId: string | null;
+    // All directory IDs in scope (selected dir + descendants). null = root (all managed studies).
+    directoryIds: string[] | null;
```
skamril, line 48: In your case `directoryId` can be removed.

```
managed: {
  directoryIds: item.id ? getDescendantIds(item.id, directories) :[],
}
```

```ts
const [directoryId] = useAppSelector((state) => getStudyFilters(state).managed.directoryIds, R.T);
```

skamril (dismissed):

webapp/src/redux/ducks/studies.ts
hdinia: It's true we could but in practice, the two fields serve different purposes:
- `directoryId` = which node is selected in the tree / where are we navigated to (used for UI - tree highlight, breadcrumbs)
- `directoryIds` = what set of directories to filter studies by (selected + all descendants - filtering only)

one tells you "what the user selected", the other tells you "what studies to show". Merging them means one field has to answer two questions which makes the code harder to read and easier to break

skamril (approved):

PR #3019, opened 2026-02-23
chore(deps): upgrade fastapi from 0.124.2 to 0.130.0

TheoPascoli:
Include performance improvements when there's a Pydantic return type

See: https://github.com/fastapi/fastapi/releases/tag/0.130.0


MartinBelthle (approved):
Nice improvement from them, I thought they already used Rust serialization

PR #3020, opened 2026-02-23
feat(scenario-builder): implement database support

TheoPascoli:


alembic/versions/c11cfe5728b0_add_ruleset_table.py
```diff
@@ -0,0 +1,154 @@
...
+        "hydroInitialLevels",
+        "hydroFinalLevels",
+        "hydroGenerationPower",
+        "shortTermStorageInflows",
+        "shortTermStorageAdditionalConstraints",
+        name="scenariotype",
+    )
+
+    op.create_table(
+        "ruleset",
```
MartinBelthle: Perhaps add a `s` to the table name

alembic/versions/c11cfe5728b0_add_ruleset_table.py
```diff
@@ -0,0 +1,154 @@
...
+        sa.PrimaryKeyConstraint("study_id", "ruleset_name", name="pk_ruleset"),
+    )
+
+    op.create_table(
+        "active_ruleset",
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("ruleset_name", sa.String(length=255), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["study_id", "ruleset_name"],
+            ["ruleset.study_id", "ruleset.ruleset_name"],
```
MartinBelthle: Funny to add the foreign key to ruleset and not Study. Don't know what are the real differences

alembic/versions/c11cfe5728b0_add_ruleset_table.py
```diff
@@ -0,0 +1,154 @@
...
+        sa.PrimaryKeyConstraint("study_id", name="pk_active_ruleset"),
+    )
+
+    op.create_table(
+        "ruleset_area",
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("ruleset_name", sa.String(length=255), nullable=False),
+        sa.Column("area_id", sa.String(length=255), nullable=False),
+        sa.Column("scenario_type", scenario_type_enum, nullable=False),
+        sa.Column("timeseries", sa.JSON(), nullable=False),
```
MartinBelthle: Not a big fan of the wording `timeseries`. What will you put here, a mapping like `{0: 3, 1: 5}` ?  To me it would be better to call it scenario_mapping or something like this

MartinBelthle: Also could it be nullable ? If someone does not define a mapping for a specific cluster, you'll store `{}` ?

alembic/versions/c11cfe5728b0_add_ruleset_table.py
```diff
@@ -0,0 +1,154 @@
...
+        sa.PrimaryKeyConstraint("study_id", "ruleset_name", "bc_group_id", name="pk_ruleset_bc_group"),
+    )
+
+    op.create_table(
+        "ruleset_area_item",
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("ruleset_name", sa.String(length=255), nullable=False),
+        sa.Column("area_id", sa.String(length=255), nullable=False),
+        sa.Column("item_id", sa.String(length=255), nullable=False),
+        sa.Column("scenario_type", scenario_type_enum, nullable=False),
```
MartinBelthle: Really personal but as we have to create different tables as we have different keys, I would prefer we remove the column `scenario_type` and create one table by type. To be also discussed with @sylvlecl

MartinBelthle (changes requested):
I think the possibility of having different rulesets and to select the active one among them makes the code harder to understand, the database bigger nad the queries slower. It's not because the simulator allows it that we should continue handle this in the webapp.
To me we should ensure with the users if they really use this feature as it would make the code easier. I asked Alexander and he does not even seem to understand that this is possible, I'll show him tomorrow and I think we should ask Benoit to do a poll in the Teams conversation with the users

MartinBelthle:
For 500 recette studies and 650 production ones I only found 2 studies where someone uses a different ruleset than `default ruleset` (same study apparently but copied).
The man just used `new ruleset` but didn't even have a `default ruleset` in his study so we could just rename this and it would be good

alembic/versions/c11cfe5728b0_add_ruleset_table.py
```diff
@@ -0,0 +1,140 @@
...
+                "st_storage_additional_constraint.constraint_id",
+            ],
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint("study_id", "area_id", "st_storage_id", "constraint_id"),
+    )
+
+
+def downgrade() -> None:
+    for table_name in reversed(_ALL_TABLES):
```
MartinBelthle: Minor: You can remove the `reversed` as we don't use it and it's a bit misleading

alembic/versions/c11cfe5728b0_add_ruleset_table.py
```diff
@@ -0,0 +1,140 @@
...
+        sa.PrimaryKeyConstraint("study_id", "area_id", "st_storage_id"),
+    )
+
+    op.create_table(
+        "scenario_storage_constraints",
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("area_id", sa.String(length=255), nullable=False),
+        sa.Column("st_storage_id", sa.String(length=255), nullable=False),
+        sa.Column("constraint_id", sa.String(length=255), nullable=False),
+        sa.Column("timeseries", sa.JSON(), nullable=False),
```
MartinBelthle: I discussed the naming `timeseries` with Sylvain and we don't think it's the right one. Moreover, for hydro levels they do not even represent a mapping towards a timeserie. I don't have an idea for a name though :/

tests/study/dao/test_database_scenario_builder_dao.py
```diff
@@ -0,0 +1,213 @@
...
+"""
+Unit tests for DatabaseScenarioBuilderDAO.
+"""
+
+from antarest.study.business.model.link_model import Link
+from antarest.study.business.model.renewable_cluster_model import RenewableCluster
+from antarest.study.business.model.scenario_builder_model import Ruleset, ScenarioType
+from antarest.study.business.model.sts_model import STStorage, STStorageAdditionalConstraint
+from antarest.study.business.model.thermal_cluster_model import ThermalCluster
+from antarest.study.dao.database.database_study_dao import DatabaseStudyDao
```
MartinBelthle, line 22: We should add tests to ensures when deleting a thermal for example, the sc builder is also empty (even if it's by construction it's nice). And also remove the TODOs inside thermal_dao and renewable_dao (at least) saying that we should perhaps do something when deleting

MartinBelthle (changes requested):
Idea for a next work: We should add a DAo method that just saves the data for one scenario type, as this would allow the front-end to be way faster when updating the scenario builder as for now we're reading all and saving all

MartinBelthle (changes requested):
I re-discussed the naming with Sylvain and he prefers to name `ts_numbers`, `value` to be as general as possible as for hydro levels it does not represent `ts_numbers` but rather a percentage to follow.

Other than that, we're good to merge

MartinBelthle (approved):

PR #3023, opened 2026-02-24
fix(sonar): try to resolve taint issues

sylvlecl:
Fixes sonar issues python:S5145 by using annotated types which ensure:
 - input expected to be UUIDs are indeed UUIDs
 - other input strings do not contain new lines


antarest/study/web/studies_blueprint.py
```diff
@@ -224,7 +239,7 @@ def count_studies(
         status_code=HTTPStatus.CREATED,
         summary="Import Study",
     )
-    def import_study(study: UploadFile, groups: str = "") -> str:
+    def import_study(study: UploadFile, groups: SanitizedStr = "") -> str:
```
TheoPascoli, line 236: You should use the `UuidStr` you introduced for `study_id` here and in other places in this file

antarest/study/web/studies_blueprint.py
```diff
@@ -46,6 +47,12 @@
 QUERY_REGEX = r"^\s*(?:\d+\s*(?:,\s*\d+\s*)*)?$"
 
 
+def _sanitize_study_id(study_id: str) -> str:
+    if study_id.isalnum():
+        return study_id
+    raise ValueError("Invalid study id")
```
TheoPascoli: This is dead code

antarest/study/web/study_data_blueprint.py
TheoPascoli: Those parameters could be `SanitizedStr` as well

TheoPascoli (changes requested):

MartinBelthle (approved):
Nice :)

antarest/study/web/studies_blueprint.py
sylvlecl: Group IDs are not UUIDs: I left it untouched

antarest/study/web/study_data_blueprint.py
sylvlecl: Indeed I changed the few remaining ": str"

TheoPascoli (approved):

PR #3024, opened 2026-02-24
feat(dao): support thematic trimming in DB

MartinBelthle:


alembic/versions/562d4e1bd95d_add_thematic_trimming_table.py
```diff
@@ -0,0 +1,34 @@
...
+        "thematic_trimming",
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("thematic_trimming", sa.JSON, nullable=False),
+        sa.ForeignKeyConstraint(
+            ["study_id"],
+            ["study.id"],
+            name=op.f("fk_thematic_trimming_study_id_study"),
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint("study_id", name=op.f("pk_thematic_trimming")),
```
TheoPascoli: Minor: you don't need to construct a string in the name

antarest/study/dao/database/database_thematic_trimming_dao.py
```diff
@@ -0,0 +1,73 @@
...
+        """Get the study ID for database queries."""
+        return self._study_id
+
+    def get_session(self) -> Session:
+        """Get the SQLAlchemy session for database operations."""
+        return self._db_session
+
+    @abstractmethod
+    def get_impl(self) -> "DatabaseStudyDao":
+        pass
```
TheoPascoli, line 40: You don't use those 3 methods you can remove them

alembic/versions/562d4e1bd95d_add_thematic_trimming_table.py
```diff
@@ -0,0 +1,34 @@
...
+
+def upgrade():
+    op.create_table(
+        "thematic_trimming",
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("thematic_trimming", sa.JSON, nullable=False),
+        sa.ForeignKeyConstraint(
+            ["study_id"],
+            ["study.id"],
+            name=op.f("fk_thematic_trimming_study_id_study"),
```
TheoPascoli: Same string stuff

antarest/study/dao/database/database_thematic_trimming_dao.py
```diff
@@ -0,0 +1,73 @@
...
+    @override
+    def save_thematic_trimming(self, trimming: ThematicTrimming) -> None:
+        session = self.get_session()
+        values = {"study_id": self.get_study_id(), "thematic_trimming": trimming.model_dump(exclude_none=True)}
+
+        try:
+            upsert_one(session, THEMATIC_TRIMMING_TABLE, values)
+        except IntegrityError:
+            # Happens if the study does not exist -> ForeignKey constraint fails
+            raise StudyNotFoundError(self.get_study_id())
```
TheoPascoli: You should keep the full stacktrace with `from e`
```suggestion
        except IntegrityError as e:
            # Happens if the study does not exist -> ForeignKey constraint fails
            raise StudyNotFoundError(self.get_study_id()) from e
```

TheoPascoli (changes requested):

antarest/study/dao/database/database_thematic_trimming_dao.py
MartinBelthle: I just keep `get_impl()` just in case as we have it in any DAO and it's not easy to understand you can do this for a new dev

TheoPascoli (approved):

PR #3027, opened 2026-02-25
feat(matrix): support `raw` operations on db studies

MartinBelthle:
Thanks to this PR, for studies stored in DB:
- The GET `/raw` endpoint returns the right value for matrices only
- It's possible to export matrices in a certain format


antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -0,0 +1,231 @@
...
+from antarest.study.business.model.xpansion_model import XpansionResourceFileType
+from antarest.study.dao.api.study_dao import ReadOnlyStudyDao
+from antarest.study.model import STUDY_VERSION_8_2, STUDY_VERSION_8_7
+
+
+class RawPathToMatrixMapper:
+    """
+    Parses a given path like `input/load/series/load_area_fr` to retrieve the corresponding matrix.
+    Used for studies stored in database as the notion of path no longer exists for them.
+    But to ensure backward compatibility for the `raw` endpoints we still have to return the matrices.
```
sylvlecl, line 27: We should document that we have chosen to do that because it's faster than the cleaner, long term alternative to have dedicated matrix endpoints, and that we should resorb that some day

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def _get_matrix_inside_input_folder(self, parts: tuple[str, ...], error_msg: str) -> pl.DataFrame:
+        study_version = self._dao.get_version()
+
+        if not parts:
+            raise ValueError(error_msg)
+
+        prefix, rest = parts[0], parts[1:]
+        if prefix == "wind":
+            if len(rest) != 2 and rest[0] != "series":
```
sylvlecl: Should be an "or" I think, otherwise we are not sure rest[0] is valid.
Same everywhere

tests/study/dao/matrix_raw_mapper/test_raw_path_to_matrix_mapper.py
```diff
@@ -0,0 +1,226 @@
...
+        mapper.get_matrix_from_path(Path("settings/generaldata"))
+
+    # User resource
+    with pytest.raises(ValueError, match=r"Path user[\\/]my_file.xlsx does not point towards a matrix."):
+        mapper.get_matrix_from_path(Path("user/my_file.xlsx"))
+
+    # Thermal.ini file
+    pattern = r"Path input[\\/]thermal[\\/]clusters[\\/]paris[\\/]list does not point towards a matrix."
+    with pytest.raises(ValueError, match=pattern):
+        mapper.get_matrix_from_path(Path(f"input/thermal/clusters/{area_id}/list"))
```
sylvlecl: We should add more invalid paths, you could use a parametzied test for this, for example "input/wind", "input/wind/series" ...
Paths that start correctly but that are not compplete

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -0,0 +1,231 @@
...
+
+        if prefix == "input":
+            return self._get_matrix_inside_input_folder(rest, basic_error_msg)
+
+        raise ValueError(basic_error_msg)
+
+    def _get_matrix_inside_user_dir(self, parts: tuple[str, ...], error_msg: str) -> pl.DataFrame:
+        """The only possible matrices are expansion weights or capacities"""
+        if len(parts) != 3 or parts[0] != "expansion" or parts[1] not in ["capa", "weights"]:
+            raise ValueError(error_msg)
```
sylvlecl: We'd probably re-use the `IncorrectPathError` as in the FS DAO, which will produce a 404

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -0,0 +1,231 @@
...
+                raise ValueError(error_msg)
+            if study_version < STUDY_VERSION_8_7:
+                return self._dao.get_constraint_values_matrix(rest[0])
+            # Based on the suffix we'll know which DAO method to use
+            mapping: dict[str, Callable[[str], pl.DataFrame]] = {
+                "lt": self._dao.get_constraint_less_term_matrix,
+                "gt": self._dao.get_constraint_greater_term_matrix,
+                "eq": self._dao.get_constraint_equal_term_matrix,
+            }
+            bc_id, suffix = rest[0][:-3], rest[0][-2:]
```
sylvlecl, line 116: If suffix is not correct we wil lhave a KeyError, we should have an IncorrectPathError instead

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -0,0 +1,231 @@
...
+
+        elif prefix == "links":
+            if study_version < STUDY_VERSION_8_2:
+                if len(rest) != 2:
+                    raise ValueError(error_msg)
+                return self._dao.get_link_series(rest[0], rest[1])
+
+            # Since v8.2 link matrices are separated in 3 different matrices, we have to determine which one was asked.
+            if len(rest) == 2:
+                return self._dao.get_link_series(rest[0], rest[1].removesuffix("_parameters"))
```
sylvlecl: if the suffix is absent, we should raise an incorrect path.
Same for the prefixes I think

sylvlecl (changes requested):
A few change requests for error cases

sylvlecl (approved):
Look good to me

PR #3028, opened 2026-02-26
chore(fastapi): align API endpoints with modern FastAPI conventions

TheoPascoli:
Fully AI made (human verified)

### DIFF
--- a/antarest/core/filesystem_blueprint.py
+++ b/antarest/core/filesystem_blueprint.py
@@ -409,7 +409,7 @@ def create_file_system_blueprint(config: Config) -> APIRouter:
         mount_dir = _get_mount_dir(fs, mount)
 
         # The following code looks weird, but it's the only way to handle exceptions in generators.
-        tasks = []
+        file_infos = []
         iterator = mount_dir.glob(path) if path else mount_dir.iterdir()
         while True:
             try:
@@ -423,9 +423,9 @@ def create_file_system_blueprint(config: Config) -> APIRouter:
                 # Unacceptable pattern: non-relative glob pattern
                 raise HTTPException(status_code=403, detail=f"Access denied to path: '{path}'. {exc}") from exc
             else:
-                tasks.append(FileInfoDTO.from_path(file_path, details=details))
+                file_infos.append(FileInfoDTO.from_path(file_path, details=details))
 
-        return tasks
+        return file_infos
 
     @bp.get(
         "/{fs}/{mount}/cat",
--- a/antarest/core/filesystem_blueprint.py
+++ b/antarest/core/filesystem_blueprint.py
@@ -14,7 +14,6 @@
 Filesystem Blueprint
 """
 
-import asyncio
 import datetime
 import os
 import shutil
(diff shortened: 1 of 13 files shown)


antarest/study/web/output_blueprint.py
```diff
@@ -50,7 +50,7 @@
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_EXPORT_FORMAT = Query(TableExportFormat.CSV, alias="format", description="Export format", title="Export Format")
+DEFAULT_EXPORT_FORMAT = Query(alias="format", description="Export format", title="Export Format")
```
sylvlecl: we should rename this, it does not define the default format anymore.

Instead we could define a TypeAlias for the annotated type

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -70,7 +70,7 @@
     ".antares": ("text/plain", "utf-8"),
 }
 
-DEFAULT_EXPORT_FORMAT = Query(TableExportFormat.CSV, alias="format", description="Export format", title="Export Format")
+DEFAULT_EXPORT_FORMAT = Query(alias="format", description="Export format", title="Export Format")
```
sylvlecl: Same, let's define a type alias instead, and it does not define anymore the default value

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -70,7 +70,7 @@
...
 PATH_TYPE = Annotated[SanitizedStr, Query(openapi_examples=get_path_examples())]
```
sylvlecl, line 76: We should have used CamelCase here since it's a type alias (not from this PR though :) )

antarest/study/web/study_data_blueprint.py
```diff
@@ -429,24 +421,30 @@ def set_thematic_trimming(uuid: UuidStr, field_values: ThematicTrimmingUpdate) -
         path="/studies/{uuid}/config/playlist/form",
         summary="Get MC Scenario playlist data for table form",
     )
-    def get_playlist(uuid: UuidStr) -> PlaylistRootModel:
+    def get_playlist(uuid: UuidStr) -> dict[int, PlaylistValues]:
```
sylvlecl: I don't think we should do it, we lose the JSON schema additions.
Same below.

At least we should use the "root model" as the `response_model` to keep the API documented

antarest/core/filesystem_blueprint.py
```diff
@@ -426,19 +423,17 @@ async def list_files(
                 # Unacceptable pattern: non-relative glob pattern
                 raise HTTPException(status_code=403, detail=f"Access denied to path: '{path}'. {exc}") from exc
             else:
-                file_info = FileInfoDTO.from_path(file_path, details=details)
-                tasks.append(file_info)
+                tasks.append(FileInfoDTO.from_path(file_path, details=details))
```
sylvlecl: This should not be named tasks, but file_infos

sylvlecl (changes requested):
Nice, minor comments

sylvlecl (approved):

PR #3033, opened 2026-02-26
feat(matrix)!: support `raw` modifications on db studies

MartinBelthle:
Using the PUT `/raw` endpoint will still succeed on DB matrices

### Breaking change

We no longer support PUT `/raw` endpoint for FS studies on the 2 matrices `k` and `translation` inside prepro folders. These files are supposed to be deprecated but it's still a breaking change


antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -9,15 +9,23 @@
...
 import polars as pl
 
 from antarest.core.exceptions import IncorrectPathError
 from antarest.study.business.model.xpansion_model import XpansionResourceFileType
-from antarest.study.dao.api.study_dao import ReadOnlyStudyDao
-from antarest.study.model import STUDY_VERSION_8_2, STUDY_VERSION_8_7
+from antarest.study.dao.api.study_dao import StudyDao
+
+
+@dataclass
```
sylvlecl: we can make it frozen

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -30,208 +38,257 @@ class RawPathToMatrixMapper:
     this class.
     """
 
-    def __init__(self, dao: ReadOnlyStudyDao) -> None:
+    def __init__(self, dao: StudyDao) -> None:
         self._dao = dao
+        self._xpansion_capa_getter = lambda m: self._dao.get_xpansion_resource(XpansionResourceFileType.CAPACITIES, m)
```
sylvlecl: no need to make it a field of that class, same for the other getter below

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -30,208 +38,257 @@ class RawPathToMatrixMapper:
     this class.
     """
 
-    def __init__(self, dao: ReadOnlyStudyDao) -> None:
+    def __init__(self, dao: StudyDao) -> None:
         self._dao = dao
```
sylvlecl: Seems not needed either to make it a field of the class, we only keep the functions below

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -30,208 +38,257 @@ class RawPathToMatrixMapper:
...
+            ),
+            RegexMatcher(
+                pattern=re.compile(
+                    r"input/st-storage/constraints/(?P<area_id>[^/]+)/(?P<storage_id>[^/]+)/rhs_(?P<constraint_id>[^/]+)"
+                ),
+                getter=self._dao.get_st_storage_additional_constraint_matrix,
+                setter=self._dao.save_st_storage_constraint_matrix,
+            ),
+            RegexMatcher(
+                pattern=re.compile(r"input/bindingconstraints/(?P<constraint_id>[^/]+)"),
```
sylvlecl: Codes saw that this will match "/intput/bindingconstraints/foo_lt" for example, whereas it should match the more specific "_lt" instead, so you should move it after the more specific ones.

Note that I think we may still have issues if someone names its BC "something_lt" ...

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -30,208 +38,257 @@ class RawPathToMatrixMapper:
...
+                getter=self._dao.get_misc_gen,
+                setter=self._dao.save_misc_gen,
+            ),
+            RegexMatcher(
+                pattern=re.compile(r"input/reserves/(?P<area_id>[^/]+)"),
+                getter=self._dao.get_reserves,
+                setter=self._dao.save_reserves,
+            ),
+            RegexMatcher(
+                pattern=re.compile(r"input/links/(?P<area_from>[^/]+)/(?P<area_to>[^/]+)_parameters"),
```
sylvlecl: Note that for studies < 8.1, the _parameters did not exist.

Looks OK to no support studies that old, but in that case we should be explicit about that higher in the stack

tests/study/dao/matrix_raw_mapper/test_raw_path_to_matrix_mapper.py
```diff
@@ -206,6 +207,228 @@ def test_nominal_cases(dao_930: DatabaseStudyDao) -> None:
...
+    series_id, df = _build_random_dataframe()
+    mapper.save_matrix_from_path(path, series_id)
+    pl.testing.assert_frame_equal(dao.get_hydro_max_daily_gen_energy(area_id), df, check_dtypes=False)
+
+    path = Path(f"input/hydro/common/capacity/maxDailyPumpEnergy_{area_id}")
+    series_id, df = _build_random_dataframe()
+    mapper.save_matrix_from_path(path, series_id)
+    pl.testing.assert_frame_equal(dao.get_hydro_max_daily_pump_energy(area_id), df, check_dtypes=False)
+
+    # todo: We're missing BC and Xpansion tests as they are not yet implemented in DB.
```
sylvlecl, line 430: aht, that's probably why the issue did not appear in test

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -9,15 +9,23 @@
...
 from antarest.core.exceptions import IncorrectPathError
 from antarest.study.business.model.xpansion_model import XpansionResourceFileType
-from antarest.study.dao.api.study_dao import ReadOnlyStudyDao
-from antarest.study.model import STUDY_VERSION_8_2, STUDY_VERSION_8_7
+from antarest.study.dao.api.study_dao import StudyDao
+
+
+@dataclass
+class RegexMatcher:
+    pattern: Any
```
sylvlecl: why not pattern type ?

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
sylvlecl: can be frozen

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -30,208 +38,257 @@ class RawPathToMatrixMapper:
...
     """
 
-    def __init__(self, dao: ReadOnlyStudyDao) -> None:
+    def __init__(self, dao: StudyDao) -> None:
         self._dao = dao
+        self._xpansion_capa_getter = lambda m: self._dao.get_xpansion_resource(XpansionResourceFileType.CAPACITIES, m)
+        self._xpansion_weights_getter = lambda m: self._dao.get_xpansion_resource(XpansionResourceFileType.WEIGHTS, m)
+        self._path_matchers = [
+            RegexMatcher(
+                pattern=re.compile(r"user/expansion/capa/(?P<file_name>[^/]+)"),
```
sylvlecl, line 45: note that we introduce a dependency from those patterns on the functions parameters names:
we really need to cover all them with tests, otherwise it will break easily since we will not be able to rely on typing to detect consistency issues if we rename parameters

sylvlecl (changes requested):
minor stuff

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
MartinBelthle: Yes. I'll add tests for this but my solution will be to not add the match depending on the version this way we avoid the issue

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
MartinBelthle: Indeed I forgot, I'll add it

antarest/study/service.py
```diff
@@ -1696,16 +1703,26 @@ def edit_study(self, uuid: str, url: str, new: SUB_JSON) -> JSON:
...
 
-        self.event_bus.push(
-            Event(
-                type=EventType.STUDY_DATA_EDITED,
-                payload=study.to_json_summary(),
-                permissions=PermissionInfo.from_study(study),
+        else:
+            self._edit_study_using_command(study=study, url=url.strip().strip("/"), data=new)
+
+            self.event_bus.push(
```
sylvlecl: should probably be outside the if/else

sylvlecl (dismissed):
Looks good, maybe just take into account the comment about the event emission

PR #3034, opened 2026-02-27
refactor(ruleset)!: study supports only one ruleset

TheoPascoli:
Breaking change: GET `/studies/{uuid}/config/scenariobuilder` now returns a single ruleset instead of a dict of rulesets. The front-end is not impacted as it only uses the per-type endpoints


antarest/study/dao/file/file_study_scenario_builder_dao.py
```diff
@@ -82,6 +58,19 @@ def _get_nb_years(self) -> int:
...
         return nb_years
 
+    @staticmethod
+    def _resolve_ruleset_name(file_study: FileStudy) -> str:
+        """
+        Determines the ruleset section name to read from the scenariobuilder file.
+        Uses "Default Ruleset" if present, otherwise falls back to the first section.
+        """
+        data = file_study.tree.get(["settings", "scenariobuilder"])
+        if DEFAULT_RULESET_NAME in data:
```
MartinBelthle: Seems you could remove this `if` block as in the end you return it and we only want to support one ruleset.

Also I think we should only consider the only ruleset is `DEFAULT_RULESET_NAME` and not consider others as there's no way to set another name for it in the app

antarest/study/dao/study_conversion/study_converter.py
```diff
@@ -70,7 +70,7 @@ def convert_study_inputs(self) -> None:
         self._convert_settings()
 
         # Scenario Builder
-        self._new_dao.save_scenario_builder(self._source_dao.get_rulesets())
+        self._new_dao.save_scenario_builder(self._source_dao.get_ruleset())
```
MartinBelthle, line 73: Makes me realize that we should modify the `GeneralFileData` writing to write `default ruleset` inside `active-rules-scenario` when buildingmode is custom as the simulator needs it I think, see JIRA ticket 3027. This will not solve the JIRA issue but it will make the study_converter succeed from DB to FS as without it the simu will fail

antarest/study/web/study_data_blueprint.py
```diff
@@ -452,11 +452,11 @@ def update_playlist(
         summary="Get MC Scenario builder config",
         response_model_exclude_none=True,
     )
-    def get_scenario_builder_config(uuid: UuidStr) -> RulesetsView:
+    def get_scenario_builder_config(uuid: UuidStr) -> RulesetView:
```
MartinBelthle, line 455: We need to ensure the front-end still works with this change (also document in the PR that this is a breaking change)

MartinBelthle (changes requested):

antarest/study/dao/file/file_study_scenario_builder_dao.py
```diff
@@ -43,36 +42,13 @@ def get_file_study(self) -> FileStudy:
...
-        except KeyError:
-            active_ruleset = default_ruleset
-        else:
-            # In some old studies, the active ruleset is stored in lowercase.
-            if not active_ruleset or active_ruleset.lower() == "default ruleset":
-                active_ruleset = default_ruleset
-        return active_ruleset
+        return parse_ruleset_from_any(scenario_builder_data, study_data.config.version)
 
     def _get_nb_years(self) -> int:
```
MartinBelthle, line 60: Not relative to this PR but we should replace this code by `self.get_impl().get_general_config().nb_years`

antarest/study/storage/rawstudy/model/filesystem/config/general.py
```diff
@@ -72,6 +76,8 @@ def from_model(cls, config: GeneralConfig, study_version: StudyVersion) -> "Gene
                 data["custom_scenario"] = config.building_mode == BuildingMode.CUSTOM
             else:
                 data["custom_ts_numbers"] = config.building_mode == BuildingMode.CUSTOM
+        if config.building_mode == BuildingMode.CUSTOM:
```
MartinBelthle, line 79: Nice thanks :). We should add a test for this just to ensure this works

MartinBelthle (changes requested):
2 last comments

MartinBelthle (approved):

PR #3048, opened 2026-03-02
chore(stopwatch): simplify stopwatch API

sylvlecl:
Minor improvements to our timer class:
 - removes the need for lambda, which shortens syntax
 - replace calls to `log_elapsed` by plain logging calls


TheoPascoli (approved):
Way better like this :)

PR #3071, opened 2026-03-03
feat(add disk usage log task): adding the disk usage log task

mehdiwahada:
What does it do ? : it puts the check disk usage into the tasks

Why it was done ? : before this method, there were 17 workers logging the same information, so we delegated it to a celery task


antarest/maintenance/tasks/disk_usage_log.py
```diff
@@ -0,0 +1,103 @@
...
+                exc_info=e,
+            )
+
+
+def disk_usage_logging(config: Config, dry_run: bool) -> DiskUsageTaskResult:
+    logger.info(f"Registering disk usage metrics (dry_run={dry_run})")
+    try:
+        with db():
+            with create_lock(db.session, LockId.DISK_USAGE):
+                if not dry_run:
```
sylvlecl: The dry_run parameter is not very useful for that task, since it does not do anything except logging.

This parameter may be useful for tasks that actually perform some (potentially dangerous) actions, like deleting stuff, so that we can log what would be done before actually doing it.

So let's remove it to make things simpler :) Let's also remove it from config.

sylvlecl (changes requested):
Looks good, we should remove the dry_run parameter though to make things even simpler, it's not very useful for that task

antarest/core/maintenance/service.py
```diff
@@ -75,37 +42,16 @@ def __init__(
         self.repo = repository
         self.event_bus = event_bus
         self.cache = cache
-        self._metrics = WorkspaceDiskUsageMetrics(REGISTRY) if config.metrics.prometheus else None
         self._init()
 
     def _init(self) -> None:
         self.thread = Thread(
```
sylvlecl: Now we should remove this thread, otherwise we will still get the logs from all workers, I will do it and merge

sylvlecl (approved):
Thanks !

PR #3072, opened 2026-03-03
test(dao): add matrix mapper tests for filesystem dao

MartinBelthle:


tests/study/dao/conftest.py
```diff
@@ -90,8 +131,15 @@ class RealCaseDBStudy:
     dataframes: list[pl.DataFrame]
 
 
-def build_real_case_db_study(dao: DatabaseStudyDao) -> RealCaseDBStudy:
-    matrix_service = dao._matrix_service
+def get_matrix_service_from_dao(dao: StudyDao) -> ISimpleMatrixService:
```
sylvlecl, line 145: would make more sense to get the matrix service from the fixture instead of relying on those hidden fields

sylvlecl (dismissed):
very minor comment, not a must

PR #3073, opened 2026-03-03
feat(ui-studies): add studies bulk move explorer

hdinia:


webapp/public/locales/en/main.json
```diff
@@ -803,7 +803,16 @@
   "studies.error.unarchiveOutput": "Failed to unarchive study output {{outputname}}",
   "studies.success.saveData": "Data saved with success",
   "studies.success.scanFolder": "Folder scan started",
-  "studies.success.moveStudy": "Study '{{study}}' was successfully moved to '{{path}}'",
+  "studies.success.moveStudies_one": "{{count}} study moved to '{{path}}'",
```
skamril, line 806: "studies.success.moveStudies_one": "Study moved to '{{path}}'"
"studies.success.moveStudies_one": "Étude déplacée vers '{{path}}'"

"studies.warning.moveStudiesPartial_one": "Study could not be moved",
"studies.warning.moveStudiesPartial_one": "L'étude n'a pas pu être déplacée"

webapp/src/components/dialogs/FormDialog.tsx
```diff
@@ -129,7 +124,7 @@ function FormDialog<TFieldValues extends FieldValues, TContext, SubmitReturnValu
         {...formProps}
         sx={{
           ".Form__Content": {
-            px: 1, // Prevent content from touching scrollbar
+            p: 1, // Prevent content from touching scrollbar
```
skamril, line 131: Not necessary to add padding to the top and bottom in all cases. Please add it directly in the view that needs it using the ".Form__Content" class.

webapp/src/routes/_authenticated/studies/-components/StudiesList/Header/index.tsx
```diff
@@ -54,8 +55,12 @@ function Header({
   const [confirmFolderScan, setConfirmFolderScan] = useState(false);
   const [isRecursiveScan, setIsRecursiveScan] = useState(false);
   const [confirmDeleteStudies, setConfirmDeleteStudies] = useState(false);
+  const [confirmMoveStudies, setConfirmMoveStudies] = useState(false);
```
skamril, line 58: `openMoveStudyDialog`

💡

```ts
  const { openDialog } = useDialog();

  const handleMoveStudies = () => {
    openDialog(({ onClose }) => (
      <MoveStudyDialog
        open
        studies={selectedStudies}
        onClose={() => {
          setSelectedStudyIds([]);
          onClose();
        }}
      />
    ));
  };
```

webapp/src/routes/_authenticated/settings/tokens/-components/dialog/TokenFormDialog/index.tsx
```diff
@@ -18,7 +18,10 @@ import { TOKEN_FORM_DEFAULT_VALUES, type TokenFormDefaultValues } from "../utils
 import TokenForm from "./TokenForm";
 
 export interface TokenFormDialogProps
-  extends Omit<FormDialogProps<TokenFormDefaultValues>, "children" | "maxWidth" | "titleIcon"> {
+  extends Omit<
+    FormDialogProps<TokenFormDefaultValues, unknown, string>,
```
skamril: No need to specify SubmitReturnValue; it’s automatically inferred from the return value of `onSubmit`.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+        />
+      </Box>
+
+      {helperText && (
+        <Typography variant="caption" color={error ? "error.main" : "text.secondary"}>
+          {helperText}
+        </Typography>
+      )}
+
+      {children}
```
skamril, line 175: A field editor should not allow children.

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+      maxWidth="sm"
+      onSubmit={handleSubmit}
+      onSubmitSuccessful={handleSubmitSuccessful}
+      onCancel={onClose}
+      allowSubmitOnPristine={allowSubmitOnPristine}
+      submitButtonIcon={<DriveFileMoveIcon />}
+      submitButtonText={t("global.move")}
+    >
+      {({ control }) => (
+        <StudyDestinationFE name="directory" control={control}>
```
skamril:
```ts
<>
  <StudyDestinationFE ... />
  <CheckBoxFE ... />
</>
```
cf. https://github.com/AntaresSimulatorTeam/AntaREST/pull/3073/changes#r2888746736

The checkbox should be hidden when we are inside a study.

skamril (changes requested):

webapp/src/routes/-shared/components/studies/StudyDestinationFE/DirectoryBreadcrumbs.tsx
```diff
@@ -0,0 +1,120 @@
...
+          </IconButton>
+        </span>
+      </Tooltip>
+
+      <Divider orientation="vertical" flexItem />
+
+      {/* Breadcrumb track + new directory input */}
+      <Box sx={breadcrumbTrackSx}>
+        {breadcrumbs.map((segment: BreadcrumbSegment, index: number) => (
+          <React.Fragment key={segment.id ?? "__root__"}>
```
skamril, line 84: `"__root__"` can be set more than once.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/DirectoryList.tsx
```diff
@@ -0,0 +1,79 @@
...
+          return (
+            <Box
+              component="li"
+              key={directory.id}
+              role="option"
+              aria-selected={false}
+              aria-label={directory.name}
+              onClick={() => onDirectoryClick(directory)}
+              onKeyDown={(event) => {
+                if (event.key === "Enter" || event.key === " ") {
```
skamril: Can you add a comment to explain `event.key === " "`

webapp/src/routes/-shared/components/studies/StudyDestinationFE/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+const StudyDestinationFEWithReactHookFormSupport = reactHookFormSupport({
+  defaultValue: ROOT_DIRECTORY,
+  preValidate: (value: DirectoryValue) =>
+    combineValidators(
+      validatePath({ allowEmpty: true, allowToStartWithSlash: false, allowToEndWithSlash: false }),
+      validateString({ specialChars: { chars: "/", mode: "allow" }, allowEmpty: true }),
+    )(value.newDirectoryPath),
+})(StudyDestinationFE);
+
+export { StudyDestinationFE };
```
skamril, line 193: No need to export StudyDestinationFE; it is used automatically when the `control` prop is not provided.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/utils.ts
```diff
@@ -0,0 +1,86 @@
...
+import type { Directory } from "@/services/api/directories/types";
+import type { DirectoryValue } from "./types";
+
+/**
+ * Builds an O(1) lookup index from a flat directory list.
+ *
+ * @param directories - Flat array of Directory objects.
+ * @returns Map of directory id → Directory.
+ */
+export function buildDirectoryIndex(directories: Directory[]): Map<string, Directory> {
```
skamril: The name is not very clear. Maybe `mapDirectoriesById`.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/utils.ts
```diff
@@ -0,0 +1,86 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+import type { Directory } from "@/services/api/directories/types";
+import type { DirectoryValue } from "./types";
+
+/**
+ * Builds an O(1) lookup index from a flat directory list.
```
skamril: `O(1)`?

webapp/src/routes/-shared/components/studies/StudyDestinationFE/utils.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Returns direct children of the given parent, sorted alphabetically (case-insensitive).
+ *
+ * @param parentId - Parent directory id, or `null` for root-level directories.
+ * @param directories - Flat array of Directory objects.
+ * @returns Sorted array of direct child directories.
+ */
+export function getDirectChildren(parentId: string | null, directories: Directory[]): Directory[] {
+  return directories
+    .filter((directory) => directory.parentId === parentId)
+    .sort((a, b) => a.name.toLowerCase().localeCompare(b.name.toLowerCase()));
```
skamril: Use `sortByName()` util function.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+  name,
+  disabled = false,
+  error = false,
+  helperText,
+  inputRef,
+  children,
+}: Props) {
+  const { data: directories } = useSuspenseQuery(directoryQueries.list());
+
+  const index = useMemo(() => buildDirectoryIndex(directories), [directories]);
```
skamril: The name is not very clear. Maybe `directoriesById`.

Also, `useMemo` can be replace by this:

```ts
const {
  data: { directories, directoriesById },
} = useSuspenseQuery({
  ...directoryQueries.list(),
  select: (directories) => ({ directories, directoriesById: mapDirectoriesById(directories) }),
});
```

webapp/src/routes/-shared/components/studies/StudyDestinationFE/types.ts
```diff
@@ -0,0 +1,35 @@
...
+export interface BreadcrumbSegment {
+  id: string | null;
+  name: string;
+  /** Whether this segment is the currently navigated directory. */
+  active?: boolean;
+}
+
+export interface DirectoryValue {
+  /** Selected directory ID, or `null` for the managed root. */
+  id: string | null;
```
skamril: Could you rename it to `currentDirectory`? Or rename `newDirectoryPath` to `newDirectoryId`?
Because the link between the id and the path is not clear.

webapp/src/routes/-shared/components/studies/dialogs/CopyStudyDialog.tsx
```diff
@@ -53,20 +58,23 @@ function CopyStudyDialog({ study, open, onClose }: Props) {
...
   ////////////////////////////////////////////////////////////////
   // Event handlers
   ////////////////////////////////////////////////////////////////
 
   const handleSubmit = ({
-    values: { studyName, destinationFolder, outputIds = [] },
+    values: { studyName, directory, outputIds = [] },
   }: SubmitHandlerPlus<DefaultValues>) => {
+    // TODO: This should be moved to the API layer when Tan stack migration is done.
+    const directoryPath = toDirectoryPath(directory, directories);
```
skamril: `directoryPath` value is not valid.

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+    directory: { id: initialDirectoryId, newDirectoryPath: "" },
+    redirect: true,
+  };
+
+  ////////////////////////////////////////////////////////////////
+  // Event Handlers
+  ////////////////////////////////////////////////////////////////
+
+  const handleSubmit = async (data: SubmitHandlerPlus<FormValues>): Promise<MoveResult> => {
+    const { directory, redirect } = formSchema.parse(data.values);
```
skamril: There is no need to check the schema here; the "Submit" button is blocked if the values are not valid.

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+    // TODO: This adapter should be moved to the API layer when Tan stack migration is done.
+    const directoryPath = toDirectoryPath(directory, directories);
+
+    // Run moves sequentially so the first call creates any new directory
+    // and subsequent calls reuse it (parallel calls would race and create duplicates).
+    const results: Array<PromiseSettledResult<unknown>> = [];
```
skamril, line 85: `Promise.allSettled()` can be used for this situation.

webapp/src/components/dialogs/FormDialog.tsx
hdinia: It's necessary it fixes a field label visibility issue, and it does not hurt

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
hdinia: did you read the comment above explain why a sequential move is required ?

webapp/src/routes/-shared/components/studies/dialogs/CopyStudyDialog.tsx
hdinia: can you explain ? (it's the path expected by the api to create 1 or more new directories)

webapp/src/routes/-shared/components/studies/StudyDestinationFE/types.ts
hdinia: `id` is the id of the current selected directory
`newDirectoryPath` is the specified (optional) path to create 1 (or more) new directory where we move studies in

webapp/src/routes/-shared/components/studies/StudyDestinationFE/index.tsx
hdinia: true, will update it

webapp/src/routes/-shared/components/studies/StudyDestinationFE/utils.ts
hdinia: space/time complexity (Big O notation)

webapp/src/routes/-shared/components/studies/StudyDestinationFE/utils.ts
hdinia: ok will see that

webapp/src/routes/-shared/components/studies/dialogs/CopyStudyDialog.tsx
skamril: The new destination is empty every time for me

webapp/src/routes/_authenticated/settings/tokens/-components/dialog/TokenFormDialog/index.tsx
hdinia: It was added to fix a type issue after using `unknown` instead of `any` on the FormDialog
You can remove it and run the lint to see the error

webapp/src/routes/-shared/components/studies/StudyDestinationFE/index.tsx
hdinia: I dont see the problem with that can you explain more ?

webapp/src/routes/_authenticated/settings/tokens/-components/dialog/TokenFormDialog/index.tsx
skamril: This is why `any` was used: to avoid forcing a fixed type for `onChange` returns value when the type is extended. It wasn't an error. The goal of `SubmitReturnValue` is to have the return type of `onChange` inside `onSubmitSuccessful`, not need to fix a type.
And TContext comes from react-hook-form, where it is typed as `any`.

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
hdinia: You can talk about the underlying problem before imposing a solution. Here, redirection does not work within a study that is the real problem. You may have a case where you want to see the directory after moving the study, and if you don't want to, you can uncheck the option.
The redirection has been corrected the rest is a matter of specifications/opinion.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/utils.ts
skamril: O(1) might not be obvious for everyone. Could you simplify the description? For example, something like: “Creates a Map indexed by directory id for fast lookup.”

webapp/src/routes/-shared/components/studies/StudyDestinationFE/DirectoryBreadcrumbs.tsx
hdinia: `segment.id ?? "__root__"` falls back to `"__root__"` only when `segment.id` is `null`

The only segment that ever has `id: null` is the root entry, which is constructed once and prepended to the array

All other segments come from `getDirectoryAncestors`, which returns real directories — all with non-null IDs

So the `"__root__"` key is tied 1-to-1 with the root directory entry, and since you can't have more than one root in a breadcrumb trail, it's safe.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/DirectoryList.tsx
hdinia: ok

webapp/src/components/dialogs/FormDialog.tsx
skamril: This fixes a label issue only in your dialog. The fix should be made in StudyDestinationFE, and it shouldn’t add extra padding to all FormDialogs. You can add an `sx` prop in StudyDestinationFE for example.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/index.tsx
skamril: Nesting a field editor inside another field editor via `children` doesn’t make sense.
It would be like putting an `<input>` inside another `<input>` (it's not possible).
These editors should be siblings instead.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/types.ts
hdinia: maybe it's better to rename the field value `destination` and inside `directoryId` and `newPath` would this be more clear ? (if not we will have things like `directory.currentDirectory`)

webapp/src/routes/-shared/components/studies/StudyDestinationFE/DirectoryBreadcrumbs.tsx
skamril: It might be clearer to either assign `"__root__"` directly to the root segment or document why this fallback is safe.
Without that context, `segment.id ?? "__root__"` can look like it may produce duplicate keys.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/index.tsx
hdinia: Technically true at runtime, the default export already delegates to the raw component when `control` is absent. But the named export is useful for consumers who need to reference the unwrapped component explicitly (e.g. for typing, testing, or re-wrapping).

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
skamril: "You can talk about the underlying problem before imposing a solution. Here, redirection does not work within a study that is the real problem" => I already explained this issue during a team call with you.

"imposing" =>  I used “should,” not “must.”

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
hdinia: Agreed that RHF blocks the button on invalid values, but `parse` also serves two other purposes: it applies the `.transform()` on `newDirectoryPath` (normalization), and it gives us a runtime validation + correct type output shape in the handler. Ideally this would be handled by a [Zod resolver at the RHF level](https://react-hook-form.com/docs/useform#resolver), which would make the manual `parse` redundant but that migration is out of scope here.

webapp/src/routes/-shared/components/studies/dialogs/CopyStudyDialog.tsx
skamril: API issue, sorry. I was reading `directory.newDirectoryPath`.

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
skamril: I wasn’t aware that Zod could also transform values; I thought it was only for schema validation.

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
skamril: I don’t see an issue with parallel calls here, since the endpoint creates the folder only if the destination doesn’t exist. The first call to be fulfilled will create it, and the others will use it.
But if you’re sure, maybe I’m missing something.

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
hdinia: Already covered on the call, the main concern is the redirection not working inside a study, that's sufficient. Hide the checkbox suggestion is a functional/UX call, please avoid those in future reviews.

webapp/src/components/dialogs/FormDialog.tsx
hdinia: Fixing it once here is better than waiting for it to surface in other dialogs and having to fix each one individually. The `p: 1` value is minimal and I'm not aware of any existing dialog where this would regress anything.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/DirectoryBreadcrumbs.tsx
hdinia: Added a comment for context

webapp/src/routes/-shared/components/studies/dialogs/CopyStudyDialog.tsx
hdinia: yes known api bug, directoryId is not updated after a copy

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
hdinia: That said, the redirect point was a good feedback that helped fix this issue, thanks

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
hdinia: Test the following workflow : 
- select multiple studies
- add a new directory destination
- submit (with redirection enabled)

-> you will notice duplicates directories with the same name + each study apart in each duplicate directory

webapp/src/routes/-shared/components/studies/StudyDestinationFE/index.tsx
hdinia: Ok I understand better thanks.
The `<input>` inside `<input>` analogy doesn't quite apply here, the `children` are rendered below the explorer panel, not inside the field control itself. The `CheckBoxFE` passed as a child has its own `name="redirect"` and `control`  it's an independent RHF field, just visually grouped with the destination picker. 
At the RHF level they're already siblings. This is purely a layout convenience so `MoveStudyDialog` doesn't need a wrapping `Box` to group them visually. 
The concern about layout responsibility is valid, but the nesting analogy doesn't apply.

webapp/src/routes/_authenticated/settings/tokens/-components/dialog/TokenFormDialog/index.tsx
hdinia: I'd argue is a good thing it makes the contract clearer. That said if you think keeping `any` for ergonomics is best I can revert that part.
Btw the `eslint` bypass should have been documented properly

webapp/src/components/dialogs/FormDialog.tsx
skamril: `p: 1` adds 8px at the top and 8px at the bottom, while the goal is more to remove extra spacing than to add it. If you place a field directly in the content, MUI already provides `margin="dense"`.
But not all FormDialog have a field directly inside them.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/index.tsx
skamril: This could also be written as:

```ts
<>
  <StudyDestinationFE ... />
  <CheckBoxFE ... />
</>
```

Introducing `children` in StudyDestinationFE just to avoid a fragment makes the JSX structure less explicit and blurs the component’s intent.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/types.ts
skamril: Yes `DirectoryDestination` is better with `newPath` or `pathToCreate` (it can be undefined also).

webapp/src/routes/_authenticated/settings/tokens/-components/dialog/TokenFormDialog/index.tsx
skamril: I think keeping `any` is a good idea in this case. The goal is only to have an inferred value. These generics are not intended to be defined directly.

webapp/src/routes/-shared/components/studies/dialogs/MoveStudyDialog/index.tsx
skamril: Ok, thanks. I thought this was handled by the backend.

skamril (approved):

PR #3075, opened 2026-03-04
feat(output): introduce first part of v2 output storage

sylvlecl:
**Implements first part of a version 2 storage for outputs.**

The storage will rely on:

- a "large file storage" for the archived form
- database for metadata and reasonable sized data like logs and variables list
- parquet files for variable values

This PR implements everything the first 2 bullet points.
It also adds a conversion endpoint between output storages.
The last one will be implemented in a separate work.

Also, tries to improve the layout of modules in output package.


antarest/lfs/__init__.py
sylvlecl, line 1: That package is meant to have an S3 implementation of the LFS interface, in the future.

It will store output archives, but should be used for study archives too.

antarest/output/filestudy/__init__.py
sylvlecl, line 1: This package is meant to contain filestudy format related functions, that may be used by any storage implementation, in particular to extract some data from the imported outputs.

antarest/output/storage/output_storage.py
sylvlecl, line 1: That package contains the interface and both implementations of the output storage.

In both implementations, I simplified naming of modules (repository, storage)

antarest/output/service.py
sylvlecl, line 1: renamed to just service, the containing package is enough information

antarest/output/routes.py
sylvlecl, line 1: renamed to routes to be consistent with fastapi wording

antarest/study/dao/database/sql_utils.py
```diff
@@ -38,55 +38,6 @@ def _update_columns(table: Table) -> list[Column[Any]]:
     return [column for column in table.columns.values() if not column.primary_key]
 
 
-def generic_upsert_multiple(
```
sylvlecl, line 41: removed this, we are not willing to support anything else than sqlite and postgresql for now, it was just a maintenance burden.

antarest/output/variable_view/__init__.py
sylvlecl, line 1: Did not change anything here, it's just moving existing functions in a better place

TheoPascoli (dismissed):
Looks good, don't forget to change the alembic down revision before you merge !

PR #3076, opened 2026-03-04
feat(ui): update MUI to v7 and MUI X Tree View to v8

skamril:


webapp/src/components/DynamicList.tsx
```diff
@@ -13,7 +13,7 @@
  */
 
 import RemoveCircleOutlineIcon from "@mui/icons-material/RemoveCircleOutline";
-import { Divider, Grid, IconButton } from "@mui/material";
+import { Divider, GridLegacy as Grid, IconButton } from "@mui/material";
```
hdinia, line 16: `GridLegacy` is removed in v7 you have to use `Grid` instead 

also in `SnackErrorMessage.tsx` `AllocationField` `CorrelationField`

webapp/src/components/TreeItemEnhanced.tsx
```diff
@@ -12,25 +12,23 @@
...
   loading?: boolean;
   disableTooltip?: boolean;
 }
 
-////////////////////////////////////////////////////////////////
-// Utils
-////////////////////////////////////////////////////////////////
+/**
+ * Utils
+ */
```
hdinia: Please keep the original comment blocks, they act as separators. 
`/** ... */` is used to document objects properties like in interfaces, or for JSdocs

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/areas/$areaId/hydro/allocation/index.tsx
```diff
@@ -14,7 +14,7 @@
 
 import Form from "@/components/Form";
 import type { SubmitHandlerPlus } from "@/components/Form/types";
-import { Grid } from "@mui/material";
+import { GridLegacy as Grid } from "@mui/material";
```
hdinia, line 17: In allocation and correlation views you have to migrate to the new `Grid` (update imports + props in the components) 

you can see the migration guide here: https://mui.com/material-ui/migration/upgrade-to-grid-v2/#how-to-upgrade

webapp/src/utils/muiUtils.ts
```diff
@@ -23,6 +23,18 @@ export function truncateTextSx(maxWidth?: number) {
...
+ * Transparentizes a color by mixing it with transparent.
+ *
+ * @param color - The color to transparentize, in any valid CSS color format (e.g., hex, rgb, rgba, hsl).
+ * Supports variables (e.g., `var(--my-color)`).
+ * @param transparency - The percentage of transparency to apply (0-100).
+ * @returns A new color string that is a mix of the original color and transparent.
+ */
+export function transparentizeColor(color: string, transparency: number) {
+  return `color-mix(in srgb, ${color}, transparent ${transparency}%)`;
+}
```
hdinia: Suggestion: the `transparency: 95` in your current code means 95% transparent that's the same as 5% opacity but inverted, that can be a readability trap

you can flip the argument to make it more obvious, and name it `withOpacity` or `applyOpacity`
```ts
export function withOpacity(color: string, opacity: number) {
  return `color-mix(in srgb, ${color} ${opacity}%, transparent)`;
}
```
```ts
transparentizeColor(color, 90) // you pass 90, but you get 10% opacity
withOpacity(color, 10)         // you pass 10, you get 10% opacity
```

hdinia (changes requested):

webapp/src/components/DynamicList.tsx
skamril: No, it hasn’t been removed. `Grid` has been renamed to `GridLegacy`, and `Grid2` has been renamed to `Grid`.

https://mui.com/material-ui/migration/upgrade-to-v7/#grid-and-grid2-renamed

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/areas/$areaId/hydro/allocation/index.tsx
skamril: https://github.com/AntaresSimulatorTeam/AntaREST/pull/3076#discussion_r2882973894

webapp/src/utils/muiUtils.ts
skamril: I see your point, but using transparency isn’t unusual. For example, Sass has a `transparentize($color, $amount)` function where a higher amount means more transparency as well, so the current API felt reasonable to me.

hdinia: Sass's `transparentize` takes values between `0 and 1`, here with 0–100  the "high number = almost invisible" relationship isn't obvious from the call site alone.

Ultimately, what we're trying to express at the call site is an opacity 
```ts
transparentizeColor(color, 97) // does 97 mean "97% opacity" or "97% transparent"? not obvious without checking
withOpacity(color, 3)          // unambiguous —> 3% opacity, no convention needed
```

webapp/eslint.config.js
```diff
@@ -136,6 +136,36 @@ export default [
...
+        },
+        {
+          selector: "TemplateElement[value.cooked=/var\\(--joy-/]",
+          message: "Don't use css vars directly. Import from `theme.vars`",
+        },
+        {
+          selector: "TemplateElement[value.cooked=/var\\(--mui-/]",
+          message: "Don't use css vars directly. Import from `theme.vars`",
+        },
+      ],
```
hdinia, line 160: The `theme.palette` → `theme.vars.palette` rules make sense 👍🏼 

However, I'd drop the 4 CSS var rules (`var(--joy-...)` / `var(--mui-...)`).
There's no realistic path where someone would hardcode these strings. There are zero occurrences in the codebase, and I don't see that changing. It adds config noise for a scenario that won't happen in practice.

hdinia (changes requested):

webapp/eslint.config.js
skamril: In fact, I already used `"var(--mui-palette-success-main)"` instead of `"success.main"` for example.

However, I’ll remove the CSS variable rules for `--joy-*`.

webapp/src/utils/muiUtils.ts
skamril: `transparentizeColor(color, 97)` clearly means “97% transparent” — that’s what the function name implies. There’s no reference to opacity here.

Sass also provides `opacify($color, $amount)` to do the same in the opposite direction.

webapp/src/components/DynamicList.tsx
hdinia: Nevermind their was a confusion with the dev branch that was using props that no longer exists, but since you updated the imports it's not an issue

webapp/src/utils/muiUtils.ts
skamril: I changed it to `function withOpacity(color: string, opacity: number)`. Both options are fine for me.

hdinia (dismissed):

hdinia (approved):

PR #3077, opened 2026-03-04
feat(comments): implements database support

TheoPascoli:


alembic/versions/8a9a91f6a2bc_add_comments_column_to_study.py
```diff
@@ -0,0 +1,27 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "8a9a91f6a2bc"
+down_revision = "c11cfe5728b0"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("study", schema=None) as batch_op:
```
sylvlecl: It's better if we use a separate table which will be dedicated to the database DAO, otherwise we will have that column for all kinds of studies whereas it won't be used for file studies

antarest/study/model.py
```diff
@@ -298,6 +299,7 @@ class Study(Base):
     version: Mapped[str] = mapped_column(String(255), index=True)
     author: Mapped[Optional[str]] = mapped_column(String(255), nullable=True)
     editor: Mapped[Optional[str]] = mapped_column(String(255), nullable=True)
+    comments: Mapped[str] = mapped_column(Text(), nullable=False, default="")
```
sylvlecl: See comment on alembic: we should not add it here, because it will not be actually used for all kinds of studies.
It will be misleading to have it here.

sylvlecl (changes requested):
We should use a separate table, not the one used for all studies metadata: otherwise good !

alembic/versions/8a9a91f6a2bc_add_comments_column_to_study.py
TheoPascoli: Ok done !

alembic/versions/8a9a91f6a2bc_add_comments_column_to_study.py
```diff
@@ -0,0 +1,31 @@
...
+down_revision = "c11cfe5728b0"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "comments",
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("comments", sa.Text(), nullable=False, server_default=""),
```
sylvlecl: I think we can remove the server_default: as we already discussed, the app is responsible for providing a default, this server default should not be needed

sylvlecl (changes requested):
Last minor comment about server default

alembic/versions/8a9a91f6a2bc_add_comments_column_to_study.py
sylvlecl: Done

sylvlecl (approved):

PR #3078, opened 2026-03-04
chore(tests): add tests dir to pythonpath for pytest

sylvlecl:
Explanation: in most dev environment, the tests directory will be identified as a source dir, and import automatically added relative to it. This causes many CI failures since in CI, we have not configured pytest to consider tests as a source dir.

This PR fixes that consistency issue.


sylvlecl:
Not a good idea after all, this could lead to ambiguities.
Look for a better IDE configuration instead

PR #3081, opened 2026-03-05
perf(commands): introduce a result object for future works

MartinBelthle:
With this work, the command returns a result object the caller can use to build its response.
It will allow the code to be more performant in future works


antarest/study/storage/variantstudy/model/command/common.py
```diff
@@ -26,8 +32,10 @@ def command_failed(message: str) -> CommandOutput:
     return CommandOutput(False, message)
 
 
-def command_succeeded(message: str, should_invalidate_cache: bool = False) -> CommandOutput:
-    return CommandOutput(True, message, should_invalidate_cache)
+def command_succeeded(
+    message: str, result: CommandApplicationResult | None, should_invalidate_cache: bool = False
```
TheoPascoli: You dont want to add a defauft parameter to result ? Instead of forcing return=None for commands that has nothing to return

TheoPascoli (changes requested):
Hello Martin, good idea !
I have a suggestion. You could use Generics to make this, this is the perfect example. You have something like :
```
@dataclass(frozen=True)
  class CommandResult(CommandApplicationResult, Generic[T]):
      data: T
```
And then you use it like: 

```
CommandResult[ThermalCluster](data=thermal)
```
So you can remove the 20+ classes thatyou created for each result.
Tell me what you think

It does not work if you plan to add specific result for some classes later on though

antarest/study/storage/variantstudy/model/command/common.py
MartinBelthle: I prefer not to. This will force devs that develop new commands to think about that. But I can change it

MartinBelthle:
> Hello Martin, good idea ! I have a suggestion. You could use Generics to make this, this is the perfect example. You have something like :
> 
> ```
> @dataclass(frozen=True)
>   class CommandResult(CommandApplicationResult, Generic[T]):
>       data: T
> ```
> 
> And then you use it like:
> 
> ```
> CommandResult[ThermalCluster](data=thermal)
> ```
> 
> So you can remove the 20+ classes thatyou created for each result. Tell me what you think
> 
> It does not work if you plan to add specific result for some classes later on though

It seems cleaner than what I did. I'll try to implement that

antarest/study/storage/variantstudy/model/command/convert_hydro_pmax.py
```diff
@@ -38,7 +38,8 @@ class ConvertHydroPmax(ICommand):
     @override
     def _apply_dao(self, study_data: StudyDao, listener: Optional[ICommandListener] = None) -> CommandOutput:
         study_data.convert_hydro_pmax(self.hydro_pmax)
-        return command_succeeded(message="Hydro pmax converted successfully.")
+        result = CommandResult[HydroPmax](data=self.hydro_pmax)
```
sylvlecl: I think you don't have to explicitly state the generic type in such cases

antarest/study/storage/variantstudy/model/command/common.py
```diff
@@ -9,15 +9,30 @@
...
+class CommandResult(CommandApplicationResult, Generic[T]):
+    data: T
 
 
 @dataclass(frozen=True)
 class CommandOutput:
     status: bool
     message: str
+    # Result of the command to be used by the managers to return the application result.
+    result: CommandApplicationResult | None = None
```
sylvlecl: if we go for generics, I don't think it's useful to keep the `CommandResult` and `CommandApplicationResult` intermediate class, let's just add a generic data here.

Note that it weakens a little the typing, when looking at that class the developer may not completely understand what it is. But seems OK.
Otherwise, we could just use a big type alias with all the classes that we want to return (`ThermalCluster | RenewableCluster | ...`).

I would go with just the generic for now

antarest/study/storage/variantstudy/model/command/common.py
```diff
@@ -9,15 +9,24 @@
...
+class CommandResult(Generic[T]):
+    data: T
 
 
 @dataclass(frozen=True)
 class CommandOutput:
     status: bool
     message: str
+    # Result of the command to be used by the managers to return the application result.
+    result: CommandResult[Any] | None = None
```
sylvlecl: I think we could just have `data: T` right here no?

sylvlecl (commented):
I think we'd better remove the remaining intermediate `CommandResult` but otherwise looks good !

PR #3086, opened 2026-03-09
fix(scenario-builder): change foreign keys

TheoPascoli:


antarest/study/dao/database/models/ruleset.py
```diff
@@ -28,7 +28,7 @@ def _create_area_scenario_table(name: str) -> Table:
         Column("study_id", String(36), nullable=False, primary_key=True),
         Column("area_id", String(255), nullable=False, primary_key=True),
         Column("value", JSON, nullable=False),
-        ForeignKeyConstraint(["study_id"], ["study.id"], ondelete="CASCADE"),
```
sylvlecl, line 31: We should probably add a test which shows that when an area is removed, the scenario builder data for that area is removed too (for ex. the load data).

antarest/tools/admin_lib.py
```diff
@@ -62,10 +62,7 @@ def reindex_table(config: Path) -> None:
 
     engine = sqlalchemy.create_engine(str(config_obj.db.db_admin_url), echo=False)
 
-    with engine.connect() as connection:
-        with connection.begin():
-            pass
-
+    with engine.connect() as connection, connection.begin():
```
smailio, line 65: hah didn't know this trick.

alembic/versions/6f8e2a9c1b3d_fix_ruleset_area_fk_to_area_table.py
```diff
@@ -0,0 +1,61 @@
...
+        sa.Column("study_id", sa.String(length=36), nullable=False),
+        sa.Column("area_id", sa.String(length=255), nullable=False),
+        sa.Column("value", sa.JSON(), nullable=False),
+        foreign_key,
+        sa.PrimaryKeyConstraint("study_id", "area_id"),
+    )
+    source_table = sa.table(table_name, *(sa.column(column_name) for column_name in columns))
+    tmp_table = sa.table(tmp_table_name, *(sa.column(column_name) for column_name in columns))
+    op.execute(sa.insert(tmp_table).from_select(columns, sa.select(*(source_table.c[column_name] for column_name in columns))))
+    op.drop_table(table_name)
```
smailio, line 50: Answers seems obvious but just to double check, deleting these rows won't cascade in other deletes right ? 

Also perf wise this migration seems heavy, but I guess there's no other way around it.

TheoPascoli: yea its heavy but  it must be this way to be compatible with sqlite and postgres. No real lighter way

sylvlecl (dismissed):

alembic/versions/6f8e2a9c1b3d_fix_ruleset_area_fk_to_area_table.py
sylvlecl: Note that we allowed ourselves to make such "breaking" migrations while the schema is not stabilized and user in prod, so no problem to do that kind of thing, we could even have been more radical and just delete everything :)

PR #3087, opened 2026-03-09
feat(study): handle directory copy

TheoPascoli:


webapp/src/routes/-shared/components/studies/dialogs/CopyStudyDialog.tsx
```diff
@@ -65,18 +72,35 @@ function CopyStudyDialog({ study, open, onClose }: Props) {
...
+      waitForCompletion: true,
+    });
+
+    if (copyTask.status !== TaskStatus.Completed || !copyTask.result?.success) {
+      throw toError(copyTask.result?.message ?? "Study copy failed");
+    }
+
+    await queryClient.refetchQueries({
+      queryKey: directoryQueries.list().queryKey,
+    });
```
hdinia: If the user copied to an existing directory, the cache is still valid so we dont need to invalidate. 
Also please use `invalidateQueries` instead of `refetchQueries` (refetch lazily via active observers, rather than forcing an immediate request)

webapp/src/routes/-shared/components/studies/dialogs/CopyStudyDialog.tsx
```diff
@@ -65,18 +72,35 @@ function CopyStudyDialog({ study, open, onClose }: Props) {
...
+
+    await dispatch(fetchStudies());
```
hdinia: We dont need to refetch studies here

webapp/src/routes/-shared/components/studies/dialogs/CopyStudyDialog.tsx
```diff
@@ -65,18 +72,35 @@ function CopyStudyDialog({ study, open, onClose }: Props) {
...
 
-    return copyStudy({
+    const copiedStudyId = await copyStudy({
       studyId: study.id,
       studyName: studyName.trim(),
       destinationFolder: directoryPath,
       outputIds,
     });
+
+    const copyTask = await getTask({
```
hdinia: This long-polls the backend until the copy task finishes. For large studies this can take some time, leaving the dialog frozen in a loading state during which the user can't close it or do anything else. The previous behavior was "submit and forget" (dialog closed immediately).

Is that intentional? I think we shouldn't block the UI. If the concern is the display of the study while the copy task is ongoing, we can handle it by exposing a dedicated field on the study model and using it on the frontend (studies list) to make it appear in a loading state while the task is pending.

hdinia (commented):
Hey, thanks for the frontend changes! 👍🏼  I do see a few points worth reconsidering though (check my comments for more context), but in the end you should be good with this code:

```ts
  const handleSubmit = async ({
    values: { studyName, destination, outputIds = [] },
  }: SubmitHandlerPlus<DefaultValues>) => {
    // TODO: This should be moved to the API layer when Tan stack migration is done.
    const directoryPath = toDirectoryPath(destination, directories);

    const taskId = await copyStudy({
      studyId: study.id,
      studyName: studyName.trim(),
      destinationFolder: directoryPath,
      outputIds,
    });

    // Poll task completion in the background, then refresh
    // the directory list so the dialog is not blocked while the task runs.
    getTask({ id: taskId, waitForCompletion: true })
      .then(() => {
        queryClient.invalidateQueries({ queryKey: directoryQueries.list().queryKey });
      })
      .catch((err) => {
        enqueueErrorSnackbar(t("studies.error.copyStudy"), toError(err));
      });

    return taskId;
  };

```

hdinia (dismissed):
Frontend part looks good to me thanks 👍🏼

hdinia (approved):

antarest/study/service.py
```diff
@@ -1233,6 +1233,7 @@ def copy_task(notifier: ITaskNotifier) -> TaskResult:
 
             study.owner = owner
             study.groups = groups
+            study.directory_id = self.directory_service.get_directory_by_path(destination_folder.as_posix())
```
sylvlecl, line 1236: Ah nice it was already available !

sylvlecl (approved):

PR #3088, opened 2026-03-09
fix(matrix): fix matrix format

TheoPascoli:
[ANT-4638]

Not sure if this is actually the bug happening in production, but it's an attempt to fix it

Bug: `get_all_matrices_on_the_filesystem` returned the stem of any file in the bucket directory (except lock files). If a file with an unknown extension was present, its stem would be treated as a matrix ID, but `infer_matrix_characteristics` couldn't find the actual file -> AssertionError. 
Fix: Only return stems of files whose extension is a known matrix format (tsv, hdf, parquet, feather).


antarest/matrixstore/repository.py
```diff
@@ -403,4 +403,5 @@ def infer_matrix_characteristics(self, matrix_id: str) -> Matrix:
         return Matrix(id=matrix_id, width=width, height=height, created_at=current_time(), version=version)
 
     def get_all_matrices_on_the_filesystem(self) -> set[str]:
-        return {f.stem for f in self.bucket_dir.iterdir() if not f.name.endswith(LOCK_SUFFIX)}
+        known_suffixes = {f".{fmt}" for fmt in InternalMatrixFormat}
+        return {f.stem for f in self.bucket_dir.iterdir() if f.suffix in known_suffixes}
```
sylvlecl: I looked up the error in production, it's a case where there is no suffix, so this should solve that issue indeed.

However it's also something that we should detect and remove, I think.

TheoPascoli: I made matrix sync detect invalid files and remove them during synchronization

sylvlecl (approved):

PR #3089, opened 2026-03-09
feat(xpansion): db dao de-normalized version

smailio:
This implementation has better perf and is less complex than the normalized implementation. 

However  even if the denormalized version seems relatively faster it's at most one ms faster than the normalized one. For such small gains I'd recommend to go with the normalized version. 

benchmark_comparison

### DIFF
--- a/antarest/study/dao/database/database_xpansion_dao.py
+++ b/antarest/study/dao/database/database_xpansion_dao.py
@@ -165,11 +165,8 @@ class DatabaseXpansionDao(XpansionDao):
 
     @override
     def checks_xpansion_settings_are_correct(self, settings: XpansionSettingsUpdate) -> None:
-        # File existence checks for additional_constraints and yearly_weights
-        # require blob storage access, which is not yet available in the database DAO.
         # TODO: validate additional_constraints and yearly_weights against blob storage
         #  once blob storage is wired up for database mode.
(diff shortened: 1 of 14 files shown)


antarest/study/dao/database/database_xpansion_dao.py
```diff
@@ -164,44 +158,22 @@ def save_xpansion_settings(self, settings: XpansionSettings) -> None:
...
-                XPANSION_SENSITIVITY_PROJECTION_TABLE.c.study_id == self._study_id
-            )
-        )
-        if sensitivity["projection"]:
-            try:
-                self._db_session.execute(
-                    insert(XPANSION_SENSITIVITY_PROJECTION_TABLE),
-                    [{"study_id": self._study_id, "candidate_name": name} for name in sensitivity["projection"]],
-                )
-            except IntegrityError:
```
smailio, line 181: The file study doesn't handle this validation. 

In the denormalized version I don't perform the validation.

However, In the normalized version the validation is enforced by the DB, I could ignore the IntegrityError just to align with the file stufy implementation but I think that would be unwise. 

Btw, the IntegrityError can be triggered both by duplicates projections lines or missing candidates, however in this PR I added validation at the model level for duplicates so we're sure at this stage of the code that if the IntegrityError it can only be the missing candidate.

sylvlecl:
We kept the normalized version, closing !
Many thanks for the benchmark

PR #3090, opened 2026-03-09
feat(xpansion): db dao normalized version

smailio:
Implements database storage for xpansion data, except for resources.

After benchmarking, it has been decided to store data such as "patterns" in a normalized way,
not as blobs, in order to benefit from FK checks and cascade deletes in particular.


antarest/study/dao/database/database_xpansion_dao.py
```diff
@@ -0,0 +1,357 @@
...
+                    & (XPANSION_CANDIDATE_TABLE.c.name == old_id)
+                )
+            )
+            assert isinstance(result, CursorResult)
+            if result.rowcount == 0:
+                raise CandidateNotFoundError(f"The candidate '{old_id}' does not exist")
+            self._db_session.execute(insert(XPANSION_CANDIDATE_TABLE).values(values))
+        else:
+            upsert_one(self._db_session, XPANSION_CANDIDATE_TABLE, values)
+        self._db_session.commit()
```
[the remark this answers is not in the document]
smailio: Nonsense

tests/study/dao/test_database_xpansion_dao.py
```diff
@@ -0,0 +1,448 @@
...
+        db_dao.delete_xpansion_candidate("cand")
+        assert db_dao.get_all_xpansion_candidates() == []
+
+        # --- delete unknown raises ---
+        with pytest.raises(CandidateNotFoundError):
+            db_dao.delete_xpansion_candidate("nonexistent")
+
+        with pytest.raises(CandidateNotFoundError):
+            db_dao.delete_xpansion_candidate("cand")
+
```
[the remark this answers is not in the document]
smailio: nonsense

antarest/study/dao/database/database_xpansion_dao.py
```diff
@@ -0,0 +1,357 @@
...
+        )
+        if sensitivity["projection"]:
+            try:
+                self._db_session.execute(
+                    insert(XPANSION_SENSITIVITY_PROJECTION_TABLE),
+                    [{"study_id": self._study_id, "candidate_name": name} for name in sensitivity["projection"]],
+                )
+            except IntegrityError:
+                self._db_session.rollback()
+                raise CandidateNotFoundError("One or more candidates in the projection do not exist")
```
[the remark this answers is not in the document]
smailio: Good point. Validation is now done at the model level for duplicates.

tests/study/dao/bench_database_xpansion_dao.py
```diff
@@ -0,0 +1,381 @@
+# Copyright (c) 2026, RTE (https://www.rte-france.com)
```
smailio: This file will be removed before merging it. But you can check how the benchmark was done if you want.

antarest/study/business/model/xpansion_model.py
```diff
@@ -282,3 +291,11 @@ class XpansionAdequacyCriterion(AntaresBaseModel, populate_by_name=True, alias_g
     stopping_threshold: float = Field(default=1e6, ge=0)
     criterion_count_threshold: float = Field(default=1, ge=0)
     patterns: list[XpansionAdequacyPattern] = Field(default_factory=list)
+
+    @field_validator("patterns")
+    @classmethod
+    def no_duplicate_areas(cls, v: list[XpansionAdequacyPattern]) -> list[XpansionAdequacyPattern]:
```
smailio, line 297: I double checked this buisness rule with Xpansion team.

antarest/study/dao/database/database_xpansion_dao.py
```diff
@@ -0,0 +1,357 @@
...
+        # Replace all patterns: delete existing rows then insert the new ones.
+        self._db_session.execute(
+            delete(XPANSION_ADEQUACY_PATTERN_TABLE).where(XPANSION_ADEQUACY_PATTERN_TABLE.c.study_id == self._study_id)
+        )
+        if criterion.patterns:
+            self._db_session.execute(
+                insert(XPANSION_ADEQUACY_PATTERN_TABLE),
+                [{"study_id": self._study_id, "area": p.area, "criterion": p.criterion} for p in criterion.patterns],
+            )
+        self._db_session.commit()
```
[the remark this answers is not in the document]
smailio: Added the check at the model level.

antarest/study/dao/database/database_xpansion_dao.py
```diff
@@ -0,0 +1,357 @@
...
+        stmt = select(XPANSION_CANDIDATE_TABLE).where(
+            (XPANSION_CANDIDATE_TABLE.c.study_id == self._study_id) & (XPANSION_CANDIDATE_TABLE.c.name == candidate_id)
+        )
+        row = self._db_session.execute(stmt).fetchone()
+        if not row:
+            raise CandidateNotFoundError(f"The candidate '{candidate_id}' does not exist")
+        return self._row_to_candidate(row)
+
+    @override
+    def save_xpansion_candidate(self, candidate: XpansionCandidate, old_id: Optional[str] = None) -> None:
```
sylvlecl, line 224: we could add some docstring to the interface method, because it's really not obvious what the "old_id" parameter is for (does not exist elsewhere)

antarest/study/dao/database/database_xpansion_dao.py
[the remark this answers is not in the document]
sylvlecl: It seems legit to me no, though?
If you want to rename a candidate from "joe" (old_id) to "bob" (candidate.name) but "bob" already exists, the insertion will raise ?

sylvlecl, line 259: suggestion to add a try/catch with rollback to handle the case where we delete the candidate but the insertion fails, to revert the deletion (AI proposition)

antarest/study/dao/database/database_xpansion_dao.py
```diff
@@ -0,0 +1,357 @@
...
+    @override
+    def delete_xpansion_candidate(self, candidate_name: str) -> None:
+        result = self._db_session.execute(
+            delete(XPANSION_CANDIDATE_TABLE).where(
+                (XPANSION_CANDIDATE_TABLE.c.study_id == self._study_id)
+                & (XPANSION_CANDIDATE_TABLE.c.name == candidate_name)
+            )
+        )
+        assert isinstance(result, CursorResult)
+        if result.rowcount == 0:
```
sylvlecl, line 270: nice optim to do it that way and not with 2 requests

antarest/study/dao/database/database_xpansion_dao.py
```diff
@@ -0,0 +1,357 @@
...
+    @abstractmethod
+    def get_impl(self) -> "DatabaseStudyDao":
+        pass
+
+    # ------------------------------------------------------------------
+    # Internal helpers
+    # ------------------------------------------------------------------
+
+    def _settings_row_exists(self) -> bool:
+        stmt = select(XPANSION_SETTINGS_TABLE.c.study_id).where(XPANSION_SETTINGS_TABLE.c.study_id == self._study_id)
```
sylvlecl, line 75: we might even use a "count" instead :)
really a detail

alembic/versions/3a7f9c2d1e8b_add_xpansion_tables.py
```diff
@@ -0,0 +1,133 @@
...
+        sa.Column("max_units", sa.Integer(), nullable=True),
+        sa.Column("max_investment", sa.Float(), nullable=True),
+        sa.Column("already_installed_capacity", sa.Integer(), nullable=True),
+        sa.Column("link_profile", sa.String(length=255), nullable=True),
+        sa.Column("already_installed_link_profile", sa.String(length=255), nullable=True),
+        sa.Column("direct_link_profile", sa.String(length=255), nullable=True),
+        sa.Column("indirect_link_profile", sa.String(length=255), nullable=True),
+        sa.Column("already_installed_direct_link_profile", sa.String(length=255), nullable=True),
+        sa.Column("already_installed_indirect_link_profile", sa.String(length=255), nullable=True),
+        sa.ForeignKeyConstraint(
```
sylvlecl, line 67: Should we have a FK constraint on the link too ?

alembic/versions/3a7f9c2d1e8b_add_xpansion_tables.py
```diff
@@ -0,0 +1,133 @@
...
+        sa.Column("relative_gap", sa.Float(), nullable=False),
+        sa.Column("relaxed_optimality_gap", sa.Float(), nullable=False),
+        sa.Column("max_iteration", sa.Integer(), nullable=False),
+        sa.Column("solver", sa.Enum("Cbc", "Coin", "Xpress", name="xpansion_solver"), nullable=False),
+        sa.Column("log_level", sa.Integer(), nullable=False),
+        sa.Column("separation_parameter", sa.Float(), nullable=False),
+        sa.Column("batch_size", sa.Integer(), nullable=False),
+        sa.Column("yearly_weights", sa.String(), nullable=False),
+        sa.Column("additional_constraints", sa.String(), nullable=False),
+        sa.Column("timelimit", sa.Integer(), nullable=False),
```
sylvlecl, line 36: a little weird to not have time_limit with underscore here but ok

alembic/versions/3a7f9c2d1e8b_add_xpansion_tables.py
sylvlecl: It seems that the FS implementation does not delete candidates when we delete a link (but it's likely something that has been forgotten), so here we are consistent with that behaviour, at least.

tests/study/dao/test_database_xpansion_dao.py
```diff
@@ -0,0 +1,448 @@
...
+    """Helper to build a minimal XpansionCandidate using the max_investment format."""
+    return XpansionCandidate(
+        name=name,
+        link=f"{area_from} - {area_to}",
+        annual_cost_per_mw=cost,
+        max_investment=5000.0,
+    )
+
+
+class TestXpansionConfiguration:
```
sylvlecl, line 67: A few tests could be added:
- tests with multiple studies to be sure that they do not leak on one another (missing in other DAOS too though ?)
- renaming a candidate to an existing name
- renaming a non existent candidate
- renaming a candidate removes it from projection:
  is it actually what we want ? Nothing is done in the filestudy, looks like a bug

sylvlecl (changes requested):
Looks very good overall, a few minor comments

antarest/study/dao/database/database_xpansion_dao.py
sylvlecl, line 224: An afterthought about this:
if we want to support a clean renaming, we'd better have an ID independant from the name.

Otherwise all data related to the renamed candidate will deleted.
(even worse in the FS implementation I think, though, because they're not even deleted, just left in place).

sylvlecl: I think we'd better remove the support for renaming for now, like for other objects ...
Let's not do it in that PR though

alembic/versions/3a7f9c2d1e8b_add_xpansion_tables.py
smailio: I added it. Hard choice between leaving DB in an inconsistent case, or having inconsistent behavior between  file study and database implementation :/

tests/study/dao/test_database_xpansion_dao.py
smailio: done

antarest/study/dao/database/database_xpansion_dao.py
smailio: You're right. Current logic to rename works but is a little weird.

antarest/study/dao/database/database_xpansion_dao.py
smailio: the count will read through all matching rows the fetch one will stop at first matching row.

antarest/study/dao/database/database_xpansion_dao.py
smailio: If the delete fails the upser won't run. Maybe I misunderstood something ?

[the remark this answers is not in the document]
smailio: Yep. I re wrote that part.

sylvlecl (dismissed):
great :)

sylvlecl (approved):

PR #3092, opened 2026-03-10
feat(ui-outputs): trigger search on input change and refactor district retrieval

skamril:
ANT-4716

### DIFF
--- a/webapp/src/redux/ducks/studyMaps.ts
+++ b/webapp/src/redux/ducks/studyMaps.ts
@@ -510,7 +510,7 @@ export const deleteStudyMapLayer = createAsyncThunk<
 });
 
 export const createStudyMapDistrict = createAsyncThunk<
-  StudyMapDistrict,
+  { studyId: StudyMetadata["id"]; district: StudyMapDistrict },
   {
     studyId: StudyMetadata["id"];
     name: StudyMapDistrict["name"];
@@ -527,7 +527,7 @@ export const createStudyMapDistrict = createAsyncThunk<
       output,
       data.comments,
     );
-    return { id, name, output, comments, areas };
+    return { studyId, district: { id, name, output, comments, areas } };
   } catch (error) {
     return rejectWithValue(error);
   }
@@ -535,11 +535,11 @@ export const createStudyMapDistrict = createAsyncThunk<
 
 export const updateStudyMapDistrict = createAsyncThunk<
   {
+    studyId: StudyMetadata["id"];
     districtId: StudyMapDistrict["id"];
     output: StudyMapDistrict["output"];
     comments: StudyMapDistrict["comments"];
     areas?: StudyMapDistrict["areas"];
-    applyFilter?: string;
   },
   {
     studyId: StudyMetadata["id"];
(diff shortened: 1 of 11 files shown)


webapp/src/redux/selectors.ts
```diff
@@ -246,6 +246,14 @@ export const getAreas = createSelector(getStudySynthesis, (synthesis) => {
   return [];
 });
 
+export const getDistricts = createSelector(getStudySynthesis, (synthesis) => {
+  if (synthesis) {
+    const districts = Object.values(synthesis.districts);
+    return sortByName(districts);
```
hdinia, line 252: Districts appear to already come sorted from the API response so the `sortByName` call here may be unnecessary. Worth confirming, but if that's the case it could be removed.

webapp/src/redux/selectors.ts
```diff
@@ -246,6 +246,14 @@ export const getAreas = createSelector(getStudySynthesis, (synthesis) => {
   return [];
 });
 
+export const getDistricts = createSelector(getStudySynthesis, (synthesis) => {
```
hdinia, line 249: It's a nice addition prevents a redundant API call 👍🏼  However I notice a potentiel bug (not due to this selector) :

`getDistricts` reads from `synthesis.districts` in the `studySyntheses` slice. 
But`createStudyMapDistrict` only updates `studyMaps.districts`. When a district is created, `synthesis.districts` is never updated, so the selector never re-runs with the new data. So new districts will never appear unless we reload the app.

To fix it i think adding dispatch `updateStudySynthesis` from within the `createStudyMapDistrict` thunk should be sufficient (not tested)

I let you test and confirm the issue, if the change involve a too large refactor you can create a follow-up PR

webapp/src/types/types.ts
```diff
@@ -298,12 +298,12 @@ export interface AreaWithId extends Area {
...
-  name?: string;
-  addAreas?: string[];
-  subtractAreas?: string[];
-  applyFilter?: string;
-  comments?: string;
+  name: string;
   output: boolean;
+  comments: string;
+  addAreas: string[];
+  subtractAreas: string[];
```
hdinia, line 309: Minor remark: The `District` type used here reflects the synthesis endpoint's internal model (`addAreas`, `subtractAreas`, `applyFilter`) rather than the cleaner `/districts` CRUD representation (`areas: string[]`). 
Ideally we'd read from the CRUD endpoint to stay consistent, but dropping synthesis from the store would require a significant refactor tied to the ongoing TanStack Query migration. 
Maybe you could add a comment explaining why synthesis is used here as a temporary source for districts?

webapp/src/redux/selectors.ts
skamril: The API does not sort the districts and returns them as an object. In JavaScript we shouldn’t rely on object property order for consistent sorting. Keeping `sortByName` guarantees the list is always ordered by name.

hdinia: Modern JS engines do preserve insertion order reliably this is part of the spec since since ES2017 for `Object` with string keys. So the fragility isn't in JavaScript's handling of object property order, it's in assuming the backend will always serialize keys in alphabetical order. Which appear to be the case here. If you prefer to keep it as a defensive safeguard it's fine, but it's indeed unnecessary if the API guarantees sort order

webapp/src/redux/selectors.ts
skamril: I fixed it for all the CRUD actions.

webapp/src/redux/selectors.ts
skamril: I’m not sure whether the API sorts by name (the id used as the index and the name can differ), but the state is now updated when a district is added or deleted without triggering a reorder each time.

webapp/src/redux/ducks/studyMaps.ts
hdinia: applyFilter input type still uses string instead of `DistrictApplyFilter`

hdinia (dismissed):

hdinia (approved):

PR #3093, opened 2026-03-10
feat(output): implement a retrocompatibility layer for /raw endpoint

TheoPascoli:


antarest/output/storage/output_storage.py
```diff
@@ -112,6 +112,7 @@ def import_output(
             output: either a path to a directory or a zip, or binary IO corresponding to the content of an archive.
             output_name_suffix: Optional name suffix to append to the output name, for example "hello" will
                                 appear at the end of the output name as "20201014-1422eco-hello"
+            logs: paths to simulation logs (out and err)
```
sylvlecl, line 166: :)

antarest/output/output_service.py
```diff
@@ -684,11 +684,36 @@ def create_aggregated_output_data_download(
...
         return download_id
 
+    def aggregate_output_data(
+        self,
+        uuid: str,
+        output_id: str,
+        query_file: QueryFileType,
+        frequency: MatrixFrequency,
+        columns_names: Sequence[str],
+        ids_to_consider: Sequence[str],
```
sylvlecl: I would prefer to have that method limited to one ID and one year, otherwise we still have the risk of use it for larger-than-memory data.
This should be enough for the compatibility layer, since it gets data for only 1 object at a time.

In that case we should probably name the method with another name like just `get_output_data` or `get_item_output_data`, since it's not really an "aggregation" anymore

antarest/output/utils.py
```diff
@@ -58,6 +58,99 @@ class MCAllLinksQueryFile(StrEnum):
...
+                break
+
+    if frequency is None or query_file is None:
+        return None
+
+    return RawOutputMatrixQuery(
+        output_id=path_parts[1],
+        query_file=query_file,
+        frequency=frequency,
+        ids_to_consider=[item_id],
```
sylvlecl: Here i's always one item --> let's remove the list

antarest/output/utils.py
```diff
@@ -58,6 +58,99 @@ class MCAllLinksQueryFile(StrEnum):
...
+    if len(path_parts) < 7 or path_parts[0] != "output":
+        return None
+
+    mc_root = path_parts[3]
+
+    if mc_root == "mc-ind":
+        if len(path_parts) < 8:
+            return None
+        try:
+            mc_years = [int(path_parts[4])]
```
sylvlecl: always one year --> let's remove the list

antarest/study/web/raw_studies_blueprint.py
```diff
@@ -187,6 +188,14 @@ def get_study_data(
         """
         logger.info(f"📘 Fetching data at {path} (depth={depth}) from study {uuid}")
 
+        if parse_raw_output_matrix_path([item for item in path.split("/") if item]) is not None and (
```
sylvlecl: we'd better move it to the study service method where we already have a similar handling for input matrices.

I think the second part of the condition is not right too, see handling of the formatted and matrix_format below:
"formatted" is kind of deprecated in favor of matrix format (probably should be marked as such), but we still support having only formatted=True in which case we serialize to JSON.

If we move this to study service, we won't have to care anymore, we will just go through the existing formatting code for matrices below

antarest/study/service.py
```diff
@@ -569,6 +572,20 @@ def archive_output(self, study_id: str, output_id: str) -> None:
     def write_output_to_dir(self, study_id: str, output_id: str, parent_dir: Path) -> None:
         raise NotImplementedError()
 
+    @abstractmethod
+    def aggregate_output_data(
```
sylvlecl: see comment about output service: we'd better restrict it and rename it too here

antarest/output/output_service.py
```diff
@@ -684,11 +684,36 @@ def create_aggregated_output_data_download(
...
+        transform_columns_headers: bool = True,
+        mc_years: Optional[Sequence[int]] = None,
+    ) -> pl.DataFrame:
+        chunks = list(
```
sylvlecl, line 700: would make sense to have a permission check here

antarest/output/output_service.py
```diff
@@ -684,11 +684,36 @@ def create_aggregated_output_data_download(
...
+                output_id=output_id,
+                query_file=query_file,
+                frequency=frequency,
+                ids_to_consider=ids_to_consider,
+                columns_names=columns_names,
+                transform_columns_headers=transform_columns_headers,
+                mc_years=mc_years,
+            )
+        )
+        return pl.concat(chunks, how="vertical_relaxed") if chunks else pl.DataFrame()
```
sylvlecl: Since we have just one item, I think we can just check there is only one DF and return it

antarest/output/filestudy/utils.py
```diff
@@ -58,6 +58,99 @@ class MCAllLinksQueryFile(StrEnum):
 
 QueryFileType: TypeAlias = MCIndAreasQueryFile | MCAllAreasQueryFile | MCIndLinksQueryFile | MCAllLinksQueryFile
 
+
+@dataclass(frozen=True)
+class RawOutputMatrixQuery:
```
sylvlecl, line 67: Nice to have that class to represent the query in a structured way

antarest/study/service.py
```diff
@@ -2592,6 +2609,46 @@ def normalize_study_by_id(self, study_id: str) -> None:
...
+        parsed: RawOutputMatrixQuery,
+        *,
+        transform_columns_headers: bool,
+    ) -> pl.DataFrame:
+        df = self._get_outputs_access().aggregate_output_data(
+            study_id=uuid,
+            output_id=parsed.output_id,
+            query_file=parsed.query_file,
+            frequency=parsed.frequency,
+            columns_names=[],  # empty means all columns
```
sylvlecl: we can remove that argument in the function definition, too

antarest/study/service.py
```diff
@@ -2592,6 +2609,46 @@ def normalize_study_by_id(self, study_id: str) -> None:
...
+            ids_to_consider=parsed.ids_to_consider,
+            transform_columns_headers=transform_columns_headers,
+            mc_years=parsed.mc_years,
+        )
+        metadata_cols = [col for col in RAW_OUTPUT_MATRIX_METADATA_COLUMNS if col in df.columns]
```
sylvlecl: This should be a responsibility of the output service instead:
since the method of the output service will take only one item ID and one year, it makes sense that it removes the related unnecessary columns too

antarest/study/service.py
```diff
@@ -2592,6 +2609,46 @@ def normalize_study_by_id(self, study_id: str) -> None:
...
+
+    def get_raw_output_matrix_dataframe(self, uuid: str, path: str) -> pl.DataFrame:
+        """
+        Returns an output matrix as a DataFrame for callers that explicitly need matrix serialization.
+        """
+        study = self.get_study(uuid)
+        assert_permission(study, StudyPermissionType.READ)
+        self.assert_study_unarchived(study)
+
+        parsed = parse_raw_output_matrix_path([item for item in path.split("/") if item])
```
sylvlecl: we'd better not parse it twice, once in the blueprint and once here.
This should disappear once we move the part which is in the blueprint to here, though

antarest/study/service.py
```diff
@@ -2607,13 +2664,21 @@ def get_raw_content(self, uuid: str, path: str, depth: int, formatted: bool) ->
         assert_permission(study, StudyPermissionType.READ)
         self.assert_study_unarchived(study)
 
+        url = [item for item in path.split("/") if item]
+
+        # Try to route output matrix requests
+        parsed = parse_raw_output_matrix_path(url)
+        if parsed is not None:
+            if formatted:
```
sylvlecl: we don't need to handle it here, it will be handled in the blueprint. Let's just return the dataframe

antarest/study/service.py
```diff
@@ -2592,6 +2609,46 @@ def normalize_study_by_id(self, study_id: str) -> None:
...
+        study = self.get_study(uuid)
+        assert_permission(study, StudyPermissionType.READ)
+        self.assert_study_unarchived(study)
+
+        parsed = parse_raw_output_matrix_path([item for item in path.split("/") if item])
+        if parsed is None:
+            raise IncorrectPathError(f"The provided path does not point to a valid output matrix: '{path}'")
+        return self._read_raw_output_matrix(uuid, parsed, transform_columns_headers=True)
+
+    def _get_raw_output_matrix_legacy_content(self, uuid: str, parsed: RawOutputMatrixQuery) -> JSON:
```
sylvlecl: I don't think we need this method: the service should just return the DF, and then the formatting is done by the web layer

antarest/output/output_service.py
```diff
@@ -684,11 +684,36 @@ def create_aggregated_output_data_download(
...
 
+    def aggregate_output_data(
+        self,
+        uuid: str,
+        output_id: str,
+        query_file: QueryFileType,
+        frequency: MatrixFrequency,
+        columns_names: Sequence[str],
+        ids_to_consider: Sequence[str],
+        transform_columns_headers: bool = True,
```
sylvlecl: should be remove for this endpoint too

sylvlecl (changes requested):
Looks pretty good already ! but a few things can be made simpler or moved

antarest/output/output_service.py
TheoPascoli: Kept: both call sites need different values: False for the legacy path (splittable column names), True for the new matrix format. Tell me if you have any idea

antarest/output/output_service.py
```diff
@@ -684,11 +685,64 @@ def create_aggregated_output_data_download(
 
         return download_id
 
+    def aggregate_output_data(
```
sylvlecl: I think this one can be removed now

antarest/study/service.py
sylvlecl: To be removed

antarest/study/service.py
sylvlecl: I still think we don't need this, it will be handled in the web layer by existing code (same as input matrices)

antarest/study/service.py
```diff
@@ -2607,13 +2654,33 @@ def get_raw_content(self, uuid: str, path: str, depth: int, formatted: bool) ->
...
+        if parsed is not None:
+            if formatted:
+                df = self._get_outputs_access().get_item_output_data(
+                    study_id=uuid,
+                    output_id=parsed.output_id,
+                    query_file=parsed.query_file,
+                    frequency=parsed.frequency,
+                    item_id=parsed.ids_to_consider,
+                    mc_year=parsed.mc_year,
+                    transform_columns_headers=False,
```
sylvlecl: don't think we need the False here, it's only meant to be used by the "download" endpoint which will not go through that method --> hence I think we can remove the parameter also from the methode

antarest/study/service.py
```diff
@@ -2600,6 +2617,20 @@ def normalize_study_by_id(self, study_id: str) -> None:
 
         self.storage_service.raw_study_service.normalize_study(study)
 
+    # def _read_raw_output_matrix(
```
sylvlecl, line 2855: to be removed

antarest/study/service.py
```diff
@@ -2615,13 +2646,30 @@ def get_raw_content(self, uuid: str, path: str, depth: int, formatted: bool) ->
...
+        if parsed is not None:
+            df = self._get_outputs_access().get_item_output_data(
+                study_id=uuid,
+                output_id=parsed.output_id,
+                query_file=parsed.query_file,
+                frequency=parsed.frequency,
+                item_id=parsed.ids_to_consider,
+                mc_year=parsed.mc_year,
+            )
+            return {
```
sylvlecl: We should just return the polars dataframe here so that it is then formatted by the web layer.
See that piece of code:
```python
if isinstance(output, pl.DataFrame):
    if matrix_format is None:
        matrix_format = MatrixFormat.JSON if formatted else MatrixFormat.PLAIN
    return matrix_format.serialize_dataframe(output)
```

Makes me think that we probably don't have integration tests that cover this correctly:
I think in the current state, the "formatting" options (formatted & matrix_format) of the endpoint will not be taken into account correctly, since they are used by the web layer only if this returns a dataframe.

--> we should add an integration test which checks the behaviour of those parameters for an output matrix (we can use the STA-mini study for that)

sylvlecl (changes requested):
I think there is still one thing to fix, so that formatting options are correctly taken into account

antarest/study/service.py
TheoPascoli: Done

PR #3094, opened 2026-03-10
feat(variant): move variants along with parent study

TheoPascoli:


antarest/study/service.py
```diff
@@ -1290,19 +1291,30 @@ def copy_task(notifier: ITaskNotifier) -> TaskResult:
...
-            study.directory_id = None
+        studies_to_move = [study, *self._get_variant_descendants(study.id)]
+        assert_permission_on_studies(studies_to_move, StudyPermissionType.WRITE)
+
+        directory_id = self.directory_service.get_directory_by_path(folder_dest) if folder_dest else None
+
+        for study_to_move in studies_to_move:
+            self._apply_study_move(study_to_move, folder_dest, directory_id)
+
+    def _get_variant_descendants(self, study_id: str) -> List[VariantStudy]:
```
sylvlecl, line 1307: we'd better have a method in the repo that would use a CTE to avoid performing one query per child

tests/integration/studies_blueprint/test_move.py
```diff
@@ -50,6 +50,55 @@ def test_move_endpoint(self, client: TestClient, internal_study_id: str, user_ac
         res.raise_for_status()
         res = client.get(f"/v1/studies/{study_id}")
         assert res.json()["folder"] is None
+        res = client.get(f"/v1/studies/{variant_id}")
+        assert res.json()["folder"] is None
+
+    def test_move_variant_is_rejected(self, client: TestClient, user_access_token: str) -> None:
+        client.headers = {"Authorization": f"Bearer {user_access_token}"}
```
sylvlecl, line 57: note that you may use the user_client fixture instead of having to request 2 fixtures and create the headers yourself

tests/integration/studies_blueprint/test_move.py
```diff
@@ -50,6 +50,55 @@ def test_move_endpoint(self, client: TestClient, internal_study_id: str, user_ac
...
+        assert res.status_code == 200
+
+        res = client.get(f"/v1/studies/{parent_id}")
+        parent = res.json()
+        res = client.get(f"/v1/studies/{variant_id}")
+        variant = res.json()
+        res = client.get(f"/v1/studies/{subvariant_id}")
+        subvariant = res.json()
+
+        assert parent["folder"] == f"project/subfolder/{parent_id}"
```
sylvlecl, line 96: completely unrelated but this makes me think that it's really weird that we have the ID here.
We should think again of removing this, now that directories are managed separately, I think there is no real risk of breaking anything.
(nothing to do in that PR of course)

tests/integration/variant_blueprint/test_variant_manager.py
```diff
@@ -283,6 +283,32 @@ def test_variant_manager(
         assert res.status_code == 404
 
 
+def test_create_variant_inherits_parent_directory_id(
+    client: TestClient,
+    admin_access_token: str,
+) -> None:
+    admin_headers = {"Authorization": f"Bearer {admin_access_token}"}
```
sylvlecl, line 290: you may use admin_client fixture instead

tests/integration/studies_blueprint/test_move.py
```diff
@@ -50,6 +50,55 @@ def test_move_endpoint(self, client: TestClient, internal_study_id: str, user_ac
...
+        res = client.post(f"/v1/studies/{study_id}/variants?name=Variant1")
+        assert res.status_code == 200
+        variant_id = res.json()
+
+        res = client.put(f"/v1/studies/{variant_id}/move", params={"folder_dest": "folder1"})
+        assert res.status_code == 400
+        assert res.json()["exception"] == "UnsupportedOperationOnThisStudyType"
+
+    def test_move_raw_propagates_to_variant_tree(self, client: TestClient, user_access_token: str) -> None:
+        client.headers = {"Authorization": f"Bearer {user_access_token}"}
```
sylvlecl, line 72: same, may use user_client

sylvlecl (changes requested):
Looks very good overall ! Minor request to add a method to get all children with only one request

sylvlecl (approved):

PR #3096, opened 2026-03-12
feat(ui-studies): add showDescendants toggle for managed tree directory filter

hdinia:
ANT-4625

  - Adds a showDescendants toggle to the managed tree, mirroring the existing folder hierarchy toggle for the external tree.
  when off -> only studies directly in the selected directory are shown
  when on -> studies in the selected directory and all descendants are shown.
  
  - Unifies `strictPath` (external) and `strictDirectory` (managed) into a single `showDescendants`: boolean field on both filter sub-objects, replacing two divergent names with one that reflects the display intent.
  
  - Collapses the two duplicate ToggleButtonGroup blocks in FilterControls into one, driven by a single showDescendants prop the correct value is resolved in the parent based on the active tree.


webapp/src/redux/middlewares/localStorageMiddleware.ts
```diff
@@ -82,11 +83,8 @@ localStorageMiddleware.startListening({
 // When a user opens a folder, it should remain open after a page refresh.
 localStorageMiddleware.startListening({
   actionCreator: updateStudyFilters,
-  effect: (action) => {
-    storage.setItem(StorageKey.StudiesFilters, (prev) => ({
-      ...prev,
-      ...action.payload,
-    }));
+  effect: (_, listenerApi) => {
+    storage.setItem(StorageKey.StudiesFilters, getStudyFilters(listenerApi.getState()));
```
skamril: The `search` filter should not be stored in localStorage. The filters from the filter panel should not be stored either (not sure if this was requested).

webapp/src/redux/ducks/studies.ts
```diff
@@ -303,7 +303,14 @@ export default createReducer(initialState, (builder) => {
       draftState.favorites = action.payload;
     })
     .addCase(updateStudyFilters, (draftState, action) => {
-      Object.assign(draftState.filters, action.payload);
+      const { managed, external, ...rest } = action.payload;
+      Object.assign(draftState.filters, rest);
```
skamril: 💡 `draftState.filters = R.mergeDeepRight(draftState.filters, action.payload)`

skamril (changes requested):

webapp/src/redux/middlewares/localStorageMiddleware.ts
hdinia: Indeed only the navigation state (activeTree, managed, external) should persist. I did a test and forgot to filter them out, good catch thanks

skamril (approved):

PR #3097, opened 2026-03-13
fix(ui-studies): cap explorer sections height to keep external tree visible

hdinia:
ANT-4654

- Cap section heights so both Managed and External tree sections are always visible in the study explorer pane, regardless of content size
- Collapsible sections : clicking a section header toggles it open/closed 
- Persist collapse state in Redux (backed by localStorage) : the managed.collapsed and external.collapsed flags survive page reloads
- Scrollbar moved inside each section : each TreeSection now owns its own CustomScrollbar instead of a single outer scrollbar wrappingboth trees
- Styles extracted : all variant styles, constants, and grid helpers moved to a new styles.ts file

>note: scroll restoration will come in a separate PR


sylvlecl (approved):
The behaviour seems really better to me :)

PR #3100, opened 2026-03-16
chore(fastapi): avoid rebuilding routes to speedup tests startup

sylvlecl:
The goal is mainly to speed up integration tests, so that we allow ourselves to rely even more on them
instead of hard to maintain unit tests at the services level, which often have little added value and/or are duplicates.

This PR uses fastapi dependency injection so that routes do not need to be re-built at each integration test execution.

At initialization time, services are set on the fastapi application state object, so that they are retrieved from there at dependency injection time.

For integration tests, the actual fastapi app is created only once. Then for each test, the services underneath are swapped for each test.

Another improvement is to initialize the admin user only once for all tests. The bcrypt hashing of the password had a significant cost.

Future possible improvement: do not unzip the whole STA-mini study and its outputs for all tests, only for those which need them.

**pytest execution on CI:**

- ubuntu: 10min -> 5min30
- windows: 19min -> 12min30


antarest/core/dependencies.py
```diff
@@ -0,0 +1,92 @@
...
+from antarest.login.service import LoginService
+from antarest.matrixstore.service import MatrixService
+from antarest.output.output_service import OutputService
+from antarest.study.directory_service import DirectoryService
+from antarest.study.service import StudyService
+from antarest.study.storage.explorer_service import Explorer
+from antarest.study.storage.rawstudy.watcher import Watcher
+
+
+def get_config(request: Request) -> Config:
```
sylvlecl: We'd better move those methods in the corresponding service modules

antarest/output/output_blueprint.py
```diff
@@ -541,6 +569,7 @@ def get_output_variables_view(
         variable_name: SanitizedStr,
         frequency: MatrixFrequency,
         type: OutputVariablesType,
+        output_service: OutputService = Depends(get_output_service),
```
sylvlecl: Maybe better to always have dependencies as the last argument

antarest/study/main.py
```diff
@@ -55,12 +47,11 @@ def build_study_service(
     study_service: Optional[StudyService] = None,
     generator_matrix_constants: Optional[GeneratorMatrixConstants] = None,
     event_bus: IEventBus = DummyEventBusService(),
-) -> StudyService:
+) -> Tuple[StudyService, DirectoryService]:
```
sylvlecl, line 50: We should move the directory service in its own build method, probably in its own python package actually.
But ok for that PR which is already large.

tests/integration/conftest.py
```diff
@@ -41,6 +48,39 @@
 RUN_ON_WINDOWS = sys.platform == "win32"
 
 
+def _render_config(config_path: Path, db_url: str, tmp_path: Path) -> None:
```
sylvlecl, line 50: Good thing to have extracted this

tests/integration/conftest.py
```diff
@@ -56,6 +96,74 @@ def initial_db_file(tmp_path_factory: TempPathFactory) -> Path:
...
+        title="AntaREST",
+        version=__version__,
+        docs_url=None,
+        root_path=config.root_path,
+        openapi_tags=tags_metadata,
+        openapi_url=f"{config.api_prefix}/openapi.json",
+    )
+
+    api_root = APIRouter(prefix=config.api_prefix)
+    ws_manager = register_all_routes(api_root, config)
```
sylvlecl: We could have the ws creation as a separate function, would probably be more clear

tests/integration/conftest.py
```diff
@@ -67,53 +175,45 @@ def db_path(tmp_path: Path, initial_db_file: Path) -> Path:
...
+    DBSessionMiddleware(None, custom_engine=per_test_engine, session_args=SESSION_ARGS)
+
+    init_admin_user(engine=per_test_engine, session_args=SESSION_ARGS, admin_password=config.security.admin_pwd)
+    services = create_services(config)
+
+    # Wire event bus to the shared websocket manager
+    connect_event_bus(services.event_bus, ws_manager)
+
+    # Store services on the shared app so Depends() resolves them
+    store_services_on_app(application, services, config)
```
sylvlecl: Now we don't use anymore the `fastapi_app` method used for real bootstrap of the app, it would be nice to share more code between the building here and the building for the real app.

antarest/login/main.py
```diff
@@ -33,11 +27,25 @@
     UserRepository,
 )
 from antarest.login.service import LoginService
-from antarest.login.web import create_login_api, create_user_api
+
+_login_service_ref: Optional[LoginService] = None
```
sylvlecl: Not great to have this global

tests/integration/conftest.py
```diff
@@ -56,6 +96,74 @@ def initial_db_file(tmp_path_factory: TempPathFactory) -> Path:
     return db_path
 
 
+@pytest.fixture(scope="session")
```
sylvlecl, line 100: Maybe the session scoped fixture could only create the routes and the connection manager, not the full app.

The creation of routes should not imply using a config which has dependencies to "per-test" temporary directories.

sylvlecl: Tested: if we don't created the full app, we don't get as much speedup

antarest/core/dependencies.py
```diff
@@ -0,0 +1,190 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+
+"""
+This module defines dependencies for all routes of the application.
+
+TODO: should probably be less "centralized" ?
```
MartinBelthle: Probably be better to at least separate methods from TypeAliases and also the `auth_required` method

antarest/dependencies.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    Returns an AuthJWT instance which can be used for authenticating the request.
+
+    Depends on config (in particular for disabling auth), and on login service to check for revoked tokens.
+    """
+
+    # TODO: that config part would better be done not on every request, not a blocker though
+    if not config.security.disabled:
+
+        @AuthJWT.load_config  # type: ignore
```
MartinBelthle, line 156: Why are we defining this here ? Shouldn't this be defined in the fast_api_jwt_auth package ?

antarest/eventbus/web.py
```diff
@@ -136,4 +132,22 @@ async def connect(
...
+    async def send_event_to_ws(event: Event) -> None:
+        event_data = event.model_dump()
+        del event_data["permissions"]
+        del event_data["channel"]
+        await manager.broadcast(to_json_string(event_data), event.permissions, event.channel)
+
     event_bus.add_listener(send_event_to_ws)
+
+
+def configure_websockets(api_root: APIRouter, event_bus: IEventBus) -> None:
```
MartinBelthle: Never used

antarest/favorite/web.py
```diff
@@ -10,49 +10,54 @@
...
 
 from antarest.core.api_types import UuidStr
-from antarest.core.config import Config
+from antarest.core.dependencies import FavoriteDirectoryServiceDep, FavoriteStudyServiceDep, auth_required
 from antarest.core.utils.web import APITag
 from antarest.favorite.model import FavoriteDirectoryDTO, FavoriteStudyDTO
-from antarest.favorite.service import FavoriteDirectoryService, FavoriteStudyService
-from antarest.login.auth import Auth, logger
+
+logger = logging.getLogger(__name__)
```
MartinBelthle, line 22: Nice

antarest/login/web.py
```diff
@@ -210,66 +196,53 @@ def get_bot(id: int, verbose: Optional[int] = None) -> BotIdentityDTO | BotDTO:
...
         resp = _generate_tokens(user, jwt_manager)
-
         return resp
 
     @bp.post(
         "/refresh",
         summary="Refresh access token",
     )
-    def refresh(jwt_manager: Annotated[AuthJWT, Depends(AuthJWT)]) -> CredentialsDTO:
+    def refresh(service: LoginServiceDep, jwt_manager: AuthDep) -> CredentialsDTO:
```
MartinBelthle, line 239: Not sure to fully understand all the changes you've done on the `Auth` could you summarize here ?

antarest/main.py
```diff
@@ -215,59 +288,70 @@ async def set_threadpool_size(app: FastAPI) -> AsyncGenerator[None, None]:
...
-            authjwt_access_token_expires=Auth.ACCESS_TOKEN_DURATION,
-            authjwt_refresh_token_expires=Auth.REFRESH_TOKEN_DURATION,
-            authjwt_cookie_csrf_protect=False,
-        )
+    application.state.ws_manager = ws_manager
 
+    # 2. Database
+    application.add_middleware(DBSessionMiddleware)
+
+    # 4. Middlewares
```
MartinBelthle: Minor but you go from 2 to 4

pyproject.toml
```diff
@@ -136,6 +136,8 @@ no_implicit_reexport = false
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
+log_cli = false  # you may set it to true for debugging purpose
```
MartinBelthle, line 139: Minor but this is the default value it seems so I don't think we need this

MartinBelthle (changes requested):

antarest/dependencies.py
MartinBelthle: Seems it's just code you moved

pyproject.toml
sylvlecl: Added this just as a helper for debugging. I find it painful to go and check the doc when I need it, but it may be only me

antarest/eventbus/web.py
sylvlecl: Indeed.
Also I refactored a little that part so that the "ConnectionManager" is now part of the state and injected as a dependency, like other services.

antarest/favorite/web.py
sylvlecl: I cannot claim I fixed it myself ^^

antarest/dependencies.py
sylvlecl: Yes. Here we just use the feature made available by the fast_api_jwt_auth, which is not great because the config is class-level, meaning it's not really contained in the app state. But not an issue in practive, we don't have mutliple apps in parallel.

antarest/login/web.py
sylvlecl: The changes are really just "moves", although I think that part would be worht a little refactoring to make it more understandable.

This `AuthDep` allows to get a `AuthJWT` (class providedd by the lib fastapi_jwt_auth), which itself can be used for several purpose:
- here it's used to generate the tokens that are sent back to the client
- in other endpoints, it is  used through the `auth_required` dependency which depends on `AuthDep` and validates the token contained in the `Request`, and sets the current user context

Finally the `AuthDep` itself depends on the config to check if security is disabled, and on the login service so that we can check the token is still in database (not revoked)

sylvlecl: Additional note:
the websocket connection also requires this dependency to check the token it receives as a message (not in thecookies of the request like standard HTTP requests)

antarest/core/dependencies.py
sylvlecl: About this, what I think after more analysis of alternatives:

having a unique file to centralize dependencies of the app seems the least bad option to me.
However, I will move it to top level alongside "main", it does not make sense to have it in the "core" package.

I would have preferred to dissociate more the **declaration** of dependencies which would go inside modules for example in `antarest.matrix.web` for `MatrixServiceDep` and its **definition**  at the app level, but it's not possible with fastapi to have this decoupling.
If we want to do that, we will need to introduce a real dependency injection library like https://dishka.readthedocs.io/en/stable/ , that would be a nice feature but not in that PR.

About the lack of typing:
to improve a little that aspect, we could add an `AppState` class to contain all services and the config, and store only this inside the starlette app.
**An opinion about this ?**

antarest/eventbus/main.py
```diff
@@ -14,25 +14,18 @@
...
 from antarest.core.config import Config
 from antarest.eventbus.business.local_eventbus import LocalEventBus
 from antarest.eventbus.business.redis_eventbus import RedisEventBus
 from antarest.eventbus.service import EventBusService
-from antarest.eventbus.web import configure_websockets
 
 
 def build_eventbus(
-    app_ctxt: Optional[AppBuildContext],
     config: Config,
```
TheoPascoli: This looks like it is not used, is it removable ?

TheoPascoli (changes requested):

antarest/dependencies.py
```diff
@@ -0,0 +1,226 @@
...
+from antarest.output.output_service import OutputService
+from antarest.service_creator import Services
+from antarest.study.directory_service import DirectoryService
+from antarest.study.service import StudyService
+from antarest.study.storage.explorer_service import Explorer
+from antarest.study.storage.rawstudy.watcher import Watcher
+
+
+@dataclass(frozen=True)
+class AppState:
```
sylvlecl, line 63: @MartinBelthle : I introduced that class, which is now the only one that needs to be set on the app state.

Improves a little the typing and the construction of the app, since there is only one "untyped" object to insert

MartinBelthle (approved):

TheoPascoli (approved):

PR #3101, opened 2026-03-16
feat(ui): add scroll position restoration

hdinia:


webapp/src/hooks/scroll/useScrollRestoration.ts
```diff
@@ -0,0 +1,61 @@
+/**
```
skamril, line 1: It’s still needed?

skamril (approved):

webapp/src/hooks/scroll/useScrollRestoration.ts
hdinia: It's not used at that moment, but good to keep for later if we need scroll restoration on any scrollable DOM element

PR #3102, opened 2026-03-16
fix(ui-studies): sync managed tree selection with navigation changes

hdinia:


skamril:
I validated it, but it would be helpful to know which root is selected as well (for example, by changing the color of the home icon for the selected one).

skamril (approved):

skamril (changes requested):
Finally, I found an issue when selecting subfolders.

Image

skamril (approved):

PR #3104, opened 2026-03-17
fix(study): incoherent state during long archive or unarchive

TheoPascoli:
- removes the use of long transactions for archival and unarchival
- adds a `/v1/studies/<id>/_repair` endpoint which can be used to repair inconsistencies between DB and file data, and possibly other stuff in the future


antarest/study/service.py
```diff
@@ -2107,21 +2148,31 @@ def unarchive(self, uuid: str) -> str:
...
+            archive_path = self.storage_service.raw_study_service.find_archive_path(study_to_unarchive)
+
+            self.storage_service.raw_study_service.unarchive(study_to_unarchive)
+
+            with db():
+                study_db = self.repository.get(uuid)
+                if study_db is None:
+                    raise StudyNotFoundError(uuid)
+                raw_study_db = assert_raw(study_db)
+                sync_raw_study_metadata(study_to_unarchive, raw_study_db)
```
sylvlecl: I **think** that session.merge can be used for that, instead of having to synchronize the state yourself, see https://docs.sqlalchemy.org/en/21/orm/session_state_management.html#merging

antarest/study/service.py
```diff
@@ -2107,21 +2148,31 @@ def unarchive(self, uuid: str) -> str:
             raise UnsupportedOperationOnThisStudyType(study.id, "unarchive", "raw")
 
         def unarchive_task(notifier: ITaskNotifier) -> TaskResult:
-            study_to_archive = self.get_study(uuid)
-            study_to_archive = assert_raw(study_to_archive)
-            self.storage_service.raw_study_service.unarchive(study_to_archive)
-            study_to_archive.archived = False
+            with db():
+                study_to_unarchive = clone_raw_study(assert_raw(self.get_study(uuid)))
```
sylvlecl, line 2174: Is it not enough to use the `session.expunge` you told me about, to make the object detached from the session and usable outside from the session ?

sylvlecl (commented):
questions ! I'm never 100% confortable with sqlalchemy ORM behaviour, but i wonder if we cannot use sqlalchemy methods expunge and merge to transfer the object from one session to the other

antarest/study/service.py
TheoPascoli: No because `expunge` only remove the object from the session. It does not turn it into a plain object. So lazy fields will fail on access (I saw it in a unit test, reading `study.workspace` that was not loaded).
Expunge would be enough if we are sure that every attribute that will be accessed later is already loaded, which is not the case here.
I could've done something like : 
```  
with db():
    study = assert_raw(self.get_study(uuid))
    _ = study.id, study.path, study.workspace, study.version, study.name
    db.session.expunge(study)
```
to load needed fields but it's too fragile to me

antarest/study/service.py
TheoPascoli: Yes indeed I can, its less explicit and intentionnal but its fine

sylvlecl (dismissed):

antarest/study/web/studies_blueprint.py
```diff
@@ -557,6 +558,14 @@ def unarchive_study(study_id: UuidStr) -> str:
         logger.info(f"Unarchiving study {study_id}")
         return study_service.unarchive(study_id)
 
+    @bp.post(
+        "/studies/{study_id}/repair",
```
sylvlecl: I think we'd better signal it as non public using the underscore convention _repair

antarest/study/service.py
```diff
@@ -2141,6 +2204,168 @@ def unarchive_task(notifier: ITaskNotifier) -> TaskResult:
             custom_event_messages=None,
         )
 
+    def repair_study(self, uuid: str, repair_request: StudyRepairRequest) -> str:
+        current_user = get_current_user()
```
sylvlecl: you may use `require_current_user` to make sure the user is indeed authenticated.

You could add a `require_admin_user` utility method to perform the admin check.

antarest/study/service.py
```diff
@@ -2141,6 +2204,168 @@ def unarchive_task(notifier: ITaskNotifier) -> TaskResult:
...
+        return report
+
+    def _repair_archive_consistency(
+        self,
+        uuid: str,
+        dry_run: bool,
+        report: StudyRepairReport,
+        notifier: ITaskNotifier,
+    ) -> None:
+        notifier.notify_message(f"Checking archive consistency for study '{uuid}'")
```
sylvlecl, line 2250: We also need to add some logs at the critical steps of the process, to enable analysis through log aggregators

antarest/study/service.py
```diff
@@ -2141,6 +2204,168 @@ def unarchive_task(notifier: ITaskNotifier) -> TaskResult:
...
+        report.issues.append(
+            StudyRepairIssue(
+                code="archive_consistency_db_fs_mismatch",
+                severity=StudyRepairSeverity.ERROR,
+                message=repair_spec.issue_message,
+            )
+        )
+        report.proposed_actions.append(proposed_action)
+        if not dry_run:
+            self._set_study_archived_state(uuid, archived=repair_spec.target_archived)
```
sylvlecl, line 2360: Here we need to add a log, and also in the cas of a dry run

antarest/study/service.py
```diff
@@ -2141,6 +2204,168 @@ def unarchive_task(notifier: ITaskNotifier) -> TaskResult:
...
+
+    def _set_study_archived_state(self, uuid: str, archived: bool) -> None:
+        with db():
+            study_db = self.repository.get(uuid)
+            if study_db is None:
+                raise StudyNotFoundError(uuid)
+            raw_study_db = assert_raw(study_db)
+            raw_study_db.archived = archived
+            self.repository.save(raw_study_db)
+            self.event_bus.push(
```
sylvlecl: Not that useful for such a feature

antarest/study/model.py
```diff
@@ -553,10 +553,47 @@ def _normalize_tags(cls, v: List[str]) -> List[str]:
...
+    ARCHIVE_CONSISTENCY = "archive_consistency"
+
+
+class StudyRepairSeverity(StrEnum):
+    INFO = "info"
+    WARNING = "warning"
+    ERROR = "error"
+
+
+class StudyRepairRequest(AntaresBaseModel):
```
sylvlecl, line 566: Let's use camel case for the models, for consistency sake

tests/integration/studies_blueprint/test_repair.py
```diff
@@ -0,0 +1,94 @@
...
+from antarest.core.tasks.model import TaskStatus
+from antarest.core.utils.fastapi_sqlalchemy import db
+from antarest.study.model import Study, StudyRepairReport
+from tests.integration.utils import wait_task_completion
+
+
+class TestRepairStudy:
+    def test_repair_archive_consistency_endpoint(
+        self,
+        client: TestClient,
```
sylvlecl, line 26: you may use the admin_client fixture instead of client + settings headers

tests/integration/studies_blueprint/test_repair.py
```diff
@@ -0,0 +1,94 @@
...
+
+        res = client.post(
+            f"/v1/studies/{study_id}/repair",
+            headers=admin_headers,
+            json={"repairs": ["archive_consistency"], "dry_run": True},
+        )
+        assert res.status_code == 200, res.json()
+        dry_run_task = wait_task_completion(client, admin_access_token, res.json())
+        assert dry_run_task.status == TaskStatus.COMPLETED, dry_run_task
+        assert dry_run_task.result is not None
```
sylvlecl, line 82: Would probably be simpler to just assert an equality with a hard coded json, same for subsequent assertions

tests/study/service/test_service.py
```diff
@@ -1931,6 +1933,157 @@ def test_task_upgrade_study(tmp_path: Path) -> None:
         )
 
 
+@with_db_context
```
sylvlecl: To me those tests are redundant with your integration test, they will be a maintainability burden.

Let's delete them if that sounds OK. 

We need to add a test for authentication on the integration side though. I also think the admin check would be better placed on the web layer, in the code, although it's not homogeneous.

sylvlecl (changes requested):
Looks good, a few comments:
- we need to add some logging
- for me we can get rid of redundant unit tests based on mocking
- we need to use camel case for consistency sake

sylvlecl (approved):

PR #3106, opened 2026-03-17
fix(output): aggregation failed on columns mismatch

sylvlecl:
In more details, the aggregation failed when a file was created with some columns missing compared to previous schema:

- 1st file with A and B
- 2nd file with B and C --> creates a new writer for schema B and C, instead of A, B, C
- 3rd file with A and B --> failure, because A is missing
-

### DIFF
--- a/antarest/core/serde/parquet_writer.py
+++ b/antarest/core/serde/parquet_writer.py
@@ -60,8 +60,7 @@ def write_dataframes_in_parquet_format_by_column_sets(
 
         first_df = _adapt_polars_schema(first_df)
         table = first_df.to_arrow()
-        current_schema = table.schema
-        current_writer = _parquet_writer(file_path, current_schema)
+        current_writer = _parquet_writer(file_path, table.schema)
         current_writer.write_table(table)
 
         while True:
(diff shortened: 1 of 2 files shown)


antarest/core/serde/parquet_writer.py
```diff
@@ -74,22 +73,21 @@ def write_dataframes_in_parquet_format_by_column_sets(
                         existing_columns.add(col)
                         new_index.append(col)
 
+                if df.columns != new_index:
```
sylvlecl, line 76: This section adapts the schema of the table to the current set of columns.
It must be done also when creating a new file, not only when appending.

tests/core/serde/test_parquet_writer.py
```diff
@@ -26,9 +26,12 @@ def test_different_columns(tmp_path: Path) -> None:
     dataframes = iter(
         [
             pl.DataFrame(data=[(1, 2), (3, 4)], schema=["A", "B"], orient="row"),
+            # will cause a new file creation with schema A,B,C
             pl.DataFrame(data=[(5, 6, 7), (8, 9, 10)], schema=["A", "B", "C"], orient="row"),
             pl.DataFrame(data=[(11, 12), (13, 14)], schema=["A", "B"], orient="row"),
+            # will cause a new file creation with schema A,B,C,D
             pl.DataFrame(data=[(15, 16, 17), (18, 19, 20)], schema=["B", "A", "D"], orient="row"),
+            pl.DataFrame(data=[(21, 22), (23, 24)], schema=["C", "D"], orient="row"),
```
sylvlecl, line 34: Failed before the fix, because the current schema was B, A, D, instead of A, B, C, D

MartinBelthle (approved):
Nice :)

PR #3107, opened 2026-03-18
fix(output): restore settings data used by one client, in /outputs API

sylvlecl:
Some of the settings removed in previous refacto was actually used by one of our clients.
Here we restore juste the necessary data, so that it's well identified.

Checked that it solves the issue with our client.


antarest/study/dtos.py
```diff
@@ -24,6 +28,32 @@
 )
 
 
+class OutputSynthesis(AntaresBaseModel):
```
sylvlecl, line 31: For the synthesis endpoint, we keep the less verbose content.

tests/integration/output_blueprint/test_output_blueprint.py
```diff
@@ -0,0 +1,176 @@
...
+    # in particular the "settings" section.
+    # If it's to be removed in the future, we need to coordinate with them.
+    assert res.json() == [
+        {
+            "archived": False,
+            "byYear": True,
+            "mode": "Economy",
+            "name": "20201014-1422eco-hello",
+            "nbYears": 1,
+            "settings": {
```
sylvlecl, line 32: This is the data that's still needed

antarest/output/storage/output_storage.py
```diff
@@ -73,6 +113,8 @@ class OutputDetails(AntaresBaseModel):
     nb_years: int
     archived: bool
 
+    settings: OutputSettings | None = Field(deprecated=True, default=None)
```
sylvlecl, line 116: deprecated to discourage its usage in new clients, we'd better have a more structured, separate endpoint

TheoPascoli (approved):

PR #3108, opened 2026-03-18
feat(disk_space_analyzer): add disk space analyzer task

mehdiwahada:
What does it do : it realizes an analysis of the disk space took by each studies and returns the number of updated studies

Why do we implement that : It is done in order to get statistics per user without scanning the disk each time and improving disk-usage endpoint performances


alembic/versions/81typ5zz6v1y_add_study_disk_space_analysis_table.py
```diff
@@ -0,0 +1,28 @@
...
+down_revision = 'ebcae82af0c3'
+branch_labels = None
+depends_on = None
+
+def upgrade():
+    op.create_table(
+        "study_disk_space_analysis",
+        sa.Column("study_id", sa.String(), nullable=False),
+        sa.Column("disk_space", sa.Integer(), nullable=False),
+        sa.Column("last_analysis_date", sa.Date(), nullable=False),
```
sylvlecl: it should be DateTime() instead of Date(), not sure it will work properly otherwise

antarest/maintenance/tasks/common.py
```diff
@@ -98,6 +98,7 @@ class LockId(IntEnum):
     WATCHER_SCAN = 1004
     VARIABLE_VIEW_GC = 1005
     TASKS_GC = 1006
+    STUDY_DISK_SPACE = 1008
```
sylvlecl: would better be moved under 1007, otherwise the next developer who adds a value will be tempted to use 1008 again

antarest/maintenance/tasks/disk_space_analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def disk_space_analysis(service: StudyService, disk_repo: StudyDiskSpaceRepository) -> DiskSpaceAnalyzerTaskResult:
+    start_time = time.time()
+    # we're giving admin access to the disk space analyzer due to the search_studies method
+    studies = service.repository.get_all(StudyFilter(access_permissions=AccessPermissions(is_admin=True)))
+    updated_studies = 0
+
+    try:
+        with db():
```
sylvlecl, line 43: Not related to your work, and we will leave it this way for that PR, but I notice that here we hold a datavase session for a potentially very long task, which we want to avoid, in general.

To remove that, we would need to refactor the lock mechanism, because for now the lock holds the session for all its lifetime.

@TheoPascoli I think we should reworkd this so that the lock only gets the session once on creation and then once on release, but not during the whole lifetime.

antarest/maintenance/tasks/disk_space_analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+        repository.save(study_disk_analysis)
+    else:
+        usage = study_service.get_disk_usage(study_id)
+        repository.update(study_id, usage)
+
+
+def disk_space_analysis(service: StudyService, disk_repo: StudyDiskSpaceRepository) -> DiskSpaceAnalyzerTaskResult:
+    start_time = time.time()
+    # we're giving admin access to the disk space analyzer due to the search_studies method
+    studies = service.repository.get_all(StudyFilter(access_permissions=AccessPermissions(is_admin=True)))
```
sylvlecl: There are chances that it  does not work since we're outside of the `with db()` block.
This is not seen in the integration test because it uses the  `with_db` decorator : I think we should remove it so that we better mimic the real runtime behaviour.

You should also perform a "real life" test starting the app and a celery worker, to make sure it works

tests/integration/maintenance_blueprint/test_disk_space_analyzer.py
```diff
@@ -0,0 +1,79 @@
...
+from tests.helpers import with_db_context
+
+
+@pytest.fixture
+def study_disk_repo() -> StudyDiskSpaceRepository:
+    return StudyDiskSpaceRepository()
+
+
+class TestDiskSpaceAnalyzerIntegration:
+    @with_db_context
```
sylvlecl: We'd better not use that decorator, because we want to check that the task gets the DB session itself

antarest/maintenance/tasks/disk_space_analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+    updated_studies = 0
+
+    try:
+        with db():
+            with create_lock(db.session, lock_id=LockId.STUDY_DISK_SPACE):
+                for study in studies:
+                    if disk_repo.get(study.id) is None:
+                        disk_space_analysis_per_study(disk_repo, service, study.id)
```
sylvlecl: we'd better add a try catch around each study, so that if the analysis fails for any random reason, like a study which has been deleted since, it does not interrrupt the whole analysis

antarest/maintenance/tasks/disk_space_analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+                        updated_studies += 1
+
+    except LockNotAcquired:
```
sylvlecl, line 76: We should add a more generic except for `Exception` and return a "failed" result in that case, see `gc_matrix` for example

antarest/study/model.py
```diff
@@ -444,6 +444,24 @@ def to_enhanced_json_summary(self) -> Any:
...
+    __tablename__ = "study_disk_space_analysis"
+
+    study_id: Mapped[str] = mapped_column(
+        String(255),
+        ForeignKey("study.id", name="fk_study_disk_space_id", ondelete="CASCADE"),
+        nullable=False,
+        primary_key=True,
+    )
+    disk_space: Mapped[int] = mapped_column(Integer, nullable=False)
+    last_analysis_date: Mapped[datetime] = mapped_column(DateTime, nullable=False)
```
sylvlecl, line 462: Here it's datetime, good

antarest/maintenance/tasks/disk_space_analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+) -> None:
+    study_disk_analysis = repository.get(study_id)
+
+    if not study_disk_analysis:
+        study_disk_analysis = StudyDiskSpaceAnalysis(
+            study_id=study_id, disk_space=study_service.get_disk_usage(study_id), last_analysis_date=datetime.now()
+        )
+        repository.save(study_disk_analysis)
+    else:
+        usage = study_service.get_disk_usage(study_id)
```
sylvlecl: We should do this **only** when the "updated_at" field of the study is more recent than the "last_analysis_date".

It's very important, otherwise we will spend a lot of time scanning the filesystem for studies that actually have not changed

alembic/versions/81typ5zz6v1y_add_study_disk_space_analysis_table.py
```diff
@@ -0,0 +1,28 @@
...
+revision = '81typ5zz6v1y'
+down_revision = 'ebcae82af0c3'
+branch_labels = None
+depends_on = None
+
+def upgrade():
+    op.create_table(
+        "study_disk_space_analysis",
+        sa.Column("study_id", sa.String(), nullable=False),
+        sa.Column("disk_space", sa.Integer(), nullable=False),
```
sylvlecl: Maye we could add the unit to the field name, otherwise it's not obvious what it is (bytes ?)

tests/integration/maintenance_blueprint/test_disk_space_analyzer.py
```diff
@@ -0,0 +1,79 @@
...
+            disk_space_analysis(service=study_service, disk_repo=study_disk_repo)
+
+            recent_analysis_date_1 = study_disk_repo.get(study_1).last_analysis_date
+            recent_analysis_date_2 = study_disk_repo.get(study_2).last_analysis_date
+            delta_1 = datetime.datetime.now() - recent_analysis_date_1
+            delta_2 = datetime.datetime.now() - recent_analysis_date_2
+
+            assert delta_1.seconds / 60 < 1
+            assert delta_2.seconds / 60 < 1
+
```
sylvlecl, line 69: We should also check that the cout of analyzed studies is 0, for the second run

antarest/study/repository.py
```diff
@@ -623,3 +624,41 @@ def get_directory_paths_bulk(self, directory_ids: List[str]) -> dict[str, str]:
...
+        session.commit()
+
+        return disk_analysis
+
+    def update(self, study_id: str, disk_space: int) -> None:
+        session = self.session
+
+        stmt = (
+            update(StudyDiskSpaceAnalysis)
+            .options(joinedload(StudyDiskSpaceAnalysis.study))
```
sylvlecl: not sure that joinedload is useful ?

sylvlecl (changes requested):
Nice! a few things to refine though

tests/integration/maintenance_blueprint/test_disk_space_analyzer.py
```diff
@@ -0,0 +1,79 @@
...
+            assert recent_analysis_date_1 == past_analysis_date_1
+            assert recent_analysis_date_2 == past_analysis_date_2
```
sylvlecl, line 71: We should also add a third phase to check that after updating a study, for example creating an area in study 1, if you re-run the analysis, the disk space is updated for that study (and not others)

tests/integration/maintenance_blueprint/test_gc_tasks.py
```diff
@@ -29,10 +29,10 @@ class TestTasksGCIntegration:
     @with_admin_user
     def test_clean_tasks_gc(self, task_service: ITaskService, task_job_repository: TaskJobRepository):
         task_1 = TaskJob(
-            id="1", status=TaskStatus.RUNNING.value, name="task_1", creation_date=datetime(2026, 1, 1, 10, 0, 0)
```
sylvlecl, line 32: We should remove changes in this file, they are not related to the PR

antarest/study/repository.py
```diff
@@ -623,3 +624,41 @@ def get_directory_paths_bulk(self, directory_ids: List[str]) -> dict[str, str]:
...
+            .where(StudyDiskSpaceAnalysis.study_id == study_id)
+            .values(disk_space=disk_space, analysis_date=datetime.now())
```
sylvlecl: the field is named `last_analysis_date`, not `analysis_date`, this will probably crash

antarest/maintenance/tasks/disk_space_analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+    start_time = time.time()
+    # we're giving admin access to the disk space analyzer due to the search_studies method
+    studies = service.repository.get_all(StudyFilter(access_permissions=AccessPermissions(is_admin=True)))
+    updated_studies = 0
+
+    try:
+        with db():
+            with create_lock(db.session, lock_id=LockId.STUDY_DISK_SPACE):
+                for study in studies:
+                    if disk_repo.get(study.id) is None:
```
sylvlecl: Note that we should check if the data must be updated even if the data is already present in DB --> this check needs to be removed

antarest/maintenance/tasks/disk_space_analyzer_task.py
```diff
@@ -0,0 +1,24 @@
...
+from antarest.maintenance.app import MaintenanceTask, TaskName, celery_app
+from antarest.maintenance.tasks.disk_space_analyzer import DiskSpaceAnalyzerTaskResult, disk_space_analysis
+
+
+@celery_app.task(base=MaintenanceTask, bind=True, name=TaskName.DISK_SPACE_ANALYZER, pydantic=True)
+def disk_space_analyzer_task(self: MaintenanceTask) -> DiskSpaceAnalyzerTaskResult:
+    """Celery task to analyze disk space usage by study."""
+    ctx = self.context
+
+    return disk_space_analysis(ctx.study_service, ctx.study_disk_space_repository)
```
sylvlecl: Like for the auto_archive_task, we should execute this as the admin user in order to be sure we have read rights on all studies:
```python
    with current_user_context(DEFAULT_ADMIN_USER):
        ...
```

alembic/versions/81typ5zz6v1y_add_study_disk_space_analysis_table.py
sylvlecl: We should use BigInteger because Integer may not be enough to represent the size of large studies (over 2G), see documentation of postrgesql:
https://www.postgresql.org/docs/current/datatype-numeric.html

antarest/study/repository.py
```diff
@@ -623,3 +624,41 @@ def get_directory_paths_bulk(self, directory_ids: List[str]) -> dict[str, str]:
...
+    def __init__(self, session: Optional[Session] = None):
+        self._session = session
+
+    @property
+    def session(self) -> Session:
+        if self._session is None:
+            return db.session
+        return self._session
+
+    def get_study(self, study_id: str) -> Optional[Study]:
```
sylvlecl: seems not usd

antarest/maintenance/tasks/disk_space_analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+    status: BackGroundTaskStatus
+    duration_seconds: float = 0
+    updated_studies: int = 0
+    reason: Optional[str] = None
+
+
+def disk_space_analysis_per_study(
+    repository: StudyDiskSpaceRepository, study_service: StudyService, study_id: str
+) -> None:
+    study_disk_analysis = repository.get(study_id)
```
sylvlecl: There is room for optimization here: we have the typical "N+1" query problem:
for example in production, if we have 10000 studies, we will perform 10000 requests, whereas we could have first retrieved data for all studies in only one query

--> let's do this instead

sylvlecl (changes requested):
More thorough review

antarest/core/config.py
```diff
@@ -215,6 +215,10 @@ class StorageConfig:
     tasks_gc_retention_days: int = 30
     tasks_gc_sleeping_time: int = 86400
     tasks_gc_dry_run: bool = False
+    disk_usage_log_sleeping_time: int = 300
+    disk_usage_log_cron: str = "0 20-23,0-7 * * * "
```
sylvlecl: let's make it every hour, not only from 20h to 7h

antarest/core/config.py
```diff
@@ -215,6 +215,10 @@ class StorageConfig:
...
+    disk_space_analyzer_sleeping_time: int = 300
+    disk_space_analyzer_cron: str = "0 20-23,0-7 * * * "
```
sylvlecl: let's make it once every night when there should be not much activity, let's say at 1am

antarest/study/model.py
```diff
@@ -444,6 +444,24 @@ def to_enhanced_json_summary(self) -> Any:
...
+
+    __tablename__ = "study_disk_space_analysis"
+
+    study_id: Mapped[str] = mapped_column(
+        String(255),
+        ForeignKey("study.id", name="fk_study_disk_space_id", ondelete="CASCADE"),
+        nullable=False,
+        primary_key=True,
+    )
+    disk_space_bytes: Mapped[int] = mapped_column(Integer, nullable=False)
```
sylvlecl: We should use BigInteger like in the migration

antarest/study/repository.py
```diff
@@ -619,3 +619,42 @@ def get_directory_paths_bulk(self, directory_ids: list[str]) -> dict[str, str]:
...
+
+        return disk_analysis
+
+    def update(self, study_id: str, disk_space: int) -> None:
+        session = self.session
+
+        stmt = (
+            update(StudyDiskSpaceAnalysis)
+            .where(StudyDiskSpaceAnalysis.study_id == study_id)
+            .values(disk_space=disk_space, last_analysis_date=datetime.now())
```
sylvlecl: The column `disk_space` does not exist: it's surprising if tests pass for the update. Should be disk_space_bytes instead

antarest/maintenance/tasks/disk_space_analyzer.py
```diff
@@ -0,0 +1,94 @@
...
+    try:
+        with db():
+            # we're giving admin access to the disk space analyzer due to the search_studies method
+            studies = service.repository.get_all(StudyFilter(access_permissions=AccessPermissions(is_admin=True)))
+            disk_analysis = disk_repo.get_all()
+            with create_lock(db.session, lock_id=LockId.STUDY_DISK_SPACE):
+                for study in studies:
+                    updated_at = study.updated_at
+                    try:
+                        list_analysis = [da for da in disk_analysis if da.study_id == study.id]
```
sylvlecl: a quite important optimization we should use:
here you have an O(n*n) complexity since you perform a linear search n times (once for each study).
Instead, we should first build a dictionary study_id -> disk_analysis outside the loop, and then look into it

antarest/maintenance/tasks/disk_space_analyzer.py
```diff
@@ -0,0 +1,94 @@
...
+
+def disk_space_analysis(service: StudyService, disk_repo: StudyDiskSpaceRepository) -> DiskSpaceAnalyzerTaskResult:
+    start_time = time.time()
+
+    updated_studies = 0
+
+    try:
+        with db():
+            # we're giving admin access to the disk space analyzer due to the search_studies method
+            studies = service.repository.get_all(StudyFilter(access_permissions=AccessPermissions(is_admin=True)))
```
sylvlecl: Those 2 requests should be done inside the "lock", otherwise we maye perform them whereas the processing will be skipped

antarest/maintenance/tasks/disk_space_analyzer_task.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Args:
+        sender: Celery app instance
+        storage: StorageConfig instance
+    """
+    if storage.disk_space_analyzer_cron:
+        try:
+            schedule = parse_cron_string(storage.disk_space_analyzer_cron)
+            sender.add_periodic_task(schedule, disk_space_analyzer_task.s(), name=TaskName.DISK_SPACE_ANALYZER)
+            logger.info(f"Disk usage logging registered with cron: {storage.disk_usage_log_cron}")
```
sylvlecl: Log must be fixed (disk usage log instead of disk space analyzer)

sylvlecl (commented):
A few more comments: some optimization and there is likely a bug where we refer to a column by a wrong name

tests/integration/maintenance_blueprint/test_disk_space_analyzer.py
```diff
@@ -0,0 +1,90 @@
...
+
+            with db():
+                area_1 = study_service.create_area(study_1, AreaCreation(name="fr"))
+                area_2 = study_service.create_area(study_1, AreaCreation(name="be"))
+
+                study_service.create_link(study_1, Link(area1=area_1.id, area2=area_2.id))
+
+            result = disk_space_analysis(service=study_service, disk_repo=study_disk_repo)
+
+            assert result.updated_studies == 1
```
sylvlecl, line 88: You should also verify that the study size and the last analysis date has changed in the table, otherwise the test does not show that the data has indeed been updated.

antarest/maintenance/tasks/disk_space_analyzer.py
```diff
@@ -0,0 +1,97 @@
...
+                # we're giving admin access to the disk space analyzer due to the search_studies method
+                studies = service.repository.get_all(StudyFilter(access_permissions=AccessPermissions(is_admin=True)))
+                disk_analysis = disk_repo.get_all()
+
+                dict_analysis = {element.study_id: element for element in disk_analysis}
+
+                for study in studies:
+                    updated_at = study.updated_at
+                    try:
+                        filtered_analysis = dict_analysis[study.id] if study.id in dict_analysis else None
```
sylvlecl: minor: it's more efficient to do `dict_analysis.get(study.id, None)`, it avoids to look in the dictionary twice (once too check the existence, then once more to get the value)

sylvlecl (commented):
just one more request to improve test and we're good

sylvlecl (approved):
Nice feature, thank you for the contribution :)

PR #3110, opened 2026-03-19
fix(study): fix raw updated time

TheoPascoli:


sylvlecl (approved):
Looks good if tests pass !

PR #3121, opened 2026-03-24
refactor(solver): use `SolverVersion` inside config instead of `str`

MartinBelthle:


tests/launcher/test_slurm_launcher.py
```diff
@@ -261,16 +261,16 @@ def test_extra_parameters(launcher_config: SlurmConfig) -> None:
 @pytest.mark.parametrize(
     "version, launcher_called, job_status",
     [
-        (840, True, JobStatus.RUNNING),
-        (860, False, JobStatus.FAILED),
+        (SolverVersion.parse(840), True, JobStatus.RUNNING),
+        (SolverVersion.parse(860), False, JobStatus.FAILED),
         pytest.param(
             999, False, JobStatus.FAILED, marks=pytest.mark.xfail(raises=VersionNotSupportedError, strict=True)
```
TheoPascoli: 999 should be `SolverVersion.parse(999)` ?

TheoPascoli (changes requested):
In `test_init_slurm_launcher_parameters`, you should call SolverVersion as well.

Is it intentionnal that you removed the fallback in `_select_best_binary` ? Previously, if the exact version wasn't  
  configured, it would pick the closest higher versio, now it raises UnsupportedStudyVersion instead

MartinBelthle:
Yes the change on `select_best_binary` is intentional as to me it doesn't make any sense to use another solver without telling the user. If we don't have the real solver version we'll raise with a clear message and he will be able to change on his side

TheoPascoli (approved):

PR #3122, opened 2026-03-25
refactor(ui-tasks): migrate tasks API to Zod/TanStack standards with GET endpoint

hdinia:


webapp/src/services/api/tasks/types.ts
```diff
@@ -12,60 +12,41 @@
...
+  getTaskParamsSchema,
+  getTasksParamsSchema,
+  taskLogSchema,
+  taskResultSchema,
+  taskSchema,
+} from "./schemas";
 
-export type TaskStatusValue = O.UnionOf<typeof TaskStatus>;
+////////////////////////////////////////////////////////////////
+// Enums
```
skamril: These are union types, not enums

webapp/src/services/api/tasks/schemas.ts
```diff
@@ -0,0 +1,89 @@
...
+// Response Schemas
+////////////////////////////////////////////////////////////////
+
+export const taskResultSchema = z
+  .object({
+    success: z.boolean(),
+    message: z.string(),
+    return_value: z.string().optional(),
+  })
+  .transform((dto) => ({
```
skamril: 💡 Consider adding `alias_generator=to_camel, populate_by_name=True` to the API model instead of handling the transformation on the frontend.

webapp/src/routes/_authenticated/studies/$studyId/-components/FreezeStudy.tsx
```diff
@@ -31,8 +31,8 @@ import { useTranslation } from "react-i18next";
 import useStudy from "../-hooks/useStudy";
 
 interface BlockingTask {
-  id: TaskDTO["id"];
-  type: TaskTypeValue;
+  id: Task["id"];
+  type: string;
```
skamril: Keep `TaskTypeValue`.

skamril (changes requested):

webapp/src/services/api/tasks/schemas.ts
```diff
@@ -0,0 +1,89 @@
...
+  id: z.string(),
+  message: z.string(),
+});
+
+export const taskSchema = z
+  .object({
+    id: z.string(),
+    name: z.string(),
+    status: z.enum(TaskStatus),
+    type: z.string().optional(),
```
skamril: `type: z.enum(TaskType).optional()`

Should this be defined in the API model as `type: TaskType | None = None`? @MartinBelthle

skamril (changes requested):

webapp/src/services/api/tasks/schemas.ts
MartinBelthle: Mmmh indeed it would be better, I tried it and had no mypy issue, we should try to do this and see it it breaks tests. If no test breaks let's do that. If it breaks more tests, identify me again i'll check

webapp/src/services/api/tasks/schemas.ts
hdinia: That's better, but I thought it might break the backend tests, I'll try

webapp/src/services/api/tasks/schemas.ts
hdinia: Updated it @MartinBelthle , I let you check the backend changes

MartinBelthle (dismissed):
Okay for the back-end part :)

webapp/src/queries/tasks/keys.ts
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+export const taskKeys = {
+  all: ["tasks"] as const,
+  lists: () => [...taskKeys.all, "list"] as const,
+  list: (params?: object) => [...taskKeys.lists(), params] as const,
+  detail: (id: string) => [...taskKeys.all, "detail", id] as const,
```
skamril, line 21: Add the valid type for `list()`:

`list: (params?: GetTasksParams)`

skamril (dismissed):

skamril (approved):

PR #3123, opened 2026-03-25
feat(output): implement parquet storage

TheoPascoli:


antarest/output/service.py
```diff
@@ -399,7 +399,9 @@ def import_output(
         # Optimized path for studies stored on external devices, that will then be unarchived there.
         # TODO: as commented elsewhere, that workflow should be refactored to not span multiple files
         if output_id and isinstance(output, Path) and output.suffix == ArchiveFormat.ZIP and auto_unzip:
-            self.unarchive_output(uuid, output_id)
+            storage = self._find_output_storage(uuid, output_id)
+            if storage.is_output_archived(uuid, output_id):
```
MartinBelthle: Why do we need this check ? There could be differences between the previous one and this one ?

antarest/output/storage/v2/converter.py
```diff
@@ -0,0 +1,388 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+
+"""
+Conversion of Antares output CSV files to parquet format.
```
MartinBelthle: Really minor but they are `.txt` files in TSV format so we should use one of the 2

antarest/output/storage/v2/converter.py
```diff
@@ -0,0 +1,388 @@
...
+    QueryFileType,
+    get_start_column,
+    normalize_df_column_names,
+    parse_output_file,
+)
+from antarest.study.model import MatrixFrequency
+
+logger = logging.getLogger(__name__)
+
+_OBJECT_TYPE_NAMES: dict[str, str] = {
```
MartinBelthle: We also use this kind of mapping inside `file_output_utils.py` and also I think we should find a way to not declare this directly and use Enums as we use for McInd and McAllAreasQueryFile

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,388 @@
...
+logger = logging.getLogger(__name__)
+
+_OBJECT_TYPE_NAMES: dict[str, str] = {
+    "details": "thermal_clusters",
+    "details-res": "renewable_clusters",
+    "details-STstorage": "short_term_storages",
+}
+
+
+def parquet_output_dir(variables_dir: Path, study_id: str, output_name: str) -> Path:
```
MartinBelthle, line 58: Minor but shouldn't this be moved inside the `storage.py` file ?

antarest/output/storage/v2/converter.py
```diff
@@ -0,0 +1,388 @@
...
+    file_type_class: type[QueryFileType] = MCIndLinksQueryFile if mc_root == MCRoot.MC_IND else MCAllLinksQueryFile
+    ref_folders = [links_path / lid for lid in link_ids]
+    combos = _discover_file_type_frequencies(ref_folders, file_type_class)
+
+    for query_file, frequency in combos:
+        file_name = _parquet_file_name(mc_root, "links", frequency)
+        _aggregate_to_parquet(output_dir, query_file, frequency, link_ids, target_dir / file_name)
+
+
+def _find_mode_dir(output_dir: Path) -> Path | None:
```
MartinBelthle: I think we should put this code in common with the current `aggregator_management.py` as it always thinks it's `economy`

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,388 @@
...
+def extract_output_to_parquet(output_dir: Path, target_dir: Path) -> None:
+    target_dir.mkdir(parents=True, exist_ok=True)
+
+    mode_dir = _find_mode_dir(output_dir)
+    if mode_dir is None:
+        logger.warning(f"No economy or adequacy directory found in {output_dir}")
+        return
+
+    for mc_root in (MCRoot.MC_IND, MCRoot.MC_ALL):
+        mc_root_path = mode_dir / str(mc_root.value)
```
MartinBelthle, line 295: Minor: no need for `str()`

antarest/output/storage/v2/converter.py
```diff
@@ -0,0 +1,388 @@
...
+    base_path: Path,
+    mc_root: MCRoot,
+    target_dir: Path,
+) -> None:
+    areas_path = base_path / "areas"
+    if not areas_path.exists():
+        return
+
+    all_ids = [d.name for d in areas_path.iterdir() if d.is_dir()]
+    area_ids = [a for a in all_ids if not a.startswith("@")]
```
MartinBelthle: Really minor but you iterate twice on the list, you could do it in one

antarest/output/storage/v2/converter.py
```diff
@@ -0,0 +1,388 @@
...
+    district_ids = [a for a in all_ids if a.startswith("@")]
```
MartinBelthle: As of today, the aggregation treats districts and areas to be the same thing. Do we want to separate these, I'm not really sure, I don't see the benefit of it as there are no districts specific endpoints for aggregation. To be discussed with @sylvlecl

antarest/output/storage/v2/converter.py
```diff
@@ -0,0 +1,388 @@
...
+
+    file_type_class: type[QueryFileType] = MCIndAreasQueryFile if mc_root == MCRoot.MC_IND else MCAllAreasQueryFile
+
+    ref_folders = [areas_path / a for a in (area_ids + district_ids)]
```
MartinBelthle: area_ids + districts_ids is just all_ids no ?

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,388 @@
...
+
+def _parquet_file_name(mc_root: MCRoot, object_type: str, frequency: MatrixFrequency) -> str:
+    return f"{mc_root.value}_{object_type}_{frequency.value}.parquet"
+
+
+_SKIPPED_QUERY_FILES = {"id"}
+"""Query file types that should not be converted to parquet (metadata files, not variable data)."""
+
+
+def _discover_file_type_frequencies(
```
MartinBelthle, line 70: Seems you could also put code in common or reuse the function `_filter_files_with_same_prefix` inside `file_output_utils.py`

antarest/output/storage/v2/converter.py
```diff
@@ -0,0 +1,388 @@
...
+            output_data = parse_output_file(file, start_col)
+        except Exception as e:
+            logger.debug(f"Skipping binding constraint {file.name}: {e}")
+            continue
+
+        df = output_data.data
+        headers = cast(MultipleOutputHeaders, output_data.headers)
+        col_names = normalize_df_column_names(mc_root, headers)
+        df.columns = col_names
+        df = df.with_row_index("timeId", offset=1)
```
MartinBelthle: timeId and mcYear are declared in `utils.py` so you could use them

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,388 @@
...
+
+    id_cols = [c for c in ["area", "link", "mcYear"] if c in combined.columns]
+    if id_cols:
+        combined = combined.sort(id_cols)
+
+    table = combined.to_arrow()
+    sorting_columns = [
+        pq.SortingColumn(table.schema.get_field_index(col)) for col in id_cols if col in table.schema.names
+    ]
+    pq.write_table(
```
MartinBelthle: Should put this method in a function as you do this twice in this file

MartinBelthle: Also you use row_group_size to 60*1024 but your table does not always have that number of rows, don't know how this works as syvlain said in the gopro page that we'd had to do a little bit of code to always have the good size

MartinBelthle (changes requested):
We should adapt the `test_aggregate_raw_data.py` test which contains a lot of cases and complicated ones to test we have the same results no matter how we store the output

tests/integration/conftest.py
```diff
@@ -55,6 +55,7 @@ def _render_config(config_path: Path, db_url: str, tmp_path: Path) -> None:
     default_workspace = tmp_path / "internal_workspace"
     ext_workspace_path = tmp_path / "ext_workspace"
     output_archive_dir = tmp_path / "output_archives"
+    output_variables_dir = tmp_path / "output_variables"
```
sylvlecl, line 58: We should add some integration test to check the behaviour.
We can probably re-use the existing ones that are quite exhaustive, and make them run on the same output but stored as "v2"

antarest/output/storage/v2/storage.py
```diff
@@ -195,7 +212,8 @@ def import_output(
 
             simulation_range = _extract_simulation_range(dir_path)
 
-            # TODO here: extract variable data
+            variables_target = parquet_output_dir(self._variables_dir, study_id, output_name)
+            extract_output_to_parquet(dir_path, variables_target)
```
sylvlecl, line 216: Note that we will also want to:
- delete the directory on archival
- re-build it on unarchival, using the content of the archive

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,388 @@
...
+
+    if mc_years and mc_root == MCRoot.MC_IND:
+        lazy = lazy.filter(pl.col("mcYear").is_in(list(mc_years)))
+
+    if columns_names:
+        schema_names = lazy.collect_schema().names()
+        selected = _filter_columns(schema_names, columns_names, mc_root, is_details)
+        lazy = lazy.select(selected)
+
+    return lazy.collect()
```
sylvlecl: We want to use the streaming engine, otherwise we will have larger than RAM data in our RAM (well, the process will be killed before this of course :) )

But this makes me think that it's not enough:
the result of "collect" will still be larger than RAM, even when using the streaming engine.
We really need to return an iterable of "small enough" dataframes.

So we need to use `collect_batches` instead which seems to do just that

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,388 @@
...
+    manager = AggregatorManager(
+        output_path=output_dir,
+        query_file=query_file,
+        frequency=frequency,
+        ids_to_consider=ids_to_consider,
+        columns_names=[],
+        transform_columns_headers=True,
+    )
+    try:
+        all_dfs = list(manager.aggregate_output_data())
```
sylvlecl: we must at all cost avoid to gather all data at once in memory, this will be larger than RAM for most real studies --> we need to keep working with iterables (generators), not lists

antarest/output/service.py
TheoPascoli: I had an issue that I cant reproduce at the moment ... I removed it.

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,388 @@
...
+    except Exception as e:
+        logger.debug(f"Skipping {query_file.value}-{frequency.value}: {e}")
+        return
+
+    if not all_dfs:
+        return
+
+    combined = pl.concat(all_dfs)
```
sylvlecl: Same here, we must not concat all DFs in memory, instead we need to stream them to the parquet file:
for this, you may re-use the `write_dataframes_stream_parquet` function / adapt it / or take inspiration from it.

sylvlecl: Actually, you will need to re-use the whole workflow
aggregate -> "write_dataframes_in_parquet_format_by_column_sets" -> "write_dataframes_stream_parquet", because otherwise you will get the same issues that we get in the aggregation methods:

not all dataframes are guaranteed to have the same columns, so we cannot just aggregate them to one parquet file in one round. We need to first write parquet files for the encountered "column sets" (like columns (A, b), then columns (A, B, C) and so on), and then write one final parquet file with all the columns.

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,388 @@
...
+def _extract_binding_constraints(
+    base_path: Path,
+    mc_root: MCRoot,
+    target_dir: Path,
+) -> None:
+    bc_path = base_path / "binding_constraints"
+    if not bc_path.exists():
+        return
+
+    for file in bc_path.iterdir():
```
sylvlecl: it seems that for mc-ind data, we don't iterate on all the years folder, so I guess we'll have only data for the first year, not for all years.

It would make sense to add the binding constraint to the existing aggregation code, to "benefit" from thhe fact that this kind of thing is already implemented there. However I know it's ugly and hard to evolve.

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+def read_output_from_parquet(
+    target_dir: Path,
+    query_file: QueryFileType,
+    frequency: MatrixFrequency,
+    ids_to_consider: Sequence[str],
+    columns_names: Sequence[str],
+    mc_years: Sequence[int] | None,
+) -> pl.DataFrame:
```
sylvlecl: See other comments:
here we lust absolutely return an iterable of dataframe (probably from collect_batches).

antarest/output/storage/v2/variables_storage.py
sylvlecl: Looks fine here to me, it's really the folder INSIDE the variables dir.

But we should probably rename the file to something like variables_storage, since it does not only handle conversion, but also reading from the files

sylvlecl (changes requested):
The structure looks good overall, but there are some important things to adapt in particular to have it working with larger than memory data, which will be the general case.
We must absolutely have functions manipulating iterables/streams of dataframes, and not concatenated dataframes, to limit memory usage.

antarest/output/storage/v2/converter.py
TheoPascoli: I created a method that a use at both places: here and in `file_output_utils`

antarest/output/storage/v2/converter.py
TheoPascoli: fixed

antarest/output/storage/v2/variables_storage.py
TheoPascoli: Pycharm not happy without str

antarest/output/storage/v2/variables_storage.py
TheoPascoli: They have differents purposes and I dont see the gain to mutualise this

antarest/output/filestudy/utils.py
```diff
@@ -63,6 +63,21 @@ class MCAllLinksQueryFile(StrEnum):
 MultipleOutputHeaders: TypeAlias = list[list[str]]
 
 
+def get_output_object_type(file_type: QueryFileType, is_link: bool) -> str:
```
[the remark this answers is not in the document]
TheoPascoli: One more useless comment like this and I uninstall you

TheoPascoli:
> We should adapt the `test_aggregate_raw_data.py` test which contains a lot of cases and complicated ones to test we have the same results no matter how we store the output

I added the unit test for the v2 storage

TheoPascoli:
> The structure looks good overall, but there are some important things to adapt in particular to have it working with larger than memory data, which will be the general case. We must absolutely have functions manipulating iterables/streams of dataframes, and not concatenated dataframes, to limit memory usage.

1. Streaming write:replaced list() + pl.concat() with the existing streaming workflow (write_by_column_sets -> yield -> stream_write), with a shutil.move shortcut for the single-file case.                                       
2. Streaming read: collect() -> collect_batches(), return Iterator[pl.DataFrame] end-to-end.                       
3. mc-ind binding constraints: iterate all year folders, not just the first one, adding a mcYear column.          
4. Archive/unarchive: delete parquet dir on archive, rebuild from archive on unarchive

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,405 @@
...
+
+    if mc_years and mc_root == MCRoot.MC_IND:
+        lazy = lazy.filter(pl.col(MCYEAR_COL).is_in(list(mc_years)))
+
+    if columns_names:
+        schema_names = lazy.collect_schema().names()
+        selected = _filter_columns(schema_names, columns_names, mc_root, is_details)
+        lazy = lazy.select(selected)
+
+    yield from lazy.collect_batches()
```
sylvlecl, line 370: Nice.
Just seeing that polars [doc](https://docs.pola.rs/api/python/dev/reference/lazyframe/api/polars.LazyFrame.collect_batches.html) warns that the feature is not stable and can be slower than "native sinks". In the future we could benchmark this, to check whether it's better to first write to an intermediate parquet file using `sink_parquet` and then read batches from it.

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,405 @@
...
+        columns_names=[],
+        transform_columns_headers=True,
+    )
+    try:
+        dataframes = manager.aggregate_output_data()
+    except Exception as e:
+        logger.debug(f"Skipping {query_file.value}-{frequency.value}: {e}")
+        return
+
+    intermediate_dir = Path(tempfile.mkdtemp())
```
sylvlecl: In consistency with what we do everywhere else for large temporary files, we should provide our own configured temporary dir here. Otherwise there is a risk that we fill up entirely the default temp dir.

To be honest I am thinking that it would a better approach to just configure the TEMPIR in our environment ... 
We should probably do that instead and get rid of the config of the temporary dir in storage

sylvlecl: Note tha we may use `with TemporaryDirectory` instead of the try/finally

antarest/output/storage/v2/variables_storage.py
```diff
@@ -0,0 +1,405 @@
...
+        output_path=output_dir,
+        query_file=query_file,
+        frequency=frequency,
+        ids_to_consider=ids_to_consider,
+        columns_names=[],
+        transform_columns_headers=True,
+    )
+    try:
+        dataframes = manager.aggregate_output_data()
+    except Exception as e:
```
sylvlecl: Seems dangerous to silently swallow this exception, in particular without a specific exception type : is it necessary ?
In production, debug logs are not enabled, so this will really be completely invisible.

tests/integration/output/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+    to their V2 names (which may differ due to timezone-dependent name reconstruction).
+    """
+    if request.param == "file":
+        return "file", {}
+
+    client.headers = {"Authorization": f"Bearer {user_access_token}"}
+
+    # Convert one at a time to track name changes reliably
+    output_name_map: dict[str, str] = {}
+    for output_id in _OUTPUT_IDS:
```
sylvlecl, line 51: The logic is hard to follow, we'd better hard code the names ourselves since we know them.
As the comment explains, we need to take into account the timezone, but we may use the `utc_to_local` function for that

tests/integration/output/test_output_variables.py
```diff
@@ -171,9 +171,9 @@ def test_get_output_variables_view(client: TestClient, user_access_token: str, i
     assert res.status_code == 404
     # Materialize the data
     task_id = client.post(f"{url}/materialize", params=query_params).json()
-    # Ask for data with materializing in process -> Should return a 404 with a status `IN_PROGRESS`
     res = client.get(f"{url}/data", params=query_params)
-    assert res.json() == {"status": "IN_PROGRESS", "task_id": task_id}
+    if res.status_code == 404:
```
sylvlecl, line 175: if not, we should assert it's the task is completed without waiting for it, and that the status code is 200.

tests/integration/output/test_aggregate_raw_data.py
```diff
@@ -25,6 +25,18 @@
 from tests.integration.assets import ASSETS_DIR as INTEGRATION_ASSETS_DIR
 from tests.integration.raw_studies_blueprint.assets import ASSETS_DIR
 
+_ID_COLUMNS = ["area", "link", "mcYear", "timeId", "cluster"]
+
+# Query file types not supported by V2 parquet storage (skipped during conversion)
+_V2_UNSUPPORTED_QUERY_FILES = {"id"}
+
+
+def _sort_by_id(df: pd.DataFrame) -> pd.DataFrame:
```
sylvlecl, line 34: It seems to me we should not need this:
we'd better have a reliable sort order in our result.

In particular, it's important that data is sorted in the parquet files we store, for retrieval performance.
But it may be important also for the clients which will perform similar queries.

sylvlecl (changes requested):
Nice :) few more comment in particular about the sort order, i will also try it on some real data which I have not done yet !

sylvlecl:
I've been able to import a large output without failure :) (Indeed I had to change my TMPDIR env var though)
Took something like 20 minutes but that was expected.

On the bright side:
 - 5 or 10x smaller output
 - around 10x faster "aggregation" endpoints 

Great !!

sylvlecl (approved):
Greaaat feature, thanks !

PR #3124, opened 2026-03-25
refactor(version)!: change format inside some endpoints

MartinBelthle:
[ANT-2959]

The version used to be returned as an integer like 800 or 920. Now it's a string like "8.0" or "9.2"

This will break the R scripts

Impacted endpoints:
- GET /studies
- PUT /studies/{uuid}
- GET /studies/{uuid}/synthesis
- GET /studies/{uuid}/parents


tests/integration/studies_blueprint/test_study_version.py
```diff
@@ -62,7 +62,7 @@ def test_nominal_case(self, client: TestClient, user_access_token: str, tmp_path
             # Gets study information
             res = client.get(f"v1/studies/{study_id}")
             res.raise_for_status()
-            assert res.json()["version"] == 920 if f == final_path else 700
+            assert res.json()["version"] == "9.2" if f == final_path else "7"
```
TheoPascoli: You should write `assert res.json()["version"] == ("9.2" if f == final_path else "7")
`

TheoPascoli (dismissed):
nice

PR #921, opened 2022-05-24
Issue 918 Update Studies page

skamril:

PR #3126, opened 2026-03-26
feat(ui-studies,ui-directories): add support for favoriting studies and managed directories via the new API

skamril:
also add a new menu to display favorites when the main menu is collapsed

ANT-8687


MartinBelthle (commented):
Okay for the back-end changes in the src code but you should modify the tests are they are not passing anymore

webapp/src/routes/_authenticated/-components/Container/FavoritesMenu/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+import { getDescendantIds } from "../../../studies/-components/StudyTree/ManagedTree/utils";
+import SidebarItem from "../SidebarItem";
+import { type Favorite } from "./types";
+import useFavorites from "./useFavorites";
+
+function FavoritesMenu() {
+  const { t } = useTranslation();
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
+  const isMainMenuOpen = useAppSelector(isMenuOpen);
+  const dispatch = useDispatch();
```
hdinia: should be useAppDispatch no

webapp/src/queries/utils.ts
```diff
@@ -12,24 +12,74 @@
...
+
+interface QueryListItemMetadata {
+  isOptimistic?: boolean;
+}
+
+type QueryListItem<T extends QueryListItemBase = QueryListItemBase> = T & {
+  _metadata?: QueryListItemMetadata;
+};
+
+type QueryList<T extends QueryListItemBase = QueryListItemBase> = Array<QueryListItem<T>>;
```
hdinia: not used

webapp/src/routes/_authenticated/-components/Container/FavoritesMenu/useFavorites.ts
```diff
@@ -0,0 +1,67 @@
...
+    storage.removeItem("studies.favorites");
+
+    if (Array.isArray(oldFavorites)) {
+      oldFavorites.forEach((value) => {
+        if (typeof value === "string") {
+          createStudyFavorite.mutate({ studyId: value });
+        }
+      });
+    }
+  });
```
hdinia, line 64: Favorites are relatively new, and very cheap to add/remove. You should drop all localStorage support to simplify the code and avoid the confusion of a frontend "migration", the consequences for users will be negligible

webapp/src/routes/-shared/components/studies/FavoriteToggle/FavoriteStudyToggle.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { studyQueries } from "@/queries/studies/queries";
+import useCreateFavoriteStudy from "@/routes/-shared/hooks/favorites/useCreateFavoriteStudy";
+import useDeleteFavoriteStudy from "@/routes/-shared/hooks/favorites/useDeleteFavoriteStudy";
+import type { FavoriteStudy } from "@/services/api/favorites/types";
+import type { StudyMetadata } from "@/types/types";
+import { useIsMutating, useSuspenseQuery } from "@tanstack/react-query";
+import * as RA from "ramda-adjunct";
+import { useCallback } from "react";
+import FavoriteButton, { type FavoriteButtonProps } from "./FavoriteButton";
+interface Props extends Omit<FavoriteButtonProps, "isFavorite" | "onClick"> {
```
hdinia, line 25: minor: missing blank line

webapp/src/services/api/favorites/index.ts
```diff
@@ -0,0 +1,55 @@
...
+import type {
+  FavoriteDirectory,
+  FavoriteDirectoryParams,
+  FavoriteStudy,
+  FavoriteStudyParams,
+} from "./types";
+
+const BASE_URL = "/v1/favorites" as const;
+const STUDIES_FAVORITES_URL = `${BASE_URL}/studies` as const;
+const DIRECTORIES_FAVORITES_URL = `${BASE_URL}/directories` as const;
```
hdinia: `BASE_URL` is never used directly it only exists to build 2 other constants, which adds indirection for no benefit. Also, `as const` on a template literal is redundant since the type is already inferred as a string literal.

Since this file is small and the URLs are only used a few times each, constants don't add much value here. I'd suggest inlining them directly in the calls

Please consider adding JSdoc (including the HTTP verb + url), to stay consistent with recent methods like in /directories, and access rapidly docs on hover where it's used

webapp/src/services/api/favorites/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  directoryName: string;
+}
+
+export interface FavoriteStudyParams {
+  studyId: string;
+}
+
+export interface FavoriteDirectoryParams {
+  directoryId: string;
+}
```
hdinia, line 31: I dont know if it's necessary to use Zod schmas and derived types here, but maybe we should for consistency ?

webapp/src/queries/utils.ts
```diff
@@ -12,24 +12,74 @@
...
+  };
+
+  return queryOptions({
+    ...options,
+    // Prevent refetching if the list is currently in an optimistic state, to avoid overwriting
+    // pending changes with server data
+    refetchOnWindowFocus: blockRefetchingIfMutating,
+    refetchOnReconnect: blockRefetchingIfMutating,
+    refetchOnMount: blockRefetchingIfMutating,
+  });
```
hdinia, line 90: (remark) the global 1min stale time is silently applied here. for favorites specifically, you'd probably want to refetch on window focus to catch changes made in other tabs ?

hdinia (changes requested):

webapp/src/routes/_authenticated/-components/Container/FavoritesMenu/useFavorites.ts
skamril: Favorites in localStorage have been around for a while (see https://github.com/AntaresSimulatorTeam/AntaREST/pull/921).

I’m a bit concerned that removing this workaround now could cause some users to lose their existing favorites. Since the migration logic is relatively small, it might be safer to keep it for now and remove it in the next release.

webapp/src/services/api/favorites/index.ts
skamril: I prefer to keep `STUDIES_FAVORITES_URL` and `DIRECTORIES_FAVORITES_URL` rather than repeating the URL in each endpoint.

webapp/src/services/api/favorites/types.ts
skamril: Maybe overkill for an internal API.

hdinia (dismissed):

hdinia (approved):

PR #3127, opened 2026-03-26
perf(dao): less db requests inside study converter for thermal

MartinBelthle:
This PR introduces new DAO methods for the thermal clusters part that allow bulk operations.
For now we suffer from N+1 queries, the following paragraph highlights what was the situation before the PR and after it.

The following values concern a typical BP study (around 50 areas with 10 thermal clusters each)
Each thermal cluster has 5 matrices (so there are in total 2500 matrices)

#### Before
- 50 inserts to save all thermal properties
- 2500 inserts for all 5 types of matrices
- If the FS study is already normalized, 2500 get to fetch all matrices in the matrixstore and 0 inserts
- Otherwise, 0 get and 2500 inserts

**Total**: 5050 operations

#### After
- 1 insert for all thermal properties
- 5 inserts for all 5 types of matrices
- If FS study is normalized, 5 get and 0 inserts
- Otherwise, 0 get and 1 insert

**Total**: 7 operations for a unnormalized study and 11 otherwise

#### Before
- 1 Get for all thermal properties
- 2500 Get for all matrices
- 2500 Get inside matrixstore (due to the MatrixNode `dump` method)
- 2500 Get inside matrixstore if the study is  denormalized (still due to the `dump` method)

**Total**: 5001 operations if normalized and 7501 otherwise

#### After
- 1 Get for all thermal properties
- 5 Get for all matrices
- 1 Get inside matrixstore

**Total**: 7 operations

| Cases | DB queries Before  | DB queries Now         |
| :------ | :--------------- |:---------------:|
| DB -> FS normalized |  5001 :red_circle:        |  7 :green_circle:        |
| DB -> FS denormalized  | 7501 :red_circle:           |  7 :green_circle:        |
| FS -> DB normalized  | 5050 :red_circle:       |  11 :green_circle:        |
| FS -> DB denormalized  | 5050 :red_circle:       |  7 :green_circle:        |


antarest/study/dao/file/file_study_thermal_dao.py
```diff
@@ -35,6 +37,14 @@
 _ALL_CLUSTERS_PATH = "input/thermal/clusters"
 
 
+SeriesMapping: TypeAlias = dict[AreaId, dict[ThermalId, SeriesId]]
+
+
+def _check_area_exists(study_data: FileStudyTreeConfig, area_id: str) -> None:
+    if area_id not in study_data.areas:
+        raise ValueError(f"The area '{area_id}' does not exist")
```
TheoPascoli: You can raise `AreaNotFound`

antarest/study/dao/file/file_study_thermal_dao.py
```diff
@@ -35,6 +37,14 @@
 _ALL_CLUSTERS_PATH = "input/thermal/clusters"
 
 
+SeriesMapping: TypeAlias = dict[AreaId, dict[ThermalId, SeriesId]]
```
TheoPascoli: You can put this in the common script. You use exactly this type without an alias in `database_thermal_dao` ligne 103 et `in_memory_study_dao` at multiple places

TheoPascoli (changes requested):
Nice improvement !

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
```diff
@@ -96,12 +96,17 @@ def _parse_dataframe(self, use_default_empty: bool) -> pl.DataFrame:
 
     @override
     def write_dataframe(self, df: pl.DataFrame) -> None:
+        if not self.config.path.parent.exists():
+            # Can happen when creating a new object and the file structure is not yet fully created
```
sylvlecl, line 100: Is this something that could not appear before ?

antarest/study/dao/file/file_study_dao.py
```diff
@@ -86,6 +87,19 @@ def __init__(
         self._file_study = study
         self._generator_matrix_constants = generator_matrix_constants
         self._blob_service = blob_service
+        self._matrix_service = generator_matrix_constants.matrix_service
```
sylvlecl: Should be provided as a dependency explicitly, not through the constants

antarest/study/dao/file/file_study_thermal_dao.py
```diff
@@ -114,48 +124,57 @@ def get_thermal_co2_cost(self, area_id: str, thermal_id: str) -> pl.DataFrame:
         return self.get_impl().get_matrix(["input", "thermal", "series", area_id, thermal_id, "CO2Cost"])
 
     @override
-    def save_thermal(self, area_id: str, thermal: ThermalCluster) -> None:
-        study_data = self.get_file_study()
-        self._update_thermal_config(study_data.config, area_id, thermal)
+    def get_all_thermals_co2_cost(self) -> SeriesMapping:
+        return self._get_thermal_matrices("input/thermal/series/{area_id}/{thermal_id}/CO2Cost")
```
sylvlecl: Very minor:
Weird to provide a str that we split right away.
We could provide a function (area_id, cluster_id) -> list instead, functions that would be defined at the top level of the module (like _get_co2_cost_path etc).
I must say it adds some "complexity" though, maybe not worth it after all

antarest/study/dao/file/file_study_dao.py
```diff
@@ -169,7 +183,8 @@ def save_matrices(self, nodes_and_matrix_ids: list[tuple[MatrixNode, str]]) -> N
 
         # First, we separate_matrices in 2 groups, the normalized and the denormalized ones.
         for matrix_node, matrix_id in nodes_and_matrix_ids:
-            if matrix_node.is_normalized:
+            matrix_id = extract_matrix_id(matrix_id)
+            if matrix_node.should_normalize_matrix():
```
sylvlecl: Feels very weird to read it from the individual node:

the "should_normalize" information is actually a study level information. If it's managed, yes, else, no.

We added it to the "matrix mapper", because the individual nodes needed that information to know how to write a matrix, but at the start, it's really a study-level behaviour.

Now here, at the study level, we get the information from the individual node-level, that itself came from the study level.

Instead, it would be more clear that we just have it at the study level from the start, for example at the DAO level when we create it

antarest/study/dao/common.py
```diff
@@ -0,0 +1,16 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+from typing import TypeAlias
+
+AreaId: TypeAlias = str
```
sylvlecl, line 14: Makes me think, I have seen recently that we may use [`NewType`](https://docs.python.org/3/library/typing.html#newtype) to improve type checking of such stuff:
it creates like a new sub-type of string that does not change anything at runtime, but allows the typechecker to check if we indeed have the right "type of string" in the right places

sylvlecl (commented):
Looks good to me overall.
Still one relatively important comment about who should be responsible of the "should normalize" information

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
MartinBelthle: Before we went through the `dump` method where we had this line `self.config.path.parent.mkdir(exist_ok=True, parents=True)` so now that we bypass it we have to do this

antarest/study/dao/file/file_study_dao.py
MartinBelthle: Agreed but I didn't want to complexify the PR review. Now that it's done I'll do this

antarest/study/dao/file/file_study_thermal_dao.py
MartinBelthle: I'll implement this because it will allow me to not duplicate the paths that are declared at mulitple places in the file

sylvlecl (approved):
Nice, thank you !!

PR #3130, opened 2026-03-30
feat(bc): implement db dao

smailio:


antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,477 @@
...
+
+    def _cleanup_scenario_builder_groups(
+        self,
+        constraints: Sequence[BindingConstraint],
+        existing: dict[str, BindingConstraint],
+    ) -> None:
+        saved_ids = {bc.id for bc in constraints}
+        groups_before = {bc.group for bc in existing.values() if bc.group}
+        groups_after = {bc.group for bc in constraints if bc.group} | {
+            bc.group for bc in existing.values() if bc.group and bc.id not in saved_ids
```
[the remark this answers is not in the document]
smailio: It's fixed. 

The problem was when all constraints were saved with groupe set to None like this 

```python
def test_resaving_null_group_constraint_does_not_wipe_default_scenario_builder_rules(
    db_dao: DatabaseStudyDao,
) -> None:
    """Saving a constraint with group=None (which reads back as 'default') must not
    delete scenario-builder rules for 'default'."""
    db_dao.save_constraints([_bc("bc1", group=None)])
    db_dao.save_scenario_builder(Ruleset(binding_constraints={"default": {"0": 1}}))

    # Re-save the same constraint with group=None — no group has changed, nothing to clean up
    db_dao.save_constraints([_bc("bc1", group=None)])
    assert db_dao.get_ruleset().binding_constraints == {"default": {"0": 1}}
```

The fix 

```python 
        if version >= STUDY_VERSION_8_7 and any(bc.group is None for bc in constraints):
            groups_after.add(DEFAULT_GROUP)
```

antarest/study/dao/database/database_area_dao.py
```diff
@@ -149,7 +149,7 @@ def get_all_areas_ui_info(self) -> dict[str, AreaUIData]:
                     "color_r": default_ui.color_r,
                     "color_g": default_ui.color_g,
                     "color_b": default_ui.color_b,
-                    "layers": " ".join(sorted([row.layer_id for row in ui_rows if row.layer_id != DEFAULT_LAYER_ID])),
+                    "layers": " ".join(sorted([row.layer_id for row in ui_rows], key=int)),
```
smailio, line 153: @TheoPascoli check this fix please. The file study dao seems to return the default layer not ignore it.

antarest/study/dao/file/file_study_constraint_dao.py
```diff
@@ -182,14 +186,15 @@ def delete_constraints(self, constraints: list[BindingConstraint]) -> None:
...
+        ids_to_delete = {c.id for c in constraints}
         deleted_binding_constraints = []
         kept_binding_constraints = []
         old_groups = set()
         for key in list(ini_content.keys()):
             constraint = parse_binding_constraint(study_data.config.version, ini_content[key])
             if constraint.group:
                 old_groups.add(constraint.group)
-            if constraint in constraints:
+            if constraint.id in ids_to_delete:
```
smailio, line 189: @MartinBelthle check this fix please. This code used to do an object comparaison, it make no sense for me so I rewrote it to compare ids instead. Maybe there was a reason ?

antarest/study/dao/file/file_study_constraint_dao.py
```diff
@@ -82,7 +83,10 @@ def get_all_constraints(self) -> dict[str, BindingConstraint]:
 
     @override
     def get_constraint(self, constraint_id: str) -> BindingConstraint:
-        return self.get_all_constraints()[constraint_id]
+        result = self.get_all_constraints()
+        if constraint_id not in result:
+            raise BindingConstraintNotFound(f"Constraint {constraint_id} not found")
```
smailio, line 76: This is to align the error raised with the database dao. It makes sense and It's more convenient for my tests.

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,484 @@
...
+
+    @override
+    def save_constraints(self, constraints: Sequence[BindingConstraint]) -> None:
+        impl = self.get_impl()
+        study_version = impl.get_version()
+        constants = impl._generator_matrix_constants
+
+        constraint_ids = [bc.id for bc in constraints]
+        # Fetch existing constraints BEFORE upserting (needed for change detection)
+        existing = self._fetch_constraints()
```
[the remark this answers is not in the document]
smailio: I disagree

antarest/study/dao/database/database_area_dao.py
TheoPascoli: Yes thanks for this

antarest/study/dao/file/file_study_constraint_dao.py
MartinBelthle: Indeed it's better to do this, don't remember why  i did this but does not seem necessary

tests/study/dao/file/conftest.py
```diff
@@ -54,3 +54,38 @@ def filestudy_dao(file_study: FileStudy, matrix_service, blob_service) -> FileSt
     constants = GeneratorMatrixConstants(matrix_service)
     constants.init_constant_matrices()
     return FileStudyTreeDao(file_study, constants, blob_service)
+
+
+@pytest.fixture
+def filestudy_dao_and_matrix_service(
```
MartinBelthle: What's the point of this fixture ? Seems you could just use the 2 separately inside your tests

tests/study/dao/file/conftest.py
```diff
@@ -54,3 +54,38 @@ def filestudy_dao(file_study: FileStudy, matrix_service, blob_service) -> FileSt
...
+def filestudy_dao_860(
+    file_study_860: FileStudy, matrix_service: ISimpleMatrixService, blob_service: IBlobService
+) -> FileStudyTreeDao:
+    constants = GeneratorMatrixConstants(matrix_service)
+    constants.init_constant_matrices()
+    return FileStudyTreeDao(file_study_860, constants, blob_service)
+
+
+@pytest.fixture
+def filestudy_dao_860_and_matrix_service(
```
MartinBelthle: Same remark

tests/study/dao/file/conftest.py
```diff
@@ -54,3 +54,38 @@ def filestudy_dao(file_study: FileStudy, matrix_service, blob_service) -> FileSt
...
+
+@pytest.fixture
+def filestudy_dao_and_matrix_service(
+    filestudy_dao: FileStudyTreeDao, matrix_service: ISimpleMatrixService
+) -> tuple[FileStudyTreeDao, ISimpleMatrixService]:
+    return filestudy_dao, matrix_service
+
+
+@pytest.fixture
+def file_study_860(tmp_path: Path, matrix_service: ISimpleMatrixService) -> FileStudy:
```
MartinBelthle: You should put code in common with the existing `file_study` fixture

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,495 @@
...
+
+
+class DatabaseBindingConstraintDao(ConstraintDao):
+    def __init__(self, study_id: str, db_session: Session) -> None:
+        self._study_id = study_id
+        self._db_session = db_session
+
+    @abstractmethod
+    def get_impl(self) -> "DatabaseStudyDao":
+        raise NotImplementedError()
```
MartinBelthle: Elsewhere, we just `pass`

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,495 @@
...
+                ConstraintTerm(
+                    weight=row.weight, offset=row.offset, data=ClusterTerm(area=row.area, cluster=row.cluster)
+                )
+            )
+
+        version = self.get_impl().get_version()
+        return {cid: self._row_to_bc(bc_rows[cid], terms[cid], version) for cid in bc_rows}
+
+    @staticmethod
+    def _row_to_bc(row: Any, terms: list[ConstraintTerm], version: Any) -> BindingConstraint:
```
MartinBelthle: Typing row as `Row[Any]` could be a bit better

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,495 @@
...
+        row = self._db_session.execute(
+            select(table).where((table.c.study_id == self._study_id) & (table.c.constraint_id == constraint_id))
+        ).fetchone()
+        if row is None:
+            raise BindingConstraintNotFound(f"Matrix for constraint {constraint_id} not found")
+        return self.get_impl().get_matrix(row.matrix_id)
+
+    def _save_bc_matrices(self, table: Table, entries: list[tuple[str, str]]) -> None:
+        rows = [
+            {"study_id": self._study_id, "constraint_id": cid, "matrix_id": mid.removeprefix(MATRIX_PROTOCOL_PREFIX)}
```
MartinBelthle: We never had to use `.removeprefix` in the DB dao before. To me it's not the DAO responsibility to do this but rather the one calling him

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,495 @@
...
+
+    @override
+    def save_constraint_greater_term_matrix(self, constraint_id: str, series_id: str) -> None:
+        self._save_bc_matrices(BINDING_CONSTRAINT_GT_MATRIX_TABLE, [(constraint_id, series_id)])
+        self._db_session.commit()
+
+    @override
+    def save_constraint_equal_term_matrix(self, constraint_id: str, series_id: str) -> None:
+        self._save_bc_matrices(BINDING_CONSTRAINT_EQ_MATRIX_TABLE, [(constraint_id, series_id)])
+        self._db_session.commit()
```
MartinBelthle, line 575: You should just commit inside the `save_bc_matrices` method

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,495 @@
...
+
+    def delete_constraint_values_matrix(self, constraint_id: str) -> None:
```
MartinBelthle: You never use these methods

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,495 @@
...
+        self._db_session.commit()
+
+    def _save_constraint_rows(self, constraints: Sequence[BindingConstraint]) -> None:
+        upsert_multiple(
+            self._db_session,
+            BINDING_CONSTRAINT_TABLE,
+            [self._bc_to_row(self._study_id, bc) for bc in constraints],
+        )
+
+    def _save_cluster_terms(self, constraints: Sequence[BindingConstraint], constraint_ids: list[str]) -> None:
```
MartinBelthle: Would be a bit better to not give `constraint_ids` as an argument but rather calculate it here. Same for the other method

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,495 @@
...
+        """Remove scenario builder rules for groups that no longer have any constraint.
+
+        Args:
+            saved: constraints being saved/upserted in this call (empty for deletes).
+            existing: all constraints as they were before this call.
+            remaining: if provided, the constraints that will remain after this call.
+                       If None, computed from ``existing`` minus ``saved`` IDs plus ``saved``.
+        """
+        version = self.get_impl().get_version()
+        if remaining is None:
```
MartinBelthle: Before v8.7 as there are no groups you can simply return there is nothing to do

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,495 @@
...
+        self._delete_bc_matrices(BINDING_CONSTRAINT_EQ_MATRIX_TABLE, [constraint_id])
+        self._db_session.commit()
+
+    @override
+    def delete_constraints(self, constraints: list[BindingConstraint]) -> None:
+        db = self._db_session
+        constraint_ids = {bc.id for bc in constraints}
+
+        # Compute orphaned groups before deleting (need the full picture first)
+        existing = self._fetch_constraints()
```
MartinBelthle: You could just execute this code after v8.7 for performance.


Also perhaps we could change the logic to simplify this.
First you delete, then if > 8.7 you get all groups you removed (based on argument `constraints`). You do a get inside BC table to see what groups you removed are in fact removed. And then you delete from sc_builder table. Seems more efficient and more understandable

MartinBelthle: I see you use the same function twice in the code so perhaps what I said does not apply except for the v8.7 part :/

MartinBelthle: Or else we could do something that is probably less performant but as the sc builder bcs are in only one table with groups the table will not be that big to fetch, so perhaps we could just do a delete where `bc_group_id` not in bc_table with the right bc_id, sounds like a join but not sure.
It would be way easier to read IMO

MartinBelthle (changes requested):
Didn't finish to read the DAO BD code as it's pretty hard to read.
Also you should add tests for the matrix_mapper and the matrix_provider

antarest/study/dao/database/database_binding_constraint_dao.py
smailio: when I do save_constraints and the operator or time step changed => I generate new matrices => I save the matrices. If I keep the commits then I'll have a commit in the middle of my save_constraints instead of at the end.

antarest/study/dao/database/database_area_dao.py
```diff
@@ -287,14 +287,14 @@ def delete_area(self, area_id: str) -> None:
         session.commit()
 
     @override
-    def save_area_ui(self, area_id: str, layer: str, area_ui_data: AreaUI) -> None:
+    def save_area_ui(self, area_id: str, layer: str, area_ui: AreaUI) -> None:
```
TheoPascoli, line 291: Nice thanks :)

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,495 @@
...
+        Step 2 : BC LEFT JOIN cluster_terms fetches cluster_terms and appends them.
+
+        Two round trips, but joining both term tables at once would produce a Cartesian product,
+        that would make us fetch more data and process more data also we'll need to handle de-duplication which would make the code less readable."""
+        db = self._db_session
+        BC = BINDING_CONSTRAINT_TABLE
+        LT = BINDING_CONSTRAINT_LINK_TERM_TABLE
+        CT = BINDING_CONSTRAINT_CLUSTER_TERM_TABLE
+
+        # Query 1: BC left join link terms
```
TheoPascoli: Import them `as BC, LT, CT` if you prefer

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,495 @@
...
+            name=row.name,
+            enabled=row.enabled,
+            time_step=row.time_step,
+            operator=row.operator,
+            comments=row.comments,
+            filter_year_by_year=row.filter_year_by_year,
+            filter_synthesis=row.filter_synthesis,
+            group=row.group,
+            terms=terms,
+        )
```
TheoPascoli: you can replace this by bc = BindingConstraint.model_construct(**get_row_representation_as_dict(row), terms=terms)

TheoPascoli (changes requested):
I haven't read everything for now

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,495 @@
...
+                select(table.c.constraint_id, table.c.matrix_id).where(
+                    (table.c.study_id == self._study_id) & (table.c.constraint_id.in_(constraint_ids))
+                )
+            ).fetchall()
+            for row in rows:
+                existing_matrix_ids.setdefault(row.constraint_id, {})[term] = row.matrix_id
+        return existing_matrix_ids
+
+    @staticmethod
+    def _default_matrix_id(
```
MartinBelthle: I just realized we should just use the null matrix in every case as this will let the Simulator do the work. 
So you can remove this and just use the null matrix

antarest/study/dao/database/database_binding_constraint_dao.py
smailio: model_construct set timestep and operator as strings, I want them to be enums, so I used model_validate, that's fine for you ?

antarest/study/dao/file/file_study_constraint_dao.py
```diff
@@ -252,31 +244,23 @@ def _remove_groups_from_scenario_builder(study_data: FileStudy, removed_groups:
 def generate_replacement_matrices(
     bc_id: str,
     study_version: StudyVersion,
-    new_time_step: BindingConstraintFrequency,
+    time_step: BindingConstraintFrequency,
```
MartinBelthle: You don't need the time_step anymore

tests/study/dao/file/conftest.py
```diff
@@ -37,18 +38,24 @@ def blob_service() -> IBlobService:
...
-@pytest.fixture
-def file_study(tmp_path: Path, matrix_service: ISimpleMatrixService) -> FileStudy:
-    study_id = "5c22caca-b100-47e7-bbea-8b1b97aa26d9"
+def build_file_study(
+    tmp_path: Path, matrix_service: ISimpleMatrixService, version: StudyVersion, suffix: str = ""
+) -> FileStudy:
+    study_id = f"5c22caca-b100-47e7-bbea-8b1b97aa26d9{suffix}"
     study_path = tmp_path.joinpath(study_id)
-    app = CreateApp(study_dir=study_path, caption="filestudy", version=STUDY_VERSION_9_3, author="Joe")
+    app = CreateApp(study_dir=study_path, caption=f"filestudy{suffix}", version=version, author="Joe")
```
MartinBelthle: Minor but why adding a suffix, you don't seem to use it ?

tests/study/dao/matrix_raw_mapper/test_raw_path_to_matrix_mapper.py
```diff
@@ -76,10 +76,14 @@ def test_get_matrix_from_path(
...
             xpansion_capacity_df,
             xpansion_weight_df,
+            bc_lt_df,
+            bc_gt_df,
+            bc_eq_df,
         ) = result.dataframes
         area_id, area2 = result.area1, result.area2
         thermal_id, renewable_id, st_storage_id = result.thermal_id, result.renewable_id, result.sts_id
         constraint_id = result.sts_constraint_id
+        bc_both_id, bc_eq_id = result.bc_both_id, result.bc_eq_id
```
MartinBelthle, line 86: You don't seem to use them

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+
+        version = self.get_impl().get_version()
+        return {cid: self._row_to_bc(bc_rows[cid], terms[cid], version) for cid in bc_rows}
+
+    @staticmethod
+    def _row_to_bc(row: Row[Any], terms: list[ConstraintTerm], version: StudyVersion) -> BindingConstraint:
+        d = get_row_representation_as_dict(row)
+        d["id"] = d.pop("constraint_id")
+        d["terms"] = terms
+        bc = BindingConstraint.model_validate(d, extra="allow")
```
MartinBelthle: Why would you do extra="allow" ?

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+        initialize_binding_constraint(bc, version)
```
MartinBelthle: Shouldn't be needed here, we save the constraint already initialized, you should fix your tests if they don't pass because of this but not the src code

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+
+    @override
+    def delete_constraints(self, constraints: list[BindingConstraint]) -> None:
+        db = self._db_session
+        constraint_ids = {bc.id for bc in constraints}
+
+        # Delete BC rows first so the table reflects the final state,
+        # then prune orphaned groups via a single subquery.
+        db.execute(delete(BC).where((BC.c.study_id == self._study_id) & (BC.c.constraint_id.in_(constraint_ids))))
+        self._cleanup_scenario_builder_groups()
```
MartinBelthle, line 590: Way better with this, thanks :)

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+
+    def _save_link_terms(self, constraints: Sequence[BindingConstraint]) -> None:
+        constraint_ids = [bc.id for bc in constraints]
+        link_terms = [
+            self._link_term_to_row(self._study_id, bc.id, term)
+            for bc in constraints
+            for term in bc.terms
+            if isinstance(term.data, LinkTerm)
+        ]
+        self._db_session.execute(
```
MartinBelthle, line 288: Same remark for link terms and cluster terms but if your list is empty, which can happend quite often you should return already before performing any DB call

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+        ).fetchone()
+        if row is None:
+            raise BindingConstraintNotFound(f"Matrix for constraint {constraint_id} not found")
+        return self.get_impl().get_matrix(row.matrix_id)
+
+    def _save_bc_matrices(self, table: Table, entries: list[tuple[str, str]]) -> None:
+        rows = [{"study_id": self._study_id, "constraint_id": cid, "matrix_id": mid} for cid, mid in entries]
+        upsert_multiple(self._db_session, table, rows)
+
+    def _delete_bc_matrices(self, table: Table, constraint_ids: list[str]) -> None:
```
MartinBelthle: Never used

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+                )
+            ).fetchall()
+            for row in rows:
+                existing_matrix_ids.setdefault(row.constraint_id, {})[term] = row.matrix_id
+        return existing_matrix_ids
+
+    def _compute_matrix_changes(
+        self,
+        constraints: Sequence[BindingConstraint],
+    ) -> tuple[dict[str, list[str]], dict[str, list[tuple[str, str]]]]:
```
MartinBelthle: Use classes or aliases but this is unreadable

MartinBelthle: Also comment what this method does as it's not easy to understand

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+    "gt": BINDING_CONSTRAINT_GT_MATRIX_TABLE,
+    "eq": BINDING_CONSTRAINT_EQ_MATRIX_TABLE,
+    "values": BINDING_CONSTRAINT_VALUES_MATRIX_TABLE,
+}
+
+
+def _get_zero_matrix_id(
+    time_step: BindingConstraintFrequency, study_version: StudyVersion, generator: GeneratorMatrixConstants
+) -> str:
+    """Return the matrix-store ID for the correctly-sized zero matrix matching the given time step."""
```
MartinBelthle: Depending on how we choose to implement the default simulator matrices in the `get` db dao methods, we could just use the `null_matrix` here.

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+
+        for bc in constraints:
+            old = existing.get(bc.id)
+            if old is None:
+                continue  # new constraint — caller handles initial matrix creation
+
+            time_step_changed = bc.time_step != old.time_step
+            operator_changed = study_version >= STUDY_VERSION_8_7 and bc.operator != old.operator
+
+            if operator_changed and not time_step_changed:
```
MartinBelthle: Does not seem you're handling the case where both changes are made with your `if` conditions

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+                new_terms = OPERATOR_MATRICES_MAP[bc.operator]
+                terms_to_add = [t for t in new_terms if t not in old_terms]
+                terms_to_del = [t for t in old_terms if t not in new_terms]
+
+                if terms_to_add:
+                    # For BOTH, lt is the canonical source (mirrors file DAO rename logic).
+                    # For single operators, there is only one term so it is unambiguous.
+                    source_term = "lt" if old.operator == BindingConstraintOperator.BOTH else old_terms[0]
+                    source_mid = existing_matrix_ids.get(bc.id, {}).get(source_term)
+                    # The command layer always initializes matrices on creation, so missing source = data corruption.
```
MartinBelthle: You could just use `[]` and remove the exception then

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+            old = existing.get(bc.id)
+            if old is None:
+                continue  # new constraint — caller handles initial matrix creation
+
+            time_step_changed = bc.time_step != old.time_step
+            operator_changed = study_version >= STUDY_VERSION_8_7 and bc.operator != old.operator
+
+            if operator_changed and not time_step_changed:
+                # Copy/move matrix data between tables without resetting values.
+                old_terms = OPERATOR_MATRICES_MAP[old.operator]
```
MartinBelthle: IMO you should not use the wording terms as it's reserved for the real terms of the constraints, or at least use `matrix_term`

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+                    for term in OPERATOR_MATRICES_MAP[old.operator]:
+                        to_delete[term].append(bc.id)
+                    for term in OPERATOR_MATRICES_MAP[bc.operator]:
+                        to_create[term].append((bc.id, zero_mid))
+
+        return to_delete, to_create
+
+    def _apply_matrix_changes(
+        self,
+        to_delete: dict[str, list[str]],
```
MartinBelthle: You should definitely use dataclasses to make this easier to read, it's already nerve wracking enough to read this functions

MartinBelthle (changes requested):

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,438 @@
...
+
+    def _cleanup_scenario_builder_groups(self) -> None:
+        """Delete scenario-builder entries for groups no longer referenced by any constraint.
+
+        Must be called after the binding_constraints table already reflects the final state
+        (i.e. after upserts in save_constraints, or after deletes in delete_constraints).
+        """
+        if self.get_impl().get_version() < STUDY_VERSION_8_7:
+            return
+        # COALESCE handles the case where group IS NULL in DB, which maps to "default".
```
MartinBelthle: This case should never happen as we're after v8.7. If it happens it means you didn't initialize you constraint before saving it but this will never happen in real case

antarest/study/dao/database/database_binding_constraint_dao.py
smailio: because of columns from JOIN (area1, areat2, etc..)

antarest/study/dao/database/database_binding_constraint_dao.py
smailio: what do you mean I don't enforce initialization on save, also this behavior is consistent with file study behavior  where initialize_binding_constraint is called on every get not on save.

antarest/study/dao/database/database_binding_constraint_dao.py
smailio: As long as the time step changes we throw everything and we replace with null matrices no ?

antarest/study/dao/database/database_binding_constraint_dao.py
smailio: For me this should never happen and we shouldn't fail silently. If this data gets corrupted I'd rather know.

antarest/study/dao/database/database_binding_constraint_dao.py
smailio: If it's a matter of mental health then okay.

antarest/study/dao/database/database_binding_constraint_dao.py
smailio: This funsion is called on every save and every delete, I do this so I don't do useless queries.

antarest/study/dao/database/database_binding_constraint_dao.py
smailio: Ah ok initialization happen in the create command. 

I don't like it that the dao is supposed to be aware of the call context but ok.

antarest/study/dao/database/database_binding_constraint_dao.py
smailio: I need to perform the delete anyway to handle case when user send me empty list for terms, we interpret this as delete all terms.

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,491 @@
...
+                matrix_type_to_add = [s for s in new_matrix_types if s not in old_matrix_types]
+                matrix_type_to_delete = [s for s in old_matrix_types if s not in new_matrix_types]
+
+                if matrix_type_to_add:
+                    # For BOTH, lt is the canonical source (mirrors file DAO rename logic).
+                    # For single operators, there is only one matrix type so it is unambiguous.
+                    source_type = (
+                        _MatrixType.LT if old.operator == BindingConstraintOperator.BOTH else old_matrix_types[0]
+                    )
+                    source_mid = existing_matrix_ids.get(bc.id, {}).get(source_type.value)
```
smailio: I proably spent too much time around this code but what it we replace this weird generic logic with explicit one 

```python
def delete(matrix_type: _MatrixType) -> None:
    changes.add_deletion(bc.id, matrix_type)

def insert(matrix_type: _MatrixType, src: str | None) -> None:
    changes.add_insertion(bc.id, matrix_type, src)

O = BindingConstraintOperator
match (old.operator, bc.operator):
    # LESS 
    case (O.LESS, O.GREATER):
        delete(_MatrixType.LT)
        insert(_MatrixType.GT, lt_mid)
    case (O.LESS, O.EQUAL):
        delete(_MatrixType.LT)
        insert(_MatrixType.EQ, lt_mid)
    case (O.LESS, O.BOTH):
        insert(_MatrixType.GT, lt_mid)
    # GREATER 
    case (O.GREATER, O.LESS):
        delete(_MatrixType.GT)
        insert(_MatrixType.LT, gt_mid)
    case (O.GREATER, O.EQUAL):
        delete(_MatrixType.GT)
        insert(_MatrixType.EQ, gt_mid)
    case (O.GREATER, O.BOTH):
        insert(_MatrixType.LT, gt_mid)
    # EQUAL
    case (O.EQUAL, O.LESS):
        delete(_MatrixType.EQ)
        insert(_MatrixType.LT, eq_mid)
    case (O.EQUAL, O.GREATER):
        delete(_MatrixType.EQ)
        insert(_MatrixType.GT, eq_mid)
    case (O.EQUAL, O.BOTH):
        delete(_MatrixType.EQ)
        insert(_MatrixType.LT, eq_mid)
        insert(_MatrixType.GT, eq_mid)
    # BOTH 
    case (O.BOTH, O.LESS):
        delete(_MatrixType.GT)
    case (O.BOTH, O.GREATER):
        delete(_MatrixType.LT)
    case (O.BOTH, O.EQUAL):
        delete(_MatrixType.LT)
        delete(_MatrixType.GT)
        insert(_MatrixType.EQ, lt_mid)
```
Obviously it's super long but imo it requires way less mental effort to understand what the code does.

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,491 @@
...
+    def __init__(self, study_id: str, db_session: Session) -> None:
+        self._study_id = study_id
+        self._db_session = db_session
+
+    @abstractmethod
+    def get_impl(self) -> "DatabaseStudyDao":
+        pass
+
+    def _fetch_constraints(
+        self, constraint_id: str | None = None, constraint_ids: list[str] | None = None
```
MartinBelthle: This should be simplified with just constraint_ids and if you want only one you give the constraint in a list. Would make the code a bit easier to read

antarest/study/dao/database/database_binding_constraint_dao.py
MartinBelthle: Ah okay I see. Thanks for the comment. Could be good to add a comment inside the code just to say this but not necessary

antarest/study/dao/database/database_binding_constraint_dao.py
MartinBelthle: You're right indeed. But then I find the logic a bit hard to understand, would you wind changing the if ? First do a if time_step changed and a elif operator_changed, I think it would be easier to understand then

antarest/study/dao/database/database_binding_constraint_dao.py
MartinBelthle: Why not, it is indeed longer but I agree with you it's more understandable. You suffered not me so the choice is yours :)

antarest/study/dao/database/database_binding_constraint_dao.py
MartinBelthle: Not sure to full grasp your answer, what I meant is that the `coalesce` should not be needed as after v8.7 groups cannot be null and at this point in the code you're sure that you're after v8.7

antarest/study/dao/database/database_binding_constraint_dao.py
MartinBelthle: I see. So let's keep it this way

tests/study/dao/conftest.py
MartinBelthle, line 207: Minor but since last PR was merged on dev, this is always `dao._matrix_service` so you can remove this utility function

tests/study/dao/conftest.py
```diff
@@ -131,6 +128,67 @@ def db_dao_920(db_session: Session, matrix_service: ISimpleMatrixService) -> Dat
...
+    else:
+        return _build_fs_dao(db_session, STUDY_VERSION_9_3, command_context, core_cache, tmp_path)
+
+
+@pytest.fixture(params=["db", "fs"], ids=["database", "filesystem"])
+def dao_860_and_matrix_service(
+    request,
+    db_session: Session,
+    matrix_service: ISimpleMatrixService,
+    command_context: "CommandContext",
```
MartinBelthle, line 182: Really minor but you don't need the `""` in this file

MartinBelthle (changes requested):
Didn't read the test files but I read everything else. Should be the last review except for the tests part

antarest/study/dao/database/database_binding_constraint_dao.py
smailio: Ah I though you were talking about the if just above. 

Yeah you're right.

tests/study/dao/matrix_raw_mapper/test_raw_path_to_matrix_mapper.py
smailio: I do bellow

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,582 @@
...
+        """
+        if self.get_impl().get_version() < STUDY_VERSION_8_7:
+            return
+        active_groups = select(BC.c.group).where(BC.c.study_id == self._study_id).distinct()
+        self._db_session.execute(
+            delete(SCENARIO_BINDING_CONSTRAINTS_TABLE).where(
+                (SCENARIO_BINDING_CONSTRAINTS_TABLE.c.study_id == self._study_id)
+                & SCENARIO_BINDING_CONSTRAINTS_TABLE.c.bc_group_id.not_in(active_groups)
+            )
+        )
```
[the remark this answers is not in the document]
smailio: We're sure it never is

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,582 @@
...
+        return {ConstraintID(cid): self._row_to_bc(bc_rows[cid], terms[cid]) for cid in bc_rows}
+
+    @staticmethod
+    def _row_to_bc(row: Row[Any], terms: list[ConstraintTerm]) -> BindingConstraint:
+        d = get_row_representation_as_dict(row)
+        d["id"] = d.pop("constraint_id")
+        d["terms"] = terms
+        # extra="allow" is required because the join query returns columns that are
+        # not fields of BindingConstraint (such as `area1` or `lt_weight`).
+        return BindingConstraint.model_validate(d, extra="allow")
```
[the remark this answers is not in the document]
smailio: image

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,582 @@
...
+    def save_constraints(self, constraints: Sequence[BindingConstraint]) -> None:
+        changes = self._compute_matrix_changes(constraints)
+        self._save_constraint_rows(constraints)
+        self._save_cluster_terms(constraints)
+        self._save_link_terms(constraints)
+        self._apply_matrix_changes(changes)
+        self._cleanup_scenario_builder_groups()
+
+        self._db_session.commit()
+
```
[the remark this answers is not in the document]
smailio: done

antarest/study/business/model/binding_constraint_model.py
```diff
@@ -262,7 +264,7 @@ def set_id(cls, data: Any) -> Any:
             data["id"] = transform_name_to_id(data["name"])
         return data
 
-    id: str
+    id: ConstraintID
     name: str
```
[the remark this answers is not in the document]
smailio: Ah I forgot to make ConstraintID inherit from SanitizedStre, I planed to initially but forgot. It's done.

tests/variantstudy/model/command/test_manage_binding_constraints_db.py
```diff
@@ -0,0 +1,230 @@
...
+                    "terms": [ConstraintTerm(weight=50, data=ClusterTerm(area="area1", cluster="cluster"))],
+                },
+                "matrices": {},
+                "command_context": command_context,
+                "study_version": study_version,
+            }
+        ).apply(db_dao_versioned)
+        assert output.status, output.message
+
+        bd1 = db_dao_versioned.get_constraint("bd 1")
```
MartinBelthle: Minor: In this file my IDE is not happy that you're not using ConstraintId but just string, could you change this ?

tests/variantstudy/model/command/test_manage_binding_constraints_db.py
```diff
@@ -0,0 +1,230 @@
+# Copyright (c) 2026, RTE (https://www.rte-france.com)
```
MartinBelthle, line 1: Nice test, I'm just not sure on where we should put it. We can keep it here for the moment

tests/variantstudy/model/command/test_update_binding_constraints_database.py
```diff
@@ -0,0 +1,344 @@
...
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _bc(name: str, **kwargs: object) -> BindingConstraint:
+    """Minimal BindingConstraint factory for use in tests."""
+    defaults: dict = {
+        "name": name,
+        "group": "default",
+        "operator": BindingConstraintOperator.EQUAL,
```
MartinBelthle: Not sure to understand the benefit of this, why not just calling BindingConstraint in your test ? You just used default values

tests/variantstudy/model/command/test_update_binding_constraints_database.py
```diff
@@ -0,0 +1,344 @@
...
+    )
+    assert cmd.apply(db_dao_87).status is False
+
+
+# ---------------------------------------------------------------------------
+# test_to_dto — pure serialisation test, identical to the FS version
+# ---------------------------------------------------------------------------
+
+
+def test_to_dto(command_context: CommandContext) -> None:
```
MartinBelthle: Does not seem necessary as it's identical and there is no mention of DB in this test

tests/study/dao/test_binding_constraint_dao.py
```diff
@@ -0,0 +1,793 @@
...
+    return BindingConstraintNotFound if isinstance(dao, DatabaseStudyDao) else ChildNotFoundError
+
+
+def _assert_reset_matrix(df: pl.DataFrame) -> None:
+    """Assert that a matrix was reset after a time-step change.
+
+    Both backends are accepted: the DB DAO stores the null matrix (empty DataFrame),
+    while the file DAO reads back a correctly-sized all-zero matrix from disk.
+    """
+    assert df.is_empty() or df.to_series(0).sum() == 0.0
```
MartinBelthle: Really minor but could be df.sum().sum() == 0 as you'll handle the case where we have the 2nd column which is filled with other values than 0.

tests/study/dao/test_binding_constraint_dao.py
```diff
@@ -0,0 +1,793 @@
...
+    dao.save_constraints([])
+    assert set(dao.get_all_constraints().keys()) == {"c1", "c2"}
+
+    # Explicit delete
+    dao.delete_constraints([c1])
+    assert "c1" not in dao.get_all_constraints()
+    assert "c2" in dao.get_all_constraints()
+
+    # Deleting a nonexistent constraint is a noop
+    dao.delete_constraints([_bc("ghost")])
```
MartinBelthle: This souldn't be the case we should raise in such case

tests/study/dao/test_binding_constraint_dao.py
```diff
@@ -0,0 +1,793 @@
...
+    r = dao.get_constraint("bc_cart")
+    assert len(r.terms) == 3
+    assert len([t for t in r.terms if isinstance(t.data, LinkTerm)]) == 2
+    assert len([t for t in r.terms if isinstance(t.data, ClusterTerm)]) == 1
+
+    # Term replacement scenario: A B C → B C' (A removed, C updated, sibling bc2 untouched)
+    bc1_v1 = _bc(
+        "bc1",
+        terms=[
+            ConstraintTerm(weight=1.0, data=LinkTerm(area1="a", area2="b")),  # A – will be removed
```
MartinBelthle: The comment on A is misleading I thought you expected it to be removed here, you can remove it as the first comment you put above is clear in itself. Same for B and C actually you can remove them too

MartinBelthle (commented):
Last round of comments, after that we're good to go :)

tests/variantstudy/model/command/test_manage_binding_constraints_db.py
smailio: I put it here because it's a clone of test_manage_binding_constraints.py. 

Unlike the dao test I couldn't parametrize it. The assert of the other test rely heavily on checking directly the files.

tests/study/dao/test_binding_constraint_dao.py
smailio: It's a no op in the file study dao, we should update it there ? Would this be a breaking change ? 

This test covers both db dao and fs dao.

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,587 @@
...
+        self._db_session.execute(
+            delete(SCENARIO_BINDING_CONSTRAINTS_TABLE).where(
+                (SCENARIO_BINDING_CONSTRAINTS_TABLE.c.study_id == self._study_id)
+                & SCENARIO_BINDING_CONSTRAINTS_TABLE.c.bc_group_id.not_in(active_groups)
+            )
+        )
+
+    def _bc_to_row(self, study_id: str, bc: BindingConstraint) -> dict[str, Any]:
+        data = bc.model_dump(exclude={"id", "terms"})
+        return {"study_id": study_id, "constraint_id": bc.id, **data}
```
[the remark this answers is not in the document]
smailio: group is never null when get_version() < STUDY_VERSION_8_7

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,587 @@
...
+        return {ConstraintId(cid): self._row_to_bc(bc_rows[cid], terms[cid]) for cid in bc_rows}
+
+    @staticmethod
+    def _row_to_bc(row: Row[Any], terms: list[ConstraintTerm]) -> BindingConstraint:
+        d = get_row_representation_as_dict(row)
+        d["id"] = d.pop("constraint_id")
+        d["terms"] = terms
+        # extra="allow" is required because the join query returns columns that are
+        # not fields of BindingConstraint (such as `area1` or `lt_weight`).
+        return BindingConstraint.model_validate(d, extra="allow")
```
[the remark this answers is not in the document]
smailio: no thanks

tests/variantstudy/model/command/test_update_binding_constraints_database.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+def test_apply_database(db_dao_87: DatabaseStudyDao, command_context: CommandContext) -> None:
+    """
+    Pre-seed bc_1 (GREATER/DAILY/old_group1) and bc_2 (LESS/HOURLY/old_group2),
+    then run UpdateBindingConstraints with new group, operator and time_step,
+    and assert the persisted state via dao.get_constraint().
+    """
+    dao = db_dao_87
+    bc_0: ConstraintId = "bc_0"
```
MartinBelthle: Nice improvement but you could do `bc_0: ConstraintId =ConstraintId("bc_0")` as my IDE is still not happy. Same for other tests

antarest/study/dao/database/database_binding_constraint_dao.py
```diff
@@ -0,0 +1,587 @@
...
+        self._db_session.commit()
+
+    @override
+    def delete_constraints(self, constraints: list[BindingConstraint]) -> None:
+        db = self._db_session
+        constraint_ids = {bc.id for bc in constraints}
+
+        # Delete BC rows first so the table reflects the final state,
+        # then prune orphaned groups via a single subquery.
+        db.execute(delete(BC).where((BC.c.study_id == self._study_id) & (BC.c.constraint_id.in_(constraint_ids))))
```
MartinBelthle, line 589: As discussed we should just add a comment saying why we're not raising when the BC does not exist.

MartinBelthle (dismissed):
Just a comment to reach the 100 :)

MartinBelthle (approved):

TheoPascoli (approved):

PR #3134, opened 2026-03-31
feat(ui-thermal): allow to collapse 'Other emission rates' fields

skamril:
add collapse support for `<Fieldset>`

### DIFF
--- a/webapp/src/components/Fieldset.tsx
+++ b/webapp/src/components/Fieldset.tsx
@@ -13,20 +13,59 @@
  */
 
 import { mergeSxProp } from "@/utils/muiUtils";
-import { Box, Divider, type BoxProps, type SxProps, type Theme } from "@mui/material";
-import * as RA from "ramda-adjunct";
+import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
+import {
+  Box,
+  Collapse,
+  Divider,
+  IconButton,
+  Stack,
+  Tooltip,
+  type BoxProps,
+  type SxProps,
+  type Theme,
+} from "@mui/material";
+import { useState } from "react";
+import { useTranslation } from "react-i18next";
 
 interface FieldsetProps {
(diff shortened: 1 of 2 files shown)


hdinia (changes requested):
Latent bug: `onClick` attached unconditionally to the legend wrapper

`onClick={handleToggleCollapse}` is always set on the `<Stack component="legend">` even when collapsible is false. 
The inner if (canCollapse) guard makes it a no-op today, but the handler still intercepts every click that bubbles up from the legend's children.

This isn't currently observable because the only collapsible fieldset in this PR has a plain string legend. But the Fieldset component is also used with interactive legends -> e.g. `StudyLaunchDialog/Fields.tsx:230` passes a `<SwitchFE>` as legend. 
If someone later adds `collapsible` to that fieldset, clicking the switch would simultaneously toggle the switch and collapse the section, with no way to prevent it.

Two minimal fixes:
```tsx
// 1. Only attach the handler when it can actually do something
onClick={canCollapse ? handleToggleCollapse : undefined}

// 2. Stop propagation on the interactive part of the legend when collapsible
<Box onClick={(e) => e.stopPropagation()}>{legend}</Box>
```

hdinia:
Non-blocking questions to raise with PO/UX 

Two small spec/UX points worth noting :

1. **Collapse state is not persisted** : if the user collapses a section, navigates away, and comes back, the section resets to its default state. Is that intentional? Depending on the answer, we may want to persist it (e.g. in `localStorage` or the URL).

2. **Validation errors inside a collapsed section silently block submission** : a user can collapse a section that contains invalid fields and hit Submit; the form won't go through but there's no visible feedback. Is this acceptable UX, or should we auto-expand sections with errors and/or surface a hint on the legend? Worth a quick alignment with the PO.

Neither is a blocker for merging, just flagging so we don't discover them later and flag them as bugs

hdinia (approved):

PR #3136, opened 2026-04-01
fix(study): allow modification on DB studies metadata

MartinBelthle:
2 endpoints that used to fail now work:
- PUT /studies/{uuid}
- PUT /studies/{uuid}/owner

### DIFF
--- a/antarest/study/business/study_interface.py
+++ b/antarest/study/business/study_interface.py
@@ -21,7 +21,7 @@ from antarest.matrixstore.service import ISimpleMatrixService
 from antarest.study.dao.api.study_dao import ReadOnlyStudyDao
 from antarest.study.dao.file.file_study_dao import FileStudyTreeDao
 from antarest.study.dao.memory.in_memory_study_dao import InMemoryStudyDao
-from antarest.study.model import StudyMetadata
+from antarest.study.model import StudyMetadataUpdate
 from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
 from antarest.study.storage.variantstudy.model.command.common import CommandOutput
 from antarest.study.storage.variantstudy.model.command.icommand import ICommand
@@ -73,7 +73,7 @@ class StudyInterface(ABC):
         raise NotImplementedError()
 
     @abstractmethod
-    def set_study_metadata(self, metadata: StudyMetadata) -> None:
+    def update_study_metadata(self, metadata: StudyMetadataUpdate) -> None:
         raise NotImplementedError()
 
 
@@ -113,7 +113,7 @@ class InMemoryStudyInterface(StudyInterface):
         return self._study_dao.read_only()
 
     @override
-    def set_study_metadata(self, metadata: StudyMetadata) -> None:
+    def update_study_metadata(self, metadata: StudyMetadataUpdate) -> None:
         self._study_dao.update_antares_file(metadata)
 
 
@@ -166,5 +166,5 @@ class FileStudyInterface(StudyInterface):
         return FileStudyTreeDao(self.file_study, self._generator_matrix_constants, self._blob_service)
 
     @override
-    def set_study_metadata(self, metadata: StudyMetadata) -> None:
+    def update_study_metadata(self, metadata: StudyMetadataUpdate) -> None:
         self._get_dao().update_antares_file(metadata)
--- a/antarest/study/business/study_interface.py
+++ b/antarest/study/business/study_interface.py
@@ -18,9 +18,10 @@ from typing_extensions import override
 
 from antarest.core.exceptions import CommandApplicationError
 from antarest.matrixstore.service import ISimpleMatrixService
-from antarest.study.dao.api.study_dao import ReadOnlyStudyDao, StudyMetadata
+from antarest.study.dao.api.study_dao import ReadOnlyStudyDao
 from antarest.study.dao.file.file_study_dao import FileStudyTreeDao
 from antarest.study.dao.memory.in_memory_study_dao import InMemoryStudyDao
+from antarest.study.model import StudyMetadata
 from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
 from antarest.study.storage.variantstudy.model.command.common import CommandOutput
 from antarest.study.storage.variantstudy.model.command.icommand import ICommand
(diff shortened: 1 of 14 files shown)


antarest/study/dao/study_conversion/study_converter.py
```diff
@@ -95,7 +95,7 @@ def _convert_settings(self) -> None:
     def _convert_xpansion(self) -> None:
         try:
             settings = self._source_dao.get_xpansion_settings()
-        except ChildNotFoundError:
+        except (ChildNotFoundError, XpansionConfigurationDoesNotExist):
```
MartinBelthle, line 98: Nothing to do with the PR but fixes an issue inside the conversion DB -> FS

antarest/study/dao/api/study_dao.py
```diff
@@ -82,6 +83,14 @@
 from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
 
 
+@dataclass(frozen=True)
+class StudyMetadata:
```
sylvlecl: we'd better define it outside from the DAO layer since this is more general metadata than the study data handled by the DAO, in model.py I guess.
Nice to have made it frozen!

antarest/study/service.py
```diff
@@ -881,34 +879,24 @@ def update_study_information(self, uuid: str, metadata_patch: StudyMetadataPatch
...
-            study_antares = self.storage_service.get_storage(study).get(study, study_antares_url)
-
-            if metadata_patch.author:
-                study_antares["author"] = metadata_patch.author
-
-            if metadata_patch.name:
-                study_antares["caption"] = metadata_patch.name
-
-            self._edit_study_using_command(study=study, url=study_antares_url, data=study_antares)
+            study_metadata = StudyMetadata(author=metadata_patch.author, name=metadata_patch.name)
```
sylvlecl: nice replacement

antarest/study/service.py
```diff
@@ -589,6 +582,11 @@ def add_commands(self, commands: Sequence[ICommand], listener: ICommandListener
         # get current user if not in session, otherwise get session user
         self._variant_service.append_commands(self._study.id, transform_command_to_dto(commands, force_aggregate=True))
 
+    @override
+    def set_study_metadata(self, metadata: StudyMetadata) -> None:
+        """We won't perform any operations on variants as the change is already reflected in the database"""
```
sylvlecl: Note that there is dirty mechanisms in the snapshot generator related to this, that we should clean up at some point:
- before generation we update files from DB, ok
- after generation, we update DB from files ... I think this should be removed. I guess it was to make metadata update work through commands, but as we have seen it should never have been this way

tests/integration/study_data_blueprint/test_area_database_mode.py
```diff
@@ -418,10 +418,22 @@ def test_area_operations_consistency_database(self, client: TestClient, user_acc
...
+    client: TestClient, user_access_token: str, admin_access_token: str
+) -> None:
+    client.headers = {"Authorization": f"Bearer {admin_access_token}"}
+    preparer = PreparerProxy(client, admin_access_token)
     study_id = preparer.create_study("MyStudy", version=870, storage_mode="database")
+    # We should be able to rename the study
+    res = client.put(f"/v1/studies/{study_id}", json={"name": "MyRenamedStudy"})
+    assert res.status_code == 200
+    res = client.get(f"/v1/studies/{study_id}")
+    assert res.json()["name"] == "MyRenamedStudy"
```
sylvlecl, line 438: nice

sylvlecl (commented):
Very minor comment about class place

antarest/study/service.py
MartinBelthle: Arf, sounds complicated I'll see what I can do

MartinBelthle: As discussed I tried to fix it and added a test to ensure the fix worked

antarest/study/service.py
```diff
@@ -917,7 +910,10 @@ def update_study_information(self, uuid: str, metadata_patch: StudyMetadataPatch
         if metadata_patch.horizon:
             study.horizon = metadata_patch.horizon
         if metadata_patch.tags is not None:
-            self.repository.update_tags(study, metadata_patch.tags)
+            self.repository.update_tags(study, metadata_patch.tags)  # This method also saves the modified study in DB
+        else:
```
sylvlecl: Will be simpler to just have the save before the tags handling.
One more request/commit but honestly here it's not critical at all

antarest/study/service.py
```diff
@@ -1818,7 +1814,13 @@ def change_owner(self, study_id: str, owner_id: int) -> None:
         study = self.get_study(study_id)
         assert_permission(study, StudyPermissionType.MANAGE_PERMISSIONS)
         self.assert_study_unarchived(study)
+
         new_owner = self.user_service.get_user(owner_id)
+        if not new_owner:
+            logger.error("user %d not found by user %s", id, get_user_id())
+            raise UserNotFoundError()
+        assert new_owner is not None
```
sylvlecl: does not seem useful, redundant with the above condition

antarest/study/business/study_interface.py
```diff
@@ -151,4 +160,11 @@ def add_commands(self, commands: Sequence[ICommand], listener: ICommandListener
...
     @override
     def get_study_dao(self) -> ReadOnlyStudyDao:
-        return FileStudyTreeDao(self.file_study, self._generator_matrix_constants, self._blob_service).read_only()
+        return self._get_dao().read_only()
+
+    def _get_dao(self) -> FileStudyTreeDao:
+        return FileStudyTreeDao(self.file_study, self._generator_matrix_constants, self._blob_service)
+
+    @override
+    def set_study_metadata(self, metadata: StudyMetadata) -> None:
```
sylvlecl: Naming: this is actually more an update, since None elements will be ignored

antarest/study/model.py
```diff
@@ -731,3 +733,11 @@ class DirectoryUpdate(AntaresBaseModel):
     def validate_name(cls, v: str | None) -> str | None:
         """Validate directory name."""
         return _validate_directory_name(v) if v is not None else v
+
+
+@dataclasses.dataclass(frozen=True)
+class StudyMetadata:
```
sylvlecl: Maybe name it "update" since None must be understood as "dont change it"

sylvlecl (dismissed):
Nice, I left some renaming suggestions that you may implement if you agree it's relevant

PR #3137, opened 2026-04-01
fix(parquet): ensure reasonably large row group size

sylvlecl:
Solves 2 issues at once:

- mainly, we observed in production that using too many row groups for one parquet file caused the metadata for those row groups to occupy a LOT of memory, which caused OOM killer to take action
- in any case, it was an anti-pattern for IO to have such small row groups (as small as 1 row), which were only the consequence of the input text files.

We now use a row group size in line with recommendations, of 1M.

We may scale it down a little if needed if:
- we find that this makes too large row groups in memory, which may happen if we get too many columns
- we find that it's more efficient to retrieve data from slightly smaller row groups

**Consequence:**
Now memory usage will not increase anymore (only marginally) with the number of rows, but only with the number of columns. The number of columns is mostly contained. It can only be increased by additional "groups" defined in input data (cluster groups etc.)

**Test results:**
 - For aggregation of links output data on a large study, resident memory:
     - before: 1.2G
     - after: 500 Mo
 - For aggregation of areas output data on a large study, resident memory:
     - before: 500 Mo
     - after: 800 Mo --> larger than for links, because there are more columns. But it will stay flat when adding more rows (more areas)

### DIFF
--- a/antarest/core/serde/parquet_writer.py
+++ b/antarest/core/serde/parquet_writer.py
@@ -35,7 +35,7 @@ class BatchParquetWriter:
 
     It's very important to use not too small row groups, in order to:
      - have efficient IO
-     - avoid excessive memory usage for metadata (as shown by experienced)
+     - avoid excessive memory usage for metadata (as shown by experience)
 
     """
 
--- a/antarest/core/serde/parquet_writer.py
+++ b/antarest/core/serde/parquet_writer.py
@@ -12,6 +12,7 @@
 
 from collections.abc import Iterator
 from pathlib import Path
+from typing import Any, Literal
 
 import pandas as pd
 import polars as pl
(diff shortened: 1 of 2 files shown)


tests/core/serde/test_parquet_writer.py
```diff
@@ -52,23 +55,18 @@ def test_different_columns(tmp_path: Path) -> None:
     )
     pd.testing.assert_frame_equal(next(dfs), expected_first_df, check_dtype=False)
 
-    expected_second_df = pd.DataFrame(data=[(5, 6, 7, np.nan), (8, 9, 10, np.nan)], columns=["A", "B", "C", "D"])
```
MartinBelthle, line 55: Not sure to understand why did you change this test ? Does not seem relative to the PR

MartinBelthle (commented):
Nice fix, just a question for the tests. Also did you see performance improvements timewise as we're performing less I/O operations ? Also not sure to understand why you use more memory for areas than before i'm probably not understanding some concept

tests/core/serde/test_parquet_writer.py
sylvlecl: That's because previously, we had one row group for each written dataframe.
Then, when reading from parquet file, we get one dataframe for each row group.

So:
- before: writing 5 DF -> 5 row groups -> reading 5 DFs
- after: writing 5 DF -> 3 row groups only (due to 3 parquet files, actually)  -> reading 3 DFs

sylvlecl:
> Nice fix, just a question for the tests. Also did you see performance improvements timewise as we're performing less I/O operations ? Also not sure to understand why you use more memory for areas than before i'm probably not understanding some concept

- Performance: no improvement in my tests, the bottleneck being reading from text files.
- About memory for areas:
  it's because in that case, the memory used for the actual data of a row group (1M rows) exceeds what was used by the 3000 metadata of row groups I had (3 areas x 1000 years).
  but it will scale better: if instead I had say 30 areas, the memory used by the metadata would be 10 times more with the old implementation. With the new implementation, we will still have row groups of 1M rows, only one at a time, so the memory used will almost not increase.

Will try to quantify it though :)

tests/core/serde/test_parquet_writer.py
MartinBelthle: Ah okay i see thanks

MartinBelthle (dismissed):

sylvlecl:
The plot below illustrates the scaling:
still using my study with 3 areas and 1000 years, the x axis here is the number of years aggregated, and the y axis is the memory used for different strategy:
- old one: you can see it scales linearly --> **will cause issues for larger studies**
- new one for 2 different row group limits, 500k or 1M:
  you see that in both cases memory increases fast until we fill up one row group, but then it stays mostly flat (increases probably a little with additional metadata of each row group)

plot

TheoPascoli (approved):

antarest/core/serde/parquet_writer.py
```diff
@@ -20,9 +21,62 @@
...
+        )
+
+        self._current_batch: list[pa.Table] = []
+        self._current_batch_size = 0
+        self._closed = False
+
+    def __enter__(self) -> "BatchParquetWriter":
+        return self
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
```
TheoPascoli, line 55: Minor: the "standard" signature for the exit is `object.__exit__(self, exc_type, exc_value, traceback)`. Idk if it changes anything


```suggestion
    def __exit__(self, exc_type: type | None, exc_val: BaseException | None, exc_tb: object) -> None:
```

antarest/core/serde/parquet_writer.py
```diff
@@ -20,9 +21,62 @@
...
-def _parquet_writer(output_file: Path, schema: pa.Schema) -> ParquetWriter:
-    return ParquetWriter(output_file, schema, compression="zstd", data_page_version="2.0")
+
+class BatchParquetWriter:
+    """
+    Wrapper around a plain parquet writer to ensure we write row groups of at least some target size.
+
+    It's very important to use not too small row groups, in order to:
+     - have efficient IO
+     - avoid excessive memory usage for metadata (as shown by experienced)
```
TheoPascoli:
```suggestion
     - avoid excessive memory usage for metadata (as shown by experience)
```

TheoPascoli (dismissed):

TheoPascoli:
Thats really nice that you spotted this :)

antarest/core/serde/parquet_writer.py
sylvlecl: I prefer to keep the "opaque" signature, I think it makes more clear that we don't use the args

PR #3138, opened 2026-04-02
refactor(variants)!: use the dao inside the snapshot generator

MartinBelthle:
**Breaking change**
The endpoint PUT `/studies/{uuid}/generate` to generate a variant no longer accepts `denormalize` as an argument

The change does not impact the R scripts (confirmed by Kamel)


antarest/study/storage/utils.py
```diff
@@ -89,9 +88,8 @@ def update_antares_info(metadata: Study, study_tree: FileStudyTree, update_autho
 
     # Update basic fields
     antares_info["caption"] = metadata.name
-    antares_info["created"] = _format_timestamp(metadata.created_at)
-    antares_info["lastsave"] = _format_timestamp(metadata.updated_at)
-    antares_info["version"] = _format_version(metadata.version)
```
sylvlecl, line 94: Question: why did you remove this ? Because it does not make sense to update it since it should be correctly defined at study creation or upgrade ?

antarest/study/storage/variantstudy/variant_command_generator.py
```diff
@@ -90,15 +88,14 @@ def _generate(
...
-    logger.info(f"{generation_type} done in {stopwatch.since_start}s")
+    logger.info(f"Variant generation done in {stopwatch.since_start}s")
     return results
 
 
 def apply_commands_to_variant(
     commands: list[list[ICommand]],
     metadata: VariantStudy,
-    study: FileStudy,
+    study: StudyDao,
```
sylvlecl, line 98: Nice, makes things more clear

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -395,61 +388,45 @@ def get_study_path(self, metadata: Study) -> Path:
             return self.find_archive_path(metadata)
         return Path(metadata.path)
 
-    def check_and_update_study_version_in_database(self, study: RawStudy) -> None:
```
sylvlecl, line 390: Wasn't used ? nice cleanup

antarest/study/storage/variantstudy/snapshot_generator.py
```diff
@@ -108,18 +108,30 @@ def generate_snapshot(
...
             logger.info(f"Applying commands to the reference study '{ref_study.id}'...")
-            results = self._apply_commands(file_study, variant_study, cmd_blocks, listener)
+            results = self._apply_commands(study_dao, variant_study, cmd_blocks, listener)
+
+            # Denormalize the variant study if asked
             if denormalize:
                 logger.info(f"Denormalizing variant study {variant_study_id}")
-                self.raw_study_service.denormalize_study(file_study)
+                # The operation only makes sense for FILEYSTEM mode
+                if variant_study.storage_mode == StorageMode.FILESYSTEM:
```
sylvlecl: All the if/else really call for a common interface.
Should normalize go into the DAO, or another more high level interface ?
Maybe DAO is OK since indeed this is about data that is managed by the DAO (study data)

antarest/study/storage/variantstudy/snapshot_generator.py
```diff
@@ -108,18 +108,30 @@ def generate_snapshot(
                 self._export_ref_study(snapshot_dir, ref_study)
 
             # The snapshot is generated, we also need to de-normalize the matrices.
```
sylvlecl, line 106: Note that for database mode, the copy which is done above should not be an export of the filestudy, but a copy of the study in database.

We should probably introduce a `copy` method in the dao factory to handle that

antarest/study/storage/variantstudy/snapshot_generator.py
```diff
@@ -108,18 +108,30 @@ def generate_snapshot(
...
-                snapshot_dir,
-                True,
-                study_id=variant_study_id,
-                output_path=snapshot_dir / OUTPUT_RELATIVE_PATH,
-                use_cache=True,
-            )
+            cmd_context = self.command_factory.command_context
+            factory: StudyFactoryDao
+            if variant_study.storage_mode == StorageMode.FILESYSTEM:
+                factory = FileStudyDaoFactory(cmd_context, self.study_factory)
```
sylvlecl: Would be best that the factory is provided in the constructor, I think, so that the snapshot generator does not need to know about those details

antarest/study/storage/variantstudy/snapshot_generator.py
```diff
@@ -108,18 +108,30 @@ def generate_snapshot(
...
-            results = self._apply_commands(file_study, variant_study, cmd_blocks, listener)
+            results = self._apply_commands(study_dao, variant_study, cmd_blocks, listener)
+
+            # Denormalize the variant study if asked
             if denormalize:
                 logger.info(f"Denormalizing variant study {variant_study_id}")
-                self.raw_study_service.denormalize_study(file_study)
+                # The operation only makes sense for FILEYSTEM mode
+                if variant_study.storage_mode == StorageMode.FILESYSTEM:
+                    file_study = self.study_factory.create_from_fs(
```
sylvlecl: Seems this will already be done when instantiating the DAO

antarest/study/storage/variantstudy/snapshot_generator.py
```diff
@@ -133,8 +145,9 @@ def generate_snapshot(
             if results.should_invalidate_cache:
                 # We need to remove the cache
                 remove_from_cache(self.cache, variant_study_id)
-            else:
-                data = FileStudyTreeConfigDTO.from_build_config(file_study.config).model_dump()
+            elif variant_study.storage_mode == StorageMode.FILESYSTEM:
+                # todo: We do not handle cache for DB mode. Same for the `add_commands` by the way.
```
sylvlecl: Indeed, should probably be hidden too inside the DAO ?

sylvlecl (commented):
Good evolution: will need further work however to make the snapshot generator work with DB study, involving more use of interfaces (existing or new ones ?) in the snapshot generator.

Do we want to merge this as a first step and postpone that to later work ?

antarest/study/storage/rawstudy/raw_study_service.py
MartinBelthle: Yes it wasn't used

antarest/study/storage/variantstudy/snapshot_generator.py
MartinBelthle: It won't at the moment as inside the `create_study_dao` I build my `file_study` with normalized matrices. It could be solved if we added the parameter `normalize_matrices: bool` or something like it inside the `create_study_dao` method.
It would also improve perfs as we won't write matrices twice in such case.

What do you think ?

antarest/study/storage/variantstudy/snapshot_generator.py
MartinBelthle: I'll introduce an `update_cache` method inside the DAO that will be empty on DB side for the moment

antarest/study/storage/variantstudy/snapshot_generator.py
MartinBelthle: Fair enough

antarest/study/storage/variantstudy/snapshot_generator.py
MartinBelthle: Now we don't need for this here anymore but the remark is still valid as inside raw_study_service and variant_study_service we do a `if DB: return`. Not sure what to do

antarest/study/storage/variantstudy/snapshot_generator.py
MartinBelthle: Indeed but I don't think I want to do this inside this PR, this will go in a future one where we actually handle the snapshots

antarest/study/storage/utils.py
MartinBelthle: Yes it's because we call this function at study copy or creation so we should not need to update the version

sylvlecl: ok!

antarest/study/storage/variantstudy/snapshot_generator.py
```diff
@@ -134,8 +122,7 @@ def generate_snapshot(
                 # We need to remove the cache
                 remove_from_cache(self.cache, variant_study_id)
             else:
-                data = FileStudyTreeConfigDTO.from_build_config(file_study.config).model_dump()
-                update_cache(self.cache, variant_study_id, data)
+                study_dao.update_cache()
```
sylvlecl, line 125: Just a note for future work:
in the end, the cache should be completely managed by the DAO:
the DAO should know when its own cache needs to be invalidated.

It means that the commands themselves should not know about it, maybe apart from the update_config which should invalidate it explicitly (or would call an "update_file" method or something like that from the DAO, which internally invalidates the cache)

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -472,8 +443,7 @@ def ignore_outputs(directory: str, _: Sequence[str]) -> Sequence[str]:
         stop_time = time.time()
         duration = f"{stop_time - start_time:.3f}"
         logger.info(f"Study '{study_dir}' exported (flat mode) in {duration}s")
-        if denormalize:
-            study = self.study_factory.create_from_fs(dest, is_study_managed, "", use_cache=False)
-            self.denormalize_study(study)
-            duration = f"{time.time() - stop_time:.3f}"
-            logger.info(f"Study '{study_dir}' denormalized in {duration}s")
+
+    def denormalize_exported_study(self, dst_dir: Path) -> None:
```
sylvlecl: Not sure of the added value of separating this from the export method ?
it **must** be called after the export, so seems it made sense to include it in the method itself.
Also, it makes less explicit that the "export flat" may leave some link files inside the exported study, which could cause errors for future developers.

antarest/study/service.py
```diff
@@ -421,7 +420,7 @@ def _upgrade_study(self) -> None:
                 )
             finally:
                 if is_study_denormalized:
-                    self.storage_service.raw_study_service.normalize_study(study_to_upgrade)
+                    self.storage_service.get_storage(study_to_upgrade).normalize_study(study_to_upgrade)
```
sylvlecl: what about having the "normalization" in the DAO ? 
Do we postpone this to later too ?

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -387,61 +380,45 @@ def get_study_path(self, metadata: Study) -> Path:
...
     @staticmethod
     def checks_antares_web_compatibility(study: Study) -> None:
         """
         A new compatibility section has been introduced with the Simulator version 9.2
         """
         pass
 
-    def normalize_study(self, study: Study | FileStudy) -> None:
+    @override
+    def normalize_study(self, study: Study) -> None:
```
sylvlecl, line 389: here too the ifs hint that we should move those methods to an abstraction like the DAO, although it will make it a leaky abstraction.

sylvlecl (approved):

PR #3141, opened 2026-04-07
feat(study): compute study synthesis from database via DAO

TheoPascoli:
Moves `get_synthesis()` from the storage layer (filesystem-based) to the DAO layer, with a full implementation in `DatabaseStudyDao` that rebuilds the synthesis from DB queries instead of reading the file tree. This removes the dependency on a generated snapshot for variants


antarest/study/dao/database/database_study_dao.py
```diff
@@ -105,6 +108,62 @@ def __init__(
...
+
+        advanced = self.get_advanced_parameters()
+        enr_modelling = EnrModelling(advanced.renewable_generation_modelling.value)
+
+        return StudyDataSynthesis.model_construct(
+            study_id=study_id,
+            version=version,
+            areas=areas,
+            districts=districts,
+            bindings=[],
```
sylvlecl, line 163: Good thing that you could not implement it, we will remove it shortly :)

tests/study/dao/test_database_synthesis_dao.py
```diff
@@ -0,0 +1,127 @@
...
+        dao.save_renewable("france", RenewableCluster(id="wind_farm", name="Wind Farm"))
+        dao.save_st_storage("germany", STStorage(id="battery", name="Battery"))
+
+        synthesis = dao.get_synthesis()
+
+        assert "france" in synthesis.areas
+        assert "germany" in synthesis.areas
+
+        france = synthesis.areas["france"]
+        assert france.name == "France"
```
sylvlecl: I think we could benefit from a plain equality comparison, to check all fields.
Same for other tests!

sylvlecl (commented):
Nice, clean and simple :)

I think it would be slightly better to test all fields equality in unit tests

sylvlecl (approved):

Issue #3118, opened 2026-03-23
Multi-Version Upgrade Does Not Complete in One Step


nikolaredstork:
## Description

The upgrade system does not correctly handle upgrades across multiple versions. When attempting to upgrade directly from an older version to a newer version that is more than one version ahead, the system only performs a single incremental upgrade instead of upgrading to the target version. This is tested on AntaREST 2.28.0
### Steps to reproduce the behavior:
1. Create a study with version 8.3.
2. Initiate an upgrade to version 8.8 (or any version more than one step ahead).

<img width="1354" height="672" alt="Image" src="https://github.com/user-attachments/assets/eca5a1de-8398-4920-bbb5-e1ff8c5d85f9" />

4. Wait for the upgrade process to complete.
5. Check the resulting version.

<img width="739" height="371" alt="Image" src="https://github.com/user-attachments/assets/d93e3e4d-aec7-4a72-b6b5-559a830fa249" />

### Describe what actually happened.
The system upgrades only one version step (from 8.3 to 8.4). To reach version 8.8, the upgrade process must be executed again.

### Describe what you expected to happen.
The system should upgrade directly from version 8.3 to version 8.8 in a single process.

TheoPascoli:
Hello @nikolaredstork, thank you for reporting this bug. It has already been identified and will be fixed in the next release

TheoPascoli:
This has been fixed by [#3142]

PR #3142, opened 2026-04-07
fix(ui-api): pass target_version as query parameter instead of json to upgrade studies

ctmbl:
Hello @TheoPascoli !

Allow me to propose this PR to solve #3118 

After some investigation it turns out that the front end sends `target_version` embedded in a JSON object :
```json
{
  "target_version" : "9.3"
}
```
while the backend expects a string, hence the bug where every upgrade of a study's version would just upgrade to the next one.

To solve this I implemented a model `StudyVersionModel` to allow fastapi to parse JSON then retrieve `target_version` as expected.

I would be very happy to get your opinion on the matter / my patch.
Feel free also to either get inspiration from this or leave me some recommendations, as you prefer :wink:

### DIFF
--- a/webapp/src/services/api/study.ts
+++ b/webapp/src/services/api/study.ts
@@ -188,8 +188,10 @@ export const unarchiveStudy = async (sid: string): Promise<void> => {
 };
 
 export const upgradeStudy = async (studyId: string, targetVersion: string): Promise<void> => {
-  await client.put(`/v1/studies/${studyId}/upgrade`, {
-    target_version: compactSemanticVersion(targetVersion),
+  await client.put(`/v1/studies/${studyId}/upgrade`, null, {
+    params: {
+      target_version: compactSemanticVersion(targetVersion),
+    },
   });
 };
 


TheoPascoli:
Hi @ctmbl, thanks for looking into this and submitting a PR

 However, we can't merge it as-is for a few reasons:

  1. Breaking change: The endpoint currently accepts `target_version` as a query parameter. Switching to a body model would break all existing clients and our integration tests.
  2. Ambiguous union type: `StudyVersionModel | str` mixes body and query parameter semantics, FastAPI doesn't resolve this reliably.
  3. Runtime error: If `target_version` is received as a plain string, `target_version.target_version` will raise an `AttributeError`.
  4. No tests: No new tests were added, and the existing ones (which use query params) would likely break.

Actually, I think the actual root cause is on the frontend side !

ctmbl:
Hi @TheoPascoli !

Tks a lot for looking so quickly into this!
Now that you're stating it I clearly see the flaws of my current solution, of course patching it on the frontend side makes much more sense.

I'll quickly push a new proposition then, but if you'd like me to close this PR and open another one just say it 😉

webapp/src/services/api/study.ts
```diff
@@ -188,9 +188,9 @@ export const unarchiveStudy = async (sid: string): Promise<void> => {
 };
 
 export const upgradeStudy = async (studyId: string, targetVersion: string): Promise<void> => {
-  await client.put(`/v1/studies/${studyId}/upgrade`, {
```
skamril, line 191:
```ts
  await client.put(`/v1/studies/${studyId}/upgrade`, null, {
    params: {
      target_version: compactSemanticVersion(targetVersion),
    },
  });
```

Axios automatically handles params encoding.

skamril (changes requested):

webapp/src/services/api/study.ts
ctmbl: Tks for the review! done in d8dbdeb

ctmbl:
Hi @skamril 

Tks for your review, I didn't know about this because, obviously, I'm not familiar with the framework 😅

I just implemented your recommendations and force-pushed a new commit!

skamril (approved):

TheoPascoli (approved):

PR #3146, opened 2026-04-09
perf(dao): less db requests for area matrices

MartinBelthle:


tests/matrixstore/test_matrix_usage_providers.py
```diff
@@ -229,6 +229,10 @@ def test_command_matrix_usage_provider_with_snapshot(
         config=Mock(),
         user_service=Mock(),
     )
+    # Initialize default matrices in the matrix_service
```
MartinBelthle: Subtle difference with the previous code but before to know if a matrix existed we relied on the matrix_service inside a node and now we rely on the file_study_dao one. In real case they are the same but in some tests they differ. And here the node matrix_service is initialized by the command_context fixture while the dao one is initialized in the test.

MartinBelthle: In the end I rewrote the setup of this test to use the same matrix_service everywhere

sylvlecl (approved):

PR #3149, opened 2026-04-10
fix(ui-matrix): allow decimal and clearable value in operation field

hdinia:
ANT-4774

### DIFF
--- a/webapp/src/components/Matrix/components/MatrixFilter/Operations.tsx
+++ b/webapp/src/components/Matrix/components/MatrixFilter/Operations.tsx
@@ -84,7 +84,9 @@ function Operations({ filter, setFilter, onApplyOperation }: OperationsProps) {
             <Box sx={{ flex: 1 }}>
               <NumberFE
                 label={t("matrix.filter.value")}
-                value={value}
+                // Falls back to empty string when value is undefined to keep
+                // the input controlled.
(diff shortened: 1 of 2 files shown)


webapp/src/components/Matrix/components/MatrixFilter/Operations.tsx
```diff
@@ -83,9 +82,8 @@ function Operations({ filter, setFilter, onApplyOperation }: OperationsProps) {
 
           {filter.operation.type !== Operation.Abs && (
             <Box sx={{ flex: 1 }}>
-              <TextField
+              <NumberFE
                 label={t("matrix.filter.value")}
-                type="number"
                 value={value}
```
skamril: > A component is changing a controlled input to be uncontrolled. This is likely caused by the value changing from a defined to undefined, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://react.dev/link/controlled-components

To prevent this error:

`value={typeof value === "undefined" ? ("" as unknown as number) : value}`

`"" as unknown as number` is not "clean", but fixing this properly would require refactoring NumberFE, which was out of scope here.

skamril (changes requested):

skamril (approved):

PR #3154, opened 2026-04-13
feat(reserves): implement global parameters

TheoPascoli:


antarest/study/business/model/reserves_global_parameters_model.py
```diff
@@ -0,0 +1,42 @@
...
+#
+# This file is part of the Antares project.
+from pydantic import ConfigDict, Field
+
+from antarest.core.serde import AntaresBaseModel
+from antarest.core.utils.string import to_camel_case
+
+
+class ReservesGlobalParameters(AntaresBaseModel):
+    model_config = ConfigDict(alias_generator=to_camel_case, extra="forbid", populate_by_name=True)
```
MartinBelthle: Really minor but we should use the pydantic `to_camel` instead of the one we defined

antarest/study/business/reserves_global_parameters_management.py
```diff
@@ -0,0 +1,60 @@
...
+        raise AreaNotFound(*invalid_areas)
+
+
+class ReservesGlobalParametersManager:
+    def __init__(self, command_context: CommandContext) -> None:
+        self._command_context = command_context
+
+    def get_reserves_global_parameters(self, study: StudyInterface, area_id: str) -> ReservesGlobalParameters:
+        _check_version(study)
+        _check_area_exists(study, area_id)
```
MartinBelthle: Here it seems we'll do this check inside the DAO so don't need to duplicate it, it's only useful for the `update` as we go through a command. Same for the `_check_version` that could even be removed from the 2 methods and just put inside a model validator inside the command

antarest/study/business/reserves_global_parameters_management.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def update_reserves_global_parameters(
+        self,
+        study: StudyInterface,
+        area_id: str,
+        parameters: ReservesGlobalParametersUpdate,
+    ) -> ReservesGlobalParameters:
+        _check_version(study)
+        _check_area_exists(study, area_id)
+        command = UpdateReservesGlobalParameters(
```
MartinBelthle, line 36: Why not creating a command that already handles updates for several areas ? We'll for sure need it for performance issue

antarest/study/dao/api/reserves_global_parameters_dao.py
```diff
@@ -0,0 +1,26 @@
...
+
+class ReadOnlyReservesGlobalParametersDao(ABC):
+    @abstractmethod
+    def get_reserves_global_parameters(self, area_id: str) -> ReservesGlobalParameters:
+        raise NotImplementedError()
+
+
+class ReservesGlobalParametersDao(ReadOnlyReservesGlobalParametersDao):
+    @abstractmethod
+    def save_reserves_global_parameters(self, area_id: str, params: ReservesGlobalParameters) -> None:
```
MartinBelthle: Same remark here, we should introduce a `get_all` method and the `save` should accept a mapping to be performant inside the `study_converter`

antarest/study/dao/file/file_study_reserves_global_parameters_dao.py
```diff
@@ -0,0 +1,53 @@
...
+    def get_file_study(self) -> FileStudy:
+        pass
+
+    @override
+    def get_reserves_global_parameters(self, area_id: str) -> ReservesGlobalParameters:
+        file_study = self.get_file_study()
+        try:
+            data = file_study.tree.get(_get_reserves_ini_path(area_id))
+        except (ChildNotFoundError, KeyError):
+            return ReservesGlobalParameters()
```
MartinBelthle: Minor but it would be preferable to do `data = {}` and always go through the parsing method as if in future versions we add data in the model we'll have the initialization done

antarest/study/storage/variantstudy/model/command/update_reserves_global_parameters.py
```diff
@@ -0,0 +1,69 @@
...
+    command_name: CommandName = CommandName.UPDATE_RESERVES_GLOBAL_PARAMETERS
+
+    area_id: str
+    parameters: ReservesGlobalParametersUpdate
+
+    @override
+    def _apply_dao(
+        self, study_data: StudyDao, listener: ICommandListener | None = None
+    ) -> CommandOutput[ReservesGlobalParameters]:
+        if self.study_version < STUDY_VERSION_10_0:
```
MartinBelthle: As said in a previous comment, you should just move this in the model validator

MartinBelthle (changes requested):
We should add integration tests and also the conversion code in the `study_converter`

antarest/study/dao/api/reserves_global_parameters_dao.py
```diff
@@ -0,0 +1,34 @@
...
+        raise NotImplementedError()
+
+    @abstractmethod
+    def save_all_reserves_global_parameters(self, mapping: dict[str, ReservesGlobalParameters]) -> None:
```
MartinBelthle: To me we should only keep this method (but keep the other one name) and declare the type `dict[str, ReservesGlobalParameters]` inside the `common.py` file to fit with the rest of the DAOs

antarest/study/dao/database/database_reserves_global_parameters_dao.py
```diff
@@ -0,0 +1,91 @@
...
+
+from antarest.study.business.model.reserves_global_parameters_model import ReservesGlobalParameters
+from antarest.study.dao.api.reserves_global_parameters_dao import ReservesGlobalParametersDao
+from antarest.study.dao.database.models.area import RESERVES_GLOBAL_PARAMETERS_TABLE
+from antarest.study.dao.database.sql_utils import upsert_one
+
+_TABLE = RESERVES_GLOBAL_PARAMETERS_TABLE
+
+
+def _convert_row_to_model(row: Any) -> ReservesGlobalParameters:
```
MartinBelthle: Minor but could be typed `Row[Any]`

antarest/study/dao/database/database_reserves_global_parameters_dao.py
```diff
@@ -0,0 +1,91 @@
...
+        study_id = self.get_study_id()
+        session = self.get_session()
+        stmt = select(_TABLE).where(_TABLE.c.study_id == study_id)
+        rows = session.execute(stmt)
+        return {row.area_id: _convert_row_to_model(row) for row in rows}
+
+    @override
+    def save_all_reserves_global_parameters(self, mapping: dict[str, ReservesGlobalParameters]) -> None:
+        for area_id, params in mapping.items():
+            self._upsert_one(area_id, params)
```
MartinBelthle: We should use `upsert_multiple` as this is highly unperformant

MartinBelthle (changes requested):

antarest/study/dao/database/database_reserves_global_parameters_dao.py
```diff
@@ -0,0 +1,88 @@
...
+        return self._db_session
+
+    @override
+    def get_reserves_global_parameters(self, area_id: str) -> ReservesGlobalParameters:
+        study_id = self.get_study_id()
+        session = self.get_session()
+        stmt = select(_TABLE).where((_TABLE.c.study_id == study_id) & (_TABLE.c.area_id == area_id))
+        row = session.execute(stmt).fetchone()
+        if not row:
+            return ReservesGlobalParameters()
```
MartinBelthle: It's a bit of a new way of doing this here. Usually when the DB is empty we raise. Because if you're asking for a fake area we will return the default value. And as you're already calling this method inside `CreateArea`, there's no reason you'll enter this case except for version issue.

Also add a test for this case

antarest/study/dao/file/file_study_reserves_global_parameters_dao.py
```diff
@@ -0,0 +1,60 @@
...
+            data = {}
+        return parse_reserves_global_parameters(data)
+
+    @override
+    def get_all_reserves_global_parameters(self) -> ReservesGlobalParametersMapping:
+        file_study = self.get_file_study()
+        return {area_id: self.get_reserves_global_parameters(area_id) for area_id in file_study.config.areas}
+
+    @override
+    def save_reserves_global_parameters(self, mapping: ReservesGlobalParametersMapping) -> None:
```
MartinBelthle, line 52: If you're trying to save data for some fake area what happens ? The `file_study.tree.save` will fail I imagine so we're good. First we should ensure all area exist this way we could raise a cleaner exception and also handle a problem that's not handled here: if the 2nd area does not exist, you'll raise but you've already modified the data for area 1.

MartinBelthle (changes requested):

MartinBelthle (approved):

PR #3161, opened 2026-04-15
refactor(dao): parametrize tests to run on FS and DB dao

smailio:


tests/study/dao/test_database_district_dao.py
```diff
@@ -45,49 +43,41 @@ def test_save_district_with_areas(self, db_dao: DatabaseStudyDao) -> None:
...
 
-    def test_get_districts_returns_all(self, db_dao: DatabaseStudyDao) -> None:
-        dao = db_dao
+    def test_get_districts_returns_all(self, dao: StudyDao) -> None:
         dao.save_district(District(id="d1", name="District 1"))
         dao.save_district(District(id="d2", name="District 2"))
 
-        districts = dao.get_districts()
-        assert len(districts) == 2
+        districts = [d for d in dao.get_districts() if d.id != "all areas"]
```
smailio, line 67: fs dao fixture is initialized with all area, db dao fixture isn't.

tests/study/dao/test_database_renewable_dao.py
```diff
@@ -82,10 +91,13 @@ def test_save_multiple_renewable_clusters(db_dao: DatabaseStudyDao) -> None:
 
     paris_clusters = dao.get_all_renewables_for_area("paris")
     assert len(paris_clusters) == 2
-    assert [c.name for c in paris_clusters] == ["Battery", "Solar PaNELS"]
-    assert [c.nominal_capacity for c in paris_clusters] == [200.0, 500.0]
+    names = {c.name for c in paris_clusters}
+    assert names == {"Battery", "Solar PaNELS"}
+    caps = {c.name: c.nominal_capacity for c in paris_clusters}
+    assert caps["Battery"] == 200.0
+    assert caps["Solar PaNELS"] == 500.0
```
smailio, line 93: Make test resilient  if the FS and DB dao return identical result but with different order.

tests/study/dao/test_database_renewable_dao.py
```diff
@@ -96,10 +101,12 @@ def test_save_multiple_renewable_clusters(db_dao: DatabaseStudyDao) -> None:
     )
     paris_clusters = dao.get_all_renewables_for_area("paris")
     assert len(paris_clusters) == 3
-    assert [c.name for c in paris_clusters] == ["Battery", "Solar PaNELS", "Wind"]
-    assert [c.nominal_capacity for c in paris_clusters] == [1000.0, 500.0, 100.0]
+    caps2 = {c.name: c.nominal_capacity for c in paris_clusters}
+    assert caps2["Battery"] == 1000.0
+    assert caps2["Solar PaNELS"] == 500.0
+    assert caps2["Wind"] == 100.0
```
smailio, line 109: also about order

antarest/study/dao/file/file_study_area_properties_dao.py
```diff
@@ -144,7 +150,7 @@ def save_area_properties(self, area_id: str, area_properties: AreaProperties) ->
         )
         if file_study.config.version >= STUDY_VERSION_8_3:
             file_study.tree.save(
-                properties.adequacy_patch_properties.model_dump(mode="json", by_alias=True),
+                properties.adequacy_patch_properties.model_dump(mode="json", by_alias=True, exclude_none=True),
```
smailio, line 153: former code save None as a string like this 

Image

antarest/study/dao/database/database_thematic_trimming_dao.py
```diff
@@ -48,7 +51,9 @@ def get_thematic_trimming(self) -> ThematicTrimming:
         row = self._db_session.execute(stmt).fetchone()
         if not row:
             raise StudyNotFoundError(study_id)
-        return ThematicTrimming.model_validate(row.thematic_trimming)
+        trimming = ThematicTrimming()
+        initialize_thematic_trimming_against_version(trimming, self.get_impl().get_version())
+        return trimming.model_copy(update=row.thematic_trimming)
```
smailio: If we don't do this, version specific fields that weren't initialized during creation would return None instead of default value. 

In practice this never happen as we always suply all  fields, so this has no functional impact today. 

However keeping both DAOs consistent prevents confusing discrepancies if either implementation is extended /used in a new module in the future

antarest/study/dao/database/database_layer_dao.py
```diff
@@ -117,7 +117,6 @@ def save_layer(self, layer: Layer) -> None:
 
         values = {"study_id": study_id, "layer_id": layer.id, "name": layer.name}
         upsert_one(session, LAYER_TABLE, values)
-        self.get_impl().save_layer_areas(layer.id, layer.areas)
```
smailio, line 120: In FS DAO we don't save areas 

```python
    @override
    def save_layer(self, layer: Layer) -> None:
        file_study = self.get_file_study()

        file_study.tree.save(layer.name, ["layers", "layers", "layers", layer.id])

```

antarest/study/dao/database/database_study_factory_dao.py
```diff
@@ -57,9 +69,14 @@ def _create_default_settings(dao: DatabaseStudyDao, study: Study) -> None:
...
     initialize_advanced_parameters_against_version(advanced_parameters, study_version)
+    if study_version >= STUDY_VERSION_9_2:
+        advanced_parameters.shedding_policy = SheddingPolicy.ACCURATE_SHAVE_PEAKS
     dao.save_advanced_parameters(advanced_parameters)
 
-    dao.save_optimization_preferences(OptimizationPreferences())
+    op = OptimizationPreferences()
+    if study_version >= STUDY_VERSION_8_4:
+        op.transmission_capacities = TransmissionCapacities.LOCAL_VALUES
+    dao.save_optimization_preferences(op)
```
smailio: default values for these field was updated. Now the default values for a db study match the ones from template studies.

antarest/study/dao/database/database_study_settings_dao.py
```diff
@@ -221,4 +239,6 @@ def get_playlist_config(self) -> Playlist:
         row = self.get_session().execute(stmt).fetchone()
         if not row:
             raise StudyNotFoundError(study_id)
-        return Playlist(years=from_json(row.years))
+        sparse = Playlist(years=from_json(row.years))
+        nb_years = self.get_general_config().nb_years
+        return _expand_playlist(sparse, nb_years)
```
smailio, line 242: front end expect all years

antarest/study/dao/file/file_study_user_resources_dao.py
```diff
@@ -44,18 +44,19 @@ def get_all_user_resources(self) -> list[UserResourceDataCreation]:
...
-                        UserResourceDataCreation(
-                            path=PurePosixPath(file.relative_to(user_path).as_posix()),
-                            resource_type=ResourceType.FILE,
-                            blob_id=blob_id,
-                        )
+                        UserResourceDataCreation(path=rel_path, resource_type=ResourceType.FILE, blob_id=blob_id)
+                    )
+                elif item.is_dir():
+                    result.append(
+                        UserResourceDataCreation(path=rel_path, resource_type=ResourceType.FOLDER, blob_id=None)
```
smailio, line 59: FS dao didn't return folder. DB dao did. DB dao is right.

antarest/study/storage/rawstudy/model/filesystem/config/files.py
```diff
@@ -467,7 +467,7 @@ def _parse_st_storage(root: Path, area: str) -> list[STStorage]:
     config_list = []
     for section, values in config_dict.items():
         try:
-            config_list.append(parse_st_storage(version, values))
+            config_list.append(parse_st_storage(version, values, section))
```
smailio, line 481: this is to initalize st storage with right id.

antarest/study/storage/rawstudy/model/filesystem/config/st_storage.py
```diff
@@ -62,8 +62,12 @@ def from_model(cls, storage: STStorage) -> "STStorageFileData":
...
 
 
-def parse_st_storage(study_version: StudyVersion, data: Any) -> STStorage:
-    storage = STStorageFileData.model_validate(data).to_model()
+def parse_st_storage(study_version: StudyVersion, data: Any, storage_id: str | None = None) -> STStorage:
+    file_data = STStorageFileData.model_validate(data)
+    model_data = file_data.model_dump(exclude_none=True)
+    if storage_id is not None:
+        model_data["id"] = storage_id
+    storage = STStorage.model_validate(model_data)
```
smailio, line 71: When we save a storage with an id that's different from its name, the STStorage id is wrongly set to the name. It results in this weird behaviro 

```python
id, name = "my_id", "my_name"
save_st_storage(STStorage(id=id, name=name)
actual = get_st_storage("my_id")
assert actual.id == "my_name" 
```

tests/study/dao/test_database_st_storage_dao.py
```diff
@@ -177,57 +177,57 @@ def test_delete_st_storage(db_dao: DatabaseStudyDao) -> None:
...
     dao.save_st_storages({"area_1": [STStorage(id="st_storage_id_1", name="st-storage-1", efficiency=0.8)]})
 
     dao.save_st_storage_additional_constraints(
         {
             "area_1": {
                 "st_storage_id_1": [
-                    STStorageAdditionalConstraint(id="constraint_id_1", name="constraint-1"),
-                    STStorageAdditionalConstraint(id="constraint_id_2", name="constraint-2"),
+                    STStorageAdditionalConstraint(name="constraint-1"),
+                    STStorageAdditionalConstraint(name="constraint-2"),
```
smailio, line 188: FS dao don't store the id at all 

Here's what aditional-constraints.ini looks like at this point 

```ini
[constraint-1]
variable = netting
operator = less
hours = []
enabled = True

[constraint-2]
variable = netting
operator = less
hours = []
enabled = True

[constraint-3]
variable = netting
operator = less
hours = []
enabled = True
```

Look at this line ini_content[constraint.name] = serialize_st_storage_additional_constraint(constraint) in the save code : 

```python
    @override
    def save_st_storage_additional_constraints(
        self, data: dict[AreaId, dict[StStorageId, list[STStorageAdditionalConstraint]]]
    ) -> None:
        study_data = self.get_file_study()

        # Hold everything in memory to validate the data before saving it.
        url_content_pairs = []

        for area_id, value in data.items():
            for storage_id, constraints in value.items():
                if not self.st_storage_exists(area_id, storage_id):
                    raise STStorageNotFound(area_id, storage_id)
                existing_constraints = self.get_st_storage_additional_constraints(area_id, storage_id)

                existing_map = {c.id: c for c in existing_constraints}
                existing_map.update({c.id: c for c in constraints})

                ini_content = {}
                for constraint_id, constraint in existing_map.items():
                    ini_content[constraint.name] = serialize_st_storage_additional_constraint(constraint) # <==== HERE we save under constraint name

                # Save into the config
                self._update_st_storage_additional_constraints_config(area_id, storage_id, constraints)

                # We have to create the folder before saving the files
                (study_data.config.study_path / "input" / "st-storage" / "constraints" / area_id / storage_id).mkdir(
                    parents=True, exist_ok=True
                )
                url = ["input", "st-storage", "constraints", area_id, storage_id, "additional_constraints"]
                url_content_pairs.append((url, ini_content))

        # Save into the files
        for url, content in url_content_pairs:
            study_data.tree.save(content, url)

```

Maybe the db dao should use the name as an id as well ?

antarest/study/dao/file/file_study_st_storage_dao.py
```diff
@@ -343,7 +349,8 @@ def delete_st_storage(self, area_id: str, storage: STStorage) -> None:
         self._remove_st_storage_from_scenario_builder(area_id, storage_id)
 
         # Deleting the short-term storage in the configuration must be done AFTER deleting the files and folders.
-        study_data.config.areas[area_id].st_storages.remove(storage)
+        st_storages = study_data.config.areas[area_id].st_storages
+        st_storages[:] = [s for s in st_storages if s.id != storage_id]
```
smailio, line 351: id comparaison safer than content comparaison 

This raises with former code 

```python
    dao.save_st_storages({"area_1": [STStorage(id="st_storage_id_1", name="st-storage-1", efficiency=0.8)]})
    dao.delete_st_storage("area_1", STStorage(id="st_storage_id_1", name="st-storage-1"))
```

antarest/study/dao/file/file_study_area_dao.py
```diff
@@ -123,10 +123,13 @@ def get_all_areas_ui_info(self) -> dict[str, AreaUIData]:
...
-            area_id: AreaUIData.model_validate(AreaUIFileData(**ui_info).to_config())
-            for area_id, ui_info in ui_info_map.items()
-        }
+        result: dict[str, AreaUIData] = {}
+        for area_id, ui_info in ui_info_map.items():
+            ui_data = AreaUIData.model_validate(AreaUIFileData(**ui_info).to_config())
+            if ui_data.layer_x:
+                ui_data.ui["layers"] = " ".join(sorted(ui_data.layer_x, key=int))
+            result[area_id] = ui_data
+        return result
```
smailio: database dao does the same. 

This is only usefull when we save layer ui info for a layer that doesn't exist. 

This never happen in practice the command performs a validation that prevents this. 

However I think it's still worth it to align the FS dao with the DB dao bevause if one day we'll use it from a new module or if we switch from the command system to another system then this dao behavior would be very confusing.

tests/study/dao/test_database_user_resources_dao.py
```diff
@@ -51,23 +48,44 @@ def test_save_user_resources_folder(db_dao: DatabaseStudyDao) -> None:
...
+        ]
+    )
+    result = sorted(dao.get_all_user_resources(), key=lambda r: str(r.path))
+    assert len(result) == 2
+    assert result[0].path == PurePosixPath("folder_path")
+    assert result[0].resource_type == ResourceType.FOLDER
+    assert result[0].blob_id is None
+    assert result[1].path == PurePosixPath("folder_path/inside.txt")
+    assert result[1].resource_type == ResourceType.FILE
+    assert result[1].blob_id == blob_id
```
smailio, line 91: now fs dao also return folders, so I added this test

tests/integration/study_data_blueprint/test_st_storage.py
```diff
@@ -405,8 +405,8 @@ def test_lifecycle__nominal(
         )
         assert res.status_code == 404
         obj = res.json()
-        assert obj["description"] == f"'{bad_area_id}' not a child of InputSTStorageClusters"
-        assert obj["exception"] == "ChildNotFoundError"
+        assert obj["description"] == f"Area is not found: '{bad_area_id}'"
```
sylvlecl, line 405: nice

antarest/study/storage/rawstudy/model/filesystem/config/st_storage.py
sylvlecl: This is a very important point, but I am not sure what to do with this:
the distinction between id and name is really broken from the start in file studies.

antares file format does not have that distinction between ID and name.
It has 2 places: the section name and the "name" property inside the section, but it only reads one of them, the other one is never read.
Then, internally, it derives an "id" from that name.
I say it picks one of the 2, but the worst part is that it's not consistent between object types (clusters, STS, ...): sometimes it picks the property, sometimes the section name.

The "ID" is important in the format, because it will appear as an identifier in other filenames.

So we need to be careful to not write a "name" inconsistent with the "id".
For ex. if we write a folder named "my_id" to reference that STS, and "my_name" in the name property, and antares picks "my_name" as the base for the ID, the study will be broken.


--> we need to decide on a consistent strategy accross all DAOs and models here

Needs to be carefully thought.
Options include:

- always enforce consistency between "id" and "name".
  One could be implemented as a calculated property instead of 2 distinct properties.
  Big drawback for the database DAO: we won't be able to rename easily, just like in the old file format.

- enforce consistency only for the FS DAO.
  In that case, we need to be careful about what we do when converting from DB study to an FS study.

- Try to use the "not used" data from the file format as the place for our "name".
  It's dangerous, since antares behaviour is not consistent about this.

- Introduce a 3rd field to distinguish between a true independent ID and the 2 fields that need to stay consistent. But that will not solve the issue of the file format, which cannot hold  an inconsistent "name" anyway

- anything else ?

antarest/study/dao/file/file_study_st_storage_dao.py
sylvlecl: Nice: but why does the delete accepts a full storage object in the first place ?

Should accept only an ID in my opinion

antarest/study/dao/database/database_study_factory_dao.py
sylvlecl: Nice: what about a PR just with that change ?

antarest/study/dao/file/file_study_area_properties_dao.py
sylvlecl: Well spotted, but seems to me that a None should not be considered a valid value for studies > 8.3 
The "initialization" is supposed to fill it

sylvlecl (commented):
Nice, a few comments as a start.
I think we could isolate the factory part in another PR to merge it separately.

There is an important and hard topic about IDs and names about which we need to carefully take a consistent decision

antarest/study/storage/rawstudy/model/filesystem/config/model.py
```diff
@@ -69,8 +69,12 @@ class LinkConfig(AntaresBaseModel, extra="ignore"):
...
     def validation(cls, values: MutableMapping[str, Any]) -> MutableMapping[str, Any]:
         # note: field names are in kebab-case in the INI file
-        filters_synthesis = values.pop("filter-synthesis", values.pop("filters_synthesis", ""))
-        filters_year = values.pop("filter-year-by-year", values.pop("filters_year", ""))
+        filters_synthesis = values.pop(
+            "filter-synthesis", values.pop("filter_synthesis", values.pop("filters_synthesis", ""))
+        )
+        filters_year = values.pop(
+            "filter-year-by-year", values.pop("filter_year_by_year", values.pop("filters_year", ""))
+        )
```
smailio: note sure about this cleanliness wise 

But due to a typo in Link model we need to support filter_synthesis argument, otherwise we won't update the config object that belongs to a file study which lead to returning a stale study synthesis.

tests/study/dao/test_database_area_dao.py
```diff
@@ -15,96 +15,73 @@
...
+def test_get_all_areas_info_returns_areas(dao: StudyDao) -> None:
     """
     Test that get_all_areas_info returns all areas for a study with original names preserved.
     """
-    dao = db_dao
     save_area(dao, "Paris")
     save_area(dao, "London")
     save_area(dao, "Berlin")
-    dao.save_thermals({"paris": [ThermalCluster(name="Gas")], "berlin": [ThermalCluster(name="Coal")]})
+    dao.save_thermals({"paris": [ThermalCluster(name="gas")], "berlin": [ThermalCluster(name="coal")]})
```
smailio: db alway lower the id, fs doesn't for retrocompatibilty

tests/study/dao/test_database_area_dao.py
```diff
@@ -211,24 +183,26 @@ def test_save_area_ui_raises_error_if_area_not_exists(db_dao: DatabaseStudyDao)
 def test_save_area_ui_raises_error_if_layer_not_exists(db_dao: DatabaseStudyDao) -> None:
     """
     Test that save_area_ui raises LayerNotFound if layer doesn't exist.
+    DB-only: FS writes layer that don't exist and changing that woule be a breaking change.
```
smailio: I mean in in ui.ini it will write 
```
[ui]
...
layers = 0 999
```

even if layer 999 doesn't exist

tests/study/dao/test_database_area_dao.py
```diff
@@ -264,58 +238,34 @@ def test_get_all_areas_ui_info_returns_all_layers(db_dao: DatabaseStudyDao) -> N
...
     Test that save_layer_areas adds and removes areas from a layer.
     """
-    dao = db_dao
-    save_area(dao, "Paris")
-    save_area(dao, "London")
-    save_area(dao, "Berlin")
+    create_area("Paris", dao)
+    create_area("London", dao)
+    create_area("Berlin", dao)
     dao.save_layer(Layer(id="1", name="Layer 1"))
```
smailio, line 250: crate_area does some initialization regarding the ui, if we don't do it this test fails for FS

tests/study/dao/test_database_layer_dao.py
```diff
@@ -63,21 +55,7 @@ def test_save_layer_creates_layer(self, db_dao: DatabaseStudyDao) -> None:
...
-        create_area("Paris", dao)
-        create_area("London", dao)
-
-        layer = Layer(id="1", name="Layer 1", areas=["paris", "london"])
-        dao.save_layer(layer)
-
-        layers = dao.get_layers()
-        layer_1 = next((layer for layer in layers if layer.id == "1"), None)
-        assert layer_1 is not None
-        assert set(layer_1.areas) == {"paris", "london"}
```
smailio, line 77: db dao don't do this anymore, it doesn't save layer_area during the save_layer. This was incosistent with db behavior.

antarest/study/dao/file/file_study_st_storage_dao.py
```diff
@@ -498,6 +508,11 @@ def _update_st_storage_config(self, area_id: str, storage: STStorage) -> None:
         if area_id not in study_data.areas:
             raise ValueError(f"The area '{area_id}' does not exist")
 
+        # Mirror the DB read path which initializes version-specific defaults
+        # (enabled, ...) so consumers reading from `config` see the same
+        # fully-populated storage as the DB DAO returns.
+        initialize_st_storage(storage, study_data.version)
```
smailio: In practice we don't need it as the command/management initialize this already, this is only for the sake of consistency. Let me know if you think I should remove this and updates the test instead.

tests/study/dao/test_database_synthesis_dao.py
```diff
@@ -15,34 +15,52 @@
...
+# DAO so synthesis assertions are backend-agnostic.
+DEFAULT_DISTRICT = District(
+    id="all areas",
+    name="All areas",
+    output=False,
+    comments="Spatial aggregates on all areas",
+    apply_filter=DistrictApplyFilter.add_all,
+)
+
+
```
smailio, line 40: fs study is always initialized with "all areas" district, we need to do that too on db study to be able to run same test on both fs and db dao.

antarest/study/dao/database/database_area_properties_dao.py
```diff
@@ -49,6 +50,8 @@ def _convert_db_properties_to_model(db_row: Any) -> AreaProperties:
         filter_by_year=parse_frequency_filters(db_row.filter_by_year),
         adequacy_patch_mode=db_row.adequacy_patch_mode,
     )
+    initialize_area_properties(props, study_version)
+    return props
```
smailio: FS does this initialization at each read

antarest/study/dao/file/file_study_thermal_dao.py
```diff
@@ -265,6 +275,11 @@ def _get_all_thermals_for_area(file_study: FileStudy, area_id: str) -> dict[str,
     def _update_thermal_config(study_data: FileStudyTreeConfig, area_id: str, thermal: ThermalCluster) -> None:
         check_area_exists(study_data, area_id)
 
+        # Mirror the DB read path which initializes version-specific defaults
+        # (nh3, so2, cost_generation, ...) so consumers reading from `config`
+        # see the same fully-populated cluster as the DB DAO returns.
+        initialize_thermal_cluster(thermal, study_data.version)
```
smailio, line 280: same than above

antarest/study/dao/database/database_study_factory_dao.py
```diff
@@ -100,6 +119,11 @@ def session(self) -> Session:
 
     @override
     def create_study_dao(self, study: Study) -> DatabaseStudyDao:
+        version = StudyVersion.parse(study.version)
+        if version not in STUDY_REFERENCE_TEMPLATES:
+            raise UnsupportedStudyVersion(
+                f"{version} is not a supported version, supported versions are: {STUDY_REFERENCE_TEMPLATES}"
+            )
         dao = DatabaseStudyDao(study.id, self.session, self._matrix_service, self._generator_matrix_constants)
```
smailio: FS dao does this check, I aligned DB DAO.

antarest/study/dao/file/file_study_user_resources_dao.py
```diff
@@ -44,18 +44,19 @@ def get_all_user_resources(self) -> list[UserResourceDataCreation]:
...
+                    blob_id = blob_service.save(item.read_bytes())
                     result.append(
-                        UserResourceDataCreation(
-                            path=PurePosixPath(file.relative_to(user_path).as_posix()),
-                            resource_type=ResourceType.FILE,
-                            blob_id=blob_id,
-                        )
+                        UserResourceDataCreation(path=rel_path, resource_type=ResourceType.FILE, blob_id=blob_id)
+                    )
+                elif item.is_dir():
```
MartinBelthle: As discussed, either we find a way to only return files and empty folders instead of rglob (not sure it exists), or we add a `and not next(item.listdir())`

tests/study/dao/test_database_scenario_builder_dao.py
```diff
@@ -11,31 +11,49 @@
...
 from tests.study.dao.utils import save_area
 
 
-def _setup_areas(dao: DatabaseStudyDao, *area_names: str) -> None:
+def _prune_empty(result: dict[str, Any]) -> dict[str, Any]:
+    """
+    Recursively remove keys whose value is an empty dict.
+
+    Normalizes FS vs DB scenario-builder output: the FS backend leaves empty
+    area dicts after all clusters are deleted, while the DB backend prunes
```
MartinBelthle: As discussed, we should have the same behvior for FS and DB meaning we should initialize missing data for the `get_scenario_by_type` method in the DB DAO

tests/study/dao/test_database_settings_dao.py
```diff
@@ -116,7 +116,16 @@ def test_nominal_case(db_dao: DatabaseStudyDao) -> None:
...
     playlist = Playlist(years={4: PlaylistValues(status=False, weight=0.2), 7: PlaylistValues(weight=1.1)})
     dao.save_playlist_config(playlist)
-    assert dao.get_playlist_config() == playlist
+    expected_playlist = Playlist(
+        years={
+            1: PlaylistValues(status=False),
+            2: PlaylistValues(status=False),
+            3: PlaylistValues(status=False),
+            4: PlaylistValues(status=False, weight=0.2),
+            7: PlaylistValues(weight=1.1),
```
MartinBelthle: Makes me realize that it does not make sense to set a Playlist for year 7 when the study has only 4 years. We should probably validate this but won't be done inside this PR

tests/study/dao/test_database_settings_dao.py
```diff
@@ -150,13 +159,13 @@ def test_get_comments_returns_empty_string_by_default(db_dao: DatabaseStudyDao)
...
     comments = "test comment study"
 
     dao.save_comments(comments)
 
     assert dao.get_comments() == comments
 
-    stmt = select(COMMENTS_TABLE.c.comments).where(COMMENTS_TABLE.c.study_id == dao.get_study_id())
-    assert db_session.execute(stmt).scalar_one() == comments
+    if isinstance(dao, DatabaseStudyDao):
+        stmt = select(COMMENTS_TABLE.c.comments).where(COMMENTS_TABLE.c.study_id == dao.get_study_id())
```
MartinBelthle: Here you keep the part of the test where we check the DB content btut everywhere else seems you removed it. I think we can remove it from here too

tests/study/dao/test_database_st_storage_dao.py
```diff
@@ -313,18 +313,19 @@ def test_st_storage_matrices_lifecycle(db_session: Session, db_dao: DatabaseStud
...
-        assert db_session.execute(select(LOWER_RULE_CURVE_TABLE)).fetchall() == []
-        assert db_session.execute(select(UPPER_RULE_CURVE_TABLE)).fetchall() == []
-        assert db_session.execute(select(INFLOWS_TABLE)).fetchall() == []
-        assert db_session.execute(select(COST_INJECTION_TABLE)).fetchall() == []
-        assert db_session.execute(select(COST_WITHDRAWAL_TABLE)).fetchall() == []
-        assert db_session.execute(select(COST_LEVEL_TABLE)).fetchall() == []
-        assert db_session.execute(select(COST_VARIATION_INJECTION_TABLE)).fetchall() == []
-        assert db_session.execute(select(COST_VARIATION_WITHDRAWAL_TABLE)).fetchall() == []
+    if isinstance(dao, DatabaseStudyDao):
+        with db_session:
```
MartinBelthle: Same remark, I'm fine with keeping this but you removed such thigns from other tests so we should do the same thing everywhere

tests/study/dao/test_database_synthesis_dao.py
MartinBelthle: In fact to me this should be done inside the DB DAO factory at study creation. This way we'll be even

tests/study/dao/test_database_thermal_dao.py
```diff
@@ -246,7 +248,53 @@ def test_thermal_matrices_lifecycle(db_session: Session, db_dao: DatabaseStudyDa
...
+        parameters=ThermalClusterCreation(name="MyThermal"),
+        command_context=command_context,
+        study_version=dao.get_version(),
+    )
+    output = command.apply(dao)
+    assert output.status  # The command should succeed even if the name is in upper case
+
+
+# ──────────────────────────────────────────────────────────────────────
+# Database-only tests
```
MartinBelthle, line 291: This is probably the proper way to do it IMO

tests/study/dao/test_database_thermal_dao.py
```diff
@@ -277,46 +325,19 @@ def test_get_thermal_matrix_raises_when_missing(db_dao: DatabaseStudyDao) -> Non
...
     ]
 
     for saver in savers:
         with pytest.raises(ThermalClusterNotFound):
             saver({"paris": {"gas": "missing-matrix-id"}})
         with pytest.raises(AreaNotFound):
             saver({"nonexistent": {"gas": "missing-matrix-id"}})
-
-
-def test_area_with_no_clusters_are_absent_from_clusters_dict(db_dao: DatabaseStudyDao) -> None:
```
MartinBelthle, line 298: What about these 2 tests ?

tests/study/dao/test_database_xpansion_dao.py
```diff
@@ -182,126 +184,135 @@ def test_save_xpansion_settings_persists_non_default_values(self, db_dao: Databa
...
-        result = db_dao.get_xpansion_settings()
+        dao.save_xpansion_settings(XpansionSettings(optimality_gap=10.0))
+        result = dao.get_xpansion_settings()
         assert result.optimality_gap == 10.0
 
-        # --- invalid projection ---
-        settings = XpansionSettings(sensitivity_config=XpansionSensitivitySettings(projection=["cand_a", "ghost_cand"]))
-        with pytest.raises(CandidateNotFoundError):
-            db_dao.save_xpansion_settings(settings)
+        # --- invalid projection: DB validates candidates exist; FS does not (gap: #TODO) ---
```
MartinBelthle: Nice to spot this. As the rest we should instead harmonize the FS code as the DB is right

tests/study/dao/test_database_xpansion_dao.py
```diff
@@ -182,126 +184,135 @@ def test_save_xpansion_settings_persists_non_default_values(self, db_dao: Databa
...
-
-        # --- rename to existing target overwrites it; no duplicate in projection ---
-        db_dao.save_xpansion_candidate(_make_candidate("bob", "bordeaux", "paris", cost=999.0), old_id="charlie")
-        assert len(db_dao.get_all_xpansion_candidates()) == 1
-        assert db_dao.get_xpansion_candidate("bob").annual_cost_per_mw == 999.0
-        proj = db_dao.get_xpansion_settings().sensitivity_config.projection
-        assert proj == ["bob"]
+            dao.get_xpansion_candidate("alice")
+        assert dao.get_xpansion_candidate("charlie").annual_cost_per_mw == 500.0
+        # DB carries the rename over to the projection; FS does not (gap: #TODO)
```
MartinBelthle: Not sure to fully understand here but as always the behaviors should be the same

tests/study/dao/conftest.py
MartinBelthle: Now that this is used inside the test_garbage_collector file, it shouldn't be private. Either make it public or use it through a fixture

tests/study/dao/test_garbage_collector.py
```diff
@@ -13,26 +13,69 @@
...
+    backend: str,
+    matrix_service: MatrixService,
+    db_session: Session,
+    blob_service: IBlobService,
+    core_cache: ICache,
+    tmp_path: Path,
+) -> StudyDao:
+    if backend == "database":
+        dao = build_db_dao(db_session, matrix_service, STUDY_VERSION_9_3)
+        ConstantsMatrixUsageProvider(dao.generator_matrix_constants, matrix_service)
```
MartinBelthle: This should already be done inside the build_db_dao

tests/study/dao/test_garbage_collector.py
```diff
@@ -236,14 +276,21 @@ def test_garbage_collection(db_dao: DatabaseStudyDao, db_session: Session, tmp_p
     pl.testing.assert_frame_equal(bc_eq, bc_eq_df, check_dtypes=False)
 
 
-def test_provider_includes_reserve_need_matrix(db_session: Session, matrix_service: ISimpleMatrixService) -> None:
-    dao = build_db_dao(db_session, matrix_service, STUDY_VERSION_10_0)
+def test_provider_includes_reserve_need_matrix(dao_10_0: StudyDao, core_cache: ICache) -> None:
```
MartinBelthle: You can add a TODO inside this test saying that it'll be removed when we'll fully support the v10.0 as it will be integrated inside the `test_garbage_collection` one

tests/study/dao/test_database_area_properties_dao.py
```diff
@@ -60,17 +47,17 @@ def test_multiple_areas(db_session: Session, db_dao: DatabaseStudyDao) -> None:
...
+    default_props = _default_props(dao)
+
     # Ensures we do not perform N+1 requests
     with DBStatementRecorder(db_session.bind) as db_recorder:
         all_properties = dao.get_all_area_properties()
-        default_props = AreaProperties()
         assert all_properties == {"paris": default_props, "london": default_props}
 
-    assert len(db_recorder.sql_statements) == 1, str(db_recorder)
+    assert len(db_recorder.sql_statements) == 2, str(db_recorder)
```
MartinBelthle: Specify that the 2nd query is to get the study version

tests/study/dao/test_database_area_properties_dao.py
```diff
@@ -80,15 +67,16 @@ def test_error_cases(db_dao: DatabaseStudyDao) -> None:
...
 
-    assert dao.get_area_properties(area_id) == AreaProperties()
+    assert dao.get_area_properties(area_id) == _default_props(dao)
 
     new_properties = AreaProperties(energy_cost_unsupplied=4.5, non_dispatch_power=False, filter_synthesis={"daily"})
     dao.save_area_properties(area_id, new_properties)
 
     # Ensures we modified the properties accordingly
-    assert dao.get_area_properties(area_id) == new_properties
+    expected = new_properties.model_copy()
```
MartinBelthle: Not sure to understand why you need to copy. Couldn't you remove this line ?

tests/study/dao/test_database_initialize_study.py
```diff
@@ -9,58 +9,96 @@
...
     initialize_thematic_trimming_against_version(expected_thematic_trimming, study_version)
 
-    assert dao.get_general_config() == expected_general_config
-    assert dao.get_optimization_preferences() == OptimizationPreferences()
-    assert dao.get_advanced_parameters() == expected_advanced_parameters
-    assert dao.get_playlist_config() == Playlist()
-    assert dao.get_timeseries_config() == TimeSeriesConfiguration()
-    assert dao.get_thematic_trimming() == expected_thematic_trimming
+    actual_general = dao.get_general_config()
+    assert actual_general == expected_general_config, explain_model_diff(actual_general, expected_general_config)
```
MartinBelthle: Minor: I like `explain_model_diff` but to me it should be used ponctually and not in the tests hard-coded, it makes the test harder to read IMO

MartinBelthle (changes requested):
I read all the tests but not the source code. Really nice work

antarest/study/dao/database/database_scenario_builder_dao.py
```diff
@@ -221,7 +229,12 @@ def get_scenario_by_type(self, scenario_type: ScenarioType) -> AnyScenarios:
             stmt = select(SCENARIO_STORAGE_CONSTRAINTS_TABLE).where(
                 SCENARIO_STORAGE_CONSTRAINTS_TABLE.c.study_id == study_id
             )
-            constraints_result: dict[str, Any] = {}
+            constraints_result: dict[str, Any] = {
+                area_id: {storage_id: {} for storage_id in storages}
+                for area_id, storages in self.get_impl().get_all_st_storages().items()
+            }
+            for area_id in self.get_impl().get_all_area_ids():
+                constraints_result.setdefault(area_id, {})
```
smailio: align DB with FS. Entires must be set for each constraint.

tests/study/dao/test_database_thermal_dao.py
smailio: they're above

antarest/study/dao/file/file_study_area_dao.py
```diff
@@ -186,25 +187,32 @@ def get_invalid_area_ids(self, areas: list[str]) -> list[str]:
...
+        try:
+            return self.get_impl().get_matrix(path_getter(area_id))
+        except ChildNotFoundError:
+            check_area_exists(self.get_file_study().config, area_id)
+            raise
+
     @override
     def get_load(self, area_id: str) -> pl.DataFrame:
-        return self.get_impl().get_matrix(_get_load_matrix_path(area_id))
+        return self._get_area_matrix(area_id, _get_load_matrix_path)
```
smailio, line 193: I just make FS dao raise same error than DB dao

antarest/study/dao/file/file_study_xpansion_dao.py
```diff
@@ -110,6 +114,38 @@ def _save_xpansion_candidates(self, candidates: list[tuple[XpansionCandidate, st
 
         self._save_candidates(existing_candidates)
 
+        if renames:
+            self._apply_projection_renames(renames)
```
smailio, line 119: aligh FS with DB. Updates should cascade.

antarest/study/dao/file/file_study_xpansion_dao.py
```diff
@@ -135,6 +176,13 @@ def get_xpansion_settings(self) -> XpansionSettings:
     def save_xpansion_settings(self, settings: XpansionSettings) -> None:
         file_study = self.get_file_study()
 
+        projection = settings.sensitivity_config.projection if settings.sensitivity_config else []
+        if projection:
+            existing_names = {c["name"] for c in self._get_all_xpansion_candidates().values()}
+            missing = [name for name in projection if name not in existing_names]
+            if missing:
+                raise CandidateNotFoundError("One or more candidates in the projection do not exist")
```
smailio, line 185: DB dao enforce it so FS should too. Plus this would make the DB dao look slower when it actually just does more work.

alembic/versions/4121d3b48393_change_reserves_format.py
```diff
@@ -5,13 +5,14 @@
...
+
 import sqlalchemy as sa
 
+from alembic import op
 
 # revision identifiers, used by Alembic.
-revision = '4121d3b48393'
-down_revision = 'rp986cf862cy'
+revision = "4121d3b48393"
+down_revision = "rp986cf862cy"
```
smailio, line 15: when i merged dev my ruff wasn't happy with single quotes

antarest/study/dao/database/database_study_factory_dao.py
```diff
@@ -56,9 +68,14 @@ def _create_default_settings(dao: DatabaseStudyDao, study_version: StudyVersion)
...
     advanced_parameters = AdvancedParameters()
     initialize_advanced_parameters_against_version(advanced_parameters, study_version)
+    if study_version >= STUDY_VERSION_9_2:
+        advanced_parameters.shedding_policy = SheddingPolicy.ACCURATE_SHAVE_PEAKS
     dao.save_advanced_parameters(advanced_parameters)
 
     optimization_preferences = OptimizationPreferences()
+    if study_version >= STUDY_VERSION_8_4:
+        # FS gets this via the empty_study template's generaldata.ini; DB has no template.
+        optimization_preferences.transmission_capacities = TransmissionCapacities.LOCAL_VALUES
```
smailio, line 79: I can't put this in `initialize_optimization_preferences_against_version` because  `initialize_optimization_preferences_against_version` is called each time we parse a file study, so on each read, and it would overwrite the default value (which is True and not None, so I can't just check is None, and I can't indentify an instance where transmission_capacities was set to True vs an instance where transmission_capacities was never updated).

antarest/study/dao/database/database_area_properties_dao.py
sylvlecl: In reality, this should never be useful for database implementation: the "initialization" goal is to replace missing data with their default values, but in database impl we should never have a missing data: the default value data should always be inserted from the start.

So I would be more in favor of removing this. If it's needed, it means there's a problem upstream.

antarest/study/dao/database/database_layer_dao.py
sylvlecl: Here we need to fix the DAO signature to make things clear.
If this method is not supposed to update areas, it must have as an argument a class which does not hold the list of areas.
We should have a simpler class, like `LayerMetadata` or something like that.

Else we should refactor the callers so that they indeed use only one method to save both metadata and areas, but that would probably require more work

antarest/study/dao/database/database_scenario_builder_dao.py
```diff
@@ -212,7 +220,7 @@ def get_scenario_by_type(self, scenario_type: ScenarioType) -> AnyScenarios:
         if scenario_type in _AREA_ITEM_TABLE_MAP:
             table, id_col = _AREA_ITEM_TABLE_MAP[scenario_type]
             stmt = select(table).where(table.c.study_id == study_id)
-            result: dict[str, Any] = {}
+            result: dict[str, Any] = {area_id: {} for area_id in self.get_impl().get_all_area_ids()}
```
MartinBelthle: Why are you initializing for 2 `if` but not for the others ? Seems you should do it for all cases

antarest/study/dao/database/database_study_settings_dao.py
```diff
@@ -50,6 +50,24 @@
...
+    Expand sparse playlist to cover years 1..nb_years, mimicking filesystem DAO semantic.
+    Missing year default status follows majority rule of saved entries (matches fs `playlist_reset`).
+    Empty playlist defaults to status=True (matches fs default when no `[playlist]` INI section).
+    """
+    if not playlist.years:
+        default_status = True
+    else:
+        activated = sum(1 for v in playlist.years.values() if v.status)
+        default_status = activated > len(playlist.years) // 2
+    expanded = dict(playlist.years)
```
MartinBelthle: Minor but why are you using `dict(...)`, playlist.years is already a dict

antarest/study/dao/database/database_study_settings_dao.py
```diff
@@ -50,6 +50,24 @@
...
+    """
+    Expand sparse playlist to cover years 1..nb_years, mimicking filesystem DAO semantic.
+    Missing year default status follows majority rule of saved entries (matches fs `playlist_reset`).
+    Empty playlist defaults to status=True (matches fs default when no `[playlist]` INI section).
+    """
+    if not playlist.years:
+        default_status = True
+    else:
+        activated = sum(1 for v in playlist.years.values() if v.status)
+        default_status = activated > len(playlist.years) // 2
```
MartinBelthle: Discussed the matter with Sylvain and we think you should always consider the `default_status` is True. It will be different from the FS but we'd actually want the FS to be like the DB for this case as it's less misleading.

You don't have to modify the FS Dao in this PR just consider the `default_status` to be True and we'll be good

antarest/study/dao/database/database_study_factory_dao.py
sylvlecl: Yes, as discussed it's the right place :)
It's the equivalent of having different values between version templates, on the FS side.

antarest/study/dao/database/database_thematic_trimming_dao.py
sylvlecl: Indeed they will be supplied on creation anyway, I am not sure we should keep this code:
it does not incentivize us to have a clean creation in the first place, and it blurs the logic of this DAO where we want complete data in the first place.

antarest/study/dao/file/file_study_reserves_global_parameters_dao.py
```diff
@@ -38,16 +38,28 @@ def get_file_study(self) -> FileStudy:
     @override
     def get_reserves_global_parameters(self, area_id: str) -> ReservesGlobalParameters:
         file_study = self.get_file_study()
+        check_area_exists(file_study.config, area_id)
```
MartinBelthle, line 41: Makes me realize that you could use this in other DAOs you've modified to simplify a bit the reading

antarest/study/dao/file/file_study_reserves_global_parameters_dao.py
```diff
@@ -38,16 +38,28 @@ def get_file_study(self) -> FileStudy:
...
         except (ChildNotFoundError, KeyError):
             data = {}
+        if GLOBAL_PARAMETERS_SECTION not in data:
+            raise ValueError(f"Reserves global parameters not found for area '{area_id}'")
         return parse_reserves_global_parameters(data)
 
     @override
     def get_all_reserves_global_parameters(self) -> ReservesGlobalParametersMapping:
         file_study = self.get_file_study()
-        return {area_id: self.get_reserves_global_parameters(area_id) for area_id in file_study.config.areas}
```
MartinBelthle, line 50: Not sure to understand why you've changed this code. Was easier to read

antarest/study/dao/file/file_study_st_storage_dao.py
```diff
@@ -440,6 +448,8 @@ def save_st_storage_additional_constraints(
 
         for area_id, value in data.items():
             for storage_id, constraints in value.items():
+                if not self.st_storage_exists(area_id, storage_id):
```
MartinBelthle: This is not performant at all. For the rest of the modification you've done I'm okay with them because they either use the config or they are in the deletion methods which are almost never used. But here I think we should be careful with performances. 

We should also use the config here or at least check the file only once per area to avoid several I/O operations on the same file (`.config` is faster we should use it and even do it once per area to avoid going through the sts list several times)

antarest/study/dao/file/file_study_st_storage_dao.py
MartinBelthle: I think we don't need this, but I'll let Sylvain give his opinion

antarest/study/storage/rawstudy/model/filesystem/config/model.py
MartinBelthle: Not sure to understand why you changed this code

antarest/study/dao/file/file_study_area_dao.py
```diff
@@ -123,10 +123,13 @@ def get_all_areas_ui_info(self) -> dict[str, AreaUIData]:
...
         # Convert to AreaUIFileData to ensure that the UI object is valid, then to Pydantic model
-        return {
-            area_id: AreaUIData.model_validate(AreaUIFileData(**ui_info).to_config())
-            for area_id, ui_info in ui_info_map.items()
-        }
+        result: dict[str, AreaUIData] = {}
+        for area_id, ui_info in ui_info_map.items():
+            ui_data = AreaUIData.model_validate(AreaUIFileData(**ui_info).to_config())
+            if ui_data.layer_x:
+                ui_data.ui["layers"] = " ".join(sorted(ui_data.layer_x, key=int))
```
MartinBelthle: This does not seem to be what we're doing in DB, I'm confused and the Area ui code sucks so much.

In DB there's no notion of `if layer_x`

To me we shouldn't modify this code to make it even harder to understand for a case that never happens.

MartinBelthle (changes requested):

sylvlecl (commented):
A few comments, could't review everything yet

antarest/study/dao/file/file_study_area_dao.py
sylvlecl: Hmmm, I mostly agree. Indeed we ensure stronger consistency in the DB implementation, with foreign keys, cascade deletes and so on, which is a good thing: we can never have data for a layer that does not exist.
But here, if files contain inconsistencies, I don't think we should try to hide them on the fly.

sylvlecl: But let's not lose time on this, it's not important.
If we want to make this right, we'll need to dig a little more: for example it may make sense to have a layer absent from the list, but still have coordinates, because maybe the user wanted to mask it temporarily without losing the coordinates.
To be honest we don't care about that kid of details for now, it has not been well thought in the past, and we'll not do it now

antarest/study/dao/file/file_study_layer_dao.py
```diff
@@ -59,10 +60,16 @@ def save_layer(self, layer: Layer) -> None:
 
     @override
     def delete_layer(self, layer_id: str) -> None:
+        if layer_id == DEFAULT_LAYER_ID:
+            raise LayerNotAllowedToBeDeleted()
```
sylvlecl, line 64: nice

antarest/study/dao/file/file_study_renewable_dao.py
```diff
@@ -101,7 +110,12 @@ def renewable_exists(self, area_id: str, renewable_id: str) -> bool:
 
     @override
     def get_renewable_series(self, area_id: str, renewable_id: str) -> pl.DataFrame:
-        return self.get_impl().get_matrix(_get_renewable_series_path(area_id, renewable_id))
+        if area_id not in self.get_file_study().config.areas:
+            raise AreaNotFound(area_id)
```
sylvlecl: Nice: seems like it's a very common check: maybe we should have a small function "check_area_exists" ?

sylvlecl: Ah indeed there's already a comment about this

antarest/study/dao/file/file_study_reserves_global_parameters_dao.py
```diff
@@ -38,16 +38,28 @@ def get_file_study(self) -> FileStudy:
     @override
     def get_reserves_global_parameters(self, area_id: str) -> ReservesGlobalParameters:
         file_study = self.get_file_study()
+        check_area_exists(file_study.config, area_id)
         try:
             data = file_study.tree.get(_get_reserves_ini_path(area_id))
         except (ChildNotFoundError, KeyError):
             data = {}
+        if GLOBAL_PARAMETERS_SECTION not in data:
```
sylvlecl: Don't we want to use the default values here instead ?

antarest/study/dao/file/file_study_reserves_global_parameters_dao.py
```diff
@@ -38,16 +38,28 @@ def get_file_study(self) -> FileStudy:
...
     def get_all_reserves_global_parameters(self) -> ReservesGlobalParametersMapping:
         file_study = self.get_file_study()
-        return {area_id: self.get_reserves_global_parameters(area_id) for area_id in file_study.config.areas}
+        result: ReservesGlobalParametersMapping = {}
+        for area_id in file_study.config.areas:
+            try:
+                data = file_study.tree.get(_get_reserves_ini_path(area_id))
+            except (ChildNotFoundError, KeyError):
+                continue
+            if GLOBAL_PARAMETERS_SECTION not in data:
```
sylvlecl: Same here, I think we want default values instead of removing the area from the map.

antarest/study/dao/file/file_study_reserves_global_parameters_dao.py
sylvlecl: Elaborating on this:
having missing data in files is not the same as having missing data in DB.

- having missing data in files must be interpreted, in general, as "use the default values". This is not our convention, this is the convention of the simulator, and we must stick with it. We don't have control on the files that we receive, so we cannot assume that everything is defined
- having missing data in DB is an application error. We should assume that we don't have missing data in our DB, because WE filled it up in the first place.

So it's natural to have differences in that case:
- for FS, we must fill missing values with default values
- for DB, we should not do it: if there is missing data, it's an application error that we must fix in the code

antarest/study/dao/file/file_study_st_storage_dao.py
sylvlecl: I think we should not do it:
what we should have is a content-check on the object which is received, which should alreay be a "full" object, otherwise we should raise.

In that particular case, I think the DB side is not right either:
- on write it should perfom the content check
- on read it should not try to fill missing data. Same pattern as other comments

antarest/study/storage/rawstudy/model/filesystem/config/model.py
sylvlecl: Seems to me the only 2 places we use that class are:
1. from the INI file (kebab case stuff):
    for that we should just refactor `_parse_links_filtering` to use `parse_link` which already handles the parsing logic
2. from the Link class:
    there we should refactor the `to_config` method to just pass explicitly arguments to the LinkConfig constructor

Then we can get rid of that obscure validator

sylvlecl (commented):
Mostly very good stuff bu I think there is a few places where the adaptation is not relevant.
In particular we need to have a clear idea of the rsponsibilities of the DAO layer vs higher layer, in particular for missing values and validation:
- DAO layer should validate its inputs against study version
- DB DAO should not bother filling missing values, because we forbid missing values
- FS DAO must fill missing values, because it's part of the format definition

antarest/study/dao/file/file_study_area_dao.py
smailio: The if is useless, it's just to avoid running the line bellow when it isn't necessary. It's usefull to overwrite `ui_data.ui["layers"]` to make sure it's synchronized with ui that were added to specific layers 

This test will fail 

```python
    create_area("Paris", dao)
    dao.save_layer(Layer(id="1", name="Layer 1"))
    dao.save_layer(Layer(id="2", name="Layer 2"))

    # Add UI
    dao.save_area_ui(
        {
            "paris": {
                "1": AreaUI(x=100, y=100, color_rgb=(255, 0, 0)),
                "2": AreaUI(x=200, y=200, color_rgb=(0, 255, 0)),
            },
        }
    )

    all_ui = dao.get_all_areas_ui_info()
    paris_ui = all_ui["paris"]
    assert paris_ui.ui["layers"] == "0 1 2" # FAIL AssertionError: assert '0' == '0 1 2'

```
But It won't fail, if  I call save_layer_area, like this : 

```python
    create_area("Paris", dao)
    dao.save_layer(Layer(id="1", name="Layer 1"))
    dao.save_layer(Layer(id="2", name="Layer 2"))

    dao.save_layer_areas("1", ["paris"]) # <=== HERE I update paris_ui.ui["layers"]
    dao.save_layer_areas("2", ["paris"])

    # Add UI
    dao.save_area_ui(
        {
            "paris": {
                "1": AreaUI(x=100, y=100, color_rgb=(255, 0, 0)),
                "2": AreaUI(x=200, y=200, color_rgb=(0, 255, 0)),
            },
        }
    )

    all_ui = dao.get_all_areas_ui_info()
    paris_ui = all_ui["paris"]
    assert paris_ui.ui["layers"] == "0 1 2" # PASS
```
That's what I was trying to explain with my initial comment, but I admit it was so badly written that myself I didnt understand what I wrote. 

My new proposal is to revert this code and do the fix in fill_area instead.

antarest/study/dao/database/database_study_settings_dao.py
smailio: to not mutate original object, I added a comment

antarest/study/dao/database/database_study_settings_dao.py
smailio: done

antarest/study/dao/file/file_study_reserves_global_parameters_dao.py
smailio: done

antarest/study/dao/file/file_study_reserves_global_parameters_dao.py
smailio: It was to apply an extra check. Turns out I didn't need to apply the check so I reverted to this.

antarest/study/dao/file/file_study_st_storage_dao.py
smailio: done

antarest/study/storage/rawstudy/model/filesystem/config/model.py
smailio: @MartinBelthle there was a bug, code used to initialize this object with a  "filter_synthesis" instead of "filters_synthesis"

@sylvlecl Yeah I tried to fix this by making as little change as possible to the code base but you're right, should have been more ambitious.

smailio: As discussed File Tree Config wasn't initializing the same way than fs dao. 

Tested with simulator, FS DAO is right. 

Aligned tree config parsing with FS DAO for links. I did the same for areas as it had the same issue.

sylvlecl (dismissed):

sylvlecl (approved):

MartinBelthle (approved):

PR #3169, opened 2026-04-16
feat(reserves): implement reserve definitions

TheoPascoli:


antarest/study/business/model/reserve_definition_model.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class ReserveDefinition(AntaresBaseModel):
+    model_config = ConfigDict(alias_generator=to_camel, extra="forbid", populate_by_name=True)
+
+    @model_validator(mode="before")
+    @classmethod
+    def set_id(cls, data: Any) -> Any:
+        if isinstance(data, dict) and "id" not in data and "name" in data:
+            data["id"] = transform_name_to_id(data["name"], lower=False)
```
smailio: why lower=False ?

antarest/core/exceptions.py
```diff
@@ -142,6 +142,12 @@ class STStorageNotFound(ClusterNotFound):
     object_name = "Short-term storage"
 
 
+class ReserveDefinitionNotFound(ClusterNotFound):
```
MartinBelthle: Minor but do we want to inherit from a `Cluster` exception for Reserves ? I think we should just inherit from HTTPException only

antarest/study/business/model/reserve_definition_model.py
MartinBelthle: Indeed we shouldn't use `lower=False`

antarest/study/business/model/reserve_definition_model.py
```diff
@@ -0,0 +1,115 @@
...
+        _check_not_reserved_name(self.name)
+        return self
+
+
+class ReserveDefinitionUpdate(AntaresBaseModel):
+    model_config = ConfigDict(alias_generator=to_camel, extra="forbid", populate_by_name=True)
+
+    @model_validator(mode="before")
+    @classmethod
+    def _ignore_name(cls, data: Any) -> Any:
```
MartinBelthle: Same here, probably copy-pasted code from thermals and renewables but we do not want this

antarest/study/business/areas/reserve_definitions_management.py
```diff
@@ -0,0 +1,79 @@
...
+            reserve_properties={area_id: {reserve_id: data}},
+            study_version=study.version,
+            command_context=self._command_context,
+        )
+        study.add_commands([command])
+        return updated
+
+    def delete_reserve_definitions(self, study: StudyInterface, area_id: str, reserve_ids: Sequence[str]) -> None:
+        commands = [
+            RemoveReserveDefinition(
```
MartinBelthle: Seems the command could handle multiple reserves as the endpoint and the manager do. Probably won't change anything for FS but will be faster for DB

antarest/study/dao/api/reserve_definition_dao.py
```diff
@@ -0,0 +1,44 @@
...
+        raise NotImplementedError()
+
+
+class ReserveDefinitionDao(ReadOnlyReserveDefinitionDao):
+    @abstractmethod
+    def save_reserve_definitions(self, data: dict[AreaId, list[ReserveDefinition]]) -> None:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def delete_reserve_definition(self, area_id: AreaId, reserve_id: ReserveDefinitionId) -> None:
```
MartinBelthle: For the same reason, we could take a list of reserve ids.

antarest/study/dao/database/models/reserve_definition.py
```diff
@@ -0,0 +1,36 @@
...
+_RESERVE_TYPE_ENUM = Enum(ReserveType, name="reservetype")
+
+RESERVE_DEFINITION_TABLE = Table(
+    "reserve_definition",
+    metadata,
+    Column("study_id", String(36), nullable=False, primary_key=True),
+    Column("area_id", String(255), nullable=False, primary_key=True),
+    Column("reserve_id", String(255), nullable=False, primary_key=True),
+    Column("name", String(255), nullable=False),
+    Column("type", _RESERVE_TYPE_ENUM, nullable=False),
```
MartinBelthle: Seems you could just use `Enum(ReserveType)`

antarest/study/dao/file/file_study_reserve_definition_dao.py
```diff
@@ -0,0 +1,138 @@
...
+    @override
+    def get_all_reserve_definitions(self) -> ReserveDefinitionsMapping:
+        file_study = self.get_file_study()
+        try:
+            data = file_study.tree.get(_ALL_RESERVES_PATH, depth=4)
+        except (ChildNotFoundError, KeyError):
+            return {}
+        result: ReserveDefinitionsMapping = {}
+        for area_id, area_data in data.items():
+            reserves_data = area_data.get("reserves", {}) if isinstance(area_data, dict) else {}
```
MartinBelthle: What could be `area_data` instead of a dict ? Not sure to understand

antarest/study/storage/variantstudy/model/command/remove_reserve_definition.py
```diff
@@ -0,0 +1,52 @@
...
+
+class RemoveReserveDefinition(ICommand):
+    """
+    Command used to remove a reserve definition from an area.
+    """
+
+    command_name: CommandName = CommandName.REMOVE_RESERVE_DEFINITION
+
+    area_id: str
+    reserve_id: str
```
MartinBelthle: We should also have command validator version 10

antarest/study/storage/rawstudy/model/filesystem/config/reserves_global_parameters.py
```diff
@@ -17,6 +17,8 @@
 from antarest.core.utils.string import to_kebab_case
 from antarest.study.business.model.reserves_global_parameters_model import ReservesGlobalParameters
 
+GLOBAL_PARAMETERS_SECTION = "globalparameters"
```
MartinBelthle: Should be moved inside the `reserve_definition_model` file

antarest/study/dao/memory/in_memory_study_dao.py
```diff
@@ -238,6 +246,8 @@ def __init__(self, version: StudyVersion, matrix_service: ISimpleMatrixService,
         self._reserves: dict[str, str] = {}
         # Reserves global parameters
         self._reserves_global_parameters: dict[str, ReservesGlobalParameters] = {}
+        # Reserve definitions (per-reserve parameters)
+        self._reserve_definitions: dict[ClusterKey, ReserveDefinition] = {}
```
MartinBelthle: Minor but we should not use `ClusterKey` but rather another name which would do the same thing

antarest/study/dao/file/file_study_reserve_definition_dao.py
```diff
@@ -0,0 +1,138 @@
...
+        for section_name, section_data in data.items()
+        if not _is_global_parameters_key(section_name)
+    }
+
+
+def _parse_reserve_sections(sections: dict[str, dict[str, object]]) -> list[ReserveDefinition]:
+    reserves: list[ReserveDefinition] = []
+    for section_name, section_data in sections.items():
+        merged = dict(section_data)
+        merged.setdefault("name", section_name)
```
MartinBelthle: Think all the method could just be
```python
return [parse_reserve_definition({"name": name, **data}) for name, data in sections.items()
```
or something like it

antarest/study/dao/file/file_study_reserve_definition_dao.py
MartinBelthle: Also I think this whole code would be easier to read if you've looped around `get_all_reserve_definitions_for_area` based on the `file_study.config.areas` object.

You'll have the same performance as these are separated files (in fact it could be a bit worse because we'd build the tree more but that should be okay)

antarest/study/dao/file/file_study_reserve_definition_dao.py
```diff
@@ -0,0 +1,138 @@
...
+                result[area_id] = {reserve.id: reserve for reserve in reserves}
+        return result
+
+    @override
+    def get_all_reserve_definitions_for_area(self, area_id: str) -> Sequence[ReserveDefinition]:
+        return _parse_reserve_sections(self._read_area_reserves(area_id))
+
+    @override
+    def get_reserve_definition(self, area_id: str, reserve_id: str) -> ReserveDefinition:
+        file_study = self.get_file_study()
```
MartinBelthle: Same here the whole code is a bit complex to read IMO.

For the same performance, we could simply use `_read_area_reserves` and choose the id the user asked

antarest/study/dao/file/file_study_reserve_definition_dao.py
```diff
@@ -0,0 +1,138 @@
...
+        return bool(section_data)
+
+    @override
+    def save_reserve_definitions(self, data: dict[AreaId, list[ReserveDefinition]]) -> None:
+        file_study = self.get_file_study()
+        for area_id, reserves in data.items():
+            check_area_exists(file_study.config, area_id)
+            for reserve in reserves:
+                section_data = serialize_reserve_definition(reserve)
+                file_study.tree.save(section_data, _reserve_section_path(area_id, reserve.id))
```
MartinBelthle: Highly inefficient if the user wants to save 50 reserves for the same area, we should do one `file_study.tree.save` per area

antarest/study/dao/database/database_reserve_definition_dao.py
```diff
@@ -0,0 +1,112 @@
...
+    def save_reserve_definitions(self, data: dict[AreaId, list[ReserveDefinition]]) -> None:
+        if not data:
+            return
+
+        values = []
+        for area_id, reserves in data.items():
+            for reserve in reserves:
+                values.append(_convert_model_to_row(self._study_id, area_id, reserve))
+
+        upsert_multiple(session=self._db_session, table=_TABLE, values=values)
```
MartinBelthle: Usually we encapsulate this inside a try, catch to raise a proper exception

tests/study/dao/file/test_filestudy_reserve_definition_dao.py
```diff
@@ -0,0 +1,218 @@
...
+from antarest.study.dao.file.file_study_dao import FileStudyTreeDao
+from antarest.study.model import STUDY_VERSION_9_3
+from antarest.study.storage.rawstudy.model.filesystem.config.files import build
+from antarest.study.storage.rawstudy.model.filesystem.factory import FileStudy
+from antarest.study.storage.rawstudy.model.filesystem.root.filestudytree import FileStudyTree
+from antarest.study.storage.variantstudy.business.matrix_constants_generator import GeneratorMatrixConstants
+from tests.study.dao.utils import save_area
+
+
+def _build_file_study_v10(tmp_path: Path, matrix_service: ISimpleMatrixService) -> FileStudy:
```
MartinBelthle: One really easier way to do this is just to use the `empty_study_930` fixture and do this:
`empty_study_930.config.version = STUDY_VERSION_10_0`.
I tried it on your branch and all the tests are good

tests/study/dao/file/test_filestudy_reserve_definition_dao.py
```diff
@@ -0,0 +1,218 @@
+# Copyright (c) 2026, RTE (https://www.rte-france.com)
```
MartinBelthle, line 1: Anis is currently trying to make tests we have use the 2 types of DAOs and I think you should do the same as almost every test should behave the same. You can use the `dao` fixture that is parametrized by nature and merge "duplicated" tests you've written

MartinBelthle (changes requested):

antarest/study/business/model/reserve_definition_model.py
```diff
@@ -0,0 +1,115 @@
...
+    model_config = ConfigDict(alias_generator=to_camel, extra="forbid", populate_by_name=True)
+
+    @model_validator(mode="before")
+    @classmethod
+    def set_id(cls, data: Any) -> Any:
+        if isinstance(data, dict) and "id" not in data and "name" in data:
+            data["id"] = transform_name_to_id(data["name"], lower=False)
+        return data
+
+    id: str
```
smailio: We can use a NewType for this Id,  That would make your ReserveDefinitionUpdates bellow more readable.

antarest/study/business/model/reserve_definition_model.py
```diff
@@ -0,0 +1,115 @@
...
+
+    type: ReserveType | None = None
+    failure_cost: Cost | None = None
+    spillage_cost: Cost | None = None
+    reference_activation_duration: Duration | None = None
+    power_activation_ratio: Ratio | None = None
+    energy_activation_ratio: Ratio | None = None
+
+
+ReserveDefinitionUpdates = dict[str, dict[str, ReserveDefinitionUpdate]]
```
smailio: You can use the AreaId type here like you did in `save_reserve_definitions

antarest/study/dao/database/database_reserve_definition_dao.py
```diff
@@ -0,0 +1,128 @@
...
+        for area_id, reserves in data.items():
+            for reserve in reserves:
+                values.append(_convert_model_to_row(self._study_id, area_id, reserve))
+        try:
+            upsert_multiple(session=self._db_session, table=_TABLE, values=values)
+        except IntegrityError as e:
+            for area_id in data:
+                if not area_exists(self._db_session, self._study_id, area_id):
+                    raise AreaNotFound(area_id) from e
+            raise
```
smailio, line 103: when that would happen ?

antarest/study/dao/database/database_reserve_definition_dao.py
```diff
@@ -0,0 +1,128 @@
...
+        self._db_session.commit()
+
+    @override
+    def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[ReserveDefinitionId]) -> None:
+        result = self._db_session.execute(
```
smailio, line 110: you can return early if reserve_ids is empty. Maybe you thought it's not worth it to do the optimisaiton as the case never occur ?

antarest/study/dao/database/database_reserve_definition_dao.py
TheoPascoli: When the IntegrityError is caused by something else than a missing area. To be sure to raise everything

antarest/study/business/model/reserve_definition_model.py
TheoPascoli: I agree but it would create a cycle ...

antarest/study/business/model/reserve_definition_model.py
```diff
@@ -0,0 +1,111 @@
...
+    DOWN = "down"
+
+
+Cost = Annotated[float, Field(ge=0)]
+Ratio = Annotated[float, Field(ge=0, le=1)]
+Duration = Annotated[int, Field(ge=0)]
+
+
+def _check_not_reserved_name(name: str) -> None:
+    if transform_name_to_id(name, lower=True) in _RESERVED_RESERVE_IDS:
```
MartinBelthle: Minor but no need for `lower=True`

antarest/study/business/model/reserve_definition_model.py
```diff
@@ -0,0 +1,111 @@
...
+    @model_validator(mode="after")
+    def _check_name(self) -> "ReserveDefinitionCreation":
+        _check_not_reserved_name(self.name)
+        return self
+
+
+class ReserveDefinitionUpdate(AntaresBaseModel):
+    model_config = ConfigDict(alias_generator=to_camel, extra="forbid", populate_by_name=True)
+
+    name: str | None = None
```
MartinBelthle: So we handle renaming a reserve ? If not we should rename the `name` attribute from this class. And if we do we should add tests (but I don't think we need to handle the renaming)

antarest/study/dao/database/database_reserve_definition_dao.py
```diff
@@ -0,0 +1,130 @@
...
+            reserve = _convert_row_to_model(row)
+            result.setdefault(row.area_id, {})[reserve.id] = reserve
+        return result
+
+    @override
+    def get_all_reserve_definitions_for_area(self, area_id: str) -> Sequence[ReserveDefinition]:
+        stmt = select(_TABLE).where((_TABLE.c.study_id == self._study_id) & (_TABLE.c.area_id == area_id))
+        rows = self._db_session.execute(stmt).fetchall()
+        if not rows:
+            validate_area_exists(self._db_session, self._study_id, area_id)
```
MartinBelthle, line 73: If the area exists but the reserve table is not filled, we won't raise here but we'll raise badly on the next line. We should do the same thing as in the following method `raise ReserveDefinitionNotFound(area_id, reserve_id)`

antarest/study/dao/file/file_study_reserve_definition_dao.py
```diff
@@ -0,0 +1,126 @@
...
+            file_study.tree.save(existing, _area_reserves_path(area_id))
+
+    @override
+    def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[ReserveDefinitionId]) -> None:
+        file_study = self.get_file_study()
+        for reserve_id in reserve_ids:
+            if not self.reserve_definition_exists(area_id, reserve_id):
+                raise ReserveDefinitionNotFound(area_id, reserve_id)
+        for reserve_id in reserve_ids:
+            file_study.tree.delete(_reserve_section_path(area_id, reserve_id))
```
MartinBelthle, line 126: Not performant as we're doing N I/O operations on the same file but the deletion almost never happens so if you can make it in one I/O it's nice otherwise not a big issue

antarest/study/storage/variantstudy/model/command/remove_reserve_definitions.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @model_validator(mode="after")
+    def _validate_version(self) -> Self:
+        if self.study_version < STUDY_VERSION_10_0:
+            raise InvalidFieldForVersionError("Reserve definitions are not valid for study version before 10.0")
+        return self
+
+    @override
+    def _apply_dao(self, study_data: StudyDao, listener: ICommandListener | None = None) -> CommandOutput[None]:
+        missing = [rid for rid in self.reserve_ids if not study_data.reserve_definition_exists(self.area_id, rid)]
```
MartinBelthle: This is also highly unefficient. In the remove command we don't need to call the `exists` as the `delete` method from the DAO will raise so I think we can remove this

tests/variantstudy/model/command/test_reserve_definition_commands.py
```diff
@@ -0,0 +1,224 @@
...
+from antarest.study.dao.memory.in_memory_study_dao import InMemoryStudyDao
+from antarest.study.model import STUDY_VERSION_10_0
+from antarest.study.storage.variantstudy.model.command.create_reserve_definition import CreateReserveDefinition
+from antarest.study.storage.variantstudy.model.command.remove_reserve_definitions import RemoveReserveDefinitions
+from antarest.study.storage.variantstudy.model.command.update_reserve_definitions import UpdateReserveDefinitions
+from antarest.study.storage.variantstudy.model.command_context import CommandContext
+from tests.study.dao.utils import save_area
+
+
+def _make_dao(version: StudyVersion = STUDY_VERSION_10_0) -> InMemoryStudyDao:
```
MartinBelthle: Would be better to really use the FS and DB DAOs if possible. In the commands tests at the same place, we only use the FS Dao with `build_dao_from_file_study` but for these the new tests we could do better and use both :)

tests/study/business/model/test_reserve_definition_model.py
```diff
@@ -0,0 +1,173 @@
...
+                ReserveDefinition(**kwargs)
+
+
+class TestReserveDefinitionCreation:
+    def test_only_name_and_type_required(self) -> None:
+        creation = ReserveDefinitionCreation(name="R1", type=ReserveType.UP)
+        assert creation.name == "R1"
+        assert creation.failure_cost is None
+
+    def test_reserved_name_rejected(self) -> None:
```
MartinBelthle: Minor but you could merge your 4 tests into a single parametrized one

MartinBelthle (changes requested):

antarest/study/business/model/reserve_definition_model.py
MartinBelthle: What about this comment ?

MartinBelthle (changes requested):

antarest/study/business/model/reserve_definition_model.py
TheoPascoli: Yes I forgot to remove the field !

MartinBelthle (approved):

PR #3170, opened 2026-04-16
fix(matrix): `/matrixindex` endpoint now works on DB studies

MartinBelthle:


antarest/study/service.py
```diff
@@ -1078,9 +1080,19 @@ def get_study_synthesis(self, study_id: str) -> StudySynthesis:
...
+    def get_input_matrix_startdate(self, study_id: str, path: str) -> MatrixIndex:
         study = self.get_study(study_id)
         assert_permission(study, StudyPermissionType.READ)
+
+        # We need to handle matrices index differently if our study is stored in DB
+        if study.storage_mode == StorageMode.DATABASE:
+            dao = self.get_study_interface(study).get_study_dao()
+            # We can give a readOnly Dao as we won't use the save methods
+            mapper = RawPathToMatrixMapper(dao)  # type: ignore
+            matrix_frequency = mapper.get_matrix_frequency_from_path(Path(path))
```
sylvlecl: Makes me realize we should use pure paths (`PurePosixPath`) as input of the mapper:
the paths here are not actual filesystem paths
(actually we should probably use that also in the API)

antarest/study/storage/utils.py
```diff
@@ -342,6 +342,28 @@ def parse_simulation_range(config: dict[str, Any]) -> SimulationRangeDefinition:
     )
 
 
+def parse_simulation_range_from_model(general: GeneralConfig) -> SimulationRangeDefinition:
```
sylvlecl: It's not really "parsing" at that point, parsing should really be reserved to transforming raw stuff (strings mainly, dicts that are just the exact image of a json/ini file).
Maybe juste get/extract

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -173,211 +178,262 @@ def _save_st_storage_cost_variation_withdrawal(area_id: str, storage_id: str, se
...
                 pattern=re.compile(r"input/reserves/(?P<area_id>[^/]+)"),
                 getter=dao.get_reserves,
                 setter=_save_reserves,
+                frequency=lambda **x: MatrixFrequency.HOURLY,
             ),
             RegexMatcher(
                 pattern=re.compile(r"input/links/(?P<area_from>[^/]+)/capacities/(?P<area_to>[^/]+)_direct"),
                 getter=dao.get_link_direct_capacities,
                 setter=_save_link_direct_capacities,
+                frequency=lambda **x: MatrixFrequency.HOURLY,
```
sylvlecl: we could define just a few functions and reference them instead of repeating identical lambdas

sylvlecl (commented):
Looks good to me, just a few code quality suggestions

antarest/study/service.py
MartinBelthle: Just changed this inside the mapper

sylvlecl (approved):

PR #3178, opened 2026-04-20
refactor(services)!: introduce skeleton for database storage service

MartinBelthle:
Breaking change:
The endpoint GET `/raw/original-file` won't work on archived studies anymore.
Shouldn't impact anyone


antarest/study/storage/abstract/abstract_service.py
```diff
@@ -0,0 +1,105 @@
...
+    Study,
+    StudyMetadataDTO,
+)
+from antarest.study.storage.study_service_interface import IStudyService
+from antarest.study.storage.utils import remove_from_cache
+
+logger = logging.getLogger(__name__)
+
+
+class AbstractService(IStudyService, ABC):
```
sylvlecl: Naming: abstract study service

antarest/study/storage/abstract/abstract_study_service.py
```diff
@@ -0,0 +1,105 @@
...
+            public_mode=study.public_mode or PublicMode.NONE,
+            horizon=study.horizon,
+            folder=folder_path or study.folder,
+            tags=[tag.label for tag in study.tags],
+            directory_id=study.directory_id,
+            parent_id=study.parent_id,
+        )
+
+    @override
+    def delete_from_filesystem(self, study: Study) -> None:
```
sylvlecl, line 80: Why "from filesystem" ?
Seems to me this should be forwarded to the underlying storage, since the deletion in DB will be different, and implementation should differ between variant and raw, since for variants we'll want to delete snapshot and commands.

antarest/study/storage/study_service_interface.py
```diff
@@ -0,0 +1,39 @@
...
+# This file is part of the Antares project.
+
+from abc import ABC, abstractmethod
+from pathlib import Path, PurePosixPath
+
+from antarest.study.dao.api.study_dao import StudyDao
+from antarest.study.model import RawStudy, Study, StudyMetadataDTO
+
+
+class IStudyService(ABC):
```
sylvlecl, line 20: We should probably find a more specific name for that hierarchy, because it's ambiguous with the `StudyService` (will think about it)

antarest/study/storage/storage_service.py
```diff
@@ -48,7 +48,7 @@ def __init__(
         self.raw_study_service = raw_study_service
         self.variant_study_service = variant_study_service
 
-    def get_storage(self, study: Study) -> IStudyStorage:
+    def get_storage(self, study: Study) -> IStudyService:
```
sylvlecl, line 51: should be renamed accordingly (but we can wait to be sure on the name we want :) )

Note: we could completely remove that class and just have a dispatching function in study service instead

antarest/study/storage/file_study_storage.py
```diff
@@ -0,0 +1,213 @@
...
+        file_study = self._get_file_study(dest_path, True, new_study.id)
+
+        update_antares_info(new_study, file_study.tree, update_author=False)
+
+        self.normalize_file_study(file_study)
+
+        return new_study
+
+    @override
+    def write_study_to_filesytem(self, study: Study) -> None:
```
sylvlecl: Seems it does not do what it says.

Not sure why we would want to denormalize a study in place ?
Seems it's called for export to a directory, but it must absolutely be denormalized in the target dir, not in the actual storage dir

antarest/study/storage/file_study_storage.py
```diff
@@ -0,0 +1,213 @@
...
+        self.denormalize_study(study)
+
+    @override
+    def write_study_for_archive(self, study: RawStudy) -> None:
```
sylvlecl: Same: it does no do what it says :) 
Here too it would be better to not denormalize in place (for ex. if it's interrupted, the study will be half denormalized)


We should probably just have an export function which handles the actual writing to a dir.

antarest/output/storage/output_storage.py
```diff
@@ -266,3 +266,9 @@ def get_logs(self, study_id: str, output_id: str, log_type: LogType) -> str:
         """
         Retrieve logs.
         """
+
+    @abstractmethod
```
sylvlecl, line 270: nice

tests/integration/conftest.py
```diff
@@ -225,3 +226,22 @@ def internal_study_fixture(
     res.raise_for_status()
     study_ids = t.cast(t.Iterable[str], res.json())
     return next(iter(study_ids))
+
+
+@pytest.fixture(scope="session")
```
MartinBelthle, line 231: These fixtures were just moved to be used by more tests

antarest/study/dao/file/file_study_factory_dao.py
```diff
@@ -36,18 +35,29 @@ def __init__(self, command_context: CommandContext, study_factory: StudyFactory,
...
+    @override
+    def get_study_dao(self, study: Study) -> FileStudyTreeDao:
+        return self._build_dao(study, create_study=False)
 
-        # If the study already exists, we won't override it but instead use the existing one to build the DAO object.
-        # This case happens in particular with variant studies snapshots.
-        if not study_path.exists():
+    def _build_dao(self, study: Study, create_study: bool) -> FileStudyTreeDao:
+        # We need to differentiate `RawStudy` and `VariantStudy` to be able to parse the config
+        if isinstance(study, RawStudy):
```
sylvlecl: I think a better separation of responsibilities would be: 
- the FS DAO just needs a path (and maybe an output path) to be able to create a correct DAO
- so, it only needs a mapping "study id -> paths" to do its work.
  --> that function should be an argument of the constructor
- who knows the mapping that should be used?
  it's either the raw service or variant service
  --> the DAO factory should be created by them, and they should provide that mapping function

Then, the methods of the factory should not receive a full SQL alchemy object, which is "dangerous" because it means that it could be modified by those functions --> the study ID should be enough, I think

antarest/study/storage/study_storage_interface.py
```diff
@@ -0,0 +1,60 @@
...
+
+from abc import ABC, abstractmethod
+from typing import BinaryIO
+
+from antarest.study.dao.api.study_dao import StudyDao
+from antarest.study.model import RawStudy, Study
+from antarest.study.storage.variantstudy.model.dbmodel import VariantStudy
+
+
+class IStudyStorage(ABC):
```
sylvlecl, line 21: We need to explain in words the responsibility/role of the interface and its methods, to make sure they are clear

antarest/study/storage/study_storage_interface.py
```diff
@@ -0,0 +1,60 @@
...
+    @abstractmethod
+    def write_study_for_archive(self, study: RawStudy) -> None:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_disk_usage(self, study: Study) -> int:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def normalize_study(self, study: Study) -> None:
```
sylvlecl, line 43: This method should be in the end an internal implementation detail of the FS implementation:
seems only used by the /normalize endpoint (OK...) and the upgrader, but the upgrader should instead call a method to export the study to a tmp working dir.

We should probably document it as "deprecated"

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -11,275 +11,130 @@
...
-        super().__init__(config=config, cache=cache)
+    def __init__(self, config: Config, study_factory: StudyFactory, cache: ICache, command_context: CommandContext):
+
+        super().__init__(cache, config)
 
         self.study_factory = study_factory
-        self._matrix_service = matrix_service
+        self._matrix_service = command_context.matrix_service
+        self._file_study_storage = FileStudyStorage(cache, config, command_context, study_factory)
+        self._database_study_storage = DatabaseStudyStorage(
```
sylvlecl, line 78: it's used only through the mapping below --> let's not add it as a member if not needed.

Would be best to not have the FS implementation as a member too, but I understand this can help at least temporarily

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -288,165 +143,21 @@ def import_study(self, metadata: RawStudy, stream: BinaryIO) -> RawStudy:
...
-        shutil.copytree(src=study_dir, dst=dest, ignore=ignore_outputs)
-
-        stop_time = time.time()
-        duration = f"{stop_time - start_time:.3f}"
-        logger.info(f"Study '{study_dir}' exported (flat mode) in {duration}s")
-
-        if denormalize:
-            study = self.study_factory.create_from_fs(dest, False, "", use_cache=False)
-            self.denormalize_file_study(study)
+        self._file_study_storage.denormalize_study(study)
```
sylvlecl, line 207: We should aim at removing such direct use of a specific implementation:
for ex. this seem used only for the study upgrade: in the end we should instead call an import function at the end of the upgrade and the denormalization will be an implementation detail of the FS storage.

Again I understand that we can keep it temporarily, but maybe leave a TODO for future PRs ?

antarest/study/storage/study_storage_interface.py
```diff
@@ -0,0 +1,60 @@
...
+        raise NotImplementedError()
+
+    @abstractmethod
+    def import_study(self, study: RawStudy, stream: BinaryIO) -> RawStudy:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def is_snapshot_up_to_date(self, study: VariantStudy) -> bool:
```
sylvlecl: After a little more thinking it seems to me it would be better that those snapshot - specific methods be isolated in a dedicated interface that will be use only by the variant service, some thing like "snapshot manager".
Those methods are only ever needed by the variant service, so it makes more sense to not mix them here where they will be accessible by the raw service.

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -11,275 +11,130 @@
...
     remove_from_cache,
-    update_antares_info,
 )
+from antarest.study.storage.variantstudy.model.command_context import CommandContext
 
 logger = logging.getLogger(__name__)
 
 
-class RawStudyService(AbstractStorageService):
+class RawStudyService(AbstractService):
```
sylvlecl: I think we start to have an ambiguity about what means "raw".

In this refacto, we consider it means a "reference" study as opposed to a variant.
But elsewhere in the app, it seems to be more or less a synonym of "filesystem" studies:
we get "raw" data etc.
And actually almost none of the "raw" API will work for DB studies.

So I think we should find another name for that "non-variant" study concept.
Maybe BaseStudy, PlainStudy, MaterialStudy ?

antarest/study/storage/study_service_interface.py
sylvlecl: Actually, could be something as simple as IStudyServiceBackend, backend is more generic than the "storage" we had before, and it says pretty well that this interface supports the features of the `StudyService` through mutiple implementations.

sylvlecl (commented):
Good direction already, I added some comments / questions mostly for discussion, so that we can make the new breakdown of our interfaces even better

antarest/study/storage/abstract/abstract_study_service.py
MartinBelthle: The DB deletion is handled via the study service and so does the output one. So the only thing left to delete here is the study folder if it exists. So I named it like this to make it clear and didn't implemented in the varaint and raw services as this code is simpler and works for every use case

PR #3181, opened 2026-04-21
perf(ui-matrix): fix paste lag and locale-aware number parsing

hdinia:


webapp/src/components/Matrix/hooks/useMatrixPasteInterceptor.ts
```diff
@@ -0,0 +1,165 @@
...
+  gridSelection,
+  gridToData,
+  getDataRowIndex,
+  onBulkPaste,
+}: UseMatrixPasteInterceptorOptions) {
+  const enabled = !readOnly && !!onBulkPaste;
+
+  // useEffectEvent keeps the handler identity stable across renders while always closing
+  // over the latest props, so we register the window listener once and never again.
+  const handlePaste = React.useEffectEvent((e: ClipboardEvent) => {
```
skamril: Please import only `useEffectEvent` rather than the whole module.

webapp/src/components/Matrix/hooks/useMatrixPasteInterceptor.ts
```diff
@@ -0,0 +1,165 @@
...
+    if (!onBulkPaste) {
```
skamril, line 61: 💡 This function is doing a few different things; worth splitting into smaller helpers to keep each part readable and testable.

webapp/src/components/Matrix/hooks/useMatrixPasteInterceptor.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  useEffect(() => {
+    if (!enabled) {
+      return;
+    }
+
+    const listener = (e: ClipboardEvent) => handlePaste(e);
+    window.addEventListener("paste", listener, { capture: true });
+    return () => window.removeEventListener("paste", listener, { capture: true });
+  }, [enabled, handlePaste]);
```
skamril: Normally `handlePaste` shouldn't be put in the deps array because it's an effect events, I don't know if it is an issue.

You can update the package "eslint-plugin-react-hooks" to v7.1.1 to support this new feature with ESLint:

"eslint-plugin-react-hooks": "7.1.1",

webapp/src/components/Matrix/hooks/useMatrixPasteInterceptor.ts
```diff
@@ -0,0 +1,165 @@
...
+    onBulkPaste(updatedData);
+  });
+
+  useEffect(() => {
+    if (!enabled) {
+      return;
+    }
+
+    const listener = (e: ClipboardEvent) => handlePaste(e);
+    window.addEventListener("paste", listener, { capture: true });
```
skamril: No need for the `listener` variable, you can pass `handlePaste` directly:

```ts
window.addEventListener("paste", handlePaste, { capture: true });
return () => window.removeEventListener("paste", handlePaste, { capture: true });
```

webapp/src/components/Matrix/shared/utils.ts
```diff
@@ -51,6 +51,64 @@ import type {
...
+
+  // Only commas -> 2+ means thousands-only.
+  if ((value.match(/,/g) ?? []).length > 1) {
+    return value.replace(/,/g, "");
+  }
+
+  // Single comma is ambiguous, 3 trailing digits -> thousands, else decimal.
+  const digitsAfterComma = value.length - lastComma - 1;
+  return digitsAfterComma === 3
+    ? value.replace(",", "") // "4,567" → "4567"
```
skamril, line 117: ⚠️ Ambiguous edge case: "4,567" and "4.567", is that 4567 or 4.567?

skamril (changes requested):

webapp/src/components/Matrix/shared/utils.ts
```diff
@@ -51,6 +51,64 @@ import type {
...
+  const cleaned = raw.replace(/\s/g, "");
+
+  if (!cleaned) {
+    return NaN;
+  }
+
+  return Number(normalizeLocaleSeparators(cleaned));
+}
+
+function normalizeLocaleSeparators(value: string): string {
```
skamril, line 76: 💡Proposition that used current local or a specified one:

```ts
function parseNumberStrWithLocale(value: string, locales?: string | string[]) {
  const parts = new Intl.NumberFormat(locales).formatToParts(1111.1);
  const group = parts.find((p) => p.type === "group")?.value; // e.g. ","
  const decimal = parts.find((p) => p.type === "decimal")?.value; // e.g. "."

  // Guard against mismatched formats:
  // e.g. value="1,234.56" with locale "de-DE" (group=".", decimal=",")
  // Without this check:
  //   replaceAll(".", "") → "1,23456"
  //   replaceAll(",", ".") → "1.23456"  ❌ (expected 1234.56)
  if (decimal && group && value.includes(group) && value.includes(decimal)) {
    // group separator must always appear before decimal separator
    if (value.lastIndexOf(group) > value.indexOf(decimal)) {
      return NaN;
    }
  }

  let newVal = value;

  if (group) {
    newVal = newVal.replaceAll(group, "");
  }

  if (decimal) {
    newVal = newVal.replaceAll(decimal, ".");
  }

  return Number(newVal);
}
```

webapp/src/components/Matrix/hooks/useMatrixPasteInterceptor.ts
hdinia: bumping eslint-plugin-react-hooks to 7.1.1 surfaces 58 errors from the new React-Compiler rules across the codebase (set-state-in-effect, "cannot access refs during render", etc.), all pre-existing. Too much scope for this PR. I added an `eslint-disable-next-line` with a TODO comment to re-enable once the plugin is bumped project-wide

webapp/src/components/Matrix/hooks/useMatrixPasteInterceptor.ts
hdinia: It's because of the eslint rule but I'll update it to allow hooks

skamril (approved):

PR #3191, opened 2026-04-24
feat(reserves): implement need matrix

TheoPascoli:
[ANT-4620]


antarest/study/dao/file/file_study_reserve_definition_dao.py
```diff
@@ -124,3 +140,67 @@ def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[Rese
...
+                matrices_mapping.setdefault(matrix_id, []).append(node)
+        self.get_impl().save_matrices(matrices_mapping)
+
+    @override
+    def delete_reserve_need(self, area_id: AreaId, reserve_id: ReserveDefinitionId) -> None:
+        study_data = self.get_file_study()
+        path = _reserve_need_matrix_path(area_id, reserve_id)
+        try:
+            study_data.tree.delete(path)
+        except (ChildNotFoundError, KeyError):
```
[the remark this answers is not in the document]
TheoPascoli: Its ok, missing matrix means the need was never created or already deleted

antarest/study/dao/database/database_reserve_definition_dao.py
```diff
@@ -130,3 +142,63 @@ def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[Rese
...
+    def get_reserve_need(self, area_id: str, reserve_id: str) -> pl.DataFrame:
+        stmt = select(_NEED_TABLE).where(
+            (_NEED_TABLE.c.study_id == self._study_id)
+            & (_NEED_TABLE.c.area_id == area_id)
+            & (_NEED_TABLE.c.reserve_id == reserve_id)
+        )
+        row = self._db_session.execute(stmt).fetchone()
+        if not row:
+            raise ReserveDefinitionNotFound(area_id, reserve_id)
+        return self.get_impl().get_matrix(str(row.matrix_id), default_empty_supplier=None)
```
MartinBelthle: No `default_empty_supplier` seems weird, we only have this for Xpansion files. Should be validated with the Simulator Team

antarest/study/dao/database/database_reserve_definition_dao.py
```diff
@@ -130,3 +142,63 @@ def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[Rese
...
+                    }
+                )
+        try:
+            upsert_multiple(session=self._db_session, table=_NEED_TABLE, values=values)
+        except IntegrityError as e:
+            for area_id, per_reserve in mapping.items():
+                if not area_exists(self._db_session, self._study_id, area_id):
+                    raise AreaNotFound(area_id) from e
+                # Otherwise, the reserve definition does not exist yet (FK violation on (study, area, reserve)).
+                for reserve_id in per_reserve:
```
MartinBelthle, line 190: All this seems a weird way to raise. Usually, we first verify all areas and then if all exist, all reserves and in the end if all exist a ValueError saying that the table is somehow badly filled

antarest/study/dao/database/database_reserve_definition_dao.py
```diff
@@ -130,3 +142,63 @@ def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[Rese
...
+            return
+        values = []
+        for area_id, per_reserve in mapping.items():
+            for reserve_id, matrix_id in per_reserve.items():
+                values.append(
+                    {
+                        "study_id": self._study_id,
+                        "area_id": area_id,
+                        "reserve_id": reserve_id,
+                        "matrix_id": extract_matrix_id(matrix_id),
```
MartinBelthle, line 180: In the rest of the code we only need this `extract_matrix_id` in the save methods of the FS Dao, we should try to stay coherent

antarest/study/dao/database/database_reserve_definition_dao.py
```diff
@@ -130,3 +142,63 @@ def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[Rese
...
+    @override
+    def delete_reserve_need(self, area_id: AreaId, reserve_id: ReserveDefinitionId) -> None:
+        result = self._db_session.execute(
+            delete(_NEED_TABLE).where(
+                (_NEED_TABLE.c.study_id == self._study_id)
+                & (_NEED_TABLE.c.area_id == area_id)
+                & (_NEED_TABLE.c.reserve_id == reserve_id)
+            )
+        )
+        assert isinstance(result, CursorResult)
```
MartinBelthle: Usually, here we do this also:
```
if result.rowcount == 0:
    raise xxx
```

antarest/study/dao/study_conversion/study_converter.py
```diff
@@ -183,6 +183,8 @@ def _convert_areas(self) -> None:
                 self._new_dao.save_reserve_definitions(
                     {area_id: list(reserves.values()) for area_id, reserves in reserve_definitions.items()}
                 )
+                # Reserve need chronicles — must be copied after the definitions (FK dependency in database backend).
+                self._new_dao.save_reserve_need(self._source_dao.get_all_reserve_needs())
```
MartinBelthle: Makes me realize we should probably add an `s` to `save_reserve_need `.

Also, now that we start to have a lot of logic you could move all this in a `_convert_reserves` method

antarest/study/storage/rawstudy/model/filesystem/config/files.py
```diff
@@ -507,6 +510,34 @@ def _parse_st_storage_additional_constraints(
     return config
 
 
+def _parse_reserves(root: Path, area: str) -> list[ReserveDefinition]:
```
MartinBelthle: Discussed this with Sylvain and he would prefer it we store in the Config only what you need so in this case seems to be only the list of reserve ids

antarest/study/storage/rawstudy/model/filesystem/root/input/reserves/area_reserves.py
```diff
@@ -13,12 +13,23 @@
...
-        return {
+        area_id = self.config.path.name
+        tree: TREE = {
             "reserves": InputReservesIni(self.config.next_file("reserves.ini")),
         }
+        for reserve in self.config.areas[area_id].reserves:
+            tree[reserve.id] = InputSeriesMatrix(
+                self.matrix_mapper,
+                self.config.next_file(f"{reserve.id}.txt"),
+                default_empty=default_scenario_hourly,
```
MartinBelthle, line 32: So there is a defauly empty. We should put this in the DB DAO. And also add tests for this in the test `test_empty_matrices.py`

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -236,6 +240,14 @@ def _get_binding_constraint_matrix_frequency(constraint_id: str) -> MatrixFreque
                 setter=_save_misc_gen,
                 frequency=_get_hourly_frequency,
             ),
+            RegexMatcher(
+                # Exclude `reserves` as a reserve_id: `input/reserves/<area>/reserves` is the
```
MartinBelthle: Makes me realize, what happens if an user created a reserve with the id `reserves` ? Is it forbidden ? And if not seems we won't be able to retrieve its matrix with this code if I understand correctly

antarest/study/storage/rawstudy/raw_path_to_matrix_mapper.py
```diff
@@ -236,6 +240,14 @@ def _get_binding_constraint_matrix_frequency(constraint_id: str) -> MatrixFreque
...
+                # reserve-definitions INI node, not a need-matrix.
+                pattern=re.compile(r"input/reserves/(?P<area_id>[^/]+)/(?P<reserve_id>(?!reserves$)[^/]+)"),
+                getter=dao.get_reserve_need,
+                setter=_save_reserve_need,
+                frequency=_get_hourly_frequency,
+            ),
             RegexMatcher(
```
MartinBelthle, line 243: Also this should be put at the end of the file with the rest of the mappers who depends on the study version

antarest/study/storage/variantstudy/model/command/remove_reserve_definitions.py
```diff
@@ -46,6 +46,8 @@ def _validate_version(self) -> Self:
 
     @override
     def _apply_dao(self, study_data: StudyDao, listener: ICommandListener | None = None) -> CommandOutput[None]:
+        for reserve_id in self.reserve_ids:
+            study_data.delete_reserve_need(self.area_id, reserve_id)
```
MartinBelthle: This should be handled in the `delete_reserve_definitions` method. This way the DB deletion will be faster and also it will be coherent with the rest of the FS and DB Daos

tests/study/dao/matrix_raw_mapper/test_raw_path_to_matrix_mapper.py
```diff
@@ -581,3 +585,36 @@ def test_version_specifics(db_dao_930: DatabaseStudyDao) -> None:
...
+        energy_activation_ratio=0.9,
+    )
+
+
+def test_reserve_need_path_get_and_save(dao_10_0: StudyDao, matrix_service: ISimpleMatrixService) -> None:
+    area_id = "paris"
+    reserve_id = "R1"
+
+    save_area(dao_10_0, area_id)
+    dao_10_0.save_reserve_definitions({area_id: [_build_reserve_definition(reserve_id)]})
```
MartinBelthle: This should rather be put in the `build_real_case_study` for other tests to use it easily

antarest/study/dao/file/file_study_reserve_definition_dao.py
```diff
@@ -124,3 +140,67 @@ def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[Rese
...
+        study_data = self.get_file_study()
+        matrix_nodes: dict[MatrixNode, tuple[str, ReserveDefinitionId]] = {}
+        for area_id, area in study_data.config.areas.items():
+            for reserve in area.reserves:
+                url = _reserve_need_matrix_path(area_id, ReserveDefinitionId(reserve.id))
+                node = study_data.tree.get_node(url)
+                assert isinstance(node, MatrixNode)
+                # Skip reserves whose matrix file was never saved (or has been deleted): neither
+                # the `.link` symlink nor the raw `.txt` file exists on disk.
+                if not node.is_normalized and not node.config.path.exists():
```
MartinBelthle: You shouldn't need to implement this logic as this should be handled in the `get_matrices_ids` method. And we don't have to do this in the other FS Daos

antarest/study/dao/file/file_study_reserve_definition_dao.py
```diff
@@ -124,3 +140,67 @@ def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[Rese
...
+        self.get_impl().save_matrices(matrices_mapping)
+
+    @override
+    def delete_reserve_need(self, area_id: AreaId, reserve_id: ReserveDefinitionId) -> None:
+        study_data = self.get_file_study()
+        path = _reserve_need_matrix_path(area_id, reserve_id)
+        try:
+            study_data.tree.delete(path)
+        except (ChildNotFoundError, KeyError):
+            pass
```
MartinBelthle, line 204: Shouldn't we raise in such case ?

MartinBelthle (changes requested):
We should add the matrix in the `StudyDatabaseMatrixUsageProvider` and add tests for it

antarest/study/dao/database/database_reserve_definition_dao.py
TheoPascoli: I tried removing it, but the integration test test_crud_on_reserve_need_matrix_via_raw fails.
Commands like CreateReserveDefinition pass URIs (matrix://...) coming from generator_matrix_constants. Without extract_matrix_id, the URI gets stored as-is, and then matrix_service.get(uri) throws a MatrixNotFound.

Other dao (thermal, link, renewable) likely have the same underlying issue, but it isn’t caught since there are no tests covering it.

antarest/study/dao/file/file_study_reserve_definition_dao.py
TheoPascoli: Like a said to copilot above, I kept it that way on purpose. Missing matrix means it was never created or already delete. I think its fine

antarest/study/dao/database/database_reserve_definition_dao.py
MartinBelthle: Okay I see. We can keep it inside this PR. I discussed the matter with Syvlain and it would be good to remove all these prefixes as it brings no value. I'll create a JIRA ticket to explain this

antarest/study/dao/file/file_study_reserve_definition_dao.py
sylvlecl: Indeed it's fine but we should do the same thing in DB DAO, which raises in that case.
We should also explain that expected behaviour in the docstring in the interface, it's part of the method contract, and a a small unit test for that :)

sylvlecl (changes requested):
1 minor inconsistency between DB anf FS to be resolved, then let's merge it

PR #3192, opened 2026-04-24
fix(xpansion): allow tree build for in-memory jobs

MartinBelthle:
The logic to know whether an Xpansion simulation failed or not has been decided by Thomas Bittar

### DIFF
--- a/antarest/study/storage/rawstudy/model/filesystem/config/files.py
+++ b/antarest/study/storage/rawstudy/model/filesystem/config/files.py
@@ -11,12 +11,12 @@
 # This file is part of the Antares project.
 
 import io
-import json
 import logging
 import re
 import tempfile
 import zipfile
 from collections.abc import Sequence
+from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
 from typing import Any, cast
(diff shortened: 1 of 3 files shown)


antarest/study/storage/rawstudy/model/filesystem/config/files.py
```diff
@@ -287,18 +284,28 @@ def parse_simulation_zip(path: Path) -> Simulation:
...
-    except FileNotFoundError:
-        return ""
-    except json.JSONDecodeError as exc:
-        raise XpansionParsingError(xpansion_json, f"invalid JSON format: {exc}") from exc
-    except KeyError as exc:
-        raise XpansionParsingError(xpansion_json, f"key '{exc}' not found in JSON object") from exc
+        version = str(obj["antares_xpansion"]["version"])
+    except Exception as e:
+        logger.warning(f"Error parsing xpansion output json file: {e}")
+        return None
```
TheoPascoli, line 305: Minor: `read_text()` raises if the file does not exist. You can remove your exist, the exception will be caught. Additionally, you could catch multiple specific errors, this `except Exception` is too broad in my opinion

TheoPascoli (changes requested):

antarest/study/storage/rawstudy/model/filesystem/config/files.py
MartinBelthle: Actually I don't want to put `read_text` with the existence test.
Because if the file simply does not exist it only means the Simulation was not an Xpansion one.

But if the `read_text` or the keys are missing we need to log the issue because it's an implicit contract with Xpansion that the file should be formatted as is. 

And I don't want to handle all possible exceptions, IMO logging them is enough as this case should never happen and if it does we'll check the logs

TheoPascoli (approved):

PR #3193, opened 2026-04-27
feat(backend_tablemode): finished implementing tablemode and testing

mehdiwahada:
What does it do ? : it implements the tablemodes, originally present in the front end, into the back end in order for the user to stock them, no matter the computer they're using and as an independent API service, not linked to any studies

Why it has been done ? : Tablemodes were, since Antares_web creation, always linked to the front end, making it difficult for the user to get their tablemode datas when connecting to another computer

How to test it ? :


antarest/tablemode/model.py
```diff
@@ -0,0 +1,84 @@
...
+    SPREAD_UNSUPPLIED_ENERGY_COST = "spreadUnsuppliedEnergyCost"
+    AVERAGE_SPILLED_ENERGY_COST = "averageSpilledEnergyCost"
+    SPREAD_SPILLED_ENERGY_COST = "spreadSpilledEnergyCost"
+    NON_DISPATCHABLE_POWER = "nonDispatchablePower"
+    DISPATCHABLE_HYDRO_POWER = "dispatchableHydroPower"
+    OTHER_DISPATCHABLE_POWER = "otherDispatchablePower"
+
+
+class TableMode(Base):
+    __tablename__ = "tablemode"
```
sylvlecl, line 197: We need to add the corresponding alembic migration

antarest/tablemode/model.py
```diff
@@ -0,0 +1,84 @@
...
+
+    table_id: Mapped[UUID] = mapped_column(Uuid(), primary_key=True, nullable=False)
+    table_name: Mapped[str] = mapped_column(String(255), nullable=False)
+    user_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("identities.id", name="fk_user_id", ondelete="CASCADE"), nullable=False
+    )
+    table_type: Mapped[str] = mapped_column(String(255), nullable=False)
```
sylvlecl: we'd better use an enum type

antarest/tablemode/model.py
```diff
@@ -0,0 +1,84 @@
...
+            table_name=self.table_name,
+            table_type=self.table_type,
+            table_columns=self.table_columns.split(","),
+        )
+
+
+class TableModeDTO(AntaresBaseModel, extra="forbid"):
+    table_id: UUID
+    table_name: str
+    table_type: str
```
sylvlecl: We should use the enum type you have defined

antarest/tablemode/model.py
```diff
@@ -0,0 +1,84 @@
...
+    table_columns: list[str]
```
sylvlecl: Here too we should use the enum type you have defined

antarest/tablemode/model.py
```diff
@@ -0,0 +1,84 @@
...
+    DISPATCHABLE_HYDRO_POWER = "dispatchableHydroPower"
+    OTHER_DISPATCHABLE_POWER = "otherDispatchablePower"
+
+
+class TableMode(Base):
+    __tablename__ = "tablemode"
+
+    table_id: Mapped[UUID] = mapped_column(Uuid(), primary_key=True, nullable=False)
+    table_name: Mapped[str] = mapped_column(String(255), nullable=False)
+    user_id: Mapped[int] = mapped_column(
```
sylvlecl, line 201: Don't forget to add an index on this column in the alembic migration, otherwise the queries will be slow

antarest/tablemode/repository.py
```diff
@@ -0,0 +1,63 @@
...
+        self._session = session
+
+    @property
+    def session(self) -> Session:
+        if self._session is None:
+            return db.session
+        return self._session
+
+    def get_all(self) -> Sequence[TableMode]:
+        stmt = select(TableMode).where(TableMode.user_id == get_user_impersonator())
```
sylvlecl, line 35: See other comments: don't forget to create an index on that column so that the request is fast

antarest/tablemode/web.py
```diff
@@ -0,0 +1,61 @@
...
+
+from antarest.core.utils.web import APITag
+from antarest.dependencies import TablemodeServiceDep, auth_required
+from antarest.tablemode.model import TableColumn, TableModeDTO, TableType
+
+
+def create_table_routes() -> APIRouter:
+    bp = APIRouter(prefix="/v1", tags=[APITag.table_mode], dependencies=[Depends(auth_required)])
+
+    @bp.post("/tablemode", summary="Create a Tablemode")
```
sylvlecl: The wording is a little weird:
the table "mode" is a feature that allows to display study data as tables.
But here what we create could be called more accurately just a table: the user defines a new table he wants to view.
--> let's use that wording in the summary instead, and in the method names too

antarest/tablemode/service.py
```diff
@@ -0,0 +1,104 @@
...
+    if table_type not in type_list:
+        raise HTTPException(status_code=http.HTTPStatus.BAD_REQUEST, detail=f"Invalid table type: {table_type}")
+
+    if not all(col in columns for col in table_columns):
+        raise HTTPException(
+            status_code=http.HTTPStatus.BAD_REQUEST, detail="Invalid table columns: " + ", ".join(table_columns)
+        )
+
+
+class TablemodeService:
```
sylvlecl: Let's use a capital M

antarest/tablemode/service.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(self, tablemode_repository: TablemodeRepository):
+        self.tablemode_repository = tablemode_repository
+
+    def get_tablemodes(self) -> list[TableModeDTO]:
```
sylvlecl: let's use just "table" instead of table mode in method names

antarest/tablemode/model.py
```diff
@@ -0,0 +1,84 @@
...
+    THERMAL = "thermals"
+    RENEWABLE = "renewables"
+    # Avoid "storages" because we may have "lt-storages" (long-term storages) in the future
+    ST_STORAGE = "st-storages"
+    # Avoid "constraints" because we may have other kinds of constraints in the future
+    BINDING_CONSTRAINT = "binding-constraints"
+    ST_STORAGE_ADDITIONAL_CONSTRAINTS = "st-storages-additional-constraints"
+
+
+class TableColumn(StrEnum):
```
sylvlecl: Actually there are many more available column types, see their definition in the variable TABLE_MODE_COLUMNS_BY_TYPE on the frontend side.

Note that the accepted column types depend on the table type !
We will need to validate that in table definitions we have consistent table types and tables columns

antarest/tablemode/service.py
```diff
@@ -0,0 +1,104 @@
...
+from uuid import UUID, uuid4
+
+from fastapi import HTTPException
+
+from antarest.login.utils import get_user_impersonator
+from antarest.tablemode.model import TableColumn, TableMode, TableModeDTO, TableType
+from antarest.tablemode.repository import TablemodeRepository
+
+
+def _check_valid_data(table_columns: list[TableColumn], table_type: str) -> None:
```
sylvlecl: See other comment: here we will need also to check that the column type is valid for the table type.
For example "nonDispatchPower" is not valid for table type LINK

antarest/tablemode/service.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Check the imput datas entered. Raise an exception if they don't match the values of the enums
+    Args:
+        table_columns: the input columns to check
+        table_type: type of the table to check
+
+    """
+    type_list = [e.value for e in TableType]
```
sylvlecl: I don't think you need those intermediate lists, no ?

antarest/tablemode/model.py
```diff
@@ -0,0 +1,84 @@
...
+    def to_dto(self) -> "TableModeDTO":
+        return TableModeDTO(
+            table_id=self.table_id,
+            table_name=self.table_name,
+            table_type=self.table_type,
+            table_columns=self.table_columns.split(","),
+        )
+
+
+class TableModeDTO(AntaresBaseModel, extra="forbid"):
```
sylvlecl, line 216: Since there is some cross-validation to make between table_type and table_columns, we could add a `@model_validator` which cross checks those fields

antarest/tablemode/model.py
```diff
@@ -0,0 +1,84 @@
...
+class TableMode(Base):
+    __tablename__ = "tablemode"
+
+    table_id: Mapped[UUID] = mapped_column(Uuid(), primary_key=True, nullable=False)
+    table_name: Mapped[str] = mapped_column(String(255), nullable=False)
+    user_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("identities.id", name="fk_user_id", ondelete="CASCADE"), nullable=False
+    )
+    table_type: Mapped[str] = mapped_column(String(255), nullable=False)
+    table_columns: Mapped[str] = mapped_column(String(255), nullable=False)
```
sylvlecl: We should not have a limit on the length of the string here, it will possibly be pretty long

antarest/tablemode/web.py
sylvlecl: Would be better to return a status code 201 instead of 200 which is the default

sylvlecl (changes requested):
A few changes required:
- adding alembic
- adding missing enum values and cross validation
- adding stronger typing
- renaming some methods

antarest/tablemode/web.py
```diff
@@ -0,0 +1,59 @@
...
+        table_name: Annotated[str, Body(), Field(min_length=1, description="Table name cannot be empty")],
+        table_type: Annotated[TableType, Body(), Field(description="Table type cannot be empty")],
+        table_columns: Annotated[
+            list[TableColumn], Body(), Field(min_length=1, description="At least one column is required")
+        ],
+    ) -> TableModeDTO:
+
+        return tablemode_service.add_table(table_name=table_name, table_type=table_type, table_columns=table_columns)
+
+    @bp.get("/tablemode", summary="Get Tablemodes")
```
sylvlecl: let's be consistent in the description and use "table" everywhere instead of "tablemode". Better also to adapt the names of the methods consistently

antarest/tablemode/model.py
```diff
@@ -0,0 +1,216 @@
...
+    table_name: str
+    table_type: TableType
+    table_columns: list[TableColumn]
+
+    @model_validator(mode="after")
+    def check_coherence(self) -> "TableModeDTO":
+        # checking the table columns match the table_type
+        if self.table_type in TABLE_TYPE_COLUMN_MAPPING:
+            columns = TABLE_TYPE_COLUMN_MAPPING[self.table_type]
+            valid_values = [column.value for column in columns]
```
sylvlecl: Let's use a set here, it's always better to check if a set contains an item, instead of a list (which has O(N) complexity).

alembic/versions/rp986cf862cy_add_tablemode_table.py
```diff
@@ -0,0 +1,40 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('tablemode',
+        sa.Column('table_id', sa.Uuid(), nullable=False),
+                sa.Column('table_name', sa.String(length=255), nullable=False),
+                sa.Column('user_id', sa.Integer(), nullable=False, index=True),
+                sa.Column('table_type', sa.String(length=255), nullable=False),
+                sa.Column('table_columns', sa.String(), nullable=False),
+                sa.PrimaryKeyConstraint('table_id', 'user_id', name="pk_table_id"),
```
sylvlecl: Here you define the PK on both table_id and user_id, while in the code it's only defined on table_id.
This will cause incorrect behaviour of sqlalchemy because it will believe that only table_id is part of the PK.

It makes sense to keep only the table_id as primary key, since the UUID will be unique accross users.

But don't forget to add an index on user_id so that queries on the ID perform well.

Also we should have a FK constraint and cascade DELETE on user_id.

antarest/tablemode/model.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+class STStorageAdditionalConstraintsColumn(StrEnum):
+    # Since v9.2
+    VARIABLE = "variable"
+    OPERATOR = "operator"
+    ENABLED = "enabled"
+
+
+TableColumn = AreaColumn | LinkColumn | ThermalColumn | RenewableColumn | STStorageColumn | BindingConstraintColumn
```
sylvlecl: STStorageAdditionalConstraintsColumn seems missing.

Also, you can add "TypeAlias" so that TableColumn is correcly recognized as a type alias by IDEs (it's only a small helper):
```python
TableColumn: TypeAlias = ..
```

antarest/tablemode/model.py
```diff
@@ -0,0 +1,216 @@
...
+    TableType.LINK: LinkColumn,
+    TableType.THERMAL: ThermalColumn,
+    TableType.RENEWABLE: RenewableColumn,
+    TableType.ST_STORAGE: STStorageColumn,
+    TableType.BINDING_CONSTRAINT: BindingConstraintColumn,
+    TableType.ST_STORAGE_ADDITIONAL_CONSTRAINTS: STStorageAdditionalConstraintsColumn,
+}
+
+
+class StorageAdditionalConstraintColumn(StrEnum):
```
sylvlecl: duplicated

alembic/versions/rp986cf862cy_add_tablemode_table.py
```diff
@@ -0,0 +1,40 @@
+"""rawstudies_folder_info
```
sylvlecl: The header is lying :)

alembic/versions/rp986cf862cy_add_tablemode_table.py
```diff
@@ -0,0 +1,40 @@
...
+
+Revision ID: 63ed81e5ce6f
+Revises: ab143a39c4db
+Create Date: 2021-09-27 18:23:44.909595
```
sylvlecl: This too should be updated.

Note that for revision creation, you can use the `alembic revision` command which should ensure such metadata is correct

antarest/tablemode/model.py
```diff
@@ -0,0 +1,216 @@
...
+    table_type: TableType
+    table_columns: list[TableColumn]
+
+    @model_validator(mode="after")
+    def check_coherence(self) -> "TableModeDTO":
+        # checking the table columns match the table_type
+        if self.table_type in TABLE_TYPE_COLUMN_MAPPING:
+            columns = TABLE_TYPE_COLUMN_MAPPING[self.table_type]
+            valid_values = [column.value for column in columns]
+            for column in self.table_columns:
```
sylvlecl: Maybe to provide an even better error message, which will mention all the wrong columns at once, you can use set operations:
```python
valid_values = {column.value for column in columns}
wrong_values = set(self.table_columns) - valid_values
if wrong_values:
    raise ValueError(f"Invalid columns {','.join(wrong_values)} for table type {self.table_type}")

```

tests/integration/tablemode_blueprint/test_tablemode_blueprint.py
```diff
@@ -0,0 +1,132 @@
...
+
+    assert res_post.status_code == 201
+    assert json_post["table_name"] == "my_table"
+    assert json_post["table_type"] == TableType.AREA
+    assert json_post["table_columns"] == [AreaColumn.DISPATCHABLE_HYDRO_POWER]
+
+    # success : updating a tablemode's table_type
+    res_put = client.put(
+        f"/v1/tablemode/{str(table_id)}",
+        json={"table_type": TableType.AREA, "table_columns": [AreaColumn.DISPATCHABLE_HYDRO_POWER]},
```
sylvlecl: here you don't actually update the data, you should put different data from the one that was already present

tests/integration/tablemode_blueprint/test_tablemode_blueprint.py
```diff
@@ -0,0 +1,132 @@
...
+            "table_columns": [AreaColumn.NON_DISPATCHABLE_POWER, AreaColumn.DISPATCHABLE_HYDRO_POWER],
+        },
+    )
+    assert res_put.status_code == 200
+    assert res_put.json()["table_id"] == table_id
+    assert res_put.json()["table_name"] == "my_table"
+    assert res_put.json()["table_type"] == TableType.AREA
+    assert res_put.json()["table_columns"] == [AreaColumn.NON_DISPATCHABLE_POWER, AreaColumn.DISPATCHABLE_HYDRO_POWER]
+
+    # success : deleting a tablemode and checking it doesn't exist anymore
```
sylvlecl, line 82: Indeed let's check that it does not appear anymore in the list of tables by issuing a new GET /tablemode :)

antarest/tablemode/service.py
```diff
@@ -0,0 +1,103 @@
...
+
+    """
+
+    column_enum = TABLE_TYPE_COLUMN_MAPPING[table_type]
+    valid_values = {col.value for col in column_enum}
+
+    invalid_columns = [col for col in table_columns if col not in valid_values]
+    if invalid_columns:
+        raise HTTPException(
+            status_code=http.HTTPStatus.BAD_REQUEST,
```
sylvlecl: we'd better issue a 422 for consistency with pydantic validation

antarest/tablemode/service.py
```diff
@@ -0,0 +1,103 @@
...
+    Args:
+        table_columns: the input columns to check
+        table_type: type of the table to check
+
+    """
+
+    column_enum = TABLE_TYPE_COLUMN_MAPPING[table_type]
+    valid_values = {col.value for col in column_enum}
+
+    invalid_columns = [col for col in table_columns if col not in valid_values]
```
sylvlecl: See other comment on the pydantic validator:
we'd better use set operations here for efficiency and readability.
Since the logic is really the same, we could also share that code in a function called in both places

tests/tablemode/test_tablemode.py
```diff
@@ -0,0 +1,224 @@
...
+    TableMode,
+    TableModeDTO,
+    TableType,
+    ThermalColumn,
+)
+from antarest.tablemode.repository import TablemodeRepository
+from antarest.tablemode.service import TableModeService
+
+
+class TestTablemode:
```
sylvlecl: In test files, we now prefer to not use classes, it does have any added value : let's delete this and keep plain test functions

tests/integration/tablemode_blueprint/test_tablemode_blueprint.py
```diff
@@ -0,0 +1,132 @@
...
+
+import uuid
+
+from starlette.testclient import TestClient
+
+from antarest.tablemode.model import AreaColumn, STStorageColumn, TableType
+
+
+def test_tablemode_success(client: TestClient, admin_access_token: str) -> None:
+    client.headers = {"Authorization": f"Bearer {admin_access_token}"}
```
sylvlecl: You can use the fixture `admin_client` to not have to do this every time

tests/integration/tablemode_blueprint/test_tablemode_blueprint.py
```diff
@@ -0,0 +1,132 @@
...
+# This file is part of the Antares project.
+
+import uuid
+
+from starlette.testclient import TestClient
+
+from antarest.tablemode.model import AreaColumn, STStorageColumn, TableType
+
+
+def test_tablemode_success(client: TestClient, admin_access_token: str) -> None:
```
sylvlecl: We should cover all possible table types with tests:
for each of the 7 possible table type we need to have at least one passing test and one failing test, so that we make sure that the column validation correspond correctly to the table type.
(OK to not test ALL possible table columns though ;) )

sylvlecl (changes requested):
Nice, we're almost there:
still a few comments to take into account
- mainly the migration which is not consistent with the code
- then better test coverage
- other comments are mainly code quality "details"

sylvlecl (approved):
Looks good, thanks :)

PR #3195, opened 2026-04-29
feat(ui-study): specific home view for archived study

skamril:
- block access to archived studies and redirect to home
- remove duplicate call for fetching variant tree


webapp/src/routes/_authenticated/studies/-components/StudiesList/StudyCardCell/GridStudyCard.tsx
```diff
@@ -119,7 +119,7 @@ function GridStudyCard({
                 <Chip label={study.workspace} size="small" sx={chipSx} />
               )}
               {study.archived && (
-                <Chip label={t("studies.archivedStudy")} size="small" color="warning" sx={chipSx} />
+                <Chip label={t("studies.archived")} size="small" color="warning" sx={chipSx} />
```
hdinia: use `study.archived` to keep the same label

webapp/src/redux/middlewares/tanstackQueryMiddleware.ts
hdinia, line 1: the middleware should not react to archive/unarchive too ?

Since `Study.archived` is part of every tree node cached variant trees will silently go stale after archive/unarchive. 
You can promote these to Redux thunks so the middleware can catch them, or call invalidateQueries directly at the call site.

webapp/src/queries/studies/keys.ts
```diff
@@ -18,4 +18,8 @@ export const studyKeys = {
   favorites: () => [...studyKeys.all(), "favoriteStudies"] as const,
   createFavorite: () => [...studyKeys.favorites(), "createFavoriteStudy"] as const,
   deleteFavorite: () => [...studyKeys.favorites(), "deleteFavoriteStudy"] as const,
+  // Variants
+  allVariants: () => [...studyKeys.all(), "variants"] as const,
+  allVariantTree: () => [...studyKeys.allVariants(), "variantTree"] as const,
+  variantTree: (studyId: string) => [...studyKeys.allVariantTree(), { studyId }] as const,
```
hdinia: Why are variant keys nested under studyKeys rather than having their own domain directory? Every other study sub-resource (areas, bindingConstraints, storages) follows the pattern queries/<domain>/keys.ts with an independent key factory. Beyond the convention mismatch, nesting here means invalidateQueries({ queryKey: studyKeys.all() }) would inadvertently bust all variant tree caches.

Suggest extracting to queries/variants/keys.ts + queries/variants/queries.ts with a root key of ["variants"], and updating the middleware to use variantKeys.allTree(). Would keep it consistent and make cache invalidation explicit.

webapp/src/routes/_authenticated/studies/$studyId/index.tsx
```diff
@@ -28,32 +28,22 @@ export const Route = createFileRoute("/_authenticated/studies/$studyId/")({
...
-
-  const variantTreeRes = usePromise(async () => {
-    const parents = await getVariantParents(study.id);
-    const root = parents.length > 0 ? parents[parents.length - 1] : study;
-    const variantTree = await getVariantTree(root.id);
-
-    return variantTree;
-  }, [study]);
+  // `isFetching` prevents rendering an outdated tree during background refetches
+  const { data: variantTree, isFetching } = useSuspenseQuery(studyQueries.variantTree(study.id));
```
hdinia: Replacing the entire tree pane with `<SimpleLoader />` whenever isFetching is true means every invalidation-triggered refetch (rename, create, delete) will flash the left panel to a spinner. This is a worse UX than showing the slightly-stale tree while the fresh data loads, which is the standard stale-while-revalidate behaviour React Query provides for free. Unless there's a specific reason to block on fresh data here, removing the isFetching branch and always rendering `<VariantsTree>` would be smoother

webapp/src/services/api/studies/variants/index.ts
hdinia, line 1: I think it was already in the legacy code, 2 synthax mixed here (export async function vs export const = async)

(optional) for consistency maybe you can add docs too

webapp/src/routes/_authenticated/studies/$studyId/-components/InformationView/index.tsx
```diff
@@ -12,114 +12,128 @@
...
+                onClick={handleUnarchiveStudy}
+                startIcon={<UnarchiveOutlinedIcon />}
+              >
+                {t("global.unarchive")}
+              </Button>
+            ) : (
+              <Button variant="contained" onClick={handleLaunchStudy} startIcon={<BoltIcon />}>
+                {t("global.launch")}
+              </Button>
+            )}
```
hdinia, line 133: Is this legacy code that have been moved ? 
Minor: an early return for the archived case would be cleaner. Zero conditionals in the main return, each state is completely explicit

hdinia (approved):

webapp/src/routes/_authenticated/studies/$studyId/-components/NavHeader/index.tsx
hdinia, line 1: `getVariantParents` is called here to extract parentStudy, but variantTree (already in cache from useSuspenseQuery) contains the full ancestor chain, the direct parent is just the node whose children includes study.id. This triggers a redundant /parents request on every study page load.

Suggest adding a `findParentInTree` utility alongside `findNodeInTree` in services/utils/index.ts and replacing the usePromise + getVariantParents call with a synchronous tree walk. The WebSocket listener can then call queryClient.invalidateQueries on the variant tree key instead of reloadStudyMetadata. Result: one /parents call per load instead of two, and usePromise is removed from this component entirely.

```ts 
  const { data: variantTree } = useSuspenseQuery(studyQueries.variantTree(study.id));
  const isExplorer = !!match;
  const parentStudy = findParentInTree(study.id, variantTree);
  const tree = findNodeInTree(study.id, variantTree);
  const variantNb = tree ? countDescendants(tree) : 0;

```

something like this should work and eliminate the double call to /parents

skamril: Nice refactor! I removed the WebSocket listener also, tanstackQueryMiddleware already does the work.

webapp/src/routes/_authenticated/studies/$studyId/index.tsx
skamril: Optimistic updates would be the cleanest fix, but we can't always get the tree cache from a study ID, so that's not really an option here.
That said, dropping the `isFetching` guard means the user could see a stale tree with no sign it's outdated, which is arguably worse than a spinner.
A middle ground: keep the tree rendered but add opacity and block user clicks while `isFetching` is true. Shows something is happening without wiping the whole panel.

webapp/src/redux/middlewares/tanstackQueryMiddleware.ts
skamril: Only the name and ID of studies are used from the tree, which is why I only invalidate tree queries when a name changes.

Otherwise, I also have to invalidate the requests for any study updates.

hdinia (dismissed):

webapp/src/redux/middlewares/tanstackQueryMiddleware.ts
hdinia: ok

hdinia (dismissed):

hdinia (approved):

PR #3196, opened 2026-04-29
fix(download): use short transactions when polling for download readiness

sylvlecl:
ANT-4939

Using a long transaction for polling block un-necessarily a DB session for a long time.

It happens to cause deadlocks from time to time with `set_ready`, probably because
of a bug in our old version of pgpool.

### DIFF
--- a/antarest/core/filetransfer/service.py
+++ b/antarest/core/filetransfer/service.py
@@ -15,7 +15,6 @@ import http
 import logging
 import os
 import tempfile
-import time
 import uuid
 from pathlib import Path
 
@@ -33,11 +32,16 @@ from antarest.core.filetransfer.repository import FileDownloadRepository
 from antarest.core.interfaces.eventbus import Event, EventType, IEventBus
 from antarest.core.model import PermissionInfo, PublicMode
 from antarest.core.requests import UserHasNotPermissionError
-from antarest.core.tasks.service import DEFAULT_AWAIT_MAX_TIMEOUT
+from antarest.core.utils.fastapi_sqlalchemy import db
 from antarest.core.utils.utils import current_time
+from antarest.core.utils.wait import wait_for
 from antarest.login.utils import get_current_user, require_current_user
 
 logger = logging.getLogger(__name__)
+
+
+DEFAULT_DOWNLOAD_POLLING_INTERVAL = 2.0
+DEFAULT_DOWNLOAD_WAIT_TIMEOUT = 3600
 
 
 class FileTransferManager:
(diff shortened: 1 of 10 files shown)


antarest/core/filetransfer/service.py
```diff
@@ -33,13 +32,18 @@
...
+from antarest.core.utils.fastapi_sqlalchemy import db
 from antarest.core.utils.utils import current_time
+from antarest.core.utils.wait import wait_for
 from antarest.login.utils import get_current_user, require_current_user
 
 logger = logging.getLogger(__name__)
 
 
+DEFAULT_DOWNLOAD_POLLING_INTERVAL = 2.0
+DEFAULT_DOWNLOAD_WAIT_TIMEOUT = 3600
```
sylvlecl, line 44: I shortened the timeout from 2d to 1h.
2d seems really a too long time to hold a session just for waiting,
and 1h should be enough in all normal cases.

antarest/core/utils/wait.py
```diff
@@ -0,0 +1,43 @@
...
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+import time
+from typing import Callable, TypeVar
+
+T = TypeVar("T")
+
+
+def wait_for(fn: Callable[[], T | None], polling_interval: float, timeout: float) -> T | None:
```
sylvlecl, line 18: Extracted that utility function that we could reuse elsewhere

antarest/core/filetransfer/service.py
```diff
@@ -203,28 +207,40 @@ def fetch_download(self, download_id: str) -> FileDownload:
...
+        polling_interval: float = DEFAULT_DOWNLOAD_POLLING_INTERVAL,
+        timeout: float = DEFAULT_DOWNLOAD_WAIT_TIMEOUT,
+    ) -> FileDownloadDTO:
+        """
+        Important:
+        we use short scoped sessions here to guarantee that sessions are not held for a long time.
+        """
+
+        def get_download_if_completed() -> FileDownloadDTO | None:
+            with db():
```
sylvlecl, line 223: This is the important part:
we now use a short lived transaction to get the download

tests/core/utils/test_db_session_middleware.py
```diff
@@ -9,59 +9,29 @@
...
 from sqlalchemy import create_engine, text
-from sqlalchemy.event import listens_for
-from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import Session, sessionmaker
 
 from antarest.core.utils.fastapi_sqlalchemy import db
-
-
-@dataclass
-class EventCounter:
```
sylvlecl, line 23: Moved that utility to `test_helpers` so that it is re-used by new test

antarest/core/filetransfer/web.py
```diff
@@ -53,7 +55,9 @@ def fetch_download(filetransfer_manager: FileTransferManagerDep, download_id: Uu
     def get_download_metadata(
         filetransfer_manager: FileTransferManagerDep,
         download_id: UuidStr,
-        wait_for_availability: bool = False,
+        wait_for_availability: Annotated[
+            bool, Query(description="If true, will wait for the download to be either ready or failed (max 1h).")
```
sylvlecl, line 59: Just added a description with the timeout

antarest/core/utils/wait.py
```diff
@@ -0,0 +1,43 @@
...
+    while time.time() < end:
+        result = fn()
+        if result is not None:
+            return result
+        time.sleep(polling_interval)
+
+    result = fn()
+    if result is not None:
+        return result
+    return None
```
smailio: why not just return fn() ?

antarest/core/filetransfer/service.py
```diff
@@ -203,28 +207,40 @@ def fetch_download(self, download_id: str) -> FileDownload:
...
                 status_code=http.HTTPStatus.UNPROCESSABLE_ENTITY,
-                detail=f"File was not successfully processed: {download.error_message}.",
+                detail=f"File was not successfully processed: {download_dto.error_message}.",
             )
 
-        if not download.ready:
-            raise HTTPException(status_code=http.HTTPStatus.EXPECTATION_FAILED, detail="File is still in process.")
-
-        return download.to_dto()
+        return download_dto
```
smailio, line 246: for the case where get_download_if_completed returned the first time, we don't check the ready attribute anymore, is this intentional ?

antarest/core/utils/wait.py
sylvlecl: ah yes it looks weird, I will change it, it was because it's not the same "None" ^^

antarest/core/filetransfer/service.py
sylvlecl: Yes: the attribute is checked inside the function itself:
if the download is not ready, it will return None and still raise that same exception.

If it's ready (and not failed), it will return a non-None DTO which will be returned to the client

tests/test_helpers/db.py
```diff
@@ -0,0 +1,50 @@
...
+        counter.commit = counter.commit + 1
+
+    @listens_for(session_factory, "after_transaction_create")
+    def on_transaction_start(session, transaction):
+        counter.transaction_start = counter.transaction_start + 1
+
+    @listens_for(session_factory, "after_transaction_end")
+    def on_transaction_end(session, transaction):
+        counter.transaction_end = counter.transaction_end + 1
+
```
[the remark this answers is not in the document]
sylvlecl: it's ok, the session factories are never shared between unit tests

tests/core/test_file_transfer.py
```diff
@@ -85,3 +89,40 @@ def test_lifecycle(tmp_path: Path) -> None:
...
+        assert download_dto.id == download.id
+        assert download_dto.ready
+
+        # We check there's been only 1 commit (set ready)
+        # and that the polling used short transactions (begin/rollback each time)
+        assert events_counter.commit == 1
+        assert events_counter.begin >= 8
+        assert events_counter.rollback >= 8
+
+        thread.join()
```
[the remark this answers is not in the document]
sylvlecl: it's OK, the test will fail anyway in that case, and the thread terminate shortly after

smailio (approved):

PR #3199, opened 2026-04-30
fix(db-dao): fix enum types definition

sylvlecl:
- postgresql has case sensitive enums. sqlalchemy uses enum names, not values, for enums representation
  --> currently, it does not work because in alembic migration we have used enum values, which use a different
        case


antarest/study/dao/database/sql_utils.py
```diff
@@ -14,16 +14,28 @@
...
-from sqlalchemy import Column, Table
+from sqlalchemy import Column, Enum, Table
 from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.dialects.postgresql.dml import Insert as PgInsert
 from sqlalchemy.dialects.sqlite import insert as sqlite_insert
 from sqlalchemy.dialects.sqlite.dml import Insert as SqliteInsert
 from sqlalchemy.orm import Session
 
 
+def enum_col(enum_class: Type[PyEnum], **kwargs: Any) -> Enum:
```
sylvlecl, line 28: Addition of that utility method to create enums that are serialized as their value to DB, instead of their name

antarest/study/dao/database/database_study_settings_dao.py
```diff
@@ -80,7 +80,7 @@ def get_impl(self) -> "DatabaseStudyDao":
 
     @override
     def save_general_config(self, config: GeneralConfig) -> None:
-        values = dict(study_id=self.get_study_id(), **config.model_dump())
+        values = dict(study_id=self.get_study_id(), **config.model_dump(mode="json"))
```
smailio: what this has to do with the enum thing ?

TheoPascoli (dismissed):

antarest/study/dao/database/database_study_settings_dao.py
sylvlecl: nothing actually, well spotted, I will remove it

PR #3200, opened 2026-05-04
fix(ui-matrix): fix column header visibility on grouped matrix columns

hdinia:


webapp/src/components/Matrix/components/MatrixGrid/index.tsx
```diff
@@ -31,15 +31,105 @@ import { useMatrixPasteInterceptor } from "../../hooks/useMatrixPasteInterceptor
...
 
+// Fonts match baseFontStyle / headerFontStyle + fontFamily from Matrix/styles.ts.
+const CELL_FONT = "13px Inter, sans-serif";
+const HEADER_FONT = "bold 11px Inter, sans-serif";
+// 2 × cellHorizontalPadding (8 px each side) + 4 px safety margin.
+const CELL_CONTENT_PADDING = 20;
+
+let _canvas: HTMLCanvasElement | null = null;
+
+function measureTextWidth(text: string, font: string): number {
```
skamril: Therei is `measureTextWidth` in domUtils.ts

webapp/src/components/Matrix/components/MatrixGrid/index.tsx
```diff
@@ -31,15 +31,105 @@ import { useMatrixPasteInterceptor } from "../../hooks/useMatrixPasteInterceptor
...
+function computeColumnWidths(
+  newData: number[][],
+  columns: readonly EnhancedGridColumn[],
+  gridToData: (item: Item) => Item | null,
+): Map<string, number> {
+  const updates = new Map<string, number>();
+
+  const aggregateTypes = columns
+    .filter((col) => col.type === Column.Aggregate)
+    .map((col) => col.id as AggregateType);
```
skamril: 💡 `R.filter` prevents using `as`

webapp/src/components/Matrix/components/MatrixGrid/index.tsx
```diff
@@ -212,7 +315,7 @@ function MatrixGrid({
     gridSelection,
     gridToData,
     getDataRowIndex,
-    onBulkPaste,
+    onBulkPaste: onBulkPaste ? handleBulkPaste : undefined,
```
skamril, line 228: Can you move the guard inside `handleBulkPaste` instead of using a ternary at the call site:

```ts
if (!onBulkPaste) { return; }
```

This makes the relationship between `onBulkPaste` and `handleBulkPaste` explicit, since the function body doesn't actually seem to need `onBulkPaste` to work.

webapp/src/components/Matrix/components/MatrixGrid/index.tsx
```diff
@@ -31,15 +31,105 @@ import { useMatrixPasteInterceptor } from "../../hooks/useMatrixPasteInterceptor
...
+    return 0;
+  }
+
+  ctx.font = font;
+  return ctx.measureText(text).width;
+}
+
+// Computes which columns need to grow to fit the pasted content.
+// Handles both data (Column.Number) and aggregate (Column.Aggregate) columns.
+function computeColumnWidths(
```
skamril: It's a large function, could we move `computeColumnWidths` to a utils file?

webapp/src/components/DataGrid.tsx
```diff
@@ -54,22 +58,25 @@ function isStringRowMarkerOptions(
...
-  onColumnResizeStart,
-  onColumnResizeEnd,
-  onGridSelectionChange,
-  enableColumnResize = true,
-  readOnly = false,
-  freezeColumns,
-  rows,
-  ...rest
-}: DataGridProps) {
+const DataGrid = forwardRef<DataGridHandle, DataGridProps>(function DataGrid(
```
skamril: In React 19, `forwardRef` is no longer necessary. Pass `ref` as a prop instead.

webapp/src/components/Matrix/components/MatrixGrid/index.tsx
```diff
@@ -202,6 +292,19 @@ function MatrixGrid({
     gridToData,
   });
 
+  const dataGridRef = useRef<DataGridHandle>(null);
+
+  const handleBulkPaste = useCallback(
+    (newData: number[][]) => {
+      onBulkPaste?.(newData);
+      const updates = computeColumnWidths(newData, columns, gridToData);
+      if (updates.size > 0) {
```
skamril, line 211: You could transform `visibleColumns` into a state and update widths here with `setVisibleColumns`, which would make the `dataGridRef.current?.resizeColumns` abstraction unnecessary.

webapp/src/components/DataGrid.tsx
```diff
@@ -82,6 +89,35 @@ function DataGrid({
     columns: CompactSelection.empty(),
   });
 
+  // Expose an imperative resize method so callers (e.g. paste handler) can grow columns
+  // without triggering a full columnsFromProps reset that would discard user resizes.
+  useImperativeHandle(ref, () => ({
+    resizeColumns(updates: Map<string, number>) {
```
skamril, line 94: There is no need to use a new function for this, updating the widths in `columns` prop is equivalent.

See https://github.com/AntaresSimulatorTeam/AntaREST/pull/3200#discussion_r3180338107

skamril (changes requested):

webapp/src/components/Matrix/components/MatrixGrid/index.tsx
hdinia: this is a minor improvement

webapp/src/components/Matrix/components/MatrixGrid/index.tsx
hdinia: visibleColumns is derived from columns via useMemo. If it becomes state, it breaks the single-source-of-truth. widths set via `setVisibleColumns` would be overwritten on the next render that re-derives from columns. The current ref-based imperative approach correctly mutates only the internal DataGrid column state, which survives prop resets. This is why the useImperativeHandle comment explicitly says "without triggering a full `columnsFromProps` reset that would discard user resizes

webapp/src/components/DataGrid.tsx
hdinia: This is incorrect. DataGrid maintains its own internal columns state `(line 86: const [columns, setColumns] = useState(initColumns))`. The `useUpdateEffect` at line 136 resets it from columnsFromProps whenever columnsFromProps changes. If you update widths in the columns prop from outside, the next re-render triggered by anything would reset them. The `useImperativeHandle` approach avoids that exact problem -> it mutates the internal state directly. The code comment at line 92-93 explicitly documents this

webapp/src/components/Matrix/components/MatrixGrid/index.tsx
hdinia: sure

webapp/src/components/Matrix/components/MatrixGrid/index.tsx
skamril: `without triggering a full columnsFromProps reset that would discard user resizes`

I forgot about it sorry.

webapp/src/components/Matrix/components/MatrixGrid/index.tsx
hdinia: The ternary is doing more than guarding the call, it controls whether the paste listener is registered at all. Look at `useMatrixPasteInterceptor:58`

skamril (approved):

PR #3202, opened 2026-05-04
chore(matrix): remove matrix prefixes and "URI mapper"

sylvlecl:
Half a fix in matrix IDs we store, half a refacto that simplifies our use of matrix IDs.

**The fix:**
- fixes the fact that we tried to store prefixed matrices in database, causing exceptions because
  we have limited matrix IDs to 64 chars.
- we now only use plain IDs in all the application, not "URIs", we only keep the prefix removal in 2 places for
  backwards compat:
    - in commands parsing
    - in link files parsing

**The refacto**:
This calls for the removal of the so called "URI mapper", since we don't use anymore the 
concept of URI. Link files storing matrix IDs are actually implementation details of the 
unique InputSeriesMatrix node type, so we move all that here:
- removes the intermediate `MatrixNode` node type, which only had one implementation,
  the `InputSeriesMatrix` node
- removes the so-called "URI mapper" for simplicity sake:
  - we never actually used the concept of "URI" for matrices, only plain IDs
  -  the separation between the "mapper" and the input series matrix node was 90% leaky,
     everything is now embedded inside it. We don"t leak anymore the concept of "link file",
     for example.


antarest/matrixstore/in_memory.py
```diff
@@ -36,8 +35,6 @@ def __init__(self) -> None:
     def add_predefined_matrix(self, matrix_factory: Callable[[], pl.DataFrame]) -> str:
         matrix_id = compute_hash(matrix_factory())
         self._predefined_matrices[matrix_id] = matrix_factory
-        # DAO methods support prefixed IDs, so we register the prefixed ID as well
```
sylvlecl, line 39: Was really weird to have to do such things

antarest/core/serde/np_array.py
```diff
@@ -37,3 +40,17 @@ def _np_to_list(array: npt.NDArray[np.float64]) -> list[float] | list[list[float
 
 
 NpArray: TypeAlias = Annotated[npt.NDArray[np.float64], PlainSerializer(_np_to_list), BeforeValidator(_list_to_np)]
+
+
+def imports_matrix_from_bytes(data: bytes) -> NpArray | None:
```
sylvlecl, line 45: moved from MatrixNode, it belongs here in serialization utils

antarest/matrixstore/service.py
```diff
@@ -141,27 +139,6 @@ def delete(self, matrix_id: str) -> None:
     def register_usage_provider(self, usage_provider: "IMatrixUsageProvider") -> None:
         raise NotImplementedError()
 
-    def get_matrix_id(self, matrix: list[list[float]] | str) -> str:
```
sylvlecl, line 144: did not really make sense anymore to have it here

antarest/matrixstore/service.py
```diff
@@ -81,8 +81,6 @@
 
 logger = logging.getLogger(__name__)
 
-MATRIX_PROTOCOL_PREFIX = "matrix://"
```
sylvlecl, line 84: Removed it from the service layer:
it's only now in lower level parsing layers for backwards compat

antarest/study/storage/rawstudy/model/filesystem/matrix/matrix_storage_context.py
```diff
@@ -0,0 +1,31 @@
...
+# This file is part of the Antares project.
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from antarest.matrixstore.service import ISimpleMatrixService
+
+
+@dataclass(frozen=True)
+class MatrixStorageContext:
```
sylvlecl, line 20: The input series matrix node now only depends on that and decides based on the is_managed flag.
Less abstraction, but the URIMapper abstraction was very very leaky anyway and was too much linked to the implementation of InputSeriesMatrix.

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
```diff
@@ -14,65 +14,226 @@
...
+def extract_matrix_id(uri: str) -> str:
     """
-    Generic node to handle input matrix behavior
+    Extract matrix ID from URL matrix://<id>.
+
+    Exists for backward compatibility reason, historically matrix IDs
+    have been written in link files with that prefix, we still
+    need to be able to parse them.
+    """
+    return uri.removeprefix(MATRIX_PROTOCOL_PREFIX)
```
sylvlecl: **Important change here:**
we always remove the possible prefix from the link file.

The reason is documented in comments.

antarest/study/storage/rawstudy/model/filesystem/matrix/input_series_matrix.py
```diff
@@ -14,65 +14,226 @@
...
         self.config.path = self.config.path.parent / self.config.path.name.removesuffix(".link")
         self.should_exist = should_exist
 
+    def _get_link_path(self) -> Path:
+        return self.config.path.parent / (self.config.path.name + ".link")
+
+    def _has_link(self) -> bool:
+        return self._get_link_path().exists()
+
+    def get_matrix_id(self) -> str | None:
```
sylvlecl, line 112: Methods of URIMapper and MatrixNode have been merged into that class, and renamed where relevant.

For example, this was `get_link_content` before, which exposed an implementation detail, it's now a more explciit `get_matrix_id`.

I tried to document in docstrings the role of each method.
They are still a bit messy to be honest, but cannot change too much

antarest/study/storage/rawstudy/model/filesystem/matrix/matrix.py
sylvlecl, line 1: Removed and merged into InputSeriesMatrix, which was the only subclass

antarest/study/storage/rawstudy/model/filesystem/matrix/output_series_matrix.py
```diff
@@ -39,8 +38,7 @@ def __init__(self, config: FileStudyTreeConfig, freq: MatrixFrequency):
 
     @override
     def get_lazy_content(self, url: list[str] | None = None, depth: int = -1, expanded: bool = False) -> str:
-        # noinspection SpellCheckingInspection
-        return add_matrix_id_prefix(self.config.path.name)
+        return f"matrix://{self.config.path.name}"
```
sylvlecl, line 41: Note that here the prefix is targetted at the front, which is the only consumer of that "lazy content".
It has nothing to do with the prefix with which the IDs were stored.

antarest/study/dao/file/file_study_constraint_dao.py
```diff
@@ -218,23 +216,22 @@ def _get_nodes_with_their_bc_id(self, term: str | None = None) -> dict[MatrixNod
...
-        matrices_mapping: dict[str, list[MatrixNode]] = {}
+        matrices_mapping: dict[str, list[InputSeriesMatrix]] = {}
 
         nodes_and_bc_ids = self._get_nodes_with_their_bc_id(term)
 
         for node, constraint_id in nodes_and_bc_ids.items():
             if constraint_id in series:
                 # We only want to save the series for given constraints and the method returned them all
                 series_id = series[constraint_id]
-                matrix_id = extract_matrix_id(series_id)
```
sylvlecl, line 236: Simplifies things quite a lot for DAO implementations: we d'ont need to handle any mysterious prefix.

tests/variantstudy/test_command_factory.py
```diff
@@ -650,8 +660,8 @@
         ),
         {
             "area_id": "area 1",
-            "inflows": "matrix://df9b25e1-e3f7-4a57-8182-0ff9791439e5",
```
sylvlecl, line 653: Checks that serialization round trip indeed removes the prefix.

antarest/study/storage/rawstudy/model/filesystem/factory.py
```diff
@@ -56,10 +54,10 @@ class StudyFactory:
 
     def __init__(
         self,
-        matrix_mapper_factory: MatrixUriMapperFactory,
+        matrix_service: ISimpleMatrixService,
```
sylvlecl, line 57: makes thinggs slightly more straight forward here

TheoPascoli (approved):

PR #3203, opened 2026-05-06
feat(binding-constraints)!: alleviate binding constraints in config by using BindingConstraintConfig

olivierPigeon-RTE:
**Breaking change:**
the /synthesis endpoint does not return any more any information about
binding constraints.

### DIFF
--- a/antarest/study/dtos.py
+++ b/antarest/study/dtos.py
@@ -16,12 +16,12 @@ from pydantic import ConfigDict
 from antarest.core.serde import AntaresBaseModel
 from antarest.core.utils.string import to_camel_case
 from antarest.output.storage.output_storage import OutputDetails, OutputStorageType
-from antarest.study.business.model.binding_constraint_model import BindingConstraint
 from antarest.study.business.model.config.general_model import Mode
 from antarest.study.business.model.district_model import District
 from antarest.study.model import StudyVersionInt
 from antarest.study.storage.rawstudy.model.filesystem.config.model import (
     AreaConfig,
+    BindingConstraintConfig,
     EnrModelling,
     FileStudyTreeConfig,
 )
@@ -64,7 +64,7 @@ class StudyDataSynthesis(AntaresBaseModel):
     version: StudyVersionInt
     districts: dict[str, District] = {}
     areas: dict[str, AreaConfig] = {}
-    bindings: list[BindingConstraint] = []
+    bindings: list[BindingConstraintConfig] = []
     enr_modelling: EnrModelling = EnrModelling.AGGREGATED
 
     @classmethod
@@ -97,7 +97,6 @@ class StudySynthesis(AntaresBaseModel):
     districts: dict[str, District] = {}
     areas: dict[str, AreaConfig] = {}
     outputs: dict[str, OutputSynthesis] = {}
-    bindings: list[BindingConstraint] = []
     enr_modelling: EnrModelling = EnrModelling.AGGREGATED
 
     @classmethod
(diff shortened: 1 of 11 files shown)


antarest/study/storage/rawstudy/model/filesystem/config/model.py
```diff
@@ -116,6 +122,34 @@ def get_file(self) -> str:
...
+    """
+    Object representing a binding constraint configuration.
+
+    Compared to the BindingConstraint class, this model does not have
+    the terms and comments fields to be lighter
+    """
+
+    model_config = ConfigDict(alias_generator=to_camel, extra="forbid", populate_by_name=True)
+
+    id: ConstraintId | None = None
```
sylvlecl: we'd better not allow None for all those fields, like in the `BindingConstraint` class: we are guaranteed they will always have a value. This will avoid to have to handle the None case elsewhere in the code

antarest/study/storage/rawstudy/model/filesystem/root/input/bindingconstraints/bindingcontraints.py
```diff
@@ -54,12 +58,12 @@ def build(self) -> TREE:
                 BindingConstraintFrequency.WEEKLY: default_bc_weekly_daily_86,
             }
             children: TREE = {
-                binding.id: InputSeriesMatrix(
+                str(binding.id): InputSeriesMatrix(
                     self.matrix_storage_context,
                     self.config.next_file(f"{binding.id}.txt"),
-                    freq=frequency_mapping[binding.time_step],
+                    freq=frequency_mapping[binding.time_step if binding.time_step else DEFAULT_TIMESTEP],
```
sylvlecl: Once `| None` are removed from the `BindingConstraintConfig` class, we will not need anymore to have the if/else.
Same in other places below

antarest/study/storage/rawstudy/model/filesystem/root/input/bindingconstraints/bindingcontraints.py
```diff
@@ -54,12 +58,12 @@ def build(self) -> TREE:
                 BindingConstraintFrequency.WEEKLY: default_bc_weekly_daily_86,
             }
             children: TREE = {
-                binding.id: InputSeriesMatrix(
+                str(binding.id): InputSeriesMatrix(
```
sylvlecl: Same, once we remove the Nones, we probably don't need anymore that `str(...)`

antarest/study/dtos.py
```diff
@@ -97,7 +97,6 @@ class StudySynthesis(AntaresBaseModel):
     districts: dict[str, District] = {}
     areas: dict[str, AreaConfig] = {}
     outputs: dict[str, OutputSynthesis] = {}
-    bindings: list[BindingConstraint] = []
```
sylvlecl, line 100: We should choose between:
- completely removing it from the synthesis: in that case we need to remove it also from the StudyDataSynthesis above, and from the corresponding model in the typescript code
- or leave it here for now

We can remove it completely in my opinion, indeed.

We need to document it as a breaking change in the release notes:
for that the convention is to use a `!` in PR title (and the merged commit) `feat(binding-constraints)!: ...`

sylvlecl (changes requested):
Looks good overall ! :) 
Requesting minor changes for simplification: not using None where not necessary, and push the cleanup of synthesis further and document the breaking change in PR title/description

sylvlecl (approved):
Thanks :)

PR #3207, opened 2026-05-12
feat(reserves): implement reserves for thermal clusters

TheoPascoli:


antarest/study/storage/rawstudy/model/filesystem/config/thermal_cluster_reserve_participation.py
```diff
@@ -0,0 +1,97 @@
...
+
+    The composite section name follows the convention ``<cluster_name><sep><reserve_id>``.
+    Splitting blindly on the separator would be ambiguous because both cluster and reserve
+    IDs may contain it; the cluster name from the content is therefore used as the
+    canonical anchor.
+
+    Returns ``None`` if the section name does not start with ``<cluster_name><sep>`` (which
+    happens for unrelated sections like ``[symmetries]``).
+    """
+    prefix = f"{cluster_name}{SECTION_SEPARATOR}"
```
sylvlecl: Unfortunately this is not how the format is specified !

The section is only the reserve name, then the thermal cluster ID is in the properties of that section.

This means that we will have duplicate section names, indeed, which is not possible with the INI reader such as implemented now.

So we will need to change the INI reader implementation to get something like a "multi-dict"

sylvlecl: We need to include a study in integration tests which will contain reserves, so that we ensure we know how to read actual data corresponding to the specified format.

antarest/study/storage/rawstudy/model/filesystem/root/input/thermal/cluster/area/reserves_ini.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Each section corresponds to a single ``(cluster, reserve)`` participation. The
+    section name uses the convention ``<cluster_id>__<reserve_id>``; this is required
+    because the standard INI parser merges duplicate section names, so a unique
+    composite key is used to address each participation independently.
+
+    Examples
+    --------
+
+    [gas_cluster__reserve_1]
```
sylvlecl: See other comment: this is not the specified format

antarest/study/dao/database/database_thermal_cluster_reserve_participation_dao.py
```diff
@@ -0,0 +1,166 @@
...
+                    select(_TABLE.c.reserve_id).where(
+                        (_TABLE.c.study_id == self._study_id)
+                        & (_TABLE.c.area_id == area_id)
+                        & (_TABLE.c.thermal_id == thermal_id)
+                    )
+                ).fetchall()
+            }
+            for rid in reserve_ids:
+                if rid not in existing:
+                    raise ThermalClusterReserveParticipationNotFound(area_id, thermal_id, rid)
```
sylvlecl: This will raise for the wrong ID: if and ID has indeed been deleted in the first request, it will not appear anymore in the second request, whereas it was correctly deleted.

sylvlecl (changes requested):
Unfortunately the parsing of INI files does not match the spec!

In thermal cluster participations files, we have repeated sections with the reserve name as the section name, not cluster__section. Then the cluster appears as a property in the section.

sylvlecl:
Spec has been heavily modified in favor of yaml files, which allow a better file structure.
Closed in favour of that new implementation in https://github.com/AntaresSimulatorTeam/AntaREST/pull/3280

PR #3210, opened 2026-05-13
feat(study): forbid invalid characters in output_suffix

olivierPigeon-RTE:


tests/launcher/test_model.py
```diff
@@ -228,3 +230,25 @@ def test_delete_job_result(self, db_session: Session) -> None:
...
+
+    def test_with_valid_output_suffix(self) -> None:
+        params = LauncherParametersDTO(output_suffix="qwertyQWERTY09&_-.")
+        self.assertEqual(params.output_suffix, "qwertyQWERTY09&_-.")
+
+    def test_output_suffix_should_should_not_contain_equals_char(self) -> None:
+        self.assertRaises(ValidationError, LauncherParametersDTO, output_suffix="test=foo")
+
+    def test_output_suffix_should_should_not_contain_slash_char(self) -> None:
+        self.assertRaises(ValidationError, LauncherParametersDTO, output_suffix="test/foo")
```
smailio: usually we use pytest for this 

```python

        with pytest.raises(ValidationError):
            LaunchParametersDTO(output_suffix="test/foo")
``

File: tests/launcher/test_model.py [hunk 2]
```diff
@@ -228,3 +230,25 @@ def test_delete_job_result(self, db_session: Session) -> None:
...
+
+    def test_output_suffix_should_should_not_contain_backslash_char(self) -> None:
+        self.assertRaises(ValidationError, LauncherParametersDTO, output_suffix=r"test\foo")
+
+    def test_output_suffix_should_should_not_contain_escaped_char(self) -> None:
+        self.assertRaises(ValidationError, LauncherParametersDTO, output_suffix=r"test\tfoo")
```
smailio: Minor, you can use parametrized tests instead of several similar tests

smailio (changes requested):

tests/launcher/test_local_launcher.py
```diff
@@ -87,7 +87,7 @@ def test_compute(tmp_path: Path, launcher_config: LocalConfig) -> None:
         xpansion_r_version=False,
         archive_output=False,
         auto_unzip=True,
-        output_suffix="",
+        output_suffix=None,
```
smailio: this is a valid suffix now so you can revert it no ?

tests/launcher/test_model.py
```diff
@@ -228,3 +230,23 @@ def test_delete_job_result(self, db_session: Session) -> None:
...
+
+    @pytest.mark.parametrize("valid_output_suffix", valid_output_suffixes)
+    def test_with_valid_output_suffixes(self, valid_output_suffix) -> None:
+        params = LauncherParametersDTO(output_suffix=valid_output_suffix)
+        assert params.output_suffix == valid_output_suffix
+
+    invalid_output_suffixes = ["/test", "test/foo"]
+
+    @pytest.mark.parametrize("invalid_output_suffix", invalid_output_suffixes)
+    def test_output_suffix_should_not_contain_slash_char(self, invalid_output_suffix) -> None:
```
smailio: you fortgot to rename this test :)

olivierPigeon-RTE: I did rename it, the output_suffix should not contain any slash, all the other characters are allowed as the study can be run safely

smailio: Right ! But I guess now it should be renamed as we reverted the regex

smailio (approved):

PR #3211, opened 2026-05-13
feat(variant): block snapshot generation when parent is archived

smailio:


antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -813,6 +834,23 @@ def copy(
         return dest_study
 
     def _safe_generation(self, metadata: VariantStudy, timeout: int = DEFAULT_AWAIT_MAX_TIMEOUT) -> None:
+        """
+        Ensure the variant snapshot exists on disk, generating it synchronously if needed.
```
sylvlecl, line 838: Great, thanks for the doc :)

sylvlecl (approved):

PR #3215, opened 2026-05-20
feat(ui-studies): add destination directory picker to import dialog

hdinia:


webapp/src/routes/-shared/components/studies/dialogs/ImportStudyDialog/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+  ////////////////////////////////////////////////////////////////
+  // Event Handlers
+  ////////////////////////////////////////////////////////////////
+
+  const handleImport: UploadDialogProps["onImport"] = async (file, onUploadProgress) => {
+    const directoryPath = toDirectoryPath(destination, directories);
+
+    await dispatch(createStudy({ file, onUploadProgress, directory: directoryPath })).unwrap();
+
+    const updatedDirectories = await refreshDirectoriesIfNeeded(
```
skamril, line 56: Why not include the refresh inside the hook directly? You can do then:

```ts
if (redirect) {
  await redirectToDestination(destination, updatedDirectories);
}
```

webapp/src/routes/-shared/components/studies/dialogs/ImportStudyDialog/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+  return (
+    <UploadDialog
+      open={open}
+      title={t("studies.importNewStudy")}
+      dropzoneText={t("studies.importHint")}
+      onCancel={onClose}
+      onImport={handleImport}
+      maxWidth="md"
+      fullWidth
+      PaperProps={{ sx: { height: "100%" } }}
```
skamril, line 80: It's not necessary to give a full height:

Image

webapp/src/components/dialogs/UploadDialog.tsx
```diff
@@ -114,7 +117,18 @@ function UploadDialog(props: UploadDialogProps) {
...
+          pt: 1,
+          display: "flex",
+          flexDirection: "column",
+          gap: 2,
+          // When extra content is provided, fill the available dialog content height
+          // so consumers can use a tall dialog without leaving empty space below.
+          ...(extraContent && { flex: 1, minHeight: 0 }),
+        }}
+      >
+        {extraContent}
```
skamril, line 131: `extraContent` should be at the end, because the main goal is to upload, the other field are optional.

Image

webapp/src/routes/-shared/components/studies/dialogs/ImportStudyDialog/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+      extraContent={
+        <>
+          <StudyDestinationFE
```
skamril, line 83: Can you add a label?

webapp/src/routes/-shared/components/studies/StudyDestinationFE/useRedirectToDestination.ts
```diff
@@ -0,0 +1,73 @@
...
+import { getDescendantIds } from "@/routes/_authenticated/studies/-components/StudyTree/ManagedTree/utils";
+import type { Directory } from "@/services/api/directories/types";
+import type { DirectoryDestination } from "./types";
+import { resolveRedirectDirectoryId } from "./utils";
+
+/**
+ * Invalidates and re-reads the directory cache when the destination created new
+ * sub-directories. Returns the original list unchanged when no refresh is needed.
+ */
+export async function refreshDirectoriesIfNeeded(
```
skamril: It would be more appropriate inside the utils file, or better yet, call it in the hook callback and remove the `export`.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/useRedirectToDestination.ts
```diff
@@ -0,0 +1,73 @@
...
+  destination: DirectoryDestination,
+  directories: Directory[],
+): Promise<Directory[]> {
+  if (!destination.newSubdirectoriesPath) {
+    return directories;
+  }
+
+  await queryClient.invalidateQueries({ queryKey: directoryQueries.list().queryKey });
+
+  return queryClient.getQueryData(directoryQueries.list().queryKey) ?? directories;
```
skamril: Why add `?? directories`? `queryClient.getQueryData()` shouldn't return a nil value.

skamril (changes requested):

webapp/src/components/dialogs/UploadDialog.tsx
skamril: I think you did it this way because uploading a file closes the dialog directly?
With extra fields this behavior is annoying, the dialog shouldn't close automatically anymore.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/useRedirectToDestination.ts
hdinia: getQueryData is typed `Directory[] | undefined`, so dropping the `??` produces a TS error. the function must return `Promise<Directory[]>`. You need some fallback. And ?? directories is actually better than ?? []

webapp/src/routes/-shared/components/studies/dialogs/ImportStudyDialog/index.tsx
hdinia: That's unnecessary here, it's obvious what the destination picker does a label will just take extra space for no benefit, and to be consistent with other dialogs that have this picker better not adding it.

webapp/src/components/dialogs/UploadDialog.tsx
hdinia: The intended flow is to select a destination first if needed, then upload. but since the destination is optional you can also just upload directly and the study lands at the root.

This is more of a UX concern than a code issue. If it turns out to be a problem in practice, users will report it and we can open a ticket to address it then. I don't think it's worth blocking the review over.

webapp/src/routes/-shared/components/studies/dialogs/ImportStudyDialog/index.tsx
hdinia: This is a minor styling point I'd consider out of scope. 

The full height is intentional as in prod the directories list is long, so the extra spacing helps the user and it stays consistent with our other dialogs that have the directories picker. Since it causes no problem I'd prefer to keep it.

webapp/src/routes/-shared/components/studies/dialogs/ImportStudyDialog/index.tsx
hdinia: The refresh/invalidation runs unconditionally, inside the callback it would only run when redirect is true. so unchecking "redirect" would leave the directory tree stale (new subdirectories not invalidated).

I'am not against doing like your example but it would imply a refactor, and affecting also the Move dialog, since it does not cause any problem I'll keep it like this for now, but thanks for the feedback.

webapp/src/routes/-shared/components/studies/StudyDestinationFE/useRedirectToDestination.ts
hdinia: Initially I wanted to call it in the hook, to encapsulate the refresh+redirect but due to `redirect` that I added later it would imply a refactor now. moving it to utils

webapp/src/routes/-shared/components/studies/dialogs/ImportStudyDialog/index.tsx
skamril: Not obvious for all users 🙂

skamril (dismissed):

skamril (approved):

PR #3220, opened 2026-05-21
feat(study): support copy, import and export on DB studies

MartinBelthle:
Tested on my local env and everything seems to work, I've even been able to run a simulation :smile:  (it failed due to output handling in the end)


antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -74,27 +75,34 @@ def __init__(
...
     @override
-    def copy(self, src_study: Study, dest_name: str, groups: list[str], destination_folder: PurePosixPath) -> RawStudy:
+    def copy(
+        self,
+        src_study: Study,
+        dest_name: str,
+        destination_folder: PurePosixPath,
+        owner: Identity,
+        groups: list[Group],
+        directory_id: str | None,
```
MartinBelthle: All these args could be a dataclass at this point.

antarest/study/web/studies_blueprint.py
```diff
@@ -269,7 +271,7 @@ def import_study(
         directory_path_sanitized = validate_folder_path(directory) if directory else ""
 
         try:
-            uuid = study_service.import_study(study.file, group_ids, directory=directory_path_sanitized)
+            uuid = study_service.import_study(study.file, group_ids, directory_path_sanitized, storage_mode)
```
sylvlecl, line 284: We need to check if the configuration allows the creation of DB studies (not enabled in production for now)

antarest/study/dao/study_conversion/study_converter.py
```diff
@@ -181,16 +177,20 @@ def _convert_areas(self) -> None:
...
+            self._new_dao.save_layer(layer)
+
         # Ui
         new_ui: AreaUiMapping = {}
         for area_id, source_ui in areas_ui.items():
             new_ui[area_id] = {}
-            for layer, x in source_ui.layer_x.items():
-                y = source_ui.layer_y[layer]
-                color = source_ui.layer_color[layer]
+            for layer_number, x in source_ui.layer_x.items():
```
sylvlecl: naming: maybe layer_id ?

antarest/study/dao/file/file_study_factory_dao.py
```diff
@@ -63,12 +63,15 @@ def _create_dao(self, metadata: StudyMetadataCreation) -> FileStudyTreeDao:
...
+        study_path = metadata.path or paths.study_path
 
         create_new_empty_study(version=metadata.version, path_study=study_path)
 
         is_study_managed = metadata.managed
-        file_study = self._study_factory.create_from_fs(study_path, is_study_managed, study_id, output_path)
+
+        # We don't want to use the cache as we're creating a new study.
+        # In particular, when launching a simulation for a DB study, the cache is filled with the old JOB path so we don't want to use it.
+        file_study = self._study_factory.create_from_fs(study_path, is_study_managed, study_id, output_path, False)
```
sylvlecl: It's not great that we don't write the study data to cache here, for file studies.

Maybe it would make sense to have a global "with_cache" flag in the FS DAO:
- for actual FS studies, we basically always want it
- for "temporary" FS studies, we will never want it

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -133,10 +134,18 @@ def __init__(
         }
 
     @override
-    def copy(self, src_study: Study, dest_name: str, groups: list[str], destination_folder: PurePosixPath) -> RawStudy:
+    def copy(
+        self,
+        src_study: Study,
+        dest_name: str,
+        destination_folder: PurePosixPath,
```
sylvlecl: It does not seem to make sense to have both the folder and the directory_id as separate args:
we should either have only one of them, or have them in a frozen dataclass, because they belong together

antarest/study/storage/utils.py
```diff
@@ -630,3 +640,4 @@ class StudyMetadataCreation:
     editor: str | None = None
     created_at: datetime | None = None
     updated_at: datetime | None = None
+    path: Path | None = None
```
sylvlecl: Kind of leaky abstraction to have it here only for a few use cases for the FS case.
We should at least document here what it will be used for. In particular, it could be confused with directories of managed studies.

Otherwise, we could just have a method not shared with the DA facotry interface which would have that specific argument, since it can only be used for FS studies

antarest/study/storage/database_storage.py
```diff
@@ -10,33 +10,45 @@
...
 
 class DatabaseStudyStorage(IStudyStorage):
-    def __init__(
-        self, config: Config, matrix_service: ISimpleMatrixService, generator_matrix_constants: GeneratorMatrixConstants
-    ):
+    def __init__(self, matrix_service: ISimpleMatrixService, study_dao_factories: dict[StorageMode, StudyFactoryDao]):
         self._matrix_service = matrix_service
-        self._generator_matrix_constants = generator_matrix_constants
-        self._config = config
+        self._dao_factories = study_dao_factories
```
sylvlecl: the 2 implementations are always referred to explicitly, it would be more clear to have 2 separate fields instead of having them in a dict

antarest/study/dao/file/file_study_factory_dao.py
sylvlecl: Not at all linked to that PR, but if we add a new class field, we could also fix this:
the FS DAO should not depend on a "command context", the concept of command is linked only to variants which is a completely independent thing.
Instead, the various services should be provided explicitly as constructor arguments

antarest/study/storage/database_storage.py
```diff
@@ -10,33 +10,45 @@
...
+        study_version = StudyVersion.parse(src_study.version)
+
+        # Build the new DB DAO
+        new_dao = self._dao_factories[StorageMode.DATABASE].get_study_dao(study_id=new_study.id, is_study_managed=True)
+
+        # Copies the inputs
+        converter = StudyConverter(
+            source_dao=source_dao, new_dao=new_dao, study_version=study_version, matrix_service=self._matrix_service
+        )
+        converter.convert_study_inputs()
```
sylvlecl: We should probably have a try/catch to handle deletion of the data if there is an exception halfway, for imports too

sylvlecl (commented):
A few comments

antarest/study/storage/variantstudy/variant_study_service.py
MartinBelthle: I created a frozen dataclass for all these args. It also improves readability

antarest/study/storage/utils.py
MartinBelthle: I introduced a `export_study` method inside the FS implem and remove the `path` attribute from the class

antarest/study/dao/file/file_study_factory_dao.py
MartinBelthle: I explicitely gave the matrix_service etc. instead of the command_context.

Also, now that the method `export_study` exists, I think the cache comment is resolved

antarest/study/model.py
```diff
@@ -768,3 +768,24 @@ class StudyMetadataUpdate:
...
     created_at: float | None = None
+
+
+@dataclasses.dataclass(frozen=True)
+class StudyMetadataCopy:
+    name: str
+    folder: PurePosixPath
+    owner: Identity
+    groups: list[Group]
+    directory_id: str | None
```
sylvlecl, line 778: Still weird to have both directory_id and folder as 2 separate fields but OK. They would better belong together in a single class, here we don't really understand that they' represent the same information.
We probably have a separate work to remove the folder field from studies, which should not exist anymore, actually

sylvlecl: Looked a little more into it:
seems we can remove folder right now, the import and creation of studies do not set it, the copy should not set it either.

antarest/study/storage/database_storage.py
```diff
@@ -63,6 +109,34 @@ def normalize_study(self, study: Study) -> None:
...
     @override
-    def import_study(self, study: RawStudy, stream: BinaryIO) -> RawStudy:
-        # TODO
-        raise NotImplementedError()
+    def import_study(self, study: RawStudy) -> None:
+        session = db.session
+
+        try:
+            # First, create the new study inside DB to avoid ForeignKey and StudyNotFound errors
+            session.add(study)
```
sylvlecl: we'd probably better go through the repo, for consistency sake with other services

antarest/study/storage/study_storage_interface.py
```diff
@@ -44,5 +48,5 @@ def normalize_study(self, study: Study) -> None:
         raise NotImplementedError()
 
     @abstractmethod
-    def import_study(self, study: RawStudy, stream: BinaryIO) -> RawStudy:
```
sylvlecl, line 47: The mechanism is very weird now:
it seems we use study.path as a temporary directory for the imported data.
Instead we should use a real temporary directory elsewhere, unrelated to that path stored in database.

Actually, study.path will have no meaning for database studies, we should probably make it optional now, and have it None for database studies.

antarest/study/storage/database_storage.py
```diff
@@ -63,6 +109,34 @@ def normalize_study(self, study: Study) -> None:
...
+            converter = StudyConverter(
+                source_dao=source_dao,
+                new_dao=new_dao,
+                study_version=source_dao.get_version(),
+                matrix_service=self._matrix_service,
+            )
+            converter.convert_study_inputs()
+
+            # Delete the source study path once the conversion is done
+            shutil.rmtree(Path(study.path))
```
sylvlecl: Should also be cleaned up in case of exception.
See other comment: we should use a real temporary directory here, not use study.path as a temporary directory

antarest/study/storage/database_storage.py
```diff
@@ -10,33 +10,60 @@
...
+            )
+            converter.convert_study_inputs()
+
+            return new_study
+
+        except Exception as e:
+            logger.error("Failed to copy study %s to %s", src_study.id, new_study.id, exc_info=e)
+            # Clean up the database if something went wrong.
+            session = db.session
+            session.delete(new_study)
```
sylvlecl: LLM tells me that in case we had an sqlalchemy exception, for example an IntegrityError, this will fail because we first need to rollback the transaction. So we'd probably better rollback first, then delete (same for import)

antarest/study/storage/utils.py
```diff
@@ -612,21 +611,12 @@ def build_raw_study_from_source(
...
         editor=get_current_user_name(),
         horizon=src_study.horizon,
-        public_mode=PublicMode.NONE if groups else PublicMode.READ,
-        groups=groups,
-        folder=str(destination_folder / dest_id),
+        public_mode=PublicMode.NONE if metadata.groups else PublicMode.READ,
+        groups=metadata.groups,
+        owner=metadata.owner,
+        directory_id=metadata.directory_id,
+        folder=str(metadata.folder / dest_id),
```
sylvlecl: See other comment: this is only used for copy to managed studies, we should not set folder

antarest/study/storage/database_storage.py
```diff
@@ -10,33 +10,60 @@
...
+from antarest.study.dao.file.file_study_factory_dao import FileStudyDaoFactory
+from antarest.study.dao.study_conversion.study_converter import StudyConverter
+from antarest.study.model import RawStudy, Study, StudyMetadataCreation
 from antarest.study.storage.study_storage_interface import IStudyStorage
-from antarest.study.storage.variantstudy.business.matrix_constants_generator import GeneratorMatrixConstants
 
 logger = logging.getLogger(__name__)
 
 
 class DatabaseStudyStorage(IStudyStorage):
```
sylvlecl, line 140: Not related to the PR, but this implementation does need to return matrices used by DB studies, otherwise matrices that are only used by those studies will be deleted at some point

antarest/study/storage/database_storage.py
```diff
@@ -10,33 +10,60 @@
...
-        self._generator_matrix_constants = generator_matrix_constants
-        self._config = config
+        self._db_dao_factory = db_dao_factory
+        self._fs_dao_factory = fs_dao_factory
 
     @override
     def copy(self, src_study: Study, new_study: RawStudy) -> RawStudy:
-        # TODO
-        raise NotImplementedError()
+        try:
```
sylvlecl, line 159: Now the error handling logic is:
 - different between DB and FS, which does not clean up anything.
    It was kind of OK before this PR, because the new_study was saved only AFTER
    actual creation of the files.
 - a little blurry between here and the service:
   it's the service which creates the new_study, but it's here that we delete it.
   It's also a little weird that we both get new_study as input, and return it as result.

So maybe we should instead leave the responsibility of everything to the storage layer so that:
- the FS keeps the behaviour of first creating files and then only save the study metadata to DB
- the DB impl keeps this cleaning logic at the end
- the interface signature makes it clear since it will not receive anymore the new_study as argument

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -74,27 +73,33 @@ def __init__(
...
         RawStudyMatrixUsageProvider(repository, self._matrix_service, self._storage_mapping)
         self.cache = cache
 
     @override
-    def copy(self, src_study: Study, dest_name: str, groups: list[str], destination_folder: PurePosixPath) -> RawStudy:
-        new_study = build_raw_study_from_source(
-            dest_name, self._config.get_workspace_path(), groups, src_study, destination_folder
-        )
+    def copy(self, src_study: Study, metadata: StudyMetadataCopy) -> RawStudy:
+        new_study = build_raw_study_from_source(src_study, self._config.get_workspace_path(), metadata)
```
sylvlecl: See here: that could be delegated to the storage layer, so that it is persisted only when the implementation decides to (at the end of the creation for FS study, in particular).

sylvlecl (changes requested):
Nice, a few more comments after slightly deeper analysis !

sylvlecl (approved):
Looks good!

PR #3221, opened 2026-05-22
feat(ui-outputs): add user column and refactor the view

skamril:
- add ID on output DTO
- add Zod schema validation on outputs API responses
- integrate TanStack Query for data fetching and caching
- add "archived" chip indicator on output items
- restrict access to archived outputs

ANT-1707


webapp/src/routes/-shared/components/studies/dialogs/ExportModal/index.tsx
```diff
@@ -134,10 +134,10 @@ export default function ExportModal(props: BasicDialogProps & Props) {
   useEffect(() => {
     (async () => {
       try {
-        const res = await getStudyOutputs(study.id);
+        const outputs = await getOutputs({ studyId: study.id });
         const tmpSynth = await getStudySynthesis(study.id);
-        setOutputList(res.map((o: OutputDetails) => ({ id: o.name, name: o.name })));
-        setCurrentOutput(res.length > 0 ? res[0].name : undefined);
+        setOutputList(outputs);
+        setCurrentOutput(outputs.length > 0 ? outputs[0].name : undefined);
```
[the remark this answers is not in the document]
skamril: legacy code

webapp/src/services/webSocket/ws.ts
```diff
@@ -302,6 +305,24 @@ function makeNotificationListener(dispatch: AppDispatch): WsEventListener {
...
       }
+      case WsEventType.StudyJobStarted:
+      case WsEventType.StudyJobStatusUpdate:
+      case WsEventType.StudyJobCompleted:
+      case WsEventType.StudyJobCancelled: {
+        const updatedJob = adaptJobDtoToJob(e.payload);
+
+        // Update job list if exists in cache
+        queryClient.setQueryData(jobQueries.list(updatedJob.studyId).queryKey, (old) => {
+          return old?.map((job) => (job.id === updatedJob.id ? updatedJob : job));
```
hdinia: **issue (blocking):** `old?.map` only *replaces* a job already in the cache. On `StudyJobStarted` the job id is brand-new, so it's silently dropped and never added. Since the outputs view now reads running jobs exclusively from this cache (no more `getJobs` on mount + reload), a freshly launched simulation won't render in the "Running jobs" section until a full refetch/navigation.

Consider appending when absent:
```ts
queryClient.setQueryData(queryKey, (old = []) =>
  old.some((j) => j.id === updatedJob.id)
    ? old.map((j) => (j.id === updatedJob.id ? updatedJob : j))
    : [updatedJob, ...old],
);
```
Please verify the live launch → running-row behavior before merge.

hdinia (commented):
Globally good but needs to verify some parts before merging

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/-hooks/useArchiveOutput.ts
```diff
@@ -0,0 +1,91 @@
...
+            output.id === outputId
+              ? createOptimisticListItem({ ...output, archived: true })
+              : output,
+          );
+        });
+      }
+
+      return { outputToArchive };
+    },
+    onSuccess: (_, variables, onMutateResult) => {
```
hdinia: **suggestion (non-blocking):** `archiveOutput` returns a *taskId* — the work is async server-side. `onSuccess` confirms `archived: true` the moment the POST resolves (task *queued*), not when the task is *done*. The websocket only `invalidateQueries(outputKeys.list)` on `StudyJobCompleted` (simulation jobs), not on archive/unarchive task completion — so if the task later fails, the UI stays wrong. Same applies to `useUnarchiveOutput`. Not a regression (the old code had the same limitation), but worth a follow-up: listen for the task's completion or `invalidateQueries` on settle.

webapp/src/services/api/studies/outputs/schemas.ts
```diff
@@ -0,0 +1,32 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+import z from "zod";
+
+const modeSchema = z.enum(["Economy", "Adequacy", "Expansion"]);
```
hdinia, line 17: **suggestion (non-blocking):** `outputSchema` and the `OutputSynthesis` interface in `types.ts` now describe the same shape, including these `mode`/`storageType` literal unions duplicated in both places. And `StudyOutput = Output | (OutputSynthesis & { id: string })` is effectively just `Output` since the two are structurally identical. Consider deriving one from the other (e.g. have `OutputSynthesis` reuse the inferred `Output` type) to keep a single source of truth.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-hooks/useStudyOutput.ts
```diff
@@ -67,7 +55,7 @@ export default function useStudyOutput({
     data: fetchedOutput,
     isLoading,
     error,
-  } = usePromise<PartialStudyOutput | undefined>(
+  } = usePromise<StudyOutput | undefined>(
```
hdinia, line 58: **suggestion (non-blocking):** the rest of the PR migrates to TanStack Query, but this hook stays on `usePromise`/`useStudySynthesis`, and `getStudyOutputById` fetches *all* outputs then filters client-side. Migrating to `outputQueries.list` + a `select` would dedupe the network call (the list is already cached/prefetched) and finish the refactor. Reasonable to defer, flagging for consistency.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/index.tsx
```diff
@@ -53,282 +46,165 @@ import {
...
+
+    if (isConfirmed) {
+      deleteOutput.mutate({ studyId, outputId: output.id });
     }
   };
 
-  const openDigestDialog = (job: Job) => {
-    setDialogState({ type: "digest", data: job });
+  const handleDownload = (job: Job) => {
+    // TODO: check if download work
```
hdinia: **nitpick:** leftover unverified `TODO: check if download work` on a user-facing button. Either verify and drop the comment, or convert it to a tracked issue before merge.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/index.tsx
```diff
@@ -53,282 +46,165 @@ import {
...
-            <DownloadIcon sx={iconStyle} onClick={() => handleDownload(output.job)} />
-          </Tooltip>
-        </Box>
-      )}
+    if (output.job?.completionDate) {
+      completionDate = convertUTCToLocalTime(output.job.completionDate);
+    }
+    // If no related job or completion date is not available, try to parse date from output name
+    else {
+      const dateMatch = output.name.match(/(\d{4})(\d{2})(\d{2})-(\d{2})(\d{2})/);
```
hdinia: **nitpick:** inconsistent UTC handling — `job.completionDate` above goes through `convertUTCToLocalTime`, but this name-parsed fallback is rendered raw. Pre-existing behavior, but the two date sources are now displayed with different timezone treatment in the same cell.

webapp/src/services/api/studies/outputs/schemas.ts
skamril: In the backend we have two models: `OutputSynthesis` and `OutputDetails` (which adds 'id' and 'settings' fields). They are fully independent from each other, neither inherits from the other. So I kept the same approach on the TypeScript side as well, to avoid accidentally adding unwanted fields.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-hooks/useStudyOutput.ts
skamril: It's updated in another branch. The redux part is not used anymore.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/index.tsx
skamril: In reality, job.completionDate cannot be empty here. I can remove the condition.

hdinia (approved):

PR #3222, opened 2026-05-22
refactor(task,download): use async endpoints for polling

sylvlecl:
Currently, 2 endpoints implement a polling logic. They can block other endpoints while performing no work.
We make them async so that fastapi thread pool is not affected anymore, and they can scale independently.


sylvlecl:
Closing for now, probably not worth the additional complexity

PR #3225, opened 2026-05-26
feat(variant): fetch whole tree in single request

smailio:


antarest/study/storage/variantstudy/repository.py
```diff
@@ -81,16 +91,19 @@ def get_ancestor_or_self_ids(self, variant_id: str) -> Sequence[str]:
...
+        result = self.session.execute(select(cte.c.id))
         return [r[0] for r in result]
 
+    def get_root_ancestor_id(self, variant_id: str) -> str | None:
+        """
+        Return the id of the topmost ancestor of `variant_id`, or `variant_id` itself if
+        it has no parent. Returns None if `variant_id` does not exist.
+        """
+        cte = self._ancestor_or_self_cte(variant_id)
+        row = self.session.execute(select(cte.c.id).where(cte.c.parent_id.is_(None))).first()
```
MartinBelthle: I think instead of using `first()`, we should use `.scalar_one_or_none()` as `first` does more than what we actually want

tests/storage/business/test_repository.py
```diff
@@ -115,3 +115,29 @@ def test_get_all_descendants(self, db_session: Session) -> None:
...
+        # Unknown id
+        assert repository.get_root_ancestor_id("nope") is None
+
+        # Standalone raw study: root is itself
+        raw_study = create_raw_study(name="Root")
+        db_session.add(raw_study)
+        db_session.commit()
+        assert repository.get_root_ancestor_id(raw_study.id) == raw_study.id
+
+        # Deep chain: root -> v1 -> v1a -> v1a1
```
MartinBelthle: Minor but for test readability, i would prefer v2 and v3 instead of v1a and v1a1

tests/storage/business/test_repository.py
```diff
@@ -115,3 +115,29 @@ def test_get_all_descendants(self, db_session: Session) -> None:
...
+        db_session.add(v1)
+        db_session.commit()
+        v1a = create_variant_study(name="v1a", parent_id=v1.id)
+        db_session.add(v1a)
+        db_session.commit()
+        v1a1 = create_variant_study(name="v1a1", parent_id=v1a.id)
+        db_session.add(v1a1)
+        db_session.commit()
+
+        for study in (v1, v1a, v1a1):
```
MartinBelthle: We should ensure this perform only one DB request. We use this usually:

 `with DBStatementRecorder(db_session.bind) as db_recorder:`

tests/storage/business/test_variant_study_service.py
```diff
@@ -116,3 +117,73 @@ def test_get_variant_children(tmp_path: Path, admin_user: Any) -> None:
...
+    repo_mock.get_root_ancestor_id.return_value = "root"
+    assert study_service.get_root_study_id("leaf") == "root"
+    repo_mock.get_root_ancestor_id.assert_called_once_with("leaf")
+
+    repo_mock.get_root_ancestor_id.return_value = None
+    with pytest.raises(StudyNotFoundError):
+        study_service.get_root_study_id("missing")
+
+
+def test_get_all_variants_children_from_root(tmp_path: Path, admin_user: Any) -> None:
```
MartinBelthle: For the first test, it does not bother me but for this one I would prefer we use a "real study"  and a "real" variant_study_service from fixtures as Mocks everywhere are hard to maintain on the long run

MartinBelthle (changes requested):
Minor comments, after that we can merge

webapp/src/services/api/studies/variants/index.ts
```diff
@@ -14,41 +14,14 @@
 
 import z from "zod";
 import client from "../../client";
-import { studySchema } from "../schemas";
-import type { Study } from "../types";
 import { variantTreeSchema } from "./schemas";
 import type { CreateVariantParams, GetVariantTreeParams } from "./types";
 
 export async function getVariantTree({ studyId, includeParents = true }: GetVariantTreeParams) {
```
skamril: You can change the param `includeParents` by `fromRoot` please.

MartinBelthle (approved):

PR #3232, opened 2026-06-01
refactor(db): introduce intermediate table for the dao

MartinBelthle:
Both upgrade and downgrade worked with sqlite on my PC


alembic/versions/844f4a7cd371_create_intermediate_table.py
```diff
@@ -0,0 +1,156 @@
...
+        metadata = sa.MetaData()
+        metadata.reflect(bind=op.get_bind(), only=[table_name])
+        table = metadata.tables[table_name]
+
+        # Generate the new foreign key constraint
+        new_fk_constraint = sa.ForeignKeyConstraint(['study_id'],[new_foreign_key], name=f'fk_{table_name}_study_id', ondelete='CASCADE')
+
+        # Create a new table with the new foreign key constraint
+        cols = [sa.Column(col.name, col.type, nullable=col.nullable, primary_key=col.primary_key) for col in table.columns]
+        op.create_table(f"{table_name}_new", *cols, new_fk_constraint)
```
sylvlecl: This will not properly re-create all constraints unfortunately:
- for example in area_ui we have a composite foreign key which will not be re-constructed
- possibly we have other constraints such as unique constraints ?

alembic/versions/844f4a7cd371_create_intermediate_table.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    Introduce an intermediate table between DAO and Study, useful for clearing variant snapshots.
+    When doing so, we'll only have to flush this table and everything will be removed by cascade.
+    """
+
+    # Create the new intermediate table
+    op.create_table(
+        INTERMEDIATE_TABLE,
+        sa.Column('study_id', sa.String(length=36), sa.ForeignKey('study.id', ondelete='CASCADE'), nullable=False, primary_key=True),
+        sa.Column('study_data_id', sa.String(length=36), nullable=False, unique=True),
```
sylvlecl: We could take the opportunity to use a real UUID actually here, instead of a string

alembic/versions/844f4a7cd371_create_intermediate_table.py
sylvlecl: Not sure there will be a better solution than recreating everything ...
At least we need to re-create foreign keys explicitly for each table, I guess

antarest/study/dao/database/models/__init__.py
```diff
@@ -9,3 +9,15 @@
 # SPDX-License-Identifier: MPL-2.0
 #
 # This file is part of the Antares project.
+from sqlalchemy import Column, ForeignKey, String, Table
+
+from antarest.dbmodel import Base
+
+metadata = Base.metadata
+
+STUDY_DATA_CONTAINER_TABLE = Table(
```
sylvlecl: maybe just "study_data" for the table naming ?

sylvlecl (changes requested):
All constraints are not properly re-constructed unfortunately !!

antarest/study/dao/database/sql_utils.py
```diff
@@ -105,8 +105,12 @@ def upsert_multiple(
             stmt = pg_insert(table).values(batch)
         else:
             stmt = sqlite_insert(table).values(batch)
-        stmt = stmt.on_conflict_do_update(
-            index_elements=key_columns,
-            set_={column: getattr(stmt.excluded, column.name) for column in update_columns},
-        )
+
+        if update_columns:
```
sylvlecl: Not sure we should keep this, since we don't have the use case anymore.
Also, if we keep it, it's still probably best to keep the `on_conflict_do_udpate` but without the set_ argument, otherwise it will raise when the primary key is already in the table, which is what we want to avoid with that upsert function

antarest/study/dao/database/models/__init__.py
MartinBelthle: I renamed the Table but kept the name `study_data_container` for the real table name. You wanted me to change it at both places ?

alembic/versions/844f4a7cd371_create_intermediate_table.py
MartinBelthle: Should be resolved

alembic/versions/844f4a7cd371_create_intermediate_table.py
MartinBelthle: Postgre supports `sa.Uuid` but Sqlite does not and uses a CHAR32 instead. So I let it like this as elsewhere we use a VARCHAR36 we would have a mismatch

antarest/study/dao/database/sql_utils.py
MartinBelthle: I removed the if and added a `or None`, it should resolve this comment

antarest/study/dao/database/models/__init__.py
sylvlecl: indeed let's be consistent all the way

sylvlecl (dismissed):
Ok for me, please review a little the part about PG migration before merging, that I just pushed  :) 
Tested on my env that it passes, but just in case ...

sylvlecl (approved):

PR #3233, opened 2026-06-01
feat(task): prevent parallel execution of same type of task

olivierPigeon-RTE:
Currently, several tasks of the same type (auto-archive, disk space analyze, disk usage, ...) can be run at the same time, which is pointless.

With this PR, it is impossible to have several tasks of the same time in parallel, but still possible to have tasks of different types.


antarest/core/utils/lock.py
```diff
@@ -35,193 +34,51 @@ class LockNotAcquired(Exception):
     pass
 
 
-class PostgresqlLockNotSupported(Exception):
-    """Raised when trying to use PostgreSQL advisory locks on a non-PostgreSQL database."""
-
-    pass
-
-
-class DistributedLock(ABC):
```
sylvlecl, line 44: We need to update the module level docstring which talks about that abstraction and the various implementations

antarest/maintenance/tasks/auto_archive_task.py
```diff
@@ -68,6 +68,7 @@ def auto_archive_task(self: MaintenanceTask) -> AutoArchiveTaskResult:
             ctx.config.storage.auto_archive_threshold_days,
             ctx.config.storage.snapshot_retention_days,
             ctx.config.storage.auto_archive_dry_run,
+            lock_folder=self.context.config.storage.tmp_dir,
```
sylvlecl, line 71: Something we could improve in the future, to not have to pass a lock_folder everywhere, is to have a Lock service to encapsulate that.

The maintenance context could then provide the lock service, without each user needing to go and get the configuration consistently with other users.

But OK for now

sylvlecl (dismissed):
Looks good to me:
we just need to fix the docstring of the lock module.

Design could be improved a little to encapsulate better the configuration of the lock dir, but let's leave that for further work

sylvlecl (commented):
Ah, seems the new arg "lock_folder" is missing for some of the tasks in test (see CI), at least for the clean_matrices

sylvlecl (approved):

PR #3237, opened 2026-06-05
test(db): add postgre migration test

MartinBelthle:


tests/core/test_alembic_migration.py
```diff
@@ -35,3 +31,19 @@ def test_alembic_migration(tmp_path: Path) -> None:
...
+    # Create a fake config file pointing towards the DB in memory
+    db_file = tmp_path / "db.sqlite"
+    db_url = f"sqlite:///{db_file}"
+
+    _checks_migration(tmp_path, db_url)
+
+
+def test_alembic_migration_postgresql(tmp_path: Path) -> None:
+    # Start a PostgreSQL container
+    with PostgresContainer("postgres:15") as postgres:
```
sylvlecl: Note that for now, we use 14 in production, maybe we'd better stick with it

sylvlecl (commented):
Great:
we probably need to disable it for windows CI, and we can use postrges 14 to stick with our current prod setup

PR #3238, opened 2026-06-08
feat(study): support archive / unarchive for DB studies

MartinBelthle:
Worked on my local env

### DIFF
--- a/antarest/study/dao/study_conversion/study_converter.py
+++ b/antarest/study/dao/study_conversion/study_converter.py
@@ -75,6 +75,9 @@ class StudyConverter:
         for district in self._source_dao.get_districts():
             self._new_dao.save_district(district)
 
+        # Comments
+        self._new_dao.save_comments(self._source_dao.get_comments())
+
     def _convert_settings(self) -> None:
         self._new_dao.save_general_config(self._source_dao.get_general_config())
         self._new_dao.save_playlist_config(self._source_dao.get_playlist_config())
--- a/antarest/study/storage/database_storage.py
+++ b/antarest/study/storage/database_storage.py
@@ -184,17 +184,22 @@ class DatabaseStudyStorage(IStudyStorage):
     @override
     def write_study_for_archive(self, study: RawStudy, dst_path: Path) -> None:
         self.export_study(study, dst_path)
-        self.clean_study_data_table(study.id)
+
+    @override
+    def remove_study_data(self, study: Study) -> None:
+        session = db.session
+        session.execute(delete(STUDY_DATA_TABLE).where(STUDY_DATA_TABLE.c.study_id == study.id))
+        session.commit()
 
     @override
     def unarchive(self, study: RawStudy, archive_path: Path) -> None:
-        self.extract_study(study, archive_path, create_study_in_db=False)
-
-    @staticmethod
-    def clean_study_data_table(study_id: str) -> None:
-        session = db.session
-        session.execute(delete(STUDY_DATA_TABLE).where(STUDY_DATA_TABLE.c.study_id == study_id))
-        session.commit()
+        try:
+            self._extract_study(study, archive_path, create_study_in_db=False)
+        except Exception as e:
+            logger.error(f"Failed to unarchive study {study.id}", exc_info=e)
+            # Clean up the database
+            self.remove_study_data(study)
+            raise e
 
     @override
     def export_study(self, study: Study, dst_path: Path) -> None:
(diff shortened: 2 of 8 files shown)


tests/integration/studies_blueprint/test_study_archive_unarchive.py
```diff
@@ -0,0 +1,106 @@
...
+# This file is part of the Antares project.
+import zipfile
+from pathlib import Path
+
+from starlette.testclient import TestClient
+
+from tests.integration.utils import wait_for
+
+
+def test_archive(client: TestClient, admin_access_token: str, tmp_path: Path, internal_study_id: str) -> None:
```
MartinBelthle, line 25: This test was just moved

antarest/study/storage/rawstudy/raw_study_service.py
```diff
@@ -153,7 +152,7 @@ def archive(self, study: RawStudy) -> None:
 
         else:
             # Remove the source study folder if everything went well
-            shutil.rmtree(Path(study.path))
+            shutil.rmtree(Path(study.path), ignore_errors=True)
```
sylvlecl: We should only remove the DB study here, otherwise we can have the case where we have an error in the archival and we have already removed the study in DB.

antarest/study/storage/database_storage.py
```diff
@@ -182,8 +183,18 @@ def copy(self, src_study: Study, metadata: StudyMetadataCopy) -> RawStudy:
 
     @override
     def write_study_for_archive(self, study: RawStudy, dst_path: Path) -> None:
-        # Nothing to do
-        pass
+        self.export_study(study, dst_path)
+        self.clean_study_data_table(study.id)
```
sylvlecl: See other comment: we cannot do it here, we are not sure that the data has safely been written to archive store.

antarest/study/storage/database_storage.py
```diff
@@ -230,20 +241,32 @@ def normalize_study(self, study: Study) -> None:
 
     @override
     def import_study(self, study: RawStudy, stream: BinaryIO) -> None:
+        try:
+            self.extract_study(study, stream, create_study_in_db=True)
+
+        except Exception as e:
+            logger.error("Failed to import study %s", str(study.path), exc_info=e)
```
sylvlecl: let's replace the %s by a f-string

antarest/study/storage/database_storage.py
```diff
@@ -230,20 +241,32 @@ def normalize_study(self, study: Study) -> None:
...
+            # Clean up the database
+            db.session.rollback()
+            self._repository.delete(study.id)
+            raise e
+
+    def extract_study(self, study: RawStudy, source: Path | BinaryIO, create_study_in_db: bool) -> None:
```
sylvlecl: let's make it _private

antarest/study/storage/database_storage.py
```diff
@@ -182,8 +183,18 @@ def copy(self, src_study: Study, metadata: StudyMetadataCopy) -> RawStudy:
...
-        pass
+        self.export_study(study, dst_path)
+        self.clean_study_data_table(study.id)
+
+    @override
+    def unarchive(self, study: RawStudy, archive_path: Path) -> None:
+        self.extract_study(study, archive_path, create_study_in_db=False)
+
+    @staticmethod
+    def clean_study_data_table(study_id: str) -> None:
```
sylvlecl: let's make it _private too

antarest/study/storage/database_storage.py
```diff
@@ -230,20 +241,32 @@ def normalize_study(self, study: Study) -> None:
 
     @override
     def import_study(self, study: RawStudy, stream: BinaryIO) -> None:
+        try:
+            self.extract_study(study, stream, create_study_in_db=True)
+
+        except Exception as e:
+            logger.error("Failed to import study %s", str(study.path), exc_info=e)
+            # Clean up the database
+            db.session.rollback()
```
sylvlecl, line 256: We need that cleanup when unarchive fails too.
Otherwise, the next unarchive will fail, because there will already be a study_data associated with that study

antarest/study/dao/study_conversion/study_converter.py
```diff
@@ -75,6 +75,9 @@ def convert_study_inputs(self) -> None:
         for district in self._source_dao.get_districts():
             self._new_dao.save_district(district)
 
+        # Comments
+        self._new_dao.save_comments(self._source_dao.get_comments())
```
sylvlecl, line 79: We should cover it in study converter tests ;)

sylvlecl (changes requested):
A  few changes requested to ensure proper error handling

antarest/study/storage/database_storage.py
MartinBelthle: I'll need this inside the snapshot cleaning so I'll keep it public. In fact I introduced a `remove_study_data` method inside the interface to use after archiving that I'll use inside the snapshot cleaning

antarest/study/storage/rawstudy/raw_study_service.py
MartinBelthle: The `else` condition is only met if no error occured, so I don't think it's an issue

antarest/study/storage/database_storage.py
```diff
@@ -182,8 +183,23 @@ def copy(self, src_study: Study, metadata: StudyMetadataCopy) -> RawStudy:
...
+        session.commit()
+
+    @override
+    def unarchive(self, study: RawStudy, archive_path: Path) -> None:
+        try:
+            self._extract_study(study, archive_path, create_study_in_db=False)
+        except Exception as e:
+            logger.error(f"Failed to unarchive study {study.id}", exc_info=e)
+            # Clean up the database
+            self.remove_study_data(study)
```
sylvlecl, line 202: Like for the other places: we'd probably better do a rollaback first, otherwise there's a chance that the delete fails

sylvlecl (changes requested):
Minor change requested on the error path again, then OK !

PR #3239, opened 2026-06-08
feat(favorite_external_directory): add external directories favorite

mehdiwahada:
What does it do ? : Implements the addition of external directories as favorite

Why it has been done ? : The external directories don't have a stable id, so we need to add their path (workspace + current path) in the favorite handling API

How to test it ? : Try to add external directories with valid / invalid workspace or paths


antarest/favorite/repository.py
```diff
@@ -137,3 +137,48 @@ def save(self, directory: FavoriteDirectory) -> FavoriteDirectory:
...
+    def get(self, workspace: str, path: str) -> FavoriteExternalDirectory | None:
+        stmt = (
+            select(FavoriteExternalDirectory)
+            .where(FavoriteExternalDirectory.user_id == get_user_impersonator())
+            .where(FavoriteExternalDirectory.workspace == workspace)
+            .where(FavoriteExternalDirectory.path == path)
+        )
+        result = self.session.execute(stmt)
+
+        return result.unique().scalar()
```
smailio: unique seems useless here ?

antarest/favorite/model.py
```diff
@@ -80,3 +81,19 @@ class FavoriteDirectory(Base):
...
+
+class FavoriteExternalDirectory(Base):
+    __tablename__ = "favorite_external_directory"
+
+    path: Mapped[str] = mapped_column(String, primary_key=True)
+    workspace: Mapped[str] = mapped_column(String, primary_key=True)
+    user_id: Mapped[int] = mapped_column(Integer, primary_key=True)
+
+    def to_dto(self) -> "FavoriteExternalDirectoryDTO":
+        return FavoriteExternalDirectoryDTO(path=PosixPath(self.path), workspace=self.workspace)
```
smailio: You should add an alembic migration for this

antarest/favorite/model.py
smailio, line 66: this is a hidden n+1 query btw

antarest/favorite/service.py
```diff
@@ -110,3 +125,40 @@ def delete_favorite(self, directory_uuid: str) -> None:
...
+    def __init__(
+        self, favorite_external_directory_repository: FavoriteExternalDirectoryRepository, workspace_config: Config
+    ):
+        self._favorite_external_directory_repository = favorite_external_directory_repository
+        self.workspaces = workspace_config.storage.workspaces
+
+    def add_favorite(self, directory_path: str, workspace: str) -> FavoriteExternalDirectoryDTO:
+
+        if workspace in self.workspaces.keys():
+            if (Path(self.workspaces[workspace].path) / directory_path).exists():
```
smailio: Mild risk of path traversal attack. Attacker can use this to extract information about filesystem. 

Ther's a utils function that you can use to remove this risk 

```python 
# antarest/study/storage/utils.py
def is_folder_safe(workspace: WorkspaceConfig, folder: str) -> bool:
    """
    Check if the provided folder path is safe to prevent path traversal attack.

    Args:
        workspace: The workspace name.
        folder: The folder path.

    Returns:
        `True` if the folder path is safe, `False` otherwise.
    """
    requested_path = workspace.path / folder
    requested_path = requested_path.resolve()
    safe_dir = workspace.path.resolve()
    # check whether the requested path is a subdirectory of the workspace
    return requested_path.is_relative_to(safe_dir)
```

antarest/favorite/service.py
```diff
@@ -110,3 +125,40 @@ def delete_favorite(self, directory_uuid: str) -> None:
             raise HTTPException(
                 status_code=http.HTTPStatus.NOT_FOUND, detail=f"Directory with id {directory_uuid} not found"
             ) from None
+
+
+class FavoriteExternalDirectoryService:
+    def __init__(
+        self, favorite_external_directory_repository: FavoriteExternalDirectoryRepository, workspace_config: Config
+    ):
+        self._favorite_external_directory_repository = favorite_external_directory_repository
```
smailio: why private member ? you didn't do this for sibling service

antarest/favorite/service.py
```diff
@@ -110,3 +125,40 @@ def delete_favorite(self, directory_uuid: str) -> None:
...
+        self.workspaces = workspace_config.storage.workspaces
+
+    def add_favorite(self, directory_path: str, workspace: str) -> FavoriteExternalDirectoryDTO:
+
+        if workspace in self.workspaces.keys():
+            if (Path(self.workspaces[workspace].path) / directory_path).exists():
+                favorite_external_directory = FavoriteExternalDirectory(
+                    path=directory_path, workspace=workspace, user_id=get_user_impersonator()
+                )
+                dto = self._favorite_external_directory_repository.save(favorite_external_directory).to_dto()
```
smailio: not important but you can normalize this path to be relative to the workspace. Just in case you recieve relative paths with .. in them or absolute paths.

antarest/favorite/repository.py
```diff
@@ -137,3 +137,48 @@ def save(self, directory: FavoriteDirectory) -> FavoriteDirectory:
...
+        session.add(fav)
+        session.commit()
+        return favorite_external_directory
+
+    def get_all(self) -> list[FavoriteExternalDirectory]:
+        stmt = select(FavoriteExternalDirectory).where(FavoriteExternalDirectory.user_id == get_user_impersonator())
+        result = self.session.execute(stmt)
+        return list(result.scalars().all())
+
+    def get(self, workspace: str, path: str) -> FavoriteExternalDirectory | None:
```
smailio: this get is only used in the delete service call. You can update the delete so that it checks the result.rowcount instead of doing a preventive check. You would save a db call most of the time a delete is called, and make the code base ligher.

antarest/favorite/web.py
```diff
@@ -61,4 +69,47 @@ def delete_favorite_directory(favorite_directory_service: FavoriteDirectoryServi
...
+    @bp.get(
+        "/favorites/external_directories",
+        summary="Listing favorite external directories for current user",
+        status_code=HTTPStatus.OK,
+    )
+    def list_favorite_external_directories(
+        favorite_external_directory_service: FavoriteExternalDirectoryServiceDep,
+    ) -> list[FavoriteExternalDirectoryDTO]:
+        logger.info("Listing favorite external directories for current user.")
+        return favorite_external_directory_service.list_favorites()
```
smailio, line 85: Ticket mentionned an aggregate endpoint that would return all type of directories, maybe that's coming in a future PR ?

smailio (changes requested):
Nice nice 

I asked some questions and made some suggestions. Should be good after those are adressed.

antarest/favorite/service.py
mehdiwahada: It's an oversight on my part, I'll correct it. Thanks for the information.

antarest/favorite/web.py
mehdiwahada: I do think it'll be in a coming PR, but I don't know when.

antarest/favorite/model.py
mehdiwahada: It'll be a good ticket for a refactoring, thanks for telling me

antarest/favorite/service.py
```diff
@@ -110,3 +126,54 @@ def delete_favorite(self, directory_uuid: str) -> None:
...
+    def add_favorite(self, directory_path: str, workspace: str) -> FavoriteExternalDirectoryDTO:
+
+        if workspace in self.workspaces.keys():
+            if not is_folder_safe(self.workspaces[workspace], directory_path):
+                raise HTTPException(
+                    status_code=http.HTTPStatus.BAD_REQUEST,
+                    detail=f"Directory {directory_path} is not safe",
+                )
+
+            relative_path = Path.relative_to(Path(directory_path), self.workspaces[workspace].path / workspace)
```
smailio: you can put this bellow the if as you don't use it elsewhere

antarest/study/storage/file_study_storage.py
```diff
@@ -143,13 +143,8 @@ def normalize_file_study(self, file_study: FileStudy) -> None:
...
         study_path = Path(study.path)
-        study_dir.rename(study_path)
-
-        # Move the "output" subfolder back to its original location as we only want to import the inputs here
-        outputs_path = study_path / "output"
-        if outputs_path.exists():
-            study_dir.mkdir()
-            outputs_path.rename(study_dir / "output")
+        # As `study_dir` and `study_path` may not be on the same fs partition, we have to do a `copytree`
+        shutil.copytree(study_dir, study_path, ignore=shutil.ignore_patterns("output"))
```
smailio: I guess this is a mege with dev that went wrong ?

antarest/favorite/repository.py
```diff
@@ -137,3 +137,39 @@ def save(self, directory: FavoriteDirectory) -> FavoriteDirectory:
...
+        if self._session is None:
+            return db.session
+        return self._session
+
+    def save(self, favorite_external_directory: FavoriteExternalDirectory) -> FavoriteExternalDirectory:
+        session = self.session
+        fav = session.merge(favorite_external_directory)
+        session.add(fav)
+        session.commit()
+        return favorite_external_directory
```
smailio: use return fav like you did in other repository class

antarest/favorite/service.py
```diff
@@ -110,3 +126,54 @@ def delete_favorite(self, directory_uuid: str) -> None:
...
+class FavoriteExternalDirectoryService:
+    def __init__(
+        self, favorite_external_directory_repository: FavoriteExternalDirectoryRepository, workspace_config: Config
+    ):
+        self.favorite_external_directory_repository = favorite_external_directory_repository
+        self.workspaces = workspace_config.storage.workspaces
+
+    def add_favorite(self, directory_path: str, workspace: str) -> FavoriteExternalDirectoryDTO:
+
+        if workspace in self.workspaces.keys():
```
smailio: there's a util function fot this `get_workspace_from_config` use it to get the workspace. It does the check you're doing, plus check againt default workspace and return a standard error message.

tests/integration/favorite_blueprint/test_favorite_external_directory.py
```diff
@@ -0,0 +1,134 @@
...
+    expected_favorite_external_directory = {"workspace": workspace_name, "path": str(path)}
+    response = admin_client.post("/v1/favorites/external_directories", params=expected_favorite_external_directory)
+    assert response.status_code == 400
+    assert response.json()["description"] == f"Directory {path} is not safe"
+
+
+# TEST GET FAVORITES EXTERNAL DIRECTORY
+def test_get_favorite_external_directory_success_get_one_favorite(admin_client: TestClient, tmp_path: Path):
+    # creating one favorite external directory and checking that it is returned
+    workspace_name = "default"
```
smailio: "default" is the name of the internal workspace so this may be confusing

tests/favorite_study/test_favorite_external_directory.py
```diff
@@ -0,0 +1,169 @@
...
+        favorite_service.delete_favorite(path=inexisting_directory_path, workspace="invalidspace")
+
+
+def test_delete_favorite_external_directory_success_deleted_one_favorite(tmp_path: Path):
+    # deleting an external directory from the favorites, and then checking that the favorite is deleted
+    config = Config(storage=StorageConfig(tmp_dir=tmp_path))
+    mock_favorite_external_directory_repository = Mock(spec=FavoriteExternalDirectoryRepository)
+    favorite_service = FavoriteExternalDirectoryService(
+        mock_favorite_external_directory_repository, workspace_config=config
+    )
```
smailio: You can create a fixture for this, you use it 3 times, or maybe there's a nuance I didn't notice ?

smailio (changes requested):
Almost there !

tests/integration/favorite_blueprint/test_favorite_external_directory.py
mehdiwahada: Since there's no way to create a workspace (I don't know how to do it), I was advised by Sylvain to use this

antarest/study/storage/file_study_storage.py
mehdiwahada: It is. Should I revert this one like it was before ? Real question.

antarest/favorite/service.py
mehdiwahada: Done, will switch default for another one.

antarest/study/storage/file_study_storage.py
smailio: Yeah should be the same than dev

antarest/favorite/service.py
mehdiwahada: The main reason I'm putting it here is that since the folder is safe, it means the path is functional. the is_folder_safe act as a security, in my opinion.

.github/workflows/main.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: Install dependencies
         run: uv sync --frozen --compile-bytecode
       - name: Test with pytest
-        run: uv run --frozen pytest --cov antarest --cov-report xml -n 4 --dist=worksteal
+        run: uv run pytest tests/integration/favorite_blueprint/test_favorite_external_directory.py #uv run --frozen pytest --cov antarest --cov-report xml -n 4 --dist=worksteal
```
smailio: you forgot this test code

antarest/favorite/model.py
```diff
@@ -80,3 +82,26 @@ class FavoriteDirectory(Base):
...
+        """
+        return PurePosixPath(path)
+
+
+class FavoriteExternalDirectory(Base):
+    __tablename__ = "favorite_external_directory"
+
+    path: Mapped[str] = mapped_column(String, primary_key=True)
+    workspace: Mapped[str] = mapped_column(String, primary_key=True)
+    user_id: Mapped[int] = mapped_column(Integer, primary_key=True)
```
MartinBelthle: As this is a foreign key you should use `ForeignKey` here

antarest/favorite/model.py
```diff
@@ -80,3 +82,26 @@ class FavoriteDirectory(Base):
...
+
+    def to_dto(self) -> "FavoriteExternalDirectoryDTO":
```
MartinBelthle: Minor but you don"t need the quotes

antarest/favorite/web.py
```diff
@@ -61,4 +69,47 @@ def delete_favorite_directory(favorite_directory_service: FavoriteDirectoryServi
         logger.info(f"Deleting directory {uuid} from favorites.")
         favorite_directory_service.delete_favorite(uuid)
 
+    @bp.get(
+        "/favorites/external_directories",
```
MartinBelthle: Ticket says to use a `-` instead of a `_` to stay coherent with the rest of the app.
Same goes for the other endpoints introduced

antarest/favorite/service.py
```diff
@@ -110,3 +126,54 @@ def delete_favorite(self, directory_uuid: str) -> None:
...
+class FavoriteExternalDirectoryService:
+    def __init__(
+        self, favorite_external_directory_repository: FavoriteExternalDirectoryRepository, workspace_config: Config
+    ):
+        self.favorite_external_directory_repository = favorite_external_directory_repository
+        self.workspace_config = workspace_config
+
+    def add_favorite(self, directory_path: str, workspace: str) -> FavoriteExternalDirectoryDTO:
+        workspace_conf = get_workspace_from_config(self.workspace_config, workspace)
+        if workspace:
```
MartinBelthle: Minor but this is unecessary as inside your endpoint you ensure the workspace arg is a string of at least one element so we never enter this case

antarest/favorite/web.py
```diff
@@ -61,4 +69,47 @@ def delete_favorite_directory(favorite_directory_service: FavoriteDirectoryServi
...
+        return favorite_external_directory_service.add_favorite(workspace=workspace, directory_path=path)
+
+    @bp.delete(
+        "/favorites/external_directories",
+        summary="Delete an external directory from the list of favorite directories",
+        status_code=HTTPStatus.ACCEPTED,
+    )
+    def delete_favorite_external_directory(
+        favorite_external_directory_service: FavoriteExternalDirectoryServiceDep,
+        workspace: Annotated[str, Field(min_length=1, examples=["workspace_name"])],
```
MartinBelthle: Seems you could introduce this as a  `WorkspaceType` and a `PathType` TypeAliased, to use them inside your 2 endpoints instead of copying the code.

antarest/favorite/service.py
```diff
@@ -110,3 +126,54 @@ def delete_favorite(self, directory_uuid: str) -> None:
...
+            if not is_folder_safe(workspace_conf, directory_path):
+                raise HTTPException(
+                    status_code=http.HTTPStatus.BAD_REQUEST,
+                    detail=f"Directory {directory_path} is not safe",
+                )
+
+            relative_path = Path.relative_to(Path(directory_path), workspace_conf.path / workspace)
```
MartinBelthle: Here, your code forces the user to give a workspace like `external` and directory_path to be `external/my_path`. But this is redundant, the user should only give `my_path` in directory_path and we should try to build the path `external/my_path` ourselves

MartinBelthle (changes requested):
We need to change the logic a bit for the `path` argument inside the endpoint

antarest/favorite/service.py
```diff
@@ -110,3 +126,52 @@ def delete_favorite(self, directory_uuid: str) -> None:
...
+
+    def list_favorites(self) -> list[FavoriteExternalDirectoryDTO]:
+        favorite_dtos = [favorite.to_dto() for favorite in self.favorite_external_directory_repository.get_all()]
+        return favorite_dtos
+
+    def delete_favorite(self, workspace: str, path: str) -> None:
+        """
+        Delete a favorite external directory from the current user list of favorites
+        Args:
+            workspace:
```
MartinBelthle: Minor but the docstring is halfway done. Either you finish it you can remove it as what the method does it pretty straight forward

tests/integration/favorite_blueprint/test_favorite_external_directory.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_add_favorite_external_directory_failure_folder_not_safe(admin_client: TestClient, tmp_path: Path):
+    # checking that the favorite external directory is not added if the path is not safe
+    workspace_name = "ext"
+    path = tmp_path / "external_workspace" / workspace_name
+    expected_favorite_external_directory = {"workspace": workspace_name, "path": path.as_posix()}
+    response = admin_client.post("/v1/favorites/external-directories", params=expected_favorite_external_directory)
+    assert response.status_code == 400
+    assert response.json()["description"] == f"Directory {path.as_posix()} is not safe"
```
MartinBelthle, line 101: Now the tests look good :)

It could be good to add one last test with 2 users to ensure that:  when user1 adds a favorite `path/to/directory`,
- 1) the 2nd user cannot see it
- 2) ensure the 2nd user is able to also add the same path `path/to/directory` to its favorites

To do this, you can probably use the `user_access_token` fixture to have a 2nd user other than the admin

tests/favorite_study/test_favorite_external_directory.py
```diff
@@ -0,0 +1,200 @@
...
+
+    # Create the service
+    service = FavoriteExternalDirectoryService(
+        favorite_external_directory_repository=mock_repository, workspace_config=config
+    )
+
+    return service, mock_repository, config, workspace_name, workspace_path
+
+
+def test_add_favorite_external_directory_failure_when_path_does_not_exist(tmp_path: Path):
```
MartinBelthle, line 61: This unit test sets Mocks and create files when it's not needed so it's not easy to understand what's happening. For instance, you can write this which is shorter and the test still works:
```python
def test_add_favorite_external_directory_failure_when_path_does_not_exist(tmp_path: Path):
    # Trying to add an external directory whose path doesn't exist to the favorites
    mock_not_a_path = Path("notapath") / "to" / "favorite" / "directory"
    workspace = "validspace"
    config = Config(storage=StorageConfig(tmp_dir=tmp_path))
    config.storage.workspaces.update({workspace: WorkspaceConfig(path=tmp_path)})

    favorite_service = FavoriteExternalDirectoryService(Mock(), workspace_config=config)

    with current_user_context(DEFAULT_ADMIN_USER):
        with pytest.raises(DirectoryNotFoundError, match=f"404: Directory '{mock_not_a_path.as_posix()}' not found"):
            favorite_service.add_favorite(mock_not_a_path.as_posix(), workspace)


```

tests/favorite_study/test_favorite_external_directory.py
```diff
@@ -0,0 +1,200 @@
...
+    favorite_service = FavoriteExternalDirectoryService(
+        mock_favorite_external_directory_repository, workspace_config=config
+    )
+
+    with current_user_context(DEFAULT_ADMIN_USER):
+        with pytest.raises(DirectoryNotFoundError, match=f"404: Directory '{mock_not_a_path.as_posix()}' not found"):
+            favorite_service.add_favorite(mock_not_a_path.as_posix(), workspace)
+
+
+def test_add_favorite_external_directory_failure_when_workspace_does_not_exist(tmp_path: Path):
```
MartinBelthle, line 75: Same remark here:

```python

def test_add_favorite_external_directory_failure_when_workspace_does_not_exist(tmp_path: Path):
    # Trying to add an external directory to a non-existing workspace
    config = Config(storage=StorageConfig(tmp_dir=tmp_path))
    path_to_favorite = PurePosixPath(tmp_path / "path" / "to" / "favorite" / "directory")
    non_existing_workspace = ""
    favorite_service = FavoriteExternalDirectoryService(Mock(), workspace_config=config)

    with pytest.raises(WorkspaceNotFound, match=f"422: Workspace {''} not found"):
        favorite_service.add_favorite(path_to_favorite.as_posix(), non_existing_workspace)


```

antarest/favorite/service.py
```diff
@@ -110,3 +126,52 @@ def delete_favorite(self, directory_uuid: str) -> None:
...
+        if workspace_conf:
+            if not is_folder_safe(workspace_conf, directory_path):
+                raise HTTPException(
+                    status_code=http.HTTPStatus.BAD_REQUEST,
+                    detail=f"Directory {directory_path} is not safe",
+                )
+
+            if (workspace_conf.path / directory_path).exists():
+                favorite_external_directory = FavoriteExternalDirectory(
+                    path=str(Path(directory_path)), workspace=workspace, user_id=get_user_impersonator()
```
MartinBelthle: Could you try to just write `path=directory_path`, and if the Windows test fail you can put back

MartinBelthle (changes requested):
Last changes

alembic/versions/80fdf2408ede_add_favorite_external_directory_table.py
```diff
@@ -0,0 +1,30 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table('favorite_external_directory',
+                sa.Column('workspace', sa.String(length=255), nullable=False),
+                sa.Column('path', sa.String(length=255), nullable=False),
+        sa.Column('user_id', sa.Integer(), nullable=False),
+                sa.ForeignKeyConstraint(['user_id'], ['users.id'], ),
```
MartinBelthle: This should be mapped to `identities.id` instead of `users.id`, as you did in the `model.py` file

MartinBelthle (commented):
1 change to be done inside the alembic migration, after this we can merge :)

MartinBelthle (dismissed):

antarest/favorite/repository.py
```diff
@@ -137,3 +137,39 @@ def save(self, directory: FavoriteDirectory) -> FavoriteDirectory:
...
+
+class FavoriteExternalDirectoryRepository:
+    def __init__(self, session: Session | None = None) -> None:
+        self._session = session
+
+    @property
+    def session(self) -> Session:
+        if self._session is None:
+            return db.session
+        return self._session
```
smailio, line 150: why you need this property ?

antarest/favorite/service.py
```diff
@@ -110,3 +126,66 @@ def delete_favorite(self, directory_uuid: str) -> None:
...
+        """
+        Add an external directory to the current user list of favorites
+        Args:
+            directory_path: the path of the external directory that we want to add as a favorite
+            workspace: the workspace of the external directory that we want to add as a favorite
+
+        Returns: the DTO of the saved favorite external directory
+
+        """
+        workspace_conf = get_workspace_from_config(self.workspace_config, workspace)
```
smailio, line 147: this function raises already if you give a wrong workspace no need for the if just bellow

antarest/favorite/model.py
```diff
@@ -21,6 +24,13 @@
 
 logger = logging.getLogger(__name__)
 
+WorkspaceType: TypeAlias = Annotated[str, Field(min_length=1)]
+
+DirectoryPathType: TypeAlias = Annotated[
+    str,
+    Field(min_length=1),
+]
+
```
smailio, line 33: nice

antarest/favorite/service.py
```diff
@@ -110,3 +126,66 @@ def delete_favorite(self, directory_uuid: str) -> None:
...
+        if workspace_conf:
+            if not is_folder_safe(workspace_conf, directory_path):
+                raise HTTPException(
+                    status_code=http.HTTPStatus.BAD_REQUEST,
+                    detail=f"Directory {directory_path} is not safe",
+                )
+
+            if (workspace_conf.path / directory_path).exists():
+                favorite_external_directory = FavoriteExternalDirectory(
+                    path=str(Path(directory_path)), workspace=workspace, user_id=get_user_impersonator()
```
smailio: How come this works on windows, shouldn't this be to_posix ? This serialize to a path with backslaches no ?

antarest/favorite/service.py
```diff
@@ -110,3 +126,66 @@ def delete_favorite(self, directory_uuid: str) -> None:
...
+                )
+                dto = self.favorite_external_directory_repository.save(favorite_external_directory).to_dto()
+                return dto
+            else:
+                raise DirectoryNotFoundError(f"{directory_path}")
```
smailio: Can you make this check and this raise at the top, so we don't have this big bloc of code nested too deep. 

```python
full_directory_path = (workspace_conf.path / directory_path)
if not full_directory_path.exists():
    raise DirectoryNotFoundError(f"{directory_path}")
favorite_external_directory = FavoriteExternalDirectory(
...

``

File: antarest/favorite/service.py (hunk 30 above)
Comment (smailio) [reply antarest/favorite/service.py]: I mean on windows the str(Path(directory_path)) would serialize to a path with backslaches.

File: antarest/favorite/model.py (hunk 29 above)
Comment (mehdiwahada) [reply antarest/favorite/model.py:33]: It was asked by Martin. Thank him for that.

File: antarest/favorite/service.py (hunk 30 above)
Comment (mehdiwahada) [reply antarest/favorite/service.py]: if done as a posix, it'll always return a path with slashes, be it on Windows or Linux, which will return errors if tested on windows as I've tested it.

File: antarest/favorite/repository.py (hunk 27 above)
Comment (mehdiwahada) [reply antarest/favorite/repository.py:150]: It's an habit, since I saw it on every repository file. We'll maybe need to talk about this to Sylvain.

File: alembic/versions/80fdf2408ede_add_favorite_external_directory_table.py [hunk 32]
```diff
@@ -0,0 +1,30 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table('favorite_external_directory',
+                sa.Column('workspace', sa.String(length=255), nullable=False),
+                sa.Column('path', sa.String(length=255), nullable=False),
+        sa.Column('user_id', sa.Integer(), nullable=False),
+                sa.ForeignKeyConstraint(['user_id'], ['identities.id'], ),
```
smailio: In model.py you declare an ondelete="CASCADE". I'm not sure if you forgot to add it here or if it isn't necessary ?

MartinBelthle: Indeed, here we should use cascade and also declare the foreign key name

antarest/favorite/service.py
MartinBelthle: Indeed, so you can remove the `if workspace_conf` and the `else raise WorkspaceNotFound`

antarest/favorite/service.py
mehdiwahada: It has been done. I'm pushing it.

smailio (approved):

PR #3243, opened 2026-06-09
refactor(output): simplify `FileStudyOutputs` class

MartinBelthle:


antarest/output/storage/v2/storage.py
```diff
@@ -262,14 +262,18 @@ def list_outputs(self, study_id: str) -> list[OutputMetadata]:
...
         Get the list of output for a study.
         """
-        metadata = self._repository.get_output_metadata(study_id, output_id)
-        if metadata is None:
-            raise OutputNotFound(output_id)
-        return _db_metadata_to_details(metadata)
+        result = []
+        for output_metadata in self._repository.search_output_metadata(study_id):
+            output_id = output_metadata.output_name
+            db_metadata = self._repository.get_output_metadata(study_id, output_id)
```
sylvlecl: Why do we need to get again the object? Cannot we use the one from the first request and avoid making other requests ?

antarest/study/storage/rawstudy/model/filesystem/root/output/simulation/mode/mcall/digest.py
```diff
@@ -86,38 +108,18 @@ def load(
         expanded: bool = False,
         formatted: bool = True,
     ) -> JSON:
-        df = self._parse_digest_file()
+        df = _parse_digest_file(self.config.path)
         return df.to_pandas().to_dict(orient="split", index=False)
 
-    def get_ui(self) -> DigestUI:
+    @staticmethod
+    def get_ui(file_path: Path) -> DigestUI:
```
sylvlecl: We could rename to "parse" something, to make it clear it parses the file.

Could be a free function in antarest.output.filestudy, too

antarest/study/storage/rawstudy/model/helpers.py
```diff
@@ -9,22 +9,23 @@
...
-            config_path = [
-                "output",
-                output_id,
-                "about-the-study",
-                "parameters",
-            ]
-            config = study.tree.get(config_path)
-            return config
-        return study.tree.get(["settings", "generaldata"])
+    def get_input_config(study_path: Path) -> JSON:
```
sylvlecl: We could use this PR to rename them to parse_*** too

Also, the wrapper class is not really useful, we can just have top level free functions ...

Also, the output part would have a better place in antarest.output.filestudy which contains utility functions for parsing file study files.

sylvlecl (commented):
Looks good overall, just suggestions about naming and place of things

antarest/output/storage/file/storage.py
```diff
@@ -422,7 +423,7 @@ def archive_study_output(self, study_id: str, output_id: str) -> None:
     def _remote_unarchive(self, output_id: str, study_outputs: FileStudyOutputs) -> None:
         dest, src = _output_paths(study_outputs.outputs_path, output_id)
         self._remote_executor.execute_remote_task(
-            f"unarchive_{study_outputs.study_workspace}",
+            f"unarchive_{output_id}",
```
sylvlecl: Actually we still need the workspace for this unfortunately :/ the remove workers rely on that naming

sylvlecl (changes requested):

antarest/output/storage/v2/storage.py
MartinBelthle: Indeed, no need for this

antarest/study/storage/rawstudy/model/filesystem/root/output/simulation/mode/mcall/digest.py
MartinBelthle: I renamed it `parse_file_for_ui` but kept it here just to centralize the logic

PR #3245, opened 2026-06-10
feat(output): add new output storage mode

MartinBelthle:
On my local env, I can now retrieve an output for my DB study, but I cannot fetch its matrices. I'll adress this issue in a next PR


antarest/output/storage/output_storage.py
```diff
@@ -36,6 +36,7 @@
 class OutputStorageType(StrEnum):
     IN_STUDY_FILE_TREE = "IN_STUDY_FILE_TREE"
     V2 = "V2"
+    OUTSIDE_STUDY_FILE_TREE = "OUTSIDE_STUDY_FILE_TREE"
```
sylvlecl: really a detail but wording "out of study" would be more consistent with "in study" (we did not use "inside")

or we could just use "file tree" ? implicitly not inside the study

tests/integration/assets/config.template.yml
```diff
@@ -21,10 +21,13 @@ storage:
     storage:
       database_mode_enabled: true
   output:
-    enable: true
-    default: false
-    archive_dir: {{output_archive_dir}}
-    variables_dir: {{output_variables_dir}}
+    default_storage_type: "IN_STUDY_FILE_TREE"
```
sylvlecl, line 24: nice, it looks a lot better

antarest/output/adapters.py
```diff
@@ -42,6 +43,18 @@ def get_outputs(self, study_id: str) -> FileStudyOutputs:
     return StudyServiceAdapter()
 
 
+def study_service_as_outside_study_file_outputs_provider(config: Config) -> IFileOutputsProvider:
```
sylvlecl: it's not anymore an adapter of the study service here, it does not use it, so we could move that to the output storage implementation.

BUT I think we should still use the study service to check permissions, actually, in that case it should indeed stay here

sylvlecl (changes requested):
Looks good overall:the only "blocking" point is that I think we should go through the study service to check permissions, for the out of study implementation

antarest/output/storage/output_storage.py
MartinBelthle: I prefer to keep it explicit so I'll use `OUT_OF_STUDY_FILE_TREE`

antarest/output/adapters.py
sylvlecl: As discussed, I was mistaken: this is already done at the service level, it's not at all the responsibility of the storage layer to perform those checks

sylvlecl (approved):

PR #3261, opened 2026-06-17
refactor(config): use pydantic when loading config

olivierPigeon-RTE:


antarest/core/config.py
```diff
@@ -156,72 +120,53 @@ class V2OutputStorageConfig:
...
 
-    @classmethod
-    def from_dict(cls, data: JSON) -> "OutOfStudyFileOutputStorageConfig":
-        defaults = cls()
-        storage_dir = Path(data.get("storage_dir", str(defaults.storage_dir)))
-        storage_dir.mkdir(parents=True, exist_ok=True)
-        return OutOfStudyFileOutputStorageConfig(storage_dir=storage_dir)
+    @model_validator(mode="after")
+    def _create_storage_dir(self) -> "OutOfStudyFileOutputStorageConfig":
+        self.storage_dir.mkdir(parents=True, exist_ok=True)
```
sylvlecl, line 133: Should be created by the storage when it's created, not just when we read the config.
But not related to that PR

sylvlecl (dismissed):
Looks good ! Thanks

PR #3275, opened 2026-06-23
feat(ui-tablemodes): use API instead of localStorage

skamril:
also create table modes inside localStorage with the API

ANT-4966


webapp/src/utils/zodUtils.ts
```diff
@@ -38,3 +38,9 @@ import { z } from "zod";
 export function nullishToOptional<T extends z.ZodTypeAny>(schema: T) {
   return schema.nullish().transform((v) => v ?? undefined);
 }
+
+export function uniqueArray<T extends z.ZodTypeAny>(schema: T) {
+  return z.array(schema).refine((values) => new Set(values).size === values.length, {
+    error: "Array must contain unique values",
+  });
+}
```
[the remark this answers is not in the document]
skamril: `message` is deprecated.

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/route.tsx
```diff
@@ -0,0 +1,82 @@
...
+    storage.removeItem("studies.model.tableMode.templates");
+
+    if (Array.isArray(oldTableModes)) {
+      oldTableModes.forEach((value) => {
+        if (tableModeCreationSchema.safeParse(value).success) {
+          createTableMode.mutate(value);
+        }
+      });
+    }
+  });
```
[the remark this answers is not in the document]
skamril: Not necessary.

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/binding-constraints/-hooks/useDeleteBindingConstraint.ts
```diff
@@ -54,7 +54,7 @@ function useDeleteBindingConstraint() {
       const nextConstraint = prevList
         ? getNextItemAfterDeletion(
             prevList,
-            prevList.findIndex((c) => c.id === constraintToDeleteId),
+            prevList.findIndex((item) => item.id === constraintToDeleteId),
```
hdinia, line 57: **nitpick:** keep short notation or use full name but describe the object (`c` or `constraint`)
`prevList.findIndex((constraint) => constraint.id === constraintToDeleteId)`

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/-utils.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import type { RouterListViewItem } from "@/components/page/list/RouterListView";
+import { isQueryListItemOptimistic } from "@/queries/utils";
+import type { TableMode } from "@/services/api/tablemode/types";
+import { linkOptions } from "@tanstack/react-router";
+import { sortByProp } from "ramda-adjunct";
+
+export function tableModesToList(tableModes: TableMode[]): RouterListViewItem[] {
+  const list = tableModes.map((tm) => ({
```
hdinia: **nitpick**: `tm` -> `tableMode`

webapp/public/locales/fr/main.json
```diff
@@ -719,11 +719,15 @@
...
-  "study.tableMode.dialog.delete.text": "Êtes-vous sûr de vouloir supprimer la table : {{name}}?",
+  "study.tableModes": "Table Modes",
+  "study.tableModes.empty": "Aucun table mode disponible",
+  "study.tableModes.dialog.add.title": "Ajouter une table mode",
+  "study.tableModes.dialog.edit.title": "Modifier une table mode",
+  "study.tableModes.dialog.delete.text": "Êtes-vous sûr de vouloir supprimer la table mode : {{name}}?",
+  "study.tableModes.detail.notFound": "Aucune table mode trouvée avec l'ID : {{id}}",
+  "study.tableModes.create.error": "Échec de la création de la table mode : {{name}}",
+  "study.tableModes.update.error": "Échec de la mise à jour de la table mode : {{name}}",
+  "study.tableModes.delete.error": "Échec de la suppression de la table mode : {{name}}",
```
hdinia: issue : fix grammatically inconsistent strings like "Aucun table mode disponible" "Aucune table mode trouvée avec l'ID : {{id}}"(feminine vs masculine)

suggestion : the spec required to rename those and add "mode" ? the previous keys where better with only "table" / "tableaux" consider dropping the suffix (question: what means "mode" in that context ? seems to have no specific sense to me)

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/route.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const { t } = useTranslation();
+  const { openDialog } = useDialog();
+  const createTableMode = useCreateTableMode();
+
+  // Workaround to migrate table modes stored in localStorage to the new backend system (since v2.32.0).
+  // It can be removed after a certain period of time when we are confident that most users have migrated.
+  useMount(() => {
+    const oldTableModes = storage.getItem("studies.model.tableMode.templates");
+    storage.removeItem("studies.model.tableMode.templates");
```
hdinia: issue : Are you sure about that ? Seems to me the "migration" deletes localStorage before the API writes succeed. if a create fails, the user's presets are gone no ? Please check and if it's confirmed remove the key only after the mutations resolve.

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/$tableModeId/-components/utils.ts
hdinia, line 1: issue: Column lists duplicated between `schemas.ts` enums and this file (TABLE_MODE_COLUMNS_BY_TYPE + TABLE_MODE_TYPES). 
Derive from the schemas to keep one source of truth. Unless you have a specific reason to do so ?

webapp/src/components/TableModeDataForm.tsx
hdinia, line 1: suggestion: `TableModeDataForm.tsx` derives gridColumns via useState + useEffect, that could cause a one-frame "No data" flash before the grid appear, you can replace with useMemo


```ts
const gridColumns = useMemo<DataGridFormProps<TableModeData>["columns"]>(() => {
  const data = res.data || {};
  const rowNames = Object.keys(data);

  if (rowNames.length === 0) {
    return [];
  }

  const columnNames = Object.keys(data[rowNames[0]]);

  return columns
    .filter((col) => columnNames.includes(col))
    .map((col) => ({ title: startCase(col), id: col }) satisfies GridColumn);
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [res.data, columnsDep]);

```

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/-hooks/useUpdateTableMode.ts
```diff
@@ -0,0 +1,84 @@
...
+      if (prevTableMode) {
+        queryClient.setQueryData(queryListKey, (old = []) => {
+          return old.map((tableMode) =>
+            tableMode.id === prevTableMode.id ? prevTableMode : tableMode,
+          );
+        });
+      }
+    },
+    onSuccess: (updatedTableMode) => {
+      const { queryKey: queryListKey } = tableModeQueries.list();
```
hdinia: nitpick : re-declares `const { queryKey: queryListKey } = tableModeQueries.list();`, shadowing the outer binding.

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/$tableModeId/-components/BaseTableModeDialog.tsx
hdinia, line 1: nitpick: `BaseTableModeDialog` exports a component named `TableModeFormDialog` file and component names disagree. Align them.

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/$tableModeId/-components/AddTableModeDialog.tsx
hdinia, line 1: nitpick: `AddTableModeDialog` puts id: "" in defaultValues and submits a full TableMode. `tableModeCreationSchema.parse` strips the stray id, so it works but the typing is looser than needed.

webapp/src/services/api/tablemode/schemas.ts
```diff
@@ -0,0 +1,251 @@
...
+      type: z.literal(type),
+      columns: columnsSchema,
+    })
+    .transform(({ type, columns }) => ({
+      table_type: type,
+      table_columns: columns,
+    }));
+}
+
+export const tableModeSchema = z.discriminatedUnion("table_type", [
```
hdinia, line 214: be aware of: you're running discriminatedUnion over transformed members. Zod 4 supports this, and the comment shows it was deliberate but just make sure there's a test exercising a round-trip parse, because discriminatedUnion over transform is the kind of thing that breaks silently on a future zod bump.

webapp/src/utils/zodUtils.ts
```diff
@@ -38,3 +38,9 @@ import { z } from "zod";
 export function nullishToOptional<T extends z.ZodTypeAny>(schema: T) {
   return schema.nullish().transform((v) => v ?? undefined);
 }
+
+export function uniqueArray<T extends z.ZodTypeAny>(schema: T) {
+  return z.array(schema).refine((values) => new Set(values).size === values.length, {
+    error: "Array must contain unique values",
+  });
```
hdinia: Nitpick : This generic seems over-broad `<T extends z.ZodTypeAny>` it advertises "works for any element" but `new Set(values)` dedupes by reference, so it silently fails to detect duplicates for arrays of objects/arrays. Here it's only ever used with string enums, so it's correct in practice, but the signature over-promises. I'd either:
- constrain it to primitives, e.g. `<T extends z.ZodType<string | number | boolean>>`, or
- leave a one-line comment that it's primitive-only.

Minor: the error string is hardcoded English (won't be i18n'd), but these schema errors aren't surfaced verbatim to users here, so that's negligible. `superRefine` with a path would tell you which element duplicated, but that's maybe overkill here.

hdinia (changes requested):

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/binding-constraints/-hooks/useDeleteBindingConstraint.ts
skamril: This is not a constraint here, but a RouterListViewItem.

webapp/src/services/api/tablemode/schemas.ts
skamril: Do you have a link?

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/route.tsx
skamril: To handle this correctly, we'd need to still show the localStorage data on error so the user doesn't recreate them manually. But that adds complexity for an edge case that's not that important, users already know table modes data is unstable and can disappear from the browser at any time, unlike now where it will be persisted in DB. Also, if there's a server error with table modes, this code won't run at all.

hdinia: > Also, if there's a server error with table modes, this code won't run at all.

It's not about the list query. If one individual create POST fails that preset is then gone from both localStorage and DB, silently. List-OK is not create-OK.

> To handle this correctly, we'd need to still show the localStorage data on error so the user doesn't recreate them manually.

No, that's a much bigger feature than what I asked. Just await the mutations and only remove the key at the end, keeping the entries that failed so they retry next mount. Don't keep the whole key on failure or you'll re-create the succeeded ones and duplicate them. ( + use `mutateAsync` to avoid a router.navigate trigger on each onMutate using forEach(...mutate))

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/$tableModeId/-components/utils.ts
skamril: Fixed

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/route.tsx
hdinia:
```ts
useMount(async () => {
  const oldTableModes = storage.getItem("studies.model.tableMode.templates");

  if (!Array.isArray(oldTableModes)) {
    return;
  }

  // Keep only entries that fail to migrate, so they retry on next mount.
  const failed: unknown[] = [];

  for (const value of oldTableModes) {
    if (!tableModeCreationSchema.safeParse(value).success) {
      continue;
    }

    try {
      await createTableMode.mutateAsync(value);
    } catch {
      failed.push(value);
    }
  }

  if (failed.length > 0) {
    storage.setItem("studies.model.tableMode.templates", failed);
  } else {
    storage.removeItem("studies.model.tableMode.templates");
  }
});
```

skamril: Yes, I understood the first comment, but in your example, if a case fails, it won't be displayed in the list even though it's still kept in localStorage.

So if needed, the user will recreate it, and it could potentially be added as a duplicate on the next mount render, resulting in a confusing duplicate for the user.

That's why I didn't overthink it, they're already aware that table modes can disappear, so I don't need to handle that error case, especially since there's also error handling to prevent certain failed cases from potentially repeating on every view launch (since I'm validating the data, any error outside of a server error will very likely always occur).

And in case of a server error, useMount wouldn't be called (so the localStorage key won't be deleted) since the GET call would block the view.

PS: adding the table modes from localStorage to the DB was not a request, I did it on my own initiative.

hdinia: So you'd rather keep a bug that can lead to silent data loss, to avoid a potential duplicate if a user re-create it which will most likely not even happen ? It's far less bad than permanent data loss, and the fix is cheap.

hdinia (dismissed):

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/$tableModeId/-components/MigrationTableModes.tsx
```diff
@@ -0,0 +1,233 @@
...
+    else {
+      storage.setItem(
+        LOCAL_STORAGE_KEY,
+        failed.map(([tableModeCreation]) => tableModeCreation),
+      );
+    }
+
+    setState({
+      status: failed.length > 0 ? "error" : "success",
+      migrated: oldTableModes.length - failed.length,
```
hdinia: migrated count includes skipped invalid entries 

Entries that fail safeParse are continued, not migrated, but also not in failed. So this expression counts them as migrated. e.g : 10 entries, 2 invalid, 0 errors -> toast says "10 migrated" when only 8 actually were.

The state.migrated incremented in the loop is already the correct value. It's intentional, or should you use that instead of re-deriving from length - failed?

hdinia (changes requested):

webapp/src/routes/_authenticated/studies/$studyId/explore/table-modes/$tableModeId/-components/MigrationTableModes.tsx
```diff
@@ -0,0 +1,233 @@
...
+          {/* Error */}
+          {state.status === "error" && (
+            <>
+              <Typography variant="body2" color="error" sx={{ mt: 1 }}>
+                {t("study.tableModes.migration.error", { count: state.failed.length })}
+              </Typography>
+              <List dense sx={{ overflow: "auto" }}>
+                {state.failed.map(([tableModeCreation, error]) => (
+                  <ListItem
+                    key={tableModeCreation.name}
```
hdinia: no `key` collision risk ?

hdinia (changes requested):

hdinia (dismissed):

hdinia (approved):

PR #3280, opened 2026-06-24
feat(reserves): support thermal reserves

MartinBelthle:
4 new endpoints introduced:
- GET `/studies/{uuid}/areas/{area_id}/reserves/symmetries/thermals`
- PUT `/studies/{uuid}/areas/{area_id}/reserves/symmetries/thermals`
- GET `/studies/{uuid}/areas/{area_id}/reserves/certifications/thermals`
- PUT `/studies/{uuid}/areas/{area_id}/reserves/certifications/thermals`


antarest/study/storage/rawstudy/model/filesystem/config/thermal_reserve_certifications.py
```diff
@@ -0,0 +1,64 @@
...
+    participation_cost_off: float
+
+    def reserve_id(self) -> ReserveDefinitionId:
+        return ReserveDefinitionId(transform_name_to_id(self.reserve))
+
+    def to_model(self) -> ThermalReserveCertification:
+        return ThermalReserveCertification.model_validate(self.model_dump(exclude={"reserve"}))
+
+
+def serialize_thermal_reserve_certifications(
```
sylvlecl: would it not be more clear to have a larger "file data" model consistent with the scope of the file (certifications + symmetries), and return that one, and use the same one when updating symmetries ?

antarest/study/business/areas/reserve_certification_management.py
```diff
@@ -0,0 +1,58 @@
...
+    def set_certifications(
+        self,
+        study: StudyInterface,
+        area_id: str,
+        data: ReserveCertifications,
+    ) -> ReserveCertifications:
+        commands: list[ICommand] = []
+
+        # Thermals part
+        if data.thermals:
```
sylvlecl, line 40: if the list is empty, we want to replace by an empty list, otherwise we will never be able to remove all certifications

antarest/study/business/areas/reserve_symmetries_management.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def get_symmetries(self, study: StudyInterface, area_id: str) -> ReserveSymmetries:
+        thermals = study.get_study_dao().get_thermal_reserve_symmetries(area_id)
+        return ReserveSymmetries(thermals=thermals)
+
+    def set_symmetries(self, study: StudyInterface, area_id: str, data: ReserveSymmetries) -> ReserveSymmetries:
+        commands: list[ICommand] = []
+
+        # Thermal part
+        if data.thermals:
```
sylvlecl: Same here: if empty we want to replace with an empty list

antarest/study/storage/variantstudy/model/command/create_area.py
```diff
@@ -108,15 +108,18 @@ def _apply_dao(self, study_data: StudyDao, listener: ICommandListener | None = N
                 study_data.save_hydro_max_hourly_pump_power({area_id: constants.get_hydro_max_hourly_pump_power()})
                 study_data.save_hydro_max_daily_gen_energy({area_id: constants.get_hydro_max_daily_gen_energy()})
                 study_data.save_hydro_max_daily_pump_energy({area_id: constants.get_hydro_max_daily_pump_energy()})
+
         # Matrices
         study_data.save_load({area_id: null_matrix})
         study_data.save_solar({area_id: null_matrix})
         study_data.save_wind({area_id: null_matrix})
+        study_data.save_misc_gen({area_id: constants.get_default_miscgen()})
```
sylvlecl, line 116: was there a bug here ?

antarest/study/dao/api/thermal_reserve_certification_dao.py
```diff
@@ -0,0 +1,33 @@
...
+# This file is part of the Antares project.
+from abc import ABC, abstractmethod
+
+from antarest.study.business.model.thermal_reserve_certification_model import (
+    ThermalReserveCertificationMapping,
+)
+from antarest.study.dao.common import AreaId
+
+
+class ReadOnlyThermalReserveCertificationDao(ABC):
```
sylvlecl: honestly we could have kept only one file for the whole reserves stuff, I think :) 
We will end up with quite a lot of files with the other stuff (hydro and STS)

antarest/study/dao/database/models/thermal_reserve_symmetries.py
```diff
@@ -0,0 +1,31 @@
...
+metadata = Base.metadata
+
+THERMAL_RESERVE_SYMMETRIES_TABLE = Table(
+    "thermal_reserve_symmetries",
+    metadata,
+    Column("study_id", String(36), nullable=False, primary_key=True),
+    Column("area_id", String(255), nullable=False, primary_key=True),
+    Column("thermal_id", String(255), nullable=False, primary_key=True),
+    Column("symmetries", String(), nullable=False),
+    ForeignKeyConstraint(
```
sylvlecl, line 26: what happens when we delete a reserve ?
It seems to me that we should delete them in symmetries too.
To implement it, maybe we should a different DB structure ?

However, maybe we also want to remove symmetries that only contain 1 item after change ... 
Don't know if it could cause an error on the simulator side.

antarest/study/dao/api/study_dao.py
```diff
@@ -883,3 +903,21 @@ def get_reserve_need(self, area_id: str, reserve_id: str) -> pl.DataFrame:
     @override
     def get_all_reserve_needs(self) -> ReserveNeedsMapping:
         return self._adaptee.get_all_reserve_needs()
+
+    @override
+    def get_all_thermal_reserve_certifications(self) -> dict[AreaId, ThermalReserveCertificationMapping]:
+        return self._adaptee.get_all_thermal_reserve_certifications()
+
+    @override
+    def get_thermal_reserve_certifications(
```
sylvlecl, line 912: We should mention what's expected when there is no certification for an existing reserve:
is the reserve absent from the mapping, or do we expect to have all existing reserves instead, with possibly empty dicts? (the second option is more clear from the interface point of view).

Same in the save method actually: do we allow to have absent reserves ? In that case does that mean that we must empty the list for that reserve ? we need to explain it in the docstring

sylvlecl: Seems to me that if a reserve is absent, we will not update it, in the DB implementation at least.
Same for the get.

It can be a little misleading, because it now looks partially like an update.

Maybe it's simpler to leave it this way but we need to document it in the docstring and probably in the API doc too

antarest/study/storage/variantstudy/model/command/replace_thermal_reserve_certifications.py
```diff
@@ -0,0 +1,77 @@
...
+    CommandName,
+    CommandOutput,
+    command_succeeded,
+)
+from antarest.study.storage.variantstudy.model.command.icommand import ICommand
+from antarest.study.storage.variantstudy.model.command_listener.command_listener import ICommandListener
+from antarest.study.storage.variantstudy.model.model import CommandDTO
+
+
+class ReplaceThermalReserveCertifications(ICommand):
```
sylvlecl, line 36: Same here, if we keep the behaviour we should document it: we could think that we will replace all the certifications of the area, whereas we will only replace the ones of reserves that are present in the mapping

sylvlecl (changes requested):
I think at least a few points need to be fixed (possibility to remove certifications and symmetries of a reserve), and other points need to be clarified (behaviour when we delete a reserve, a cluster, and handling of missing reserves in dicts).

antarest/study/storage/variantstudy/model/command/create_area.py
MartinBelthle: Not at all, it was to separate the reserves code from the rest a bit more clearly

antarest/study/dao/database/database_reserve_symmetries_dao.py
```diff
@@ -0,0 +1,127 @@
...
+            self._db_session.execute(stmt)
+            # Then, insert the new values
+            if values:
+                self._db_session.execute(insert(_THERMAL_TABLE), values)
+        except IntegrityError as e:
+            thermals = {area_id: list(thermal_dict) for area_id, thermal_dict in data.items()}
+            self.get_impl().raise_the_right_thermal_exception(thermals, exc=e)
+        self._db_session.commit()
+
+    @staticmethod
```
sylvlecl: Implementation detail: there is no real benefit of using `@staticmethod` when we can just have a free function, easier to call

antarest/study/dao/api/reserve_symmetries_dao.py
```diff
@@ -0,0 +1,35 @@
...
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_thermal_reserve_symmetries(self, area_id: AreaId) -> dict[ThermalId, ReserveSymmetries]:
+        raise NotImplementedError()
+
+
+class ReserveSymmetriesDao(ReadOnlyReserveSymmetriesDao):
+    @abstractmethod
+    def save_thermal_reserve_symmetries(self, data: ThermalReserveSymmetriesMapping) -> None:
```
sylvlecl, line 42: We still need to document the behaviour which is really not obvious:
- if an area is absent from the dict, its data is not modified
- if a thermal cluster is absent from the inner dict, its data is removed

antarest/study/dao/api/reserve_symmetries_dao.py
```diff
@@ -0,0 +1,35 @@
...
+from antarest.study.dao.common import (
+    AreaId,
+    ThermalId,
+    ThermalReserveSymmetriesMapping,
+)
+
+
+class ReadOnlyReserveSymmetriesDao(ABC):
+    @abstractmethod
+    def get_all_thermal_reserve_symmetries(self) -> ThermalReserveSymmetriesMapping:
```
sylvlecl, line 24: We need to document that areas and clusters that have no data will be absent from the returned mappings, it's not obvious

antarest/study/storage/rawstudy/model/filesystem/config/thermal_reserve_certifications.py
```diff
@@ -0,0 +1,64 @@
...
+from antarest.study.business.model.reserve_definition_model import ReserveDefinitionId
+from antarest.study.business.model.thermal_reserve_certification_model import ThermalReserveCertification
+from antarest.study.storage.rawstudy.model.filesystem.config.identifier import transform_name_to_id
+
+
+def parse_thermal_reserves_certifications(
+    data: dict[str, Any],
+) -> dict[ReserveDefinitionId, dict[str, ThermalReserveCertification]]:
+    result: dict[ReserveDefinitionId, dict[str, ThermalReserveCertification]] = {}
+    existing_thermals = set()
```
sylvlecl: This seems not used

antarest/study/dao/memory/in_memory_study_dao.py
```diff
@@ -1684,3 +1706,43 @@ def save_solar(self, series: AreaSeriesMapping) -> None:
...
+    @override
+    def get_all_thermal_reserve_symmetries(self) -> ThermalReserveSymmetriesMapping:
+        return self._thermal_reserve_symmetries
+
+    @override
+    def get_thermal_reserve_symmetries(self, area_id: AreaId) -> dict[ThermalId, ReserveSymmetries]:
+        return self._thermal_reserve_symmetries[area_id]
+
+    @override
+    def save_thermal_reserve_symmetries(self, data: ThermalReserveSymmetriesMapping) -> None:
```
sylvlecl, line 1745: the behaviour is not consistent with other DAOs which replace all thermal clusters data for each area

antarest/study/dao/memory/in_memory_study_dao.py
```diff
@@ -1684,3 +1706,43 @@ def save_solar(self, series: AreaSeriesMapping) -> None:
...
+        self, area_id: AreaId
+    ) -> dict[ReserveDefinitionId, dict[ThermalId, ThermalReserveCertification]]:
+        result: dict[ReserveDefinitionId, dict[ThermalId, ThermalReserveCertification]] = {}
+        for key, certification in self._thermal_reserve_certifications.items():
+            if key.area_id == area_id:
+                result.setdefault(ReserveDefinitionId(key.reserve_id), {})[key.thermal_id] = certification
+        return result
+
+    @override
+    def save_thermal_reserve_certifications(self, data: dict[AreaId, ThermalReserveCertificationMapping]) -> None:
```
sylvlecl, line 1729: Same here, it's not consistent with other implementations

antarest/study/dao/memory/in_memory_study_dao.py
```diff
@@ -1684,3 +1706,43 @@ def save_solar(self, series: AreaSeriesMapping) -> None:
...
+
+    @override
+    def get_thermal_reserve_symmetries(self, area_id: AreaId) -> dict[ThermalId, ReserveSymmetries]:
+        return self._thermal_reserve_symmetries[area_id]
+
+    @override
+    def save_thermal_reserve_symmetries(self, data: ThermalReserveSymmetriesMapping) -> None:
+        for area_id, thermal_dict in data.items():
+            for thermal_id, symmetries in thermal_dict.items():
+                self._thermal_reserve_symmetries[area_id][thermal_id] = symmetries
```
sylvlecl, line 1748: this will raise when there is not already some data for `area_id` (no dictionary yet)

sylvlecl (changes requested):
Minor stuff: we need to document the saving behaviour, and there are some issues in the in-mem implementation.

Since it's not used, we may decide to just not implement it for now, but we'd better not keep bugs in it

PR #3281, opened 2026-06-25
feat(ui-reserves): add read-only mode to GroupedDataTable and reserves view

hdinia:


webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/areas/$areaId/reserves/general.tsx
```diff
@@ -74,6 +76,13 @@ function ReservesGeneral() {
   const { studyId, areaId } = Route.useParams();
   const queryClient = useQueryClient();
 
+  // Reserves are editable only when enabled in the optimization configuration
+  // ("includeReserves"). Otherwise the table is shown in read-only mode.
+  const { data: showReserves } = usePromise(
```
skamril: ❌ Use `usePromiseWithSnackbarError` to display an error if the call fails.

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/areas/$areaId/reserves/general.tsx
```diff
@@ -144,6 +153,7 @@ function ReservesGeneral() {
       key={`${studyId}-${areaId}`}
       data={rows}
       columns={columns}
```
skamril: ❌ During loading, the view becomes uneditable with no feedback given to the user.
You can add the prop `isLoading`:

```ts
const { data: showReserves, isLoading } = usePromiseWithSnackbarError(...);

// ...

return (
  <GroupedDataTable 
    ...
    isLoading={isLoading}
   />
  );
```

webapp/src/components/GroupedDataTable/index.tsx
```diff
@@ -221,6 +223,10 @@ function GroupedDataTable<TGroups extends string[], TData extends RowData<TGroup
     muiTableBodyRowProps: ({ row }) => {
       const isPending = isPendingRow(row.original);
 
+      if (readOnly) {
+        return {};
+      }
+
       return {
         onClick: () => {
           if (isPending) {
```
skamril, line 232: 🧼 Can you move this condition with the `readOnly` condition above.

webapp/src/components/GroupedDataTable/index.tsx
```diff
@@ -263,7 +270,7 @@ function GroupedDataTable<TGroups extends string[], TData extends RowData<TGroup
             startIcon={<ContentCopyIcon />}
             variant="outlined"
             onClick={() => setOpenDialog("duplicate")}
-            disabled={table.getSelectedRowModel().rows.length !== 1}
+            disabled={readOnly || table.getSelectedRowModel().rows.length !== 1}
```
skamril, line 273: 💡 For me, in read-only mode, the button should not be displayed.

skamril (changes requested):

webapp/src/components/GroupedDataTable/index.tsx
hdinia: Indeed buttons are visible but disabled here on purpose : "these actions exist but aren't available"

skamril (approved):

PR #3282, opened 2026-06-25
feat(launcher): handle gracefully empty output

smailio:
ANT-4636


MartinBelthle (approved):
If you find a way to add a test that ensures that we don't add the trace in the log it would be good, but it is perhaps hard to do

PR #3283, opened 2026-06-26
feat(launcher): save launcher loads in db

olivierPigeon-RTE:


antarest/launcher/adapters/database_launcher_load_dao.py
```diff
@@ -0,0 +1,37 @@
...
+    from antarest.launcher.service import LauncherService
+
+
+class DataBaseLauncherLoadDao:
+    def update_launcher_load(self, launcher_load: LauncherLoad) -> None:
+        db.session.merge(launcher_load)
+        db.session.commit()
+
+    def get_launchers_loads(self) -> list[LauncherLoad]:
+        return db.session.query(LauncherLoad).all()
```
smailio: this is never used except in tests

antarest/launcher/model.py
```diff
@@ -357,6 +357,44 @@ class LauncherLoadDTO(AntaresBaseModel, extra="forbid", alias_generator=to_camel
     )
 
 
+class LauncherLoad(Base):
```
smailio, line 358: You need an alembic migration for this

antarest/launcher/adapters/database_launcher_load_dao.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def get_launcher_load(self, launcher_name: str) -> LauncherLoad | None:
+        return db.session.get(LauncherLoad, launcher_name)
+
+    def update_all_launcher_loads(self, service: "LauncherService") -> None:
+        for launcher_id, dto in service.get_all_loads().items():
+            self.update_launcher_load(LauncherLoad.from_dto(dto, launcher_id))
```
smailio: you do one commit per object here

Also I suggess that instead of doing eveything here you move the logic that get all loads in launcher_service and here you only take care of the update, thus this repository won't hold a reference to LauncherService.

antarest/launcher/adapters/database_launcher_load_dao.py
```diff
@@ -0,0 +1,37 @@
...
+from typing import TYPE_CHECKING
+
+from antarest.core.utils.fastapi_sqlalchemy import db
+from antarest.launcher.model import LauncherLoad
+
+if TYPE_CHECKING:
+    from antarest.launcher.service import LauncherService
+
+
+class DataBaseLauncherLoadDao:
```
smailio: This should be renamed to LauncherLoadRepository and moved to antarest/launcher/repository.py to match other classes that serve the same purpose

antarest/launcher/service.py
```diff
@@ -624,8 +627,17 @@ def get_load(self, launcher_id: str | None) -> LauncherLoadDTO:
         if launcher is None:
             raise InvalidConfigurationError(launcher_id)
 
+        load = self.launcher_load_dao.get_launcher_load(launcher_id)
+        if load is not None:
+            return load.to_dto()
```
smailio: Seems that a savec launcher load never goes stale. It isn't an issue if we're super confident that the celery task will be always running, I just want to make sure this scenario was considered. 

If I understood well, then the code bellow will never be called in practice, because as soon as we store a load entry in DB once it'll be served forever untill it's replaced by another db entry.

antarest/maintenance/tasks/launcher_load_task.py
```diff
@@ -0,0 +1,69 @@
...
+        return LauncherLoadTaskResult(
+            status=BackGroundTaskStatus.ERROR,
+            duration_seconds=0,
+            error="Launcher not found",
+        )
+    start_time = time.time()
+    try:
+        with db():
+            dao = DataBaseLauncherLoadDao()
+            dao.update_all_launcher_loads(launcher_service)
```
smailio: launcher_service already holds a reference to a DataBaseLauncherLoadDao why do you build one here ?

antarest/maintenance/app.py
```diff
@@ -53,6 +53,7 @@ class TaskName(StrEnum):
     TASKS_CLEANER = "tasks_cleaner"
     DISK_SPACE_ANALYZER = "disk_space_analyzer"
     DISK_USAGE = "disk_usage"
+    LAUNCHER_LOAD = "launcher_load"
```
smailio: This sounds like you wand to load a launcher. May I ssuggest something alon "LAUNCHER_CACHE" ?

antarest/launcher/service.py
```diff
@@ -624,8 +627,17 @@ def get_load(self, launcher_id: str | None) -> LauncherLoadDTO:
...
+
+        logger.info("No cached load for launcher '%s', querying live", launcher_id)
         return launcher.get_load()
 
+    def get_all_loads(self) -> dict[str, LauncherLoadDTO]:
+        """Get the load of all configured launchers."""
+        return {launcher_id: launcher.get_load() for launcher_id, launcher in self.launchers.items()}
```
smailio: get_load  can raise on connection fail. When one launcher is down, this whole function fail. 

In practice the get_load won't connect directly to launcher as I explained in the comment above so maybe this is not a big deal.

smailio (changes requested):

smailio (approved):

PR #3292, opened 2026-07-02
refactor(variant)!: remove legacy endpoints

MartinBelthle:
Removed endpoints:
- GET **"/studies/{uuid}/commands/_matrices"**
- GET **"/studies/{uuid}/commands/{cid}"**
- PUT **"/studies/{uuid}/commands/{cid}/move"**
- PUT **"/studies/{uuid}/commands/{cid}"**

Modified endpoints:
- POST **"/studies/{uuid}/commands"** does not return command ids when applied on variants anymore

### NOTE:
Discussed the breaking changes with Kamel and it should not impact the R scripts


webapp/src/services/api/variant.ts
```diff
@@ -72,11 +71,6 @@ export const getCommands = async (studyId: string): Promise<CommandDTO[]> => {
   return res.data;
 };
 
-export const exportCommandsMatrices = async (studyId: string): Promise<FileDownloadTask> => {
```
sylvlecl, line 75: We should also remove the other unused wrappers: getCommand, moveCommand, updateCommand

antarest/study/web/variant_blueprint.py
```diff
@@ -181,49 +159,7 @@ def append_command(study_service: StudyServiceDep, uuid: UuidStr, command: Comma
...
-        logger.info(f"Moving command {cid} to index {index} for variant study {uuid}")
-        variant_study_service = study_service.storage_service.variant_study_service
-        sanitized_cid = sanitize_string(cid)
-        variant_study_service.move_command(uuid, sanitized_cid, index)
-
-    @bp.put(
-        "/studies/{uuid}/commands/{cid}",
-        summary="Move a command to an other index",
-    )
-    def update_command(
```
sylvlecl, line 216: Unfortunately it seems used by the front end: we should remove it from there, then

antarest/study/web/variant_blueprint.py
```diff
@@ -181,49 +159,7 @@ def append_command(study_service: StudyServiceDep, uuid: UuidStr, command: Comma
         logger.info(f"Appending new command to variant study {uuid}")
         variant_study_service = study_service.storage_service.variant_study_service
         internal_command = variant_study_service.convert_commands(uuid, [command])[0]
-        return variant_study_service.append_command(uuid, internal_command)
-
-    @bp.get(
-        "/studies/{uuid}/commands/{cid}",
```
sylvlecl, line 187: Same here: there is a button which uses it in the front end, let's remove it too

sylvlecl (changes requested):
nice cleanup bu we need to remove a little more things on the frontend side for individual commands: namely the download and upload buttons.

Thinking that those endpoints actually can make sense in the future:
in other tools it's common to be able to edit one modification.
But we could re-introduce it in the future when actually needed, maybe that would be part of larger modifications.

PR #3293, opened 2026-07-02
feat(favorite_aggregate): added the aggregate method, checking for another tests to be done

mehdiwahada:
What does it do ? : It returns the 3 current type of favorites (studies, directories and external directories)

Why it has been done ? : It has been done in order to not be forced to call 3 GET methods to get every favorites

How do you test it ? : 
- Test 1 : you must create one favorite of each type (studies, directories and external directories), call the aggregate list method and verify that each of them are returned
  - After that, you must verify that any other user does not get the aggregate list, except if they add those same favorites

- Test 2 : you must call the aggregate list method and verify that nothing has been returned


antarest/favorite/model.py
```diff
@@ -118,3 +118,9 @@ class FavoriteExternalDirectory(Base):
 
     def to_dto(self) -> FavoriteExternalDirectoryDTO:
         return FavoriteExternalDirectoryDTO(path=PurePosixPath(self.path), workspace=self.workspace)
+
+
+class FavoriteAggregateDTO(AntaresBaseModel, extra="forbid"):
```
sylvlecl: We should use the camel case serialization, like for the inner DTOs

tests/integration/favorite_blueprint/test_favorite_aggregate.py
```diff
@@ -0,0 +1,94 @@
...
+    fav_study_response = admin_client.post(f"v1/favorites/studies/{study_response.json()}")
+    fav_directory_response = admin_client.post(f"v1/favorites/directories/{dir_id}")
+    fav_ext_directory_response = admin_client.post(
+        "v1/favorites/external-directories", params=expected_favorite_external_directory
+    )
+    aggregate_res = admin_client.get("/v1/favorites").json()
+
+    assert aggregate_res == {
+        "directories": [fav_directory_response.json()],
+        "external_directories": [fav_ext_directory_response.json()],
```
sylvlecl: see Here, we should have `externalDirectories`, not `external_directories`, for consistency with inner objects (and the API in general)

tests/integration/favorite_blueprint/test_favorite_aggregate.py
```diff
@@ -0,0 +1,94 @@
...
+    # Create directory structure 'foo/bar' before moving the study
+    res = admin_client.post(
+        "/v1/directories",
+        json={"name": "foo"},
+    )
+    assert res.status_code == 201, res.json()
+
+    directory_id = res.json()["id"]
+    directory_res = admin_client.post(f"/v1/favorites/directories/{directory_id}")
+    assert directory_res.status_code in [200, 201, 204]
```
sylvlecl: We should assert an exact status code here ( 201 I guess ?)

tests/integration/favorite_blueprint/test_favorite_aggregate.py
```diff
@@ -0,0 +1,94 @@
...
+    study_response = admin_client.post(
+        "/v1/studies",
+        params={"name": "Test Study for Favorites"},
+    )
+    assert study_response.status_code == 201
+    study_id = study_response.json()
+
+    # Add the study to favorites
+    study_res = admin_client.post(f"/v1/favorites/studies/{study_id}")
+    assert study_res.status_code in [200, 201, 204]
```
sylvlecl: Wed'better assert an exact status code.
The reason: we don't want that a change in our API (status code changing from 201 to 204 for example), silently passes our tests

sylvlecl (changes requested):
Thanks for the PR @mehdiwahada !
Very minor change request about naming convention and details in tests, otherwise it looks good :)

sylvlecl (approved):
Looks good thanks!
I just changed the status code of deletes from 202 to 204 which is more standard

PR #3295, opened 2026-07-07
refactor(ui-output): update the view

skamril:
- enhance 'Variable per variable' mode
- More concise filters

ANT-4287 
ANT-4412


webapp/src/components/Matrix/components/MatrixFilter/index.tsx
```diff
@@ -203,4 +196,4 @@ function MatrixFilter(
   );
 }
 
-export default forwardRef(MatrixFilter);
+export default MatrixFilter;
```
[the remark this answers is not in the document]
skamril: deprecated

webapp/src/components/Matrix/components/MatrixFilter/types.ts
```diff
@@ -48,11 +48,16 @@ export interface FilterCriteria {
...
+export interface MatrixFilterHandle {
+  toggleFilter: () => void;
+}
+
 export interface MatrixFilterProps {
   dateTime?: DateTimes;
   isTimeSeries: boolean;
   timeFrequency?: TimeFrequencyType;
   readOnly?: boolean;
+  ref?: React.Ref<MatrixFilterHandle>;
```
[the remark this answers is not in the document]
skamril: deprecated

webapp/src/components/Matrix/components/FilterableMatrixGrid.tsx
```diff
@@ -13,51 +13,34 @@
...
-import MatrixFilter, { type MatrixFilterHandle } from "./MatrixFilter";
+import MatrixFilter from "./MatrixFilter";
 
+import type { MatrixFilterProps } from "./MatrixFilter/types";
 import MatrixGrid, { type MatrixGridProps } from "./MatrixGrid";
 
 export interface FilterableMatrixGridProps extends MatrixGridProps {
   timeFrequency?: TimeFrequencyType;
+  ref?: MatrixFilterProps["ref"];
 }
```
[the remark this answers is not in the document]
skamril: deprecated

webapp/src/components/Matrix/components/FilterableMatrixGrid.tsx
```diff
@@ -131,4 +114,4 @@ function FilterableMatrixGrid(
   );
 }
 
-export default forwardRef(FilterableMatrixGrid);
+export default FilterableMatrixGrid;
```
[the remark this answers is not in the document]
skamril: deprecated

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-hooks/useOutputFilters.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+import { useContext } from "react";
+import OutputContext from "../-components/OutputMatrixViewer/OutputContext";
+
+function useOutputFilters() {
```
skamril: Rename useOutputContext

hdinia (changes requested):
🔴 Blocking

1. Columns filter shows wrong data under the filtered headers
📍 webapp/src/.../OutputMatrixViewer/OutputMatrix.tsx:111 (also lines 162–164)

filteredColumns filters the header list, but the full matrixResult.data is passed to the grid. useColumnMapping.gridToData maps the Nth Number column to data column N sequentially, so after filtering, the kept headers display values from the first N original data columns. 

Example: a matrix with stats `exp|min|max`, filter stats to `max` → columns labeled "max" render the `exp` values. The deleted code in the old OutputMatrixViewer/index.tsx handled this explicitly by remapping rows through headerIndices (filteredData = data.map(row => headerIndices.map(i => row[i]))). That remapping must be restored, e.g.:

```ts
const keptIndices = useMemo(() => /* indices of filteredColumns in data.columns */, [...]);
const filteredData = useMemo(() => data.map(row => keptIndices.map(i => row[i])), [...]);
```

2. "Process" (materialization) feature deleted (button left dead)

You can re-wire the materialization flow 

---
🟠 Issue

3. dataType is reset to "General values" whenever the options array identity changes
📍 webapp/src/.../OutputMatrixViewer/Header/ResultFilters/index.tsx:64

setDataType(dataTypeOptions[0].value) runs unconditionally when dataTypeOptions changes. In variable-per-variable mode getDataTypeOptions returns a new filtered array on every recompute, so selecting a different area resets the user's Display choice (e.g. "Thermal plants" → "General values") even though it's still valid. Guard it: only reset when the current dataType is not present in the new options.

4. useOutput — dead check placed after use, and error-path behavior change vs deleted useStudyOutput
📍 webapp/src/.../outputs/$outputId/-hooks/useOutput.ts:41

- The if (!outputId) throw sits after useSuspenseQuery already used outputId; it's also dead code since the route param is typed and guaranteed. Remove it (or move above the query if you keep it).
- Behavior change: the deleted useStudyOutput returned undefined gracefully (with a documented workaround for the variant-generation websocket bug). the new hook throws during render when the output isn't in the list. The always-fetch-via-react-query approach preserves the workaround's spirit, but please confirm the route error boundary renders something recoverable for a just-generated variant output (race between task completion and list refresh) rather than crashing the whole page.

---
🟡 Suggestion

5. Filename typo
📍 webapp/src/.../OutputMatrixViewer/Header/ColumnsFIlter.tsx — rename ColumnsFIlter.tsx → ColumnsFilter.tsx (capital "I")


6. Naming mismatches around the new context
📍 webapp/src/.../outputs/$outputId/-hooks/useOutputFilters.ts:19 
the error message says "within a <FiltersContext>" but the provider is `OutputContext`, and the hook exposes far more than filters (grid ref, loaded state)  `useOutputContext` would be more accurate.

7.📍 webapp/src/.../OutputMatrixViewer/Header/index.tsx:58 — `handleToggleColumnsFilter` actually toggles the matrix data filter drawer (`matrixGridRef.current?.toggleFilter()`), while the columns drawer is opened by `toggleColumnsFilter`. consider a rename e.g. `handleToggleDataFilter`.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-hooks/useStudyOutput.ts
```diff
@@ -1,100 +0,0 @@
...
-      }
-
-      // BUG: For variants, after a generation job is initiated, we must await the job
-      // completion to display results. However, the WebSocket listener that should notify
-      // us of job completion never gets triggered for this particular case. This causes
-      // variant outputs to remain outdated/unavailable until the app is manually refreshed.
-      // As a workaround, we attempt to fetch the output directly.
-      //
-      // TODO: The API should ensure that the WebSocket listener is properly set up to notify
-      // us of job completion.
```
hdinia, line 75: 🟡 Suggestion
Have you tested this case is still properly handled after the refactor ?

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-components/OutputMatrixViewer/OutputVariableMatrix.tsx
```diff
@@ -0,0 +1,170 @@
...
+      }}
+      ifRejected={(err) => {
+        const error = isAxiosError(err) ? err.response?.data : undefined;
+        const status = error?.status;
+        const taskId = error?.task_id;
+
+        // NOT_FOUND status with no task ID means data not materialized yet
+        if (status === "NOT_FOUND" && taskId === null) {
+          // TODO: update the status to "NOT_MATERIALIZED" + handle the ongoing materialization
+          // state using the taskId and the "IN_PROGRESS" status
```
hdinia: 🟡 Suggestion
maybe we can drop this TODO, it requires backend changes for a minor rename

webapp/src/utils/apiUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+export function sanitizeJsonResponse<T = unknown>(response: any): T {
+  if (typeof response !== "string") {
+    return response;
+  }
+
+  // Replace invalid JSON literals with valid JSON strings
+  const sanitized = response.replace(/NaN/g, '"NaN"').replace(/Infinity/g, '"Infinity"');
+
+  return JSON.parse(sanitized);
+}
```
hdinia: 🟡 Suggestion
`replace(/NaN/g, '"NaN"')` also rewrites occurrences inside string values (a variable/area name containing "NaN" or "Infinity" would corrupt the JSON), and `-Infinity` becomes `-"Infinity"`, which is invalid JSON and makes JSON.parse throw. Pydantic's `ser_json_inf_nan="constants"` does emit `-Infinity`. 

(You can also fix the doc comment (backend emits {"value": NaN}, not {"value": "NaN"}). )

It's pre-existing logic, but now is the moment to fix it since it's being centralized.

hdinia: (It's an edge case)

skamril: https://github.com/AntaresSimulatorTeam/AntaREST/pull/3295#discussion_r3534569524

hdinia (changes requested):
- 1 🔴 Feedback: "Process" (materialization) button dead

Status : still not fixed

- 2 🟠 Feedback: `useOutput.ts`: dead check after use + error-path behavior change vs old useStudyOutput

Status: Not fixed. `useOutput.ts` is identical to before the review, the if (!outputId) throw is still after `useSuspenseQuery` uses it, and it still throws synchronously if output isn't in the list yet, with no confirmation/handling of the variant-generation race the old hook papered over (see comment https://github.com/AntaresSimulatorTeam/AntaREST/pull/3295#discussion_r3587579718).

- 3 🟡 Feedback: Rename ColumnsFIlter.tsx → ColumnsFilter.tsx (typo)

Status: Not fixed. Renamed to ColumnsFIlters.tsx but the capital-I typo is still there

 - 4 🟡 Feedback: useOutputFilters/<FiltersContext> naming mismatch
 
Status: Fixed. Hook renamed to useOutputContext, error message now says <OutputContext>. (Minor: the file itself is still named useOutputFilters.ts.)

webapp/src/components/Matrix/shared/types.ts
```diff
@@ -117,10 +109,10 @@ export interface MatrixDataDTO {
   index: number[];
 }
 
-export interface ResultMatrixDTO {
+export interface MatrixResultDTO {
```
hdinia: `MatrixResultDTO` = "the result of a matrix (operation)" -> reads as: a DTO representing a result

`ResultMatrixDTO` = "a matrix of results" -> reads as: a DTO representing a matrix, whose contents are results. This implies the DTO is structured as a matrix (2D array/grid) containing result values.

keep `ResultMatrixDTO`

hdinia (changes requested):

webapp/src/components/Matrix/shared/types.ts
skamril: So I rename MatrixDataDTO -> DataMatrixDTO and MatrixAggregates -> AggregatesMatrix? 
To remain consistent.

webapp/src/routes/_authenticated/studies/$studyId/explore/outputs/$outputId/-hooks/useStudyOutput.ts
skamril: If there is an issue, shouldn't the fix be made in the outputs list view instead of here?

webapp/src/components/Matrix/shared/types.ts
hdinia: No those are correct, just revert this one

skamril: Following the logic in your first comment, MatrixDataDTO would technically be inconsistent too, since it has the same structure as ResultMatrixDTO. So whichever convention we pick should logically apply to both.
That said, since renaming ResultMatrixDTO is out of scope for this PR, I'll leave it as is for now.

hdinia (approved):

PR #3296, opened 2026-07-07
refactor(archive)!: drop `py7zr` library

MartinBelthle:
### Rationale

`py7zr` is slower and less stable when compared to `7z` CLI.
In particular, in versions 0.x it did not handle well the absence of root directory.
It had us introduce a fake `.` folder, which raises errors when upgrading to `py7zr` 1.x.
Switching to `7z` CLI only is more reliable while allowing to keep the feature in most 
environments.

### Breaking change

- We no longer support importing matrix datasets archived in 7zip format
- For the Desktop version, you'll need `7z` installed on your computer to import studies in 7z format or to unarchive old studies


antarest/core/utils/archives.py
```diff
@@ -222,10 +211,19 @@ def read_original_file_in_archive(archive_path: Path, posix_path: str) -> bytes:
...
+
     elif archive_path.suffix == ArchiveFormat.SEVEN_ZIP:
-        with py7zr.SevenZipFile(archive_path, mode="r") as szf:
-            output: bytes = szf.read([posix_path])[posix_path].read()
-            return output
+        if _has_7z():
+            logger.info("Using 7z CLI to read data inside archive %s", archive_path)
+            with tempfile.TemporaryDirectory() as target_dir:
+                try:
+                    run(["7z", "e", str(archive_path), f"-o{target_dir}", "-y", posix_path], check=True)
```
sylvlecl: You need to use "x", not "e". "e" extract files in a flat way (no subdirectories).

Possibly, we could have a small utility function since we use that command line in 2 places

sylvlecl (changes requested):
Nice cleanup, seems there is an issue with using "7z e" instead of "7z x" though

PR #3300, opened 2026-07-08
feat(launcher): add launch scheduling

smailio:


antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -501,8 +511,10 @@ def _run_study(
                         )
                     _override_solver_version(study_path, version)
 
+                    if run_at is not None:
+                        append_log(launch_uuid, f"Study scheduled to start at {run_at} UTC")
```
smailio: Should I add Paris time between parenthesis ? As UTC time can confuse user.

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -663,6 +687,16 @@ def _remove_study_from_workspace_db(self, study_name: str) -> None:
...
 
 
+def _format_slurm_begin(run_at: datetime | None) -> str | None:
+    """
+    Format a scheduled start time for the SLURM `sbatch --begin` option.
+    """
+    if run_at is None:
+        return None
+    minutes = round((run_at - current_time()).total_seconds() / 60)
+    return f"now+{max(0, minutes)}minutes"
```
smailio: This is my solution to sumbit the job without caring about the time zone of the slurm server. 

If I don't do this we'd need either a new config attribute or we need to always make sure that clusters are configured with same time zone than our backend.

antarest/launcher/service.py
```diff
@@ -266,17 +298,20 @@ def run_study(
...
             job_status=JobStatus.PENDING,
             launcher=launcher,
             launcher_params=launcher_parameters.model_dump_json() if launcher_parameters else None,
             owner_id=(owner_id or None),
+            scheduled_at=scheduled_at,
         )
         self.job_result_repository.save(job_status)
 
-        self.launchers[launcher].run_study(study_uuid, job_uuid, solver_version, launcher_parameters)
+        self.launchers[launcher].run_study(study_uuid, job_uuid, solver_version, launcher_parameters, scheduled_at)
```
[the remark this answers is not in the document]
smailio: meh then we'll have to decide how we identify the launcher, I can't think of a clean way to do it or add a new attribute which is too much of a change for almost no benefit.

antarest/launcher/service.py
```diff
@@ -97,6 +98,13 @@ def __init__(self, engine: str):
 LAUNCHER_PARAM_NAME_SUFFIX = "output_suffix"
 EXECUTION_INFO_FILE = "execution_info.ini"
 
+MAX_SCHEDULE_HORIZON = timedelta(days=15)
+
+
+class InvalidScheduleTime(HTTPException):
```
sylvlecl: Nice to have thought of that limit !
The value seems relevant to me too, we may want to schedule simulations for the end of the week, but not too much later

antarest/launcher/service.py
```diff
@@ -237,19 +245,43 @@ def _assert_launcher_is_initialized(self, launcher: str) -> None:
...
+    def _normalize_scheduled_at(run_at: datetime) -> datetime:
+        """
+        Validate a requested scheduled start time and convert it to a naive UTC datetime.
+
+        A naive `run_at` is assumed to already be UTC; an aware one is converted to UTC.
+
+        Raises:
+            InvalidScheduleTime: if the time is not in the future or beyond the allowed horizon.
+        """
+        if run_at.tzinfo is not None:
```
sylvlecl, line 232: We may have a utility function in core for that, like "to_utc_naive" or something like that

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
sylvlecl: We cannot hard code it to "Paris", for sure.
We could use the local timezone of the server. maybe that would be consistent with other places, although it's not great. Maybe you could try to find in the codebase if we already log some datetimes and how

antarest/launcher/service.py
```diff
@@ -95,6 +96,13 @@ def __init__(self, engine: str):
 LAUNCHER_PARAM_NAME_SUFFIX = "output_suffix"
 EXECUTION_INFO_FILE = "execution_info.ini"
 
+MAX_SCHEDULE_HORIZON = timedelta(days=15)
```
smailio, line 101: Nobody asked for this but seems reasonable

antarest/launcher/service.py
```diff
@@ -95,6 +96,13 @@ def __init__(self, engine: str):
...
+
+
+class InvalidScheduleTime(HTTPException):
+    def __init__(self, message: str) -> None:
+        super().__init__(HTTPStatus.BAD_REQUEST, message)
+
```
smailio: I was hesitant to create an exceptions.py file

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -651,6 +676,17 @@ def _remove_study_from_workspace_db(self, study_name: str) -> None:
         self.data_repo_tinydb.db.remove(tinydb.where(pk_name) == study_name)
 
 
+def _format_slurm_begin(run_at: datetime | None) -> str | None:
```
sylvlecl: It's smart to do that.
Can this cause issues ? Maybe, because the actual submission to slurm may be delayed to a later "now", because of the time required to zip and send the studies, before actually submitting them.
Therefore I wonder if we'd not better just assume a consistent timezone between the server and the slurm cluster (it's possible that we already assume that, by the way, if we read some time information from the slurm commands outputs).

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -371,6 +375,11 @@ def _check_studies_state(self) -> None:
                 else:
                     # study.started => still running
                     # study.finished => waiting for ZIP + logs retrieval (or failure)
+                    if study.started and study.name not in self._running_notified:
+                        # The job left the SLURM queue (it was PENDING, possibly held by `--begin`) and
+                        # actually started computing, so flip it to RUNNING (only once).
```
sylvlecl, line 380: On restart of the server, this will be called again for all running computations: is that an issue ?
The update to RUNNING state seems idempotent, so it looks OK, hope it will not be too fragile in the long run.

sylvlecl (commented):
Looks good to me overall:
I wonder about the "now+..." implementation detail, I worry that it may schedule the computation for quite a long time after the actual requested time because of pre-processing time (zipping and sending to cluster)

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
sylvlecl: As you just told me:
a good solution would be to implement that logic inside the launcher lib, so that it's done at the actual submit time.

This will solve that delay issue.

--> we'll need a new release of the lib

antarest/launcher/adapters/slurm_launcher/slurm_launcher.py
```diff
@@ -561,18 +578,26 @@ def _apply_params(self, launcher_params: LauncherParametersDTO) -> argparse.Name
                 # could even lead to security breaches
                 raise ValueError("Other options cannot contain a single quote, you should use double quotes instead")
 
+            launcher_args.begin = _format_slurm_begin(run_at)
             return launcher_args
 
-        return self.launcher_args
+        arguments = self.launcher_args
+        arguments.begin = _format_slurm_begin(run_at)
```
sylvlecl: This mutates the shared default value --> we need to copy it before mutating it

sylvlecl (changes requested):
One more comment: there is a potential bug on mutating the default value

uv.lock
```diff
@@ -999,7 +999,6 @@ wheels = [
...
     { url = "https://files.pythonhosted.org/packages/09/16/2c3792cba130000bf2a31c5272999113f4764fd9d874fb257ff588ac779a/greenlet-3.2.4-cp311-cp311-manylinux2014_aarch64.manylinux_2_17_aarch64.whl", hash = "sha256:1ee8fae0519a337f2329cb78bd7a8e128ec0f881073d43f023c7b8d4831d5246", size = 632472, upload-time = "2025-08-07T13:42:55.044Z" },
     { url = "https://files.pythonhosted.org/packages/ae/8f/95d48d7e3d433e6dae5b1682e4292242a53f22df82e6d3dda81b1701a960/greenlet-3.2.4-cp311-cp311-manylinux2014_ppc64le.manylinux_2_17_ppc64le.whl", hash = "sha256:94abf90142c2a18151632371140b3dba4dee031633fe614cb592dbb6c9e17bc3", size = 644646, upload-time = "2025-08-07T13:45:26.523Z" },
-    { url = "https://files.pythonhosted.org/packages/d5/5e/405965351aef8c76b8ef7ad370e5da58d57ef6068df197548b015464001a/greenlet-3.2.4-cp311-cp311-manylinux2014_s390x.manylinux_2_17_s390x.whl", hash = "sha256:4d1378601b85e2e5171b99be8d2dc85f594c79967599328f95c1dc1a40f1c633", size = 640519, upload-time = "2025-08-07T13:53:13.928Z" },
```
smailio, line 1002: I use uv 0.11.33 ir removed some wheels after I added antarest-launcher==1.16.0 

I don't think it's an issue, let me know.

sylvlecl (approved):
Looks good thanks :)

PR #3305, opened 2026-07-15
fix(variant): solve snapshot concurrency issues

MartinBelthle:


alembic/versions/e56b1130bc1f_create_commands_list_version_table.py
```diff
@@ -0,0 +1,60 @@
...
+revision = 'e56b1130bc1f'
+down_revision = '665f7b1d7575'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # First, perform the operation with the least amount of risk
+    with op.batch_alter_table('variant_study_snapshot', schema=None) as batch_op:
+        batch_op.add_column(sa.Column("version", sa.Integer(), nullable=False, server_default="0"))
```
sylvlecl: We don't need a server default, the applciation code is responsible for the inital value

antarest/study/dao/database/models/area.py
```diff
@@ -19,9 +19,9 @@
 
 from sqlalchemy import Boolean, Column, Float, ForeignKey, ForeignKeyConstraint, Integer, String, Table
 
+from antarest.core.utils.sql_utils import enum_col
```
sylvlecl, line 22: Good thing to have moved that module :)

tests/core/utils/test_sql_utils.py
```diff
@@ -16,7 +16,7 @@
 from sqlalchemy.exc import CompileError
 from sqlalchemy.orm import Session, sessionmaker
 
-from antarest.study.dao.database.sql_utils import upsert_multiple, upsert_one
+from antarest.core.utils.sql_utils import upsert_multiple, upsert_one
```
sylvlecl, line 19: we'd better move that test file to keep consistent with the production code layout

antarest/study/service.py
```diff
@@ -533,10 +533,11 @@ def add_commands(self, commands: Sequence[ICommand], listener: ICommandListener
     @override
     def update_study_metadata(self, metadata: StudyMetadataUpdate) -> None:
         """
-        We update the last modification date in DB.
+        We increment the `commands_list_version` table.
```
sylvlecl: Note that the increment will not be needed for DB studies:
the only reason we need regeneration for file studies is that the metadata is duplicated in the files, but it's not the case in DB.
We can make the comment more detailed here so that the reader understands why we are doing this, which can look kindf of weird otherwise

antarest/dbmodel.py
```diff
@@ -9,10 +9,17 @@
...
 
 from antarest.core.persistence import Base as PersistenceBase
 from antarest.core.tasks.model import TaskJob  # noqa: F401
 from antarest.launcher.model import JobResult  # noqa: F401
 from antarest.login.model import Identity  # noqa: F401
 
 Base = PersistenceBase
+
+
+def get_row_representation_as_dict(row: Row[Any]) -> dict[str, Any]:
```
sylvlecl, line 24: Loooks like we can achieve the same result using the following sqlalchemy function:
```
for row in result.mappings():
    row["id"]
```

antarest/study/storage/variantstudy/repository.py
```diff
@@ -132,25 +139,111 @@ def get_all_command_blocks(self) -> list[CommandBlock]:
...
+    def initialize_commands_list_version_table(self, variant_id: str) -> None:
+        self._save_commands_list_version(variant_id, 0)
+
+    def _save_commands_list_version(self, variant_id: str, version: int) -> None:
+        session = self.session
+        upsert_one(session, COMMANDS_LIST_VERSION_TABLE, {"variant_id": variant_id, "version": version})
+        session.commit()
+
+    def increment_commands_list_version(self, variant_id: str) -> None:
+        current_version = self.get_commands_list_version(variant_id)
```
sylvlecl: THere is room for concurrent updates of the version here:
the version must be protected with a "for update", otherwise it can be read from another thread and updated with the same value in that other thread

antarest/study/storage/variantstudy/repository.py
```diff
@@ -132,25 +139,111 @@ def get_all_command_blocks(self) -> list[CommandBlock]:
...
+        cmd_blocks = []
+        version_found = False
+        for row in rows:
+            row_as_dict = get_row_representation_as_dict(row)
+            cmd_blocks.append(row_as_dict["CommandBlock"])
+            if row_as_dict["variant_id"] == last_child:
+                version_found = True
+                version = row_as_dict["version"]
+
+        if not version_found:
```
sylvlecl: We should be able to handle that case with an outer join in the initial query.
This will avoid an additional query, and more importantly, this will correctly lock the version, unlike the second query here

antarest/study/storage/variantstudy/model/dbmodel.py
```diff
@@ -166,3 +181,17 @@ class VariantStudy(Study):
     @override
     def __str__(self) -> str:
         return super().__str__() + f", snapshot={self.snapshot}"
+
+
+COMMANDS_LIST_VERSION_TABLE = Table(
```
sylvlecl: note that I think that we could have implemented it also using the ORM approach, by adding the version as a relation inside the VariantStudy class.
Could make the query code simpler, because then we will not mix together ORM and "core" APIs.

Ok this way, though

antarest/study/storage/variantstudy/snapshot/file_snapshot_manager.py
```diff
@@ -29,14 +29,6 @@ class FileSnapshotManager(ISnapshotManager):
...
 
-    @override
-    def is_snapshot_up_to_date(self, study: VariantStudy) -> bool:
-        return self.has_snapshot(study) and (study.snapshot.created_at >= study.updated_at)
-
-    @override
-    def has_snapshot(self, study: VariantStudy) -> bool:
-        return (study.snapshot is not None) and (get_snapshot_dir(study) / "study.antares").is_file()
-
     @override
```
sylvlecl, line 32: Nice, indeed it does not make sense anymore to handle those here, we have the same versioning system whatever the underlying storage

antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
```diff
@@ -188,86 +185,59 @@ def search_ref_study(
...
+                    if command_block.id == last_executed_cmd_id:
+                        last_exec_index = command_block.index
+                        return RefStudySearchResult(
+                            ref_study=current_variant,
+                            cmd_blocks=cmd_blocks_w_version.commands[last_exec_index + 1 :],
+                            force_regenerate=False,
+                            version=cmd_blocks_w_version.version,
+                        )
+
+        # Final case: The variant has no snapshot, or its `last_executed_command` does not exist anymore.
```
sylvlecl, line 249: Note that the case "we have a snapshot AND the last command does not exist anymore" must not happen:
when we remove commands (or other change which is not juste a simple append), the snapshot is invalidated.

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -257,26 +267,38 @@ def replace_commands(self, study_id: str, commands: list[CommandDTO]) -> str:
     def _modify_commands(self, study: VariantStudy, commands: list[CommandDTO], replace_commands: bool) -> list[str]:
         command_objs = self._check_commands_validity(study.id, commands)
         validated_commands = transform_command_to_dto(command_objs, commands)
+
+        current_commands = self._get_study_commands(study.id)
```
sylvlecl: Maybe we should make more explicit that this locks the commands, otherwise future devs can misuse that method. It will also make the intent more explicit

alembic/versions/e56b1130bc1f_create_commands_list_version_table.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Gathers all variant study ids
+    result = session.query(table('variantstudy', column('id'))).all()
+    if not result:
+        # Means there are no variant studies in the database
+        return
+    data_to_insert = [{"variant_id": variant_id, "version": 0} for variant_id, in result]
+
+    # Insert values to fill the newly created table
+    sql = sa.text("INSERT INTO commands_list_version (variant_id, version) VALUES (:variant_id, :version)")
```
sylvlecl: We can do a single query to do everything on the database side:
```sql

INSERT INTO commands_list_version (variant_id, version)
SELECT id, 0
FROM variant_study;
```

antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
```diff
@@ -188,86 +185,59 @@ def search_ref_study(
...
+        # We search for a variant with an up-to-date snapshot to use it as a reference study.
+        # If no such variant is found, we use the root study as a reference study.
+
+        variant_ids = [v.id for v in descendants]
+        ref_study = root_study
+
+        # Iterate in reverse order to find the most recent variants first.
+        for k, variant in enumerate(descendants[-2::-1]):
+            # todo: this performs N+1 queries, we should probably introduce a new method to avoid this.
```
sylvlecl: Having everything joined-loaded in the single query `get_study_tree` would solve this: we could get a consistent image of all parents, versions and commands in one query.
(Probably easier to implement with the ORM approach)

sylvlecl (changes requested):
Looks pretty good!
2 places where a lock on the version needs to be added though.

Also, not a must-do, but maybe for that feature the ORM approach would have made the code simpler, and in particular getting the whole "tree" of variants (with version and commands) at once

antarest/dbmodel.py
MartinBelthle: Not really sure where I should put the code you propose :/

antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
MartinBelthle: The snapshot is not invalidated anymore when removing a command, I changed this. Only the children snapshots are invalidated. Now I just change the command ids of all commands that were after the one removed. It will be more performant than the old code.

See `remove_command` method inside `variant_study_service.py`

sylvlecl: Ah! ok sounds good

alembic/versions/e56b1130bc1f_create_commands_list_version_table.py
MartinBelthle: Not sure to understand. The table already contains data and I typed the column as not-nullable so I need a value to fill it. Perhaps this comment was meant for the new table introduce where I can indeed remove the `server_default` parameter

sylvlecl: Ah indeed. We could remove the server default in a second operation, to be consistent with the decision of having no server defaults (the app is responsible to write some data). Clearly not very important though

MartinBelthle: Okay I see. I'll let this as is, as we have the same issue for existing alembic migrations (storage_mode, study_version in commands ...) so we'd better handle all of them at once in a next alembic migration if we want

antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
```diff
@@ -46,10 +45,41 @@ class RefStudySearchResult(NamedTuple):
     """
 
     ref_study: Study
-    cmd_blocks: Sequence[CommandBlock]
+    cmd_blocks: list[CommandBlock]
+    version: int
     force_regenerate: bool = False
 
 
+def _get_sorted_command_blocks(variants: Sequence[VariantStudy]) -> list[CommandBlock]:
```
sylvlecl, line 53: the commands are already sorted by index in the variant study class, I think we can safely remove that

antarest/study/storage/variantstudy/repository.py
```diff
@@ -132,25 +139,102 @@ def get_all_command_blocks(self) -> list[CommandBlock]:
...
+    def get_commands_list_version(self, variant_id: str) -> int:
+        """
+        This method should only be used to increment the commands' list version later.
+        That is why it locks the `CommandsListVersion` table using a `with_for_update` clause.
+        """
+        stmt = select(CommandsListVersion.version).where(CommandsListVersion.variant_id == variant_id).with_for_update()
+        version: int = self.session.execute(stmt).scalar_one()
+        return version
+
+    def save_commands_list_version(self, variant_id: str, commands: CommandBlocksWithVersion) -> None:
```
sylvlecl: would probably make the code even simpler to always use the variant study to get and set version + commands. It may have a minore drawback which is that we may get the variant study data while we don't really need them, bu not a big issue

antarest/study/storage/variantstudy/repository.py
```diff
@@ -132,25 +139,102 @@ def get_all_command_blocks(self) -> list[CommandBlock]:
...
+    def get_command_blocks_with_associated_version(self, variant_id: str) -> CommandBlocksWithVersion:
+        """
+        This method performs a single JOIN query to ensure that the 2 separated infos (version and cmd blocks) are synchronized.
+        """
+        query = (
+            select(CommandsListVersion, CommandBlock)
+            .outerjoin(CommandBlock, CommandBlock.study_id == CommandsListVersion.variant_id)
+            .where(CommandsListVersion.variant_id == variant_id)
+            .order_by(CommandBlock.index)
+            .with_for_update()
```
sylvlecl: AI tells me that postgresql does not accept "for update" on such a join.

Indeed anyway, we should only lock the version row.

Therefore we should restrict the for update to the version table, or perform 2 queries, if really needed.

PS:
indeed I get "FOR UPDATE cannot be applied to the nullable side of an outer join".
Not sure which one is the nullable side ... in any case we only need to lock the version side

antarest/study/storage/variantstudy/variant_study_service.py
```diff
@@ -516,6 +559,7 @@ def create_variant_study(self, uuid: str, name: str) -> VariantStudy:
             storage_mode=study.storage_mode,
         )
         self.repository.save(variant_study)
+        self.repository.initialize_commands_list_version_table(new_id)
```
sylvlecl: we'd better put the version in the study definition above, this way it will be written in the same commit

sylvlecl (changes requested):
One blocking point:
one of the "for update" does not work on postgresql

Not a must do: we could probably remove more repository code by just reading and writing the commands and the version through the variant study

alembic/versions/e56b1130bc1f_create_commands_list_version_table.py
```diff
@@ -0,0 +1,51 @@
...
+# revision identifiers, used by Alembic.
+revision = 'e56b1130bc1f'
+down_revision = '665f7b1d7575'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # First, perform the operation with the least amount of risk
+    with op.batch_alter_table('variant_study_snapshot', schema=None) as batch_op:
```
sylvlecl: Actually, we cannot presume that the existing snapshot is up to date, we'd better just invalidate all of them

sylvlecl (approved):
Good thank you !
Just fixed the postgresql-related issue + migration

PR #3314, opened 2026-07-24
fix(variant): snapshot save full lineage information

sylvlecl:
The goal is to solve race conditions when modifying parents of a study while a snapshot 
is being generated.

 - save the full "lineage" information together with the snapshot, instead
   of only the commands version
 - also fixes the behaviour in case of crash during snapshot generation:
   we must not consider the modified snapshot data as consistent with the snapshot metadata,
   therefore we now invalidate it at generation start
 - finally, if the source snapshot has been changed during the generation, we restart it
   (optimistic concurrency handling: in a large majority of cases, source will not have changed)

Removes permission checks from snapshot generation, it does not belong here and adds
burden to the study tree query with 2 additional joins

### DIFF
--- a/antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
+++ b/antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
@@ -65,6 +65,10 @@ def _get_aggregated_command_blocks(variants: Sequence[VariantStudy]) -> list[Com
     return [cmd for variant in variants for cmd in variant.commands]
 
 
+def _copy_command_blocks(command_blocks: Sequence[CommandBlock]) -> list[CommandBlock]:
+    return [CommandBlock(**command_block.to_dict()) for command_block in command_blocks]
+
+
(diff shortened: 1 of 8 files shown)


antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
```diff
@@ -112,10 +159,26 @@ def generate_snapshot(
...
-        ref_study = search_result.ref_study
-        cmd_blocks = search_result.cmd_blocks
+        variant_study = descendants[-1]
+        previous_last_executed_command = (
+            variant_study.snapshot.last_executed_command if variant_study.snapshot is not None else None
+        )
+        cmd_blocks = _copy_command_blocks(search_result.cmd_blocks)
+        ref_study_id = search_result.ref_study.id
+
+        # Persist this before modifying snapshot data. Commands are copied
```
sylvlecl, line 163: weird that we need to reload those again ?
we may get different ones from above, in addition to issuing more queries

sylvlecl:
Closing, implemented in separate PRs

PR #3319, opened 2026-07-27
perf(variant): get variant lineage in 1 query

sylvlecl:
The retrieval of ancestors was performed in 2 steps without any reason
(first get ancestors IDs, then get the data).

Also removes the permission check during generation:
permissions should be handled upstream, not at generation time. It does not make sense to verify
all parents every time we make a change to the study.


antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
```diff
@@ -50,7 +48,7 @@ class RefStudySearchResult(NamedTuple):
     force_regenerate: bool = False
 
 
-def _get_sorted_command_blocks(variants: Sequence[VariantStudy]) -> list[CommandBlock]:
+def _aggregate_command_blocks(variants: Sequence[VariantStudy]) -> list[CommandBlock]:
```
sylvlecl, line 51: Just simplified a bit the iteration logic here (I think ?).

antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
```diff
@@ -107,7 +105,6 @@ def generate_snapshot(
         logger.info(f"Generating variant study snapshot for '{variant_study_id}'")
 
         root_study, descendants = self._retrieve_descendants(variant_study_id)
-        assert_permission_on_studies([root_study, *descendants], StudyPermissionType.READ)
         if root_study.archived:
             raise UnsupportedOperationOnArchivedStudy(root_study.id)
         search_result = self.search_ref_study(root_study, descendants, from_scratch=from_scratch)
```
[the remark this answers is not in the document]
sylvlecl: This is on purpose, permissions need to be reviewed to be managed at the tree level.
It's acceptable that for that edge case, the user will be able to generate the variant although the parent ownership has changed. Permissions are enforced at variant creation time.

antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
```diff
@@ -251,8 +240,7 @@ def search_ref_study(
         # If no such variant is found, we use the root study as a reference study.
 
         ref_study = root_study
-        # Give the list in reverse order to find the most recent variants first.
-        ref_variant_study, commands = _find_last_snapshot_up_to_date(descendants[-2::-1])
+        ref_variant_study, commands = _find_last_snapshot_up_to_date(descendants[:-1])
```
MartinBelthle, line 246: Really minor but I think now you're unecessary checking the first variant again here even thought we checked it earlier

antarest/study/storage/variantstudy/repository.py
MartinBelthle: This is now unused you can remove it

antarest/study/storage/variantstudy/repository.py
```diff
@@ -202,25 +209,32 @@ def is_snapshot_up_to_date(self, study_id: str) -> bool:
...
+            cte, Study.id == cte.c.parent_id
+        )
+        cte = cte.union_all(recursive_q)
 
-        join_query = [
+        study_w_p = with_polymorphic(Study, "*")
+
+        stmt = select(study_w_p).join(cte, study_w_p.id == cte.c.id).order_by(cte.c.depth.desc())
+        stmt = stmt.options(
             joinedload(study_w_p.owner),
```
MartinBelthle, line 216: Seems we don't need to fetch the `owner` and `groups` anymore if we do not perform a permission check

MartinBelthle (dismissed):

antarest/study/storage/variantstudy/repository.py
sylvlecl: Thought the same: actually one purpose of this was to be able to remove the joinedload on those tables.

I tried to replace the `joinedload` with `raiseload` which raises if the groups are lazily loaded later (allows to identify if there are unwanted N+1 queries).
It raised when saving the variant later on, because the variant saving includes refreshing the groups.
It may be something that should be changed, but I preferred to not touch that in that PR.

antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
sylvlecl: The `-1` excludes the last variant (which is the one we're generating for), so I think it's OK ?
I changed the logic so that we always keep descendants in the "natural" order from parent to child, not reversing it here.

MartinBelthle: Indeed my bad

antarest/study/storage/variantstudy/repository.py
MartinBelthle: Oh okay complex. Perhaps we should comment it with a TODO as it's not that easy to understand

sylvlecl: Done

antarest/study/storage/variantstudy/snapshot/snapshot_generator.py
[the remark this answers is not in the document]
sylvlecl: Added a comment

PR #3320, opened 2026-07-28
fix(variant): invalidate snapshot during generation

sylvlecl:
We now invalidate the previous snapshot at the start of generation.

This will avoid to keep the previous snapshot as "valid" if the
generation starts modifying it and is interrupted, for example by a
process crash. Could not find an easy way to unit test that.

Also set expire_on_commit=False in unit tests, in consistence with the
actual application configuration.


tests/study/test_repository.py
MartinBelthle: Minor but we should update this comment too

tests/study/test_repository.py
```diff
@@ -1133,16 +1133,16 @@ def test_update_tags(
...
     with DBStatementRecorder(db_session.bind) as db_recorder:
         repository.update_tags(study, ["Tag1", "Tag2"])
-    assert len(db_recorder.sql_statements) == 6, str(db_recorder)
+    assert len(db_recorder.sql_statements) == 4, str(db_recorder)
 
     # Check that when we change the tags to ["TAG1", "Tag3"],
     # "Tag1" is preserved, "Tag2" is deleted and "Tag3" is created
     # 1- finding existing tags requires 1 query
-    # 2- updating the study tags requires 4 queries (2 selects, 2 inserts, 1 delete)
+    # 2- updating the study tags requires 2 queries (2 inserts, 1 delete)
```
MartinBelthle: Minor but it's 3

tests/variantstudy/model/test_variant_model.py
```diff
@@ -295,10 +295,9 @@ def test_command_same_author(
         with current_user_context(jwt_user):
             with DBStatementRecorder(db.session.bind) as db_recorder:
                 variant_study_service.get_commands(variant_study.id)
-                # 2 queries must be executed:
+                # Only 1 query must be executed:
```
MartinBelthle, line 300: The test was here to ensure we do not perform N+1 queries.

But I think you only have 1 request now because you're running the `get_commands` inside the `with_current_user` context.

In production, this would perform 2 queries. I think we should find a way to really test what we wanted to do initially

sylvlecl: Actually, the comment was wrong: there was 2 requests only because "variant.id" was issuing a select, because the variant object was expired.

The function `get_commands` was already issuing only 1 request, which retrieves everything needed.

I still isolated that code block in a new "with db()" to make sure the session is independent from code above

MartinBelthle (approved):

PR #3321, opened 2026-07-28
feat(launcher): slurm add oversubscribe

smailio:


antarest/launcher/model.py
```diff
@@ -697,6 +697,33 @@ def apply_update_solver_presets(
     return SolverPresetsDB.from_model(updated_dto)
 
 
+class SlurmRuntimeConfig(AntaresBaseModel):
```
sylvlecl, line 700: We could add some docstring to explain what is its role, without having to go and read the code where it's used

sylvlecl (commented):
Looks good to me: maybe we can just add a docstring to the config class to explain a little what it's used for ?

sylvlecl (dismissed):

PR #3328, opened 2026-08-04
fix(maintenance): fix multiple services creation

sylvlecl:
Services should be created only once.

Also, aligns naming of all services into `Services` dataclass.


antarest/maintenance/context.py
```diff
@@ -50,31 +48,30 @@ def create(cls, config: "Config") -> "MaintenanceContext":
 
         engine = init_db_engine(config, auto_upgrade_db=False)
         init_db_singleton(custom_engine=engine, session_args=SESSION_ARGS)
-        core_services = create_core_services(config=config)
         services = create_services(config=config)
```
sylvlecl, line 51: In this method invocation we re-created the core services, effectively creating duplicates.

I got an error on my local env with metrics being added multiple times.

sylvlecl:
Should not be needed following https://github.com/AntaresSimulatorTeam/AntaREST/pull/3331

PR #3330, opened 2026-08-04
feat(output_disk_usage): added the output disk usage in database 

mehdiwahada:
What does it do ? : it adds an output table in the database and the selected output is deleted when archived, unarchived, converted or deleted (yes)

Why ? : Because the disk usage is recalculated each time we're asking for him and it is optimal to stock and get him when needed.


alembic/versions/6012b0407e38_add_output_table_in_db.py
```diff
@@ -0,0 +1,32 @@
...
+down_revision = "3855b61c3232"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "output",
+        sa.Column("output_id", sa.String(), nullable=False),
+        sa.Column("study_id", sa.String(), nullable=False),
```
sylvlecl, line 23: we'd better have the study_id first in the columns and in the primary key:
logically outputs belong to a study, therefore when sorting the primary keys we should have study_id first.

antarest/output/model.py
```diff
@@ -93,3 +97,26 @@ class OutputVariablesViewStatus(StrEnum):
...
+    __tablename__ = "output"
+
+    study_id: Mapped[str] = mapped_column(
+        String(),
+        ForeignKey("study.id", name="fk_output_study_id", ondelete="CASCADE"),
+        primary_key=True,
+        nullable=False,
+    )
+    output_id: Mapped[str] = mapped_column(String(), primary_key=True, nullable=False)
+    disk_space_bytes: Mapped[int] = mapped_column(Integer(), nullable=True, default=None)
```
sylvlecl: Should be BigInteger in consistence with the alembic migration.
Also, if it's nullable, the type should be `Mapped[int | None]`

antarest/output/repository.py
```diff
@@ -0,0 +1,48 @@
...
+    @property
+    def session(self) -> Session:
+        if self._session is None:
+            return db.session
+        return self._session
+
+    def save(self, output_metadata: Output) -> Output:
+        session = self.session
+        output = session.merge(output_metadata)
+        session.add(output)
```
sylvlecl: add is actually redundant with merge, we can remove it

antarest/output/service.py
```diff
@@ -912,7 +920,13 @@ def get_logs(self, study_id: str, output_id: str, log_type: LogType) -> str:
...
-        return self._find_output_storage(study_id, output_id).get_disk_usage(study_id, output_id)
+        output = self._output_repository.get(study_id, output_id)
+        if output and output.disk_space_bytes is not None:
+            return output.disk_space_bytes
+        else:
+            disk_usage = self._find_output_storage(study_id, output_id).get_disk_usage(study_id, output_id)
+            self._output_repository.save(Output(study_id=study_id, output_id=output_id, disk_space_bytes=disk_usage))
+            return disk_usage
 
     def convert_output(self, study_id: str, output_id: str, storage_type: OutputStorageType) -> None:
```
sylvlecl, line 931: copilot identified that I missed some cases where we need to remove the cached size:

- when we call convert_output: we change the storage format, so the size will likely change
- when we delete the output, we also need to delete the associated entry

antarest/output/service.py
```diff
@@ -335,6 +339,8 @@ def unarchive_output_task(notifier: ITaskNotifier) -> TaskResult:
             custom_event_messages=None,
         )
 
+        self._output_repository.delete(study_id, output_id)
```
sylvlecl: This must be done at the end of the task execution, not just after submitting the task like here:
it means this need to be moved inside `unarchive_output_task` above, after the unarchival has actually completed.

Otherwise, the following thing can happen:
1. someone starts an unarchival --> the disk usage is remove right away
2. someone calls get_disk_usage --> disk usage is re-computed on a partially unarchived study, and written to database
3. the unarchival task completes: now its size has probably changed a lot compared to what we have cached in 2

Same thing for the archive method below

antarest/output/model.py
```diff
@@ -93,3 +97,26 @@ class OutputVariablesViewStatus(StrEnum):
...
+    study: Mapped["Study"] = relationship(Study)
+
+    def to_dto(self) -> "OutputDTO":
+        return OutputDTO(output_id=self.output_id, study_id=self.study_id, disk_space_bytes=self.disk_space_bytes)
+
+
+class OutputDTO(AntaresBaseModel, extra="forbid"):
```
sylvlecl: for now that class does not seem to be used: let's remove it

sylvlecl (changes requested):
Looks mostly good :) a few fixes still required before we can merge !

sylvlecl (dismissed):
Good thank you :)

PR #3337, opened 2026-08-06
fix(output): search digest in correct folder in adequacy

sylvlecl:
The search of the digest was hard-coded in "economy" folder, we now correctly look 
into adequacy instead where relevant

ANT-5331


tests/integration/output/test_digest.py
```diff
@@ -9,50 +9,94 @@
...
+import zipfile
+from pathlib import Path
 
+import pytest
 from starlette.testclient import TestClient
 
+from antarest.core.utils.archives import ArchiveFormat, archive_dir
 
-class TestDigest:
-    def test_get_digest_endpoint(self, client: TestClient, user_access_token: str, internal_study_id: str) -> None:
```
sylvlecl, line 17: All that looks removed just because I deleted the containing class, but only the end of the test is actually changed

smailio (approved):
Solution is straightforward and good test coverage

PR #3342, opened 2026-08-10
fix(output): fix "output download" when areas have different variables

sylvlecl:
Different areas of the study may have different variables, in particular when they contain
different thermal cluster groups, or STS groups, etc.

In that case, the "output download" feature is expected to return only the relevant variables
for each area, not their union.

The implementation implies to move the responsibility of building that object to the storage
level, because different storages will store the information about area variables in a different
way.

**This will require a future evolution in parquet implementation to store more metadata**

Also makes the feature more efficient (0.7s -> 0.25s on one year of a large study).


tests/output/filestudy/data/20260810-1420eco-thermal_groups/checkIntegrity.txt
```diff
@@ -0,0 +1,8 @@
+-1.58056824843973e+04
```
MartinBelthle, line 1: Shouldn't all these files be zipped in a single output archive as we do elsewhere in the tests ?
This way if we want to add big files for future tests, the dev won't have to change this

antarest/output/filestudy/download.py
```diff
@@ -0,0 +1,87 @@
...
+)
+from antarest.study.storage.utils import get_start_date
+
+
+def build_matrix_aggregation_result(output_path: Path, data_selection: StudyDownloadDTO) -> MatrixAggregationResultDTO:
+    """
+    Build a MatrixAggregationResultDTO from the given output path and data.
+    """
+    if data_selection.type == StudyDownloadType.DISTRICT:
+        raise ValueError("DISTRICT download type is not supported anymore")
```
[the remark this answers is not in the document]
MartinBelthle: Districts were handled as areas before and it used to work (tested on recette env). At the very least it's a breaking change that we should mark and notify the Imagrid team.

IMO we shouldn't change this

antarest/output/filestudy/model.py
```diff
@@ -78,32 +98,61 @@ def get_output_object_type(
...
     """
     Serializes multi-indexed column headers into a single string, concatenating with " % " as a separator.
     """
-    data.headers = [" % ".join(col) for col in data.headers]
+    if isinstance(output_header, str):
+        raise ValueError("Header concatenation is only supported for tuple headers.")
+    return " % ".join(output_header)
 
 
 def split_concatenated_columns_from_dataframe(df: pd.DataFrame) -> Iterator[TimeSerie]:
```
MartinBelthle: Not used anymore

MartinBelthle (changes requested):
First comments, i didn't read the whole thing as it's hard to

antarest/output/filestudy/variables.py
```diff
@@ -88,7 +88,7 @@ def _read_headers_only(
     """
 
     with open(file_path, encoding="utf-8") as file:
-        output_headers = parse_headers(file, start_column)
+        output_headers = [v.to_tuple() for v in parse_headers(file, start_column)]
```
MartinBelthle: We should try to get rid ot the `.to_tuple()` method.
You've introduced a dataclass which is clearer so we should use it all the way rather than using `col[0]` as we did before

For instance the `normalize_df_column_names` function could be a method of the `VariableDescription` dataclass that would take a `McRoot` as an argument

antarest/output/filestudy/download.py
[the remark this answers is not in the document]
sylvlecl: Indeed; I think I got confused with the areas endpoint where we deprecated that.
Probably not used though, but I'll restore it

tests/output/filestudy/data/20260810-1420eco-thermal_groups/checkIntegrity.txt
sylvlecl: I think it's better to keep the unzipped versions here.
If you're worried about the disk space of the repo, git already stores files compressed, so that will not be a problem.

Keeping them as text allows to better understand the test, and also possible future changes (they will be easier to review and understand).
If we get to many files, that could start to hurt, but there are really only a few here.

antarest/output/filestudy/variables.py
sylvlecl: I am working on this, but we aware that it will make the review a little harder !
At first, I tried to limit the scope of changes to make it easier.
But ok to push it more in that PR, on my side :)

antarest/output/filestudy/download.py
```diff
@@ -0,0 +1,87 @@
...
+        year = file_data.file.metadata.year
+        df = file_data.data.data
+
+        # TODO: better handling of the link case, with 2 separate strings instead of that arbitrary formatting
+        if data_selection.type == StudyDownloadType.LINK:
+            element_name = "^".join(element_name.split(" - "))
+
+        for var_index, var in enumerate(file_data.data.headers):
+            if data_selection.columns and var.name not in data_selection.columns:
+                continue
```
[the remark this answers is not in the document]
sylvlecl: Historically the code was this:
```python
    if data.columns and len(data.columns) > 0 and column_name not in data.columns:
        continue
```

So this just restores the old behaviour where we did not use the "aggregation" function, which is finally not appropriate. Let's keep it this way.

antarest/output/filestudy/variables.py
sylvlecl: I removed most usage of to_tuple, it's only used now to create tuples for a pandas dataframe index.

I removed the argument `transform_column_headers` which does not make sense any more, too.

antarest/output/filestudy/download.py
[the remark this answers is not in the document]
sylvlecl: I reverted that change, added a test with a district and made it pass

antarest/output/filestudy/variables.py
MartinBelthle: I'll perhaps regret it when I'll review it but I think it's for the better :)

antarest/output/filestudy/model.py
```diff
@@ -57,8 +53,58 @@ class MCAllLinksQueryFile(StrEnum):
 
 QueryFileType: TypeAlias = MCIndAreasQueryFile | MCAllAreasQueryFile | MCIndLinksQueryFile | MCAllLinksQueryFile
 
-SingleOutputHeaders: TypeAlias = list[str]
-MultipleOutputHeaders: TypeAlias = list[list[str]]
+AggregationType: TypeAlias = Literal["mc-all", "mc-ind"]
+
+
+@dataclass(frozen=True, slots=True)
```
MartinBelthle, line 62: Didn't know `slots` argument, seems nice, did you see real gains in performance while using it ?

antarest/output/filestudy/model.py
```diff
@@ -78,39 +124,27 @@ def get_output_object_type(
             return "areas"
 
 
+C = TypeVar("C")
+C2 = TypeVar("C2")
+
+
 @dataclass
```
MartinBelthle: Still worth using dataclass with a in-house init method ?

antarest/output/filestudy/model.py
```diff
@@ -78,39 +124,27 @@ def get_output_object_type(
...
 
-def concatenate_dataframe_multi_indexed_columns(data: OutputDataFrame) -> None:
-    """
-    Serializes multi-indexed column headers into a single string, concatenating with " % " as a separator.
-    """
-    data.headers = [" % ".join(col) for col in data.headers]
+    data: pl.DataFrame
+    headers: list[C]
 
+    def __init__(self, data: pl.DataFrame, headers: Sequence[C]):
```
MartinBelthle: Seems headers is always given as a list in the constructor. So doing list(headers) does a shallow copy that we don't need. Could improve perfs a bit as we have dataframes with long headers

antarest/output/filestudy/variables.py
```diff
@@ -93,10 +92,10 @@ def _read_headers_only(
     if "details" in file_type.value:
         cols_mapping: dict[str, set[str]] = {}
         for col in output_headers:
-            cols_mapping.setdefault(col[0], set()).add(col[1])
+            cols_mapping.setdefault(col.name, set()).add(col.unit or " ")
```
MartinBelthle: You should use `unit_repr`

antarest/output/filestudy/download.py
```diff
@@ -0,0 +1,85 @@
...
+            ts_data = element_results.setdefault(
+                element_name, TimeSeriesData(type=data_selection.type, name=element_name, data={})
+            )
+            numerical_data = df.to_series(var_index).cast(float).to_list()
+            ts_data.data.setdefault(str(year), []).append(
+                TimeSerie(name=var.name, unit=var.unit or " ", data=numerical_data)
```
MartinBelthle: Same here we should use `unit_repr`

antarest/output/filestudy/iteration.py
```diff
@@ -0,0 +1,219 @@
...
+    OutputDataFrame,
+    QueryFileType,
+    VariableDescription,
+)
+from antarest.study.model import MatrixFrequency
+
+logger = logging.getLogger(__name__)
+
+
+def _find_mode_dir(output_dir: Path) -> Path:
```
MartinBelthle: This function exists almost identically in output/utils.py, we should put code in common

antarest/output/filestudy/download.py
```diff
@@ -0,0 +1,85 @@
...
+        df = file_data.data.data
+
+        # TODO: better handling of the link case, with 2 separate strings instead of that arbitrary formatting
+        if data_selection.type == StudyDownloadType.LINK:
+            element_name = "^".join(element_name.split(" - "))
+
+        for var_index, var in enumerate(file_data.data.headers):
+            if data_selection.columns and var.name not in data_selection.columns:
+                continue
+            ts_data = element_results.setdefault(
```
MartinBelthle, line 73: **Minor**:   

Not really used to using the return of `setdefault` but it seems it returns the `TimeSeriesData` object.
I would personally prefer you'd first fill the entire `TImeSeriesData` object before inserting it in the dict

antarest/output/filestudy/iteration.py
```diff
@@ -0,0 +1,219 @@
...
+    Carries information about the current state of a matrix which was read from an output file, and possibly already
+    transformed (column fitlers, pivots, ...).
+    """
+
+    file: OutputFile
+    data: OutputDataFrame[VariableDescription]
+
+
+def _filter_files(folder_path: Path, ids: set[str]) -> list[str]:
+    # Areas names filtering
```
MartinBelthle: Seems that it's used for both areas and links.

antarest/output/filestudy/iteration.py
```diff
@@ -0,0 +1,219 @@
...
+    filtered = sorted([d.name for d in folder_path.iterdir()])
```
MartinBelthle, line 89: I think we do not need to sort here as the sorting is done inside aggregation.

If it's not needed we could use a set to find ids faster. Perhaps the overhead of building a set is too big compared to a list in most cases I don't know. Apparently it's 2 times faster to create a list compared to a set. But if the user has 100 areas and he gives you 50, we'll see the difference.

antarest/output/filestudy/iteration.py
```diff
@@ -0,0 +1,219 @@
...
+) -> list[OutputFile]:
+    mode_dir = _find_mode_dir(output_path)
+    mc_ind_path = mode_dir / MCRoot.MC_IND.value
+    if not mc_ind_path.exists():
+        raise OutputSubFolderNotFound(output_path.name, f"{mode_dir.name}/mc-ind")
+
+    output_type = "areas" if isinstance(query_file, MCIndAreasQueryFile) else "links"
+
+    # Monte Carlo years filtering
+    if not mc_ind_path.is_dir():
```
MartinBelthle: How could this be as you raised above. It will mean it's a file. Seems we don't need to handle this case

antarest/output/filestudy/iteration.py
```diff
@@ -0,0 +1,219 @@
...
+        return []
+    all_mc_years = [d.name for d in mc_ind_path.iterdir()]
+    if mc_years:
+        all_mc_years = [y for y in all_mc_years if int(y) in frozenset(mc_years)]
```
MartinBelthle, line 112: Why using a frozenset compared to a classic one ?

antarest/output/filestudy/iteration.py
```diff
@@ -0,0 +1,219 @@
...
+    def parse(self) -> "OutputFileData":
+        data = parse_output_file(self.path, get_start_column(self.metadata.frequency))
+        return OutputFileData(file=self, data=data)
+
+
+@dataclass(frozen=True)
+class OutputFileData:
+    """
+    Carries information about the current state of a matrix which was read from an output file, and possibly already
+    transformed (column fitlers, pivots, ...).
```
MartinBelthle: Seems that it's always the original content unfiltered, unpivot

antarest/output/filestudy/iteration.py
```diff
@@ -0,0 +1,219 @@
...
+
+@dataclass(frozen=True, order=True)
+class OutputFileMetadata:
+    """
+    Attributes of one output matrix file.
+    Uniquely identifies a file in the output.
+    """
+
+    year: int | Literal["mc-all"]
+    file_type: QueryFileType
```
MartinBelthle, line 60: Seems unused for now. Perhaps it will be used in a future dev. For now i think we can remove it

MartinBelthle (changes requested):
Read everything, i think the code is clearer with the classes you've introduced thanks :).

Didn't grasp how you let go of the `transform_column_headers` arg but it's nice. I imagine it's due to the new `VariableDescription` class you're using that you can choose in the end how to handle the data

Also makes me realize I did not even check the fix you made. I don't understand what was wrong with the old code :/, could you comment on the part of the code that was wrong to help me ?

antarest/output/filestudy/model.py
sylvlecl: No to be honest I am pretty sure it does not make a real difference here

antarest/output/filestudy/model.py
sylvlecl: The issue is with "covariance": we cannot give list[str] where list[VariableDescription | str] is expected, which caused issues.

However, I changed it: now the field is typed as a `Sequence` instead of `list`, which remvoes the problem and the need for the initializer override

antarest/output/filestudy/model.py
sylvlecl: Still brings some features (str representations, equality etc) in particular when "frozen".
Actually I meant it to be frozen: it is now

antarest/output/filestudy/download.py
sylvlecl: Not possible, the `TimeSeriesData` object will be filled by multiple iterations of this loop.
The setdefault simply say: if it does not exist yet create it, else use the one that is already here.

Avoids having several lines just for the logic "if element_name in dict then get it else create it"

antarest/output/filestudy/iteration.py
sylvlecl: generally speaking it makes more sense to use a frozenset when no mutation is foreseen, which is the case here, but there is no other reason.
I would rather question all other places where we use a `set` whereas we should maybe use `frozenset` :)

antarest/output/filestudy/iteration.py
sylvlecl: Indeed, this is inherited from uncommitted changes that may not end up in the code, I will change that

antarest/output/filestudy/iteration.py
sylvlecl: I'd prefer to keep it, this way we have all the metadata that characterizes the file, not only part of it, which is the goal of that class

antarest/output/filestudy/iteration.py
sylvlecl: Indeed

antarest/output/filestudy/iteration.py
sylvlecl: It's better to have a deterministic behaviour for the functions in that module.
If we don't sort here, they will not be stable.

The selected IDs is already a set, so iterating over the list of files and seach inside the set have reasonable complexity. WIll really not be a bottleneck anyway here

antarest/output/filestudy/iteration.py
sylvlecl: indeed I improved the docstring

antarest/output/filestudy/iteration.py
sylvlecl: I merged them into `filestudy.model` which is the more "generic" place for now for filestudy specific functions.

Remove `output.utils`, it was the last function inside it.

MartinBelthle (approved):

PR #3346, opened 2026-08-11
refactor(db-study): add an intermediate study_data_id as a bigint

sylvlecl:
Both:
- a design improvement, because now study data actually get a proper identifier,
  distinct from the study ID, which makes more sense because they don't share
  their lifetime: for instance, a variant snapshot will be deleted independently from
  its containing study.
- will avoid repeating 36 char strings thousands or millions of times in DB tables,
  improving disk and memory efficiency

Implemented with LLM support, and reviewed carefully.

Implementation note:
for postgresql the migration APPENDS the new study_data_id to existing columns,
therefore it will appear last. It should not have negative impacts about performance
or disk usage, the important point being that it's the first column in primary keys.
This avoids re-creating all tables from scratch.

### DIFF
--- a/antarest/study/dao/database/models/user_resources.py
+++ b/antarest/study/dao/database/models/user_resources.py
@@ -14,11 +14,12 @@
 SQLAlchemy Core table definitions for user resources storage.
 """
 
-from sqlalchemy import Column, ForeignKeyConstraint, String, Table
+from sqlalchemy import Column, ForeignKeyConstraint, Index, String, Table
 
 from antarest.core.utils.sql_utils import enum_col
 from antarest.dbmodel import Base
 from antarest.study.business.model.user_model import ResourceType
+from antarest.study.dao.database.models import study_data_id_col
 
 metadata = Base.metadata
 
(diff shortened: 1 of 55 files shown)


antarest/study/dao/database/models/user_resources.py
```diff
@@ -27,12 +28,15 @@
...
     Column("name", String(255), nullable=False),
     Column("parent_id", String(36), nullable=True),
     Column("resource_type", _RESOURCE_TYPE_ENUM, nullable=False),
     Column("blob_id", String(64), nullable=True),
-    ForeignKeyConstraint(["study_id"], ["study_data.study_id"], ondelete="CASCADE"),
+    ForeignKeyConstraint(["study_data_id"], ["study_data.study_data_id"], ondelete="CASCADE"),
     ForeignKeyConstraint(["parent_id"], ["user_resources.id"], ondelete="CASCADE"),
+    # This is the only data table whose study key is not the leading column of the primary key,
+    # so it needs an explicit index: neither PostgreSQL nor SQLite indexes foreign keys.
+    Index("ix_user_resources_study_data_id", "study_data_id"),
```
smailio, line 41: is it normal in the first place that it wasn't part of the primary key ?

sylvlecl: Very good point: I don't think so (this should be handled separately though, of course)

PR #3348, opened 2026-08-11
fix(ui): stabilize debounced output export in ExportModal

hdinia:
## Fix broken debounce in ExportModal
[(Sonar issue) ](https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&branch=refactor%2Fadd-study-data-id-bigint&id=AntaresSimulatorTeam_api-iso-antares&open=AZ_xFSavG2oxJMR_1auA)

`exportOutput` was created with `lodash/debounce` directly in the component body, so every render produced a fresh instance with a clean timer. 
Since clicking Export closes the dialog (triggering a re-render), the guard never actually suppressed anything, a double-click during the close transition could fire the export API twice.

Fixed by switching to the existing `useDebounce` hook, which memoizes the
debounce instance and always invokes the latest callback via a ref.

Resolves the Sonar warning on this file.

### DIFF
--- a/webapp/src/routes/-shared/components/studies/dialogs/ExportModal/index.tsx
+++ b/webapp/src/routes/-shared/components/studies/dialogs/ExportModal/index.tsx
@@ -14,12 +14,12 @@
 
 import BasicDialog, { type BasicDialogProps } from "@/components/dialogs/BasicDialog";
 import SelectSingle from "@/components/SelectSingle";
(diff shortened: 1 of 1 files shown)


sylvlecl (approved):
Thanks!

PR #3351, opened 2026-08-13
fix(study): path is actually nullable now

sylvlecl:
study.path has always been nullable in database, but it was not correctly typed
in our code.
Now, path is indeed actually null for DB studies, which breaks some features
such as the study scan.

Needs to be fixed before making it available in prod.


antarest/study/model.py
```diff
@@ -311,7 +311,7 @@ class Study(Base):
     created_at: Mapped[datetime | None] = mapped_column(DateTime, nullable=True, index=True)
     updated_at: Mapped[datetime | None] = mapped_column(DateTime, nullable=True, index=True)
     last_access: Mapped[datetime | None] = mapped_column(DateTime, nullable=True)
-    path: Mapped[str] = mapped_column(String())
+    path: Mapped[str | None] = mapped_column(String(), nullable=True)
```
sylvlecl, line 314: Note that it was already nullable which is the default, juste making it explicit

sylvlecl:
Closing, will be handled in another clean branch

PR #3354, opened 2026-08-13
feat(reserves): support short term storage reserves

olivierPigeon-RTE:


antarest/study/business/model/reserve_certification_model.py
```diff
@@ -9,30 +9,59 @@
...
 
 from antarest.core.serde import AntaresBaseModel
 from antarest.study.business.model.reserve_definition_model import ReserveDefinitionId
+from antarest.study.dao.common import AreaAssetId
 
 Cost = Annotated[float, Field(ge=0)]
 Power = Annotated[float, Field(ge=0)]
 
+
+class ReserveCertification(AntaresBaseModel):
```
sylvlecl: From a design point of view, it's an over-use of inheritance here:
- we never use interchangeably both types of certifications, so having a parent class does not really make sense. We should not use inheritance just because some properties are common "by chance" between 2 classes. 
- it's actually more code and harder to follow since the actual implementation classes are not self contained anymore

Generally speaking, inheritance is often not very interesting for pure dataclasses

antarest/study/storage/rawstudy/model/filesystem/config/reserve_participations.py
```diff
@@ -0,0 +1,296 @@
...
+    ReserveCertification,
+    StorageReserveCertification,
+    ThermalReserveCertification,
+)
+from antarest.study.business.model.reserve_definition_model import ReserveDefinitionId
+from antarest.study.business.model.reserve_symmetries_model import ReserveSymmetries, ReserveSymmetry
+from antarest.study.storage.rawstudy.model.filesystem.config.identifier import transform_name_to_id
+
+
+class _AreaAssetCertification(ABC, AntaresBaseModel):
```
sylvlecl: Same comment here: it's an over-use of inheritance, let's use plain dataclasses, not an ABC for this

antarest/study/storage/rawstudy/model/filesystem/config/reserve_participations.py
```diff
@@ -0,0 +1,296 @@
...
+    reserve: str
+
+    @abstractmethod
+    def to_model(self) -> Any:
```
sylvlecl: See here, the abstraction does not make much sense since we don't want the same return type (could be made generic to have strong type checking, but really it's not worth it, far simpler to keep classes separate)

antarest/study/storage/rawstudy/model/filesystem/config/reserve_participations.py
```diff
@@ -0,0 +1,296 @@
...
+
+    @override
+    def set_id(self, id: str) -> None:
+        self.storage = id
+
+
+ParticipationT = TypeVar("ParticipationT", bound=Participation[Any])
+
+
+class _AreaAssetParticipationFileData(ABC, AntaresBaseModel, Generic[ParticipationT]):
```
sylvlecl, line 119: here too the inheritance looks a little overkill, the class contains mostly class functions that could be free functions 

On the other side, it works, probably not worth changing the implementation now

antarest/study/storage/variantstudy/model/command/replace_st_storage_reserve_certifications.py
```diff
@@ -0,0 +1,79 @@
...
+    # Command parameters
+    # ==================
+
+    area_id: AreaId
+    certifications: StorageReserveCertificationMapping
+
+    @model_validator(mode="after")
+    def _validate_version(self) -> Self:
+        if self.study_version < STUDY_VERSION_10_0:
+            msg = "Short-term storage reserve certifications are not valid for study version before 10.0"
```
sylvlecl: Actually, after discussion with atnares-solver team, the reserves model will only be valid from 10.2.
We should update it, also in other places where we version-check

antarest/study/web/study_data_blueprint.py
```diff
@@ -1698,14 +1700,64 @@ def get_reserve_certifications(
...
         return study_service.reserve_certifications_manager.set_thermal_certifications(study_interface, area_id, data)
 
+    @bp.get(
+        path="/studies/{uuid}/areas/{area_id}/reserves/symmetries/storages",
+        summary="Fetch all short-term storages reserve symmetries for a given area",
+    )
+    def get_st_storage_reserve_symmetries(
+        study_service: StudyServiceDep, uuid: UuidStr, area_id: SanitizedStr
+    ) -> dict[StorageId, ReserveSymmetries]:
+        logger.info("Fetching reserve symmetries for study '%s' and area '%s'", uuid, area_id)
```
sylvlecl: weird that we did not used f-strings in those logs

antarest/study/business/model/reserve_certification_model.py
```diff
@@ -9,30 +9,59 @@
...
 
-class ThermalReserveCertification(AntaresBaseModel):
+class ThermalReserveCertification(
+    ReserveCertification,
+):
     model_config = ConfigDict(alias_generator=to_camel, extra="forbid", populate_by_name=True)
 
     max_power: Power = 0.0
     max_power_off: Power = 0.0
-    participation_cost: Cost = 0.0
```
sylvlecl, line 33: (will also change the serialization order by the way, not a breaking change because it's equivalent, but makes more sense to pack together the participation*** properties)

antarest/core/exceptions.py
```diff
@@ -865,3 +865,9 @@ class ThermalReserveCertificationNotFound(HTTPException):
     def __init__(self, area_id: str, thermal_id: str, reserve_ids: set[str]):
         msg = f"Certifications for reserve(s) '{reserve_ids}' on thermal cluster '{thermal_id}' not found in area '{area_id}'"
         super().__init__(HTTPStatus.NOT_FOUND, msg)
+
+
+class ReserveCertificationNotFound(HTTPException):
```
sylvlecl, line 870: The previous exception does not seem used anymore

antarest/study/dao/api/common.py
```diff
@@ -9,26 +9,30 @@
...
+)
 from antarest.study.business.model.reserve_definition_model import ReserveDefinitionId
 from antarest.study.business.model.reserve_symmetries_model import ReserveSymmetries
-from antarest.study.business.model.thermal_reserve_certification_model import ThermalReserveCertificationMapping
-from antarest.study.dao.common import AreaId, ThermalId
+from antarest.study.dao.common import AreaAssetId, AreaId
 
 
-def check_thermal_symmetries_are_certified(
+def check_symmetries_are_certified(
```
sylvlecl, line 23: Good thing that you have been able to re-use that

antarest/study/dao/database/database_reserve_certification_dao.py
```diff
@@ -139,3 +150,103 @@ def _raise_the_right_thermal_reserve_exception(
...
+        values = []
+        for area_id, reserves_dict in data.items():
+            for reserve_id, storage_dict in reserves_dict.items():
+                for storage_id, certification in storage_dict.items():
+                    values.append(
+                        _convert_st_storage_model_to_row(self._study_id, area_id, storage_id, reserve_id, certification)
+                    )
+        return values
+
+    def __clean_db(
```
sylvlecl: not sure why we use double underscores for those methods ?

antarest/study/dao/file/file_study_reserve_certification_dao.py
```diff
@@ -61,28 +71,79 @@ def get_thermal_reserve_certifications(self, area_id: AreaId) -> ThermalReserveC
...
+            self._check_all_thermals_exist_in_area(area_id, reserves_dict)
 
             yaml_content = get_thermal_reserve_participations_as_yaml_content(area_id, file_study)
             symmetries = parse_thermal_reserves_symmetries(yaml_content)
             new_content = serialize_thermal_reserve_participations(symmetries, reserves_dict)
 
             # Saves the content into the YAML file
             file_study.tree.save(new_content, get_thermal_reserve_path(area_id))
+
+    @staticmethod
```
sylvlecl: we'd better prefer free functions over staticmethods, it just make the call site more verbose with no real added value

antarest/study/dao/file/file_study_reserve_certification_dao.py
```diff
@@ -61,28 +71,79 @@ def get_thermal_reserve_certifications(self, area_id: AreaId) -> ThermalReserveC
...
+    def _check_reserves_exist(
+        area_id: str,
+        file_study: FileStudy,
+        reserves_dict: Mapping[ReserveDefinitionId, Mapping[str, ReserveCertification]],
+    ) -> None:
+        existing_reserve_ids = file_study.config.areas[area_id].reserves
+        invalid_reserves: set[str] = set(reserves_dict) - set(existing_reserve_ids)  # type: ignore
```
sylvlecl: probably better have a cast instead of type ignore on the whole line, really a detail though

antarest/study/storage/rawstudy/model/filesystem/root/input/st_storage/clusters/area/area.py
```diff
@@ -11,28 +11,23 @@
...
-        super().__init__(matrix_storage_context, config)
-        self.area = area
-
     @override
     def build(self) -> TREE:
         # Each area has a folder containing a file named "list.ini"
-        # If the area does not have any short term storage cluster, the file is empty.
+        # If the area does not have any short-term storage cluster, the file is empty.
         children: TREE = {"list": InputSTStorageAreaList(self.config.next_file("list.ini"))}
+        if self.config.version >= STUDY_VERSION_10_0:
```
sylvlecl: another place where we'll need to go to 10.2

antarest/study/storage/rawstudy/model/filesystem/root/input/st_storage/clusters/area/area.py
```diff
@@ -11,28 +11,23 @@
...
+            from antarest.study.storage.rawstudy.model.filesystem.yaml_file_node import YAMLFileNode
```
sylvlecl: should go at the top of the file

antarest/study/storage/rawstudy/model/filesystem/root/input/st_storage/clusters/clusters.py
```diff
@@ -17,11 +17,11 @@
 
 
 class InputSTStorageClusters(FolderNode):
-    # Each area has it own folder named after the area id.
+    # Each area has its own folder named after the area id.
     @override
     def build(self) -> TREE:
         children: TREE = {
-            a: InputSTStorageArea(self.matrix_storage_context, self.config.next_file(a), area=a)
+            a: InputSTStorageArea(self.matrix_storage_context, self.config.next_file(a))
```
sylvlecl, line 24: good thing to have remove the area arg!

antarest/study/web/study_data_blueprint.py
```diff
@@ -1698,14 +1700,64 @@ def get_reserve_certifications(
...
-    def save_reserve_certifications(
+    def save_thermal_reserve_certifications(
         study_service: StudyServiceDep, uuid: UuidStr, area_id: SanitizedStr, data: ThermalReserveCertificationMapping
     ) -> ThermalReserveCertificationMapping:
         logger.info("Saving reserve certifications for study '%s' and area '%s'", uuid, area_id)
         study = study_service.check_study_access(uuid, StudyPermissionType.WRITE)
         study_interface = study_service.get_study_interface(study)
         return study_service.reserve_certifications_manager.set_thermal_certifications(study_interface, area_id, data)
 
+    @bp.get(
```
sylvlecl, line 1711: Just noticing in that PR that we don't cover those endpoints with integration tests.

Maybe the reason is that we were not provided yet with a 10.2 test study that already contains reserves, but we could already test them by creating somee reserves and some participations ourselves

sylvlecl (changes requested):
Looks mostly good already, thanks !

Some details in the comments, not necessarily all to be handled:
I think we should at least remove the inheritance in the ReserveCertification model though (not necessarily in the models use for parsing files).
I think we can already add some integration tests too, although we don't have a 10.2 study file.

antarest/study/business/model/reserve_certification_model.py
olivierPigeon-RTE: This class is used to call common functions on any type of reserve certification, creating a type with`ReserveCertification = ThermalReserveCertification | StorageReserveCertification`  would have been cleaner indeed

antarest/study/web/study_data_blueprint.py
olivierPigeon-RTE: what is the best way to create a v10.2 study using a client, knowing that for now, the most recent supported version is 9.3?

antarest/study/dao/database/database_reserve_definition_dao.py
```diff
@@ -132,11 +132,22 @@ def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[Rese
 
         self._db_session.commit()
 
-        # We have to delete the reserve symmetries associated with the reserve as we do not have a foreign key constraint
-        symmetries_dict = self.get_impl().get_thermal_reserve_symmetries(area_id)
-        new_symmetries = remove_reserve_symmetries_by_cascade(symmetries_dict, set(reserve_ids))
-        if new_symmetries is not None:
-            self.get_impl().save_thermal_reserve_symmetries({area_id: new_symmetries})
+        self._delete_symmetries_associated_with_deleted_reserve(area_id, reserve_ids)
```
smailio, line 133: shouldn't the commit just above be just bellow this line ?

antarest/study/dao/database/database_reserve_symmetries_dao.py
```diff
@@ -48,8 +57,36 @@ def _convert_model_to_row(
...
+    values = {
+        "study_id": study_id,
+        "area_id": area_id,
+        "st_storage_id": st_storage_id,
+        "symmetries": json.dumps(symmetries),
+    }
+    return values
+
+
+def _convert_all_rows_to_model(rows: Sequence[Row[tuple[Any]]], id_field_name: str) -> dict[Any, Any]:
```
smailio: I think you can't do much to make rows type more explicit but `dict[Any, Any]` could become  `dict[AreaId , ThermalReserveCertification|SSTReserveCertification]` no ? 

We could also define a ReserveCertification interface

antarest/study/dao/study_conversion/study_converter.py
```diff
@@ -187,7 +187,7 @@ def _convert_areas(self) -> None:
         self._new_dao.save_misc_gen(self._source_dao.get_all_misc_gen())
 
         # Reserves
-        if self._study_version >= STUDY_VERSION_10_0:
+        if self._study_version >= STUDY_VERSION_10_2:
             self._convert_reserves()
```
smailio: you need to update _convert_reserves to convert sst certifications and symmetries

antarest/study/storage/rawstudy/model/filesystem/config/reserve_participations.py
```diff
@@ -0,0 +1,289 @@
...
+
+    @classmethod
+    def _reorganize_certifications(
+        cls, certifications: Mapping[ReserveDefinitionId, Mapping[str, ReserveCertification]]
+    ) -> dict[str, dict[ReserveDefinitionId, ReserveCertification]]:
+        # Reorganize certifications to order them by id.
+
+        reorganized_certifications: dict[str, dict[ReserveDefinitionId, ReserveCertification]] = {}
+        for reserve_id, value in certifications.items():
+            for thermal_id, certification in value.items():
```
smailio: it is no longer true to say this is a thermal_id is it ?

antarest/study/storage/rawstudy/model/filesystem/config/reserve_participations.py
```diff
@@ -0,0 +1,289 @@
...
+
+    @classmethod
+    def _iterate_through_symmetries(
+        cls,
+        participations: list[Any],
+        symmetries: dict[str, list[list[ReserveDefinitionId]]],
+        certifications: dict[str, dict[ReserveDefinitionId, ReserveCertification]],
+    ) -> None:
+        for cluster_id, reserve_symmetries in symmetries.items():
+            certifs = certifications.pop(cluster_id, {})
```
smailio: this function is called _iterate_through and it mutates the given certification argument, this can be confusing

Also it's maybe worth it to add some doc to explain this function's purpose

antarest/study/storage/rawstudy/model/filesystem/config/reserve_participations.py
```diff
@@ -0,0 +1,289 @@
...
+                reorganized_certifications.setdefault(thermal_id, {})[reserve_id] = certification
+        return reorganized_certifications
+
+    @classmethod
+    def from_model(
+        cls,
+        symmetries: dict[str, ReserveSymmetries],
+        certifications: Mapping[ReserveDefinitionId, Mapping[str, ReserveCertification]],
+    ) -> Self:
+        thermal_certifications = cls._reorganize_certifications(certifications)
```
smailio: not thermal_certifications anymore is it ?

smailio (changes requested):
Nice work as this improves the reserve module design, it provides a solid foundation to build the hydro part on top of that 👍

antarest/study/dao/database/database_reserve_symmetries_dao.py
olivierPigeon-RTE: The best I can do is change the output type to `dict[Any, list[list[ReserveDefinitionId]]]`, but as the input rows have 'Any' type, the key of the output dict has to be Any as well

antarest/study/web/study_data_blueprint.py
sylvlecl: As we have discussed:
we'll add integration tests later, when we update the antares-study-versions library to support 10.2

antarest/study/dao/study_conversion/study_converter.py
smailio: Actually this may break now. 

This convert is done before the SST convert. Which is doe l222.

It means that during this convert, you may reference SST that don't exist yet. 

This isn't introduced in this PR and it wasn't an issue prior because this function only used to convert thermals, which are converted before this is called. 

The solution is to move the convert reserves bellow convert storage.

antarest/study/dao/database/database_reserve_certification_dao.py
```diff
@@ -80,54 +106,38 @@ def get_thermal_reserve_certifications(self, area_id: AreaId) -> ThermalReserveC
...
-            stmt = delete(_THERMAL_TABLE).where(
-                (_THERMAL_TABLE.c.study_data_id == self._study_data_id) & (_THERMAL_TABLE.c.area_id.in_(area_ids))
-            )
-            self._db_session.execute(stmt)
-            # Then, insert the new values
-            if values:
-                self._db_session.execute(insert(_THERMAL_TABLE), values)
+            self._clean_db(_THERMAL_TABLE, new_certifications)
+            self._insert_data_to_table(_THERMAL_TABLE, values)
         except IntegrityError as e:
```
smailio, line 131: rollback is missing

antarest/study/dao/database/database_reserve_certification_dao.py
```diff
@@ -142,3 +152,105 @@ def _raise_the_right_thermal_reserve_exception(
...
+    ) -> NoReturn:
+        self._raise_exception_if_missing_area(data)
+        self._raise_exception_if_missing_reserve(data)
+
+        all_existing_st_storage = self.get_impl().get_all_st_storages()
+        invalid_st_storage_dict = {}
+        for area_id, reserves_dict in data.items():
+            for st_storage_ids in reserves_dict.values():
+                if invalid_st_storage := set(st_storage_ids) - set(all_existing_st_storage.get(area_id, [])):
+                    invalid_st_storage_dict[area_id] = invalid_st_storage
```
smailio: When, in a same area,  several symmetries have invalid storage ids only the last one is saved. 

`invalid_st_storage_dict[area_id] = invalid_st_storage`this line do overwrite. 

Same issue for thermal.

smailio (changes requested):

smailio (dismissed):

smailio (approved):

Issue #3359, opened 2026-08-14
Is matrix_gc_dry_run: true in the reference config deliberate?


insatomcat:
`resources/deploy/config.prod.yaml` ships `matrix_gc_dry_run: true`, while `core/config.py` defaults it to `False` and neither `blob_gc_dry_run` nor `variable_view_gc_dry_run` is set. The only collector the reference deployment starts is therefore also the only one neutralised, and enabling the others as they stand would start them destructive while that one is not. The Celery route inherits this, since the tasks read the same settings.

Is that `true` deliberate? Concretely: is the matrix GC's set of usage providers considered complete enough to run it destructive in production, or is the line a leftover?

Asking because it decides what an operator should do: leave the collectors in dry run permanently, or turn them off once a few cycles look right. I run beat plus worker with all four `*_dry_run` at `true` and have no way to tell from the outside.

PR #3360, opened 2026-08-14
fix(services): wire the variable_view_gc module into the worker

insatomcat:
`Module.VARIABLE_VIEW_GC` is advertised as a valid `--module` value, since argparse takes its choices from the `Module` enum in `antarest/main.py`, but `singleton_services._init` handles five of the six modules it is supposed to start and never builds that one. A worker started with it gets an empty service list, then `start_all_services` loops forever on its `time.sleep(2)`. The container stays up, looks healthy from the outside, logs nothing and collects nothing.

`create_variable_view_gc` already exists in `service_creator.py` and only needs the configuration, so what was missing is the dispatch block, placed next to the `blob_gc` one it mirrors.

Why it matters beyond the module itself: rows in `output_variables_views` are only ever deleted by this collector, and each row pins its matrix through `OutputVariablesMatrixUsageProvider`. As long as the collector cannot run outside the API process, those matrices stay in use and never become eligible for the matrix garbage collector either.

### Test

`tests/test_singleton_services.py` did not exist, the module had no coverage at all. The new test dispatches every member of `Module` except `APP` and asserts that each one produces exactly the service its factory returns. Without the fix, only the `variable_view_gc` case fails, and a module added to the enum later cannot stay unwired silently.

### Context

Found while deploying Antares Web 2.33 from the reference compose file, in the same pass as #3355, #3356, #3357, #3358 and #3359. #3359 asks whether `blob_gc` and `variable_view_gc` should run in the reference deployment. Whatever is decided there, `variable_view_gc` cannot be deployed as a container until this is fixed.

### DIFF
--- a/antarest/singleton_services.py
+++ b/antarest/singleton_services.py
@@ -25,6 +25,7 @@ from antarest.service_creator import (
     create_blob_gc,
     create_core_services,
     create_matrix_gc,
+    create_variable_view_gc,
     create_watcher,
     init_db_engine,
 )
@@ -53,6 +54,10 @@ def _init(config_file: Path, services_list: list[Module]) -> list[IService]:
     if Module.BLOB_GC in services_list:
         blob_gc = create_blob_gc(config, core_services.blob_service)
         services.append(blob_gc)
+
+    if Module.VARIABLE_VIEW_GC in services_list:
+        variable_view_gc = create_variable_view_gc(config)
+        services.append(variable_view_gc)
 
     if Module.AUTO_ARCHIVER in services_list:
         auto_archive_service = AutoArchiveService(core_services.study_service, core_services.output_service, config)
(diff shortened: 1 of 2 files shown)


sylvlecl:
@insatomcat Note that now the favored deployment schema would be to have all those background tasks handled by some celery worker(s), for example using the subcommands `start.sh celery-beat` and `start.sh celery-worker` of the startup script

sylvlecl:
> @insatomcat Note that now the favored deployment schema would be to have all those background tasks handled by some celery worker(s), for example using the subcommands `start.sh celery-beat` and `start.sh celery-worker` of the startup script

I must say the `docker-compose.yml`, which actually is not used in production, is now quite outdated, as you have seen.
It should not be used as a reference for production deployment.
The doc in docs/DEPLOY.md is not up to date either, we should update it, in particular to mention the celery worker for background tasks.

sylvlecl (approved):

PR #3363, opened 2026-08-14
fix(celery): make the beat schedule path configurable

insatomcat:
`celery beat` writes its schedule to `celerybeat-schedule`, a path relative to the working directory. The image sets no `WORKDIR`, so that directory is `/`, which a container that does not run as root cannot write. Beat does not degrade into a scheduler without persistence, it exits during startup, and comes back under any restart policy without ever scheduling anything.
Note that running the container as an unprivileged user is the usual hardening for a production deployment (this is how I ran into this).
`CELERYBEAT_SCHEDULE` points that file at a writable path. It is only passed when the variable is set, so the command line is unchanged for a deployment that does not define it.

A second commit lists the variables the script reads in its header comment.

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -75,8 +75,16 @@ else
   case "$1" in
     celery-beat)
       echo "Starting Celery Beat scheduler..."
+      # Celery writes its schedule to `celerybeat-schedule`, relative to the
+      # working directory: set CELERYBEAT_SCHEDULE when that directory is not
+      # writable.
+      EXTRA_OPTS=()
+      if [ -n "${CELERYBEAT_SCHEDULE:-}" ]; then
+        EXTRA_OPTS+=(--schedule="$CELERYBEAT_SCHEDULE")
+      fi
       exec celery -A antarest.maintenance.app:celery_app beat \
         --loglevel=info \
+        "${EXTRA_OPTS[@]}" \
         --pidfile=/tmp/celerybeat.pid
       ;;
     celery-worker)
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -18,6 +18,19 @@
 #    configure some load balancing upstream, in that case.
 #    Number of workers is defined by ANTARES_NB_WORKERS env variable,
 #    or equal to 2*cpu + 1.
+#
+# Environment variables:
+#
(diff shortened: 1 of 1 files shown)


scripts/start.sh
```diff
@@ -75,8 +75,14 @@ else
   case "$1" in
     celery-beat)
       echo "Starting Celery Beat scheduler..."
+      # --schedule is not optional here. Celery defaults it to
```
sylvlecl: AI-based contributions are OK, but please at least review them before submitting:
here the comment is really too long and is only related to your setup (running this script inside a docker image, with a non root env, with "/" as the working dir).

In our prod setup the --schedule option is actually optional, since we use a different docker image, which makes that comment misleading.

Seems OK to use `/tmp/celerbeat-schedule` as a temp directory but I would prefer to keep the previous behaviour when the env var is not defined to not impact our prod env.

sylvlecl (changes requested):

scripts/start.sh
insatomcat: Hi @sylvlecl, my bad, I run my containers as a non-root user, which is how I hit the bug, and I just assumed that was everyone's production target.
The comment now states the condition instead of claiming the option is mandatory, and `--schedule` is only passed when `CELERYBEAT_SCHEDULE` is set, so an unset variable leaves the command line exactly as it is today. I shortened the description as well.

scripts/start.sh
```diff
@@ -75,8 +75,16 @@ else
   case "$1" in
     celery-beat)
       echo "Starting Celery Beat scheduler..."
+      # Celery writes its schedule to `celerybeat-schedule`, relative to the
+      # working directory: set CELERYBEAT_SCHEDULE when that directory is not
+      # writable.
+      SCHEDULE_OPT=()
```
sylvlecl: If you're introducing an array for optional additional options, let's make it more generic than just "SCHEDULE_OPT" so that we can re-use it later for similar needs.

The supported environment variables should be documented either in the comment at the top of the script or in a help message (which does not exist as of now).

insatomcat: good point, I renamed it to EXTRA_OPTS.
As requested, I documented the supported env variables at the top of the script (all of them, not just the new one), in a separate commit.

sylvlecl (approved):

PR #3366, opened 2026-08-20
fix(reserves): remove orphan symmetries

MartinBelthle:
Fix 2 issues:
- When removing a thermal cluster, orphan symmetries were not removed for FS storage mode (same issue for sts)
- When removing a certification, orphan symmetries were not removed for both storage modes (same issue for sts)


antarest/study/dao/database/database_reserve_certification_dao.py
```diff
@@ -116,113 +69,118 @@ def save_thermal_reserve_certifications(
...
+        reserve_type: ReserveObjectType,
+        old_certifications: dict[AreaId, dict[ReserveDefinitionId, dict[str, Any]]],
+        new_certifications: dict[AreaId, dict[ReserveDefinitionId, dict[str, Any]]],
+    ) -> None:
         values = []
-        for area_id, reserves_dict in data.items():
-            for reserve_id, storage_dict in reserves_dict.items():
-                for storage_id, certification in storage_dict.items():
+        for area_id, reserves_dict in new_certifications.items():
+            for reserve_id, thermal_dict in reserves_dict.items():
```
olivierPigeon-RTE: I think this method is not specific to thermal cluster, I guess you mean `area_asset_dict` or `asset_dict`. Same thing for `thermal_id` below

PR #3369, opened 2026-08-25
perf(output): avoid rezipping output for managed studies

MartinBelthle:
Solves the performance issue inside [ANT-5556] but not the design flaws

### DIFF
--- a/antarest/launcher/service.py
+++ b/antarest/launcher/service.py
@@ -550,18 +550,25 @@ class LauncherService:
             is_study_managed = is_managed(study)
 
         zip_path: Path | None = None
-        if not is_study_managed:
-            # Optimized path for studies stored on external devices, that will then be unarchived there.
-            # TODO: that whole optimization path should be refactored to:
-            #       - be more explicit
-            stopwatch = StopWatch()
-            logger.info("Re zipping output for transfer")
-            zip_path = output_true_path.parent / f"{output_true_path.name}.zip"
-            archive_dir(output_true_path, target_archive_path=zip_path, archive_format=ArchiveFormat.ZIP)
-            logger.info(f"Zipped output for job {job_id} in {stopwatch}s")
-            final_output_path = zip_path
-        else:
+
+        if is_zip(output_true_path):
+            # Possible if the option `-z` was used to run the solver.
             final_output_path = output_true_path
+
+        else:
+            if is_study_managed and job_launch_params.auto_unzip:
+                # Nothing to do, the output is already unarchived.
+                final_output_path = output_true_path
+            else:
+                # For studies stored on external devices, it's faster to re-zip the output for transfer and unarchive it there.
+                # Also, for managed studies when the user did not ask for auto-unzip, we'd better re-zip the output here instead of copying the tree and then zip it.
+                # TODO: that whole optimization should be refactored to be more explicit
+                stopwatch = StopWatch()
+                logger.info("Re zipping output for transfer")
+                zip_path = output_true_path.parent / f"{output_true_path.name}.zip"
+                archive_dir(output_true_path, target_archive_path=zip_path, archive_format=ArchiveFormat.ZIP)
+                logger.info(f"Zipped output for job {job_id} in {stopwatch}s")
+                final_output_path = zip_path
 
         with db():
             try:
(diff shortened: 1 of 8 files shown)


antarest/output/storage/file/abstract_storage.py
```diff
@@ -189,7 +189,7 @@ def _import_zip_as_archived(
     output_full_name = extract_output_name(output_zip_path, output_name_suffix)
     final_path = _archived_output_path(study_outputs_path, output_full_name)
     study_outputs_path.mkdir(exist_ok=True)
-    shutil.copyfile(output_zip_path, final_path)
+    shutil.move(output_zip_path, final_path)
```
olivierPigeon-RTE, line 192: The docstring says this function copies the zip, is it replaced by a "move" to save space? If yes, you should update the docstring

olivierPigeon-RTE (changes requested):

antarest/output/storage/file/abstract_storage.py
MartinBelthle: Indeed, I'll change the docstring. I changed the code as `move` can be inistantaneous if done on the same FS which should be the case for us

antarest/launcher/service.py
```diff
@@ -575,11 +580,7 @@ def _import_output(
                         logs=additional_logs,
                     )
             except StudyNotFoundError:
```
smailio, line 589: Now this exception is only reached in the rare case where the study gets deleted after line 546, right ?

antarest/launcher/service.py
```diff
@@ -532,23 +533,27 @@ def _import_output(
...
+
             self._save_solver_stats(job_result, output_true_path)
 
         zip_path: Path | None = None
-        # Optimized path for studies stored on external devices, that will then be unarchived there.
-        # TODO: that whole optimization path should be refactored to:
-        #       - be more explicit
-        #       - not affect internal studies
-        if job_launch_params.archive_output:
+        if not is_study_managed:
```
smailio: Why this isn't `not is_study_managed and  job_launch_params.archive_output` ? I get that archive_output is depreceated but should still be supported untill it's removed no ?

smailio: Also if we do this `if not is_study_managed or not job_launch_params.auto_unzip` then we won't need to run an archive on it later, I mean we could get rid of : 

```python
            if output.is_dir() and not auto_unzip:
                # This only happens for managed studies (not obvious due to the leaky workflow)
                self.archive_output(uuid, output_id)
```

antarest/launcher/service.py
smailio: This will error for zipped output of non managed study. 

The zip_path gets unlinked during the import by `_import_zip_as_archived`.

Call chain : 
`l:575 output_service.import_output => _import_zip_as_archived => shutil.move(output_zip_path, final_path)`

Found with AI tbh

antarest/launcher/service.py
```diff
@@ -532,23 +533,27 @@ def _import_output(
...
             if not output_true_path.is_dir() and not is_zip(output_true_path):
                 raise NoValidOutputError(f"No valid output for job {job_id}: {output_true_path}")
 
+            study_id = job_result.study_id
+            study = db.session.get(Study, study_id)
+            if study is None:
+                return self._import_fallback_output(job_id, output_true_path, job_launch_params.output_suffix)
+            is_study_managed = is_managed(study)
+
             self._save_solver_stats(job_result, output_true_path)
```
smailio: Now we don't save stats when `study is None` this line should be moved above.

smailio (changes requested):

antarest/launcher/service.py
MartinBelthle: The `archive_output` did not make any sense (always True, non-editable in the front). The only parameter that matters is `auto_unzip` where if it is False the output is left archived and otherwise it is automatically unarchived. So I deprecated it.

MartinBelthle: And for the 2nd comment, the `if` actually matters as now the output always arrives as a folder inside the study for managed studies (except for the -z option but we do not use it) so if the user did not ask for automatic unzip we have to re-zip it

PR #3370, opened 2026-08-25
feat(studies): make path nullable for database studies

olivierPigeon-RTE:


alembic/versions/1bb1c26c70d8_make_studies_path_nullable.py
```diff
@@ -0,0 +1,25 @@
...
+# revision identifiers, used by Alembic.
+revision = '1bb1c26c70d8'
+down_revision = '40f4391430e7'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("study", schema=None) as batch_op:
+        batch_op.alter_column("path", existing_type=sa.String(), nullable=True)
```
MartinBelthle, line 21: If we want to put the `path` to None for DB studies, we should also put it to None for existing studies

antarest/output/adapters.py
```diff
@@ -35,8 +36,13 @@ class StudyServiceAdapter(IFileOutputsProvider):
         @override
         def get_outputs(self, study_id: str) -> FileStudyOutputs:
             metadata = study_service.get_study(study_id)
+            if metadata.path is None:
+                # Point to a path that is guaranteed not to exist for database studies
+                outputs_path = study_service.config.storage.tmp_dir / "no-file-outputs-for-database-studies" / study_id
```
MartinBelthle: Should not happen and the solution is more than weird. This `if` should just be removed

antarest/study/storage/file_study_storage.py
```diff
@@ -58,8 +63,10 @@ def __init__(
     def copy(self, src_study: Study, metadata: StudyMetadataCopy) -> RawStudy:
         new_study = build_raw_study_from_source(src_study, self._config.get_workspace_path(), metadata)
 
+        new_study_path = check_study_path(new_study)
```
MartinBelthle: Really minor but no need to introduce a variable here, you can do it in one line as you did in the rest of the code

antarest/study/service.py
```diff
@@ -1092,12 +1092,23 @@ def sync_studies_on_disk(
         now = current_time()
         clean_up_missing_studies_threshold = now - timedelta(days=MAX_MISSING_STUDY_TIMEOUT)
         all_studies = self.repository.get_all_raw()
+
+        # Database-mode studies (workspace == DEFAULT_WORKSPACE_NAME) have no path on disk (path is None),
```
MartinBelthle: No need for this comment. The real reason is that as we're syncing studies on disk, we only want unmanaged studies. The DB studies are just a subset of the managed studies that do not make sense to fetch in the first place

antarest/study/service.py
```diff
@@ -1092,12 +1092,23 @@ def sync_studies_on_disk(
...
+        # so they must be excluded before any Path(raw_study.path) is computed below.
+        all_studies = [study for study in all_studies if not is_managed(study)]
         if directory:
             if recursive:
-                all_studies = [raw_study for raw_study in all_studies if directory in Path(raw_study.path).parents]
+                all_studies = [
+                    raw_study
+                    for raw_study in all_studies
+                    if raw_study.path and directory in Path(raw_study.path).parents
```
MartinBelthle: Would be clearer to use `check_study_path` here to raise a clear Exception instead of silently skipping them

antarest/study/service.py
```diff
@@ -1092,12 +1092,23 @@ def sync_studies_on_disk(
...
+                ]
             else:
-                all_studies = [raw_study for raw_study in all_studies if directory == Path(raw_study.path).parent]
-        all_studies = [study for study in all_studies if study.workspace != DEFAULT_WORKSPACE_NAME]
+                all_studies = [
+                    raw_study
+                    for raw_study in all_studies
+                    if raw_study.path and directory == Path(raw_study.path).parent
```
MartinBelthle: Same remark

tests/study/service/test_service.py
```diff
@@ -514,6 +515,43 @@ def test_partial_sync_studies_from_disk() -> None:
     )
 
 
+# noinspection PyArgumentList
+def test_partial_sync_studies_from_disk_ignores_database_mode_studies() -> None:
```
MartinBelthle: As said in a previous comment you can adapt your test to say that we only ignore managed studies

tests/study/service/test_service.py
```diff
@@ -598,6 +636,39 @@ def test_create_study(tmp_path: Path, raw_study_service: RawStudyService) -> Non
     factory.create_study_dao.assert_called_once()
 
 
+# noinspection PyArgumentList
+@with_db_context
+def test_new_study_path_should_be_none_if_in_db(
```
MartinBelthle: I would personnaly prefer a integration test for this, where we create studies with both storage modes and ensure the DB is None and the FS has a value.

Makes me realize to check the path in the integration test we'll probably have to do a `with db` to really check the DB value as the endpoint do not send the info i think

MartinBelthle (changes requested):

antarest/output/adapters.py
olivierPigeon-RTE: actually the case happens when we want to access archives and ouputs of a db study. In these use cases, a filesystem behaviour is used to organize files but the study still has no path. If I remove the `if`, some integration tests fail (for instance `test_db_study_properties_edition_and_deletion`)

To be honest I don't  like this line neither but I don't know how to do it cleaner

MartinBelthle: Okay I see now. So what we should do if possible instead is do this in the if:
`outputs_path = config.storage.output.out_of_study.storage_dir / study_id` as we do in the other `get_outputs` right below

antarest/output/service.py
```diff
@@ -401,6 +401,7 @@ def import_output(
         # for now we just check after the fact that this output_id was actually not conflicting,
         # should be improved in the future so that all storages have access to the already existing IDs.
         for other_storage in [s for s in self._storages if s is not storage]:
+            # TODO: Ignore OutputStorageType.IN_STUDY_FILE_TREE storage for db studies as they do not have a file tree
```
MartinBelthle: You can remove this TODO as it's more than just that we should also not check OUT_OF_STUDY for FS studies. The real thing is said in the previous comment `all storages should have access to the already existing IDs.`

tests/integration/studies_blueprint/test_create_study.py
```diff
@@ -106,6 +108,12 @@ def test_create_study_with_path(
         assert res.status_code == 200
         study = res.json()
         assert study["name"] == "test-study"
+        assert study["folder"] == f"project/subfolder/{study_id}"
+
+        if storage_mode == "database":
```
MartinBelthle: Could also check that if storage_mode is FS we see an actual path that is not None

alembic/versions/1bb1c26c70d8_make_studies_path_nullable.py
```diff
@@ -0,0 +1,36 @@
...
+    # Set the path to None for existing database-mode studies
+    study_table = sa.table("study", sa.column("path", sa.String()), sa.column("storage_mode", sa.String()))
+    op.execute(study_table.update().where(study_table.c.storage_mode == "DATABASE").values(path=None))
+
+def downgrade():
+
+    # Set the path to the study ID for existing database-mode studies
+    study_table = sa.table("study", sa.column("id", sa.String()), sa.column("path", sa.String()))
+    op.execute(
+        study_table.update().where(study_table.c.path.is_(None)).values(path=study_table.c.id)
```
smailio, line 32: this used to be workspace/study_id 

this is how it was set in study/service.py `study_path = self.config.get_workspace_path() / sid`

tests/study/service/test_service.py
MartinBelthle: Do you still need this test now that you've added the integration one ? IMO we don't

tests/study/service/test_service.py
```diff
@@ -514,6 +515,43 @@ def test_partial_sync_studies_from_disk() -> None:
     )
 
 
+# noinspection PyArgumentList
+def test_partial_sync_studies_from_disk_ignores_managed_studies() -> None:
```
MartinBelthle: This test' title does not really describe what it does. 
Either we add a managed study with storage mode FS and we ensure it's not modifed either or we rename the test

MartinBelthle (changes requested):
Last round of review

tests/integration/studies_blueprint/test_create_study.py
```diff
@@ -32,7 +32,6 @@ class TestCreateStudy:
             "8.5",
             "8.6",
             "8.7",
-            "8.7",
```
MartinBelthle, line 32: This was actually meant to be v8.8 i imagine. We'd better just put 8.8 instead of removing the duplicata

MartinBelthle (dismissed):

MartinBelthle (approved):

smailio (changes requested):

alembic/versions/1bb1c26c70d8_make_studies_path_nullable.py
olivierPigeon-RTE: @MartinBelthle do you agree? Getting the workspace path from the migration file will not be trivial as it not stored in the study table. We can save the pathes before migration to restore them on downgrade or make a migration that cannot be fully downgraded

PR #3373, opened 2026-08-28
feat(reserves): add support for hydro

smailio:


antarest/study/dao/api/common.py
```diff
@@ -101,33 +106,81 @@ def check_st_storage_symmetries_integrity(
...
+        # There is no asset id here: an area owns exactly one long-term storage.
+        for symmetry in symmetries:
+            for reserve_id in symmetry:
+                if reserve_id not in existing_certifications[area_id]:
+                    if not study_dao.reserve_definition_exists(area_id, reserve_id):
+                        raise ReserveDefinitionNotFound(area_id, reserve_id)
+                    raise ReserveCertificationNotFound(area_id, "hydro", None, {reserve_id})
+
+
+def remove_reserves_from_symmetries(
```
smailio, line 141: I renamed remove_reserve_symmetry_by_cascade to remove_reserves_from_symmetries_dict because I find the laiter more straightforward. The concept of cascade isn't intrinsic to this function, as it's written it can be called merely to delete reserves from symmetries. 

I also make a light refactoring, I created `remove_reserves_from_symmetries` which has the code that was in `remove_reserve_symmetry_by_cascade` inner loop. Now `remove_reserves_from_symmetries_dict` just unpack the dicts passed to it and calls `remove_reserves_from_symmetries`.

Also when a reserves is removed and leaves a symmetry empty, we used to leave an empty list now we don't keep the list at all. 

```python
        # before 
        symmetries_dict = {"th1": [["r1", "r2"], ["r3", "r2"]], "th2": [["r3", "r4", "r5"]]}
        result = remove_reserve_symmetry_by_cascade(symmetries_dict, {"r3"})
        assert result == {"th1": [["r1", "r2"], []], "th2": [["r4", "r5"]]}

        # after
        symmetries_dict = {"th1": [["r1", "r2"], ["r3", "r2"]], "th2": [["r3", "r4", "r5"]]}
        result = remove_reserves_from_symmetries_dict(symmetries_dict, {"r3"})
        assert result == {"th1": [["r1", "r2"]], "th2": [["r4", "r5"]]}

```

antarest/study/dao/study_conversion/study_converter.py
```diff
@@ -274,7 +274,12 @@ def _convert_certifications_reserves(self) -> None:
         if st_storage_certitifcations:
             self._new_dao.save_st_storage_reserve_certifications(st_storage_certitifcations)
 
-    def _convert_resvers_symmetries(self) -> None:
+        # Hydro (long-term storage) certifications
+        hydro_certifications = self._source_dao.get_all_hydro_reserve_certifications()
+        if hydro_certifications:
+            self._new_dao.save_hydro_reserve_certifications(hydro_certifications)
```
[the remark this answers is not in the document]
smailio: There are no test for other reserves neither

antarest/study/dao/database/database_reserve_symmetries_dao.py
```diff
@@ -115,3 +122,44 @@ def _save_reserve_symmetries(self, area_ids: set[str], table: Table, values: lis
...
+    @override
+    def get_hydro_reserve_symmetries(self, area_id: AreaId) -> ReserveSymmetries:
+        table = HYDRO_RESERVE_SYMMETRIES_TABLE
+        stmt = select(table).where((table.c.study_data_id == self._study_data_id) & (table.c.area_id == area_id))
+        row = self._db_session.execute(stmt).fetchone()
+        return convert_row_to_symmetries(row) if row is not None else []
+
+    @override
+    def save_hydro_reserve_symmetries(self, data: HydroReserveSymmetriesMapping) -> None:
+        values = []
```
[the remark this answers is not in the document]
smailio: Fair point, error messages are meant to help user fix his requests. Performint a silent no-op would be confusing.

antarest/study/dao/database/database_reserve_symmetries_dao.py
```diff
@@ -115,3 +122,47 @@ def _save_reserve_symmetries(self, area_ids: set[str], table: Table, values: lis
...
+                }
+            )
+
+        if values:
+            # Check foreign keys integrity for values to insert
+            check_hydro_symmetries_integrity(self.get_impl(), data)
+        else:
+            # We need this check to avoid performing a silent no-op
+            # User should know when they send an invalid area
+            validate_areas_exist(self._db_session, self._study_data_id, set(data))
```
smailio, line 160: In addition to what the comment says, this additional check make the DB dao behavior consistent with FS dao.

antarest/study/dao/file/common.py
```diff
@@ -88,3 +88,16 @@ def get_st_storage_reserve_participations_as_yaml_content(area_id: AreaId, file_
 
 def get_thermal_reserve_participations_as_yaml_content(area_id: AreaId, file_study: FileStudy) -> dict[str, Any]:
     return _get_participations(area_id, file_study, get_thermal_reserve_path)
+
+
+def get_hydro_reserve_path(area_id: str) -> list[str]:
+    return ["input", "hydro", "reserves", area_id, "reserve-participations"]
+
+
+def get_hydro_reserve_participations_as_yaml_content(area_id: AreaId, file_study: FileStudy) -> dict[str, Any]:
```
MartinBelthle, line 97: Seems you could use the `_get_participations`

antarest/study/dao/common.py
```diff
@@ -47,4 +47,6 @@
 ReserveNeedsMapping: TypeAlias = dict[AreaId, dict[ReserveDefinitionId, SeriesId]]
 ThermalReserveSymmetriesMapping: TypeAlias = dict[AreaId, dict[ThermalId, ReserveSymmetries]]
 STStorageReserveSymmetriesMapping: TypeAlias = dict[AreaId, dict[StStorageId, ReserveSymmetries]]
+# The long-term storage (hydro) is modelled per area, so its symmetries have no asset dimension.
```
MartinBelthle, line 50: IMO we don't need to document this everywhere, every time I see a comment on this I think it's gonna be interesting and then I'm disappointed as it's just the same comment with a slightly different wording.

 We should just pick one place. Same goes for the certifications

antarest/study/storage/rawstudy/model/filesystem/root/input/hydro/reserves/reserves.py
```diff
@@ -0,0 +1,26 @@
...
+from antarest.study.storage.rawstudy.model.filesystem.root.input.hydro.reserves.area import InputHydroReservesArea
+
+
+class InputHydroReserves(FolderNode):
+    # Each area has its own folder named after the area id.
+    @override
+    def build(self) -> TREE:
+        return {
+            area_id: InputHydroReservesArea(self.matrix_storage_context, self.config.next_file(area_id))
+            for area_id in self.config.area_names()
```
MartinBelthle, line 25: Not a big fan of using `area_names` method when using ids. Just self.config.areas does the job

antarest/study/dao/database/database_reserve_definition_dao.py
```diff
@@ -133,21 +136,28 @@ def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[Rese
         reserve_ids_to_delete = set(reserve_ids)
         self.delete_orphan_thermal_symmetries(area_id, reserve_ids_to_delete)
         self.delete_orphan_st_storage_symmetries(area_id, reserve_ids_to_delete)
+        self.delete_orphan_hydro_symmetries(area_id, reserve_ids_to_delete)
         self._db_session.commit()
 
+    def delete_orphan_hydro_symmetries(self, area_id: str, reserves: set[ReserveDefinitionId]) -> None:
+        # An area owns exactly one long-term storage, so there is no asset to iterate over.
```
MartinBelthle, line 143: I'm gonna crash out with these comments

antarest/study/storage/rawstudy/model/filesystem/config/reserve_participations.py
```diff
@@ -299,3 +300,72 @@ def serialize_st_storage_reserve_participations(
...
+    @classmethod
+    def from_model(cls, symmetries: ReserveSymmetries, certifications: HydroReserveCertificationMapping) -> Self:
+        participation: dict[str, Any] = {}
+        if certifications:
+            participation["certifications"] = [
+                {"reserve": reserve_id, **certification.model_dump()}
+                for reserve_id, certification in certifications.items()
+            ]
+            # Drop symmetries whose reserves lost their certification. Same filtering rule as
+            # `_build_participations_from_symmetries`, except that the key is left out entirely
```
MartinBelthle, line 349: Not sure to understand what you mean with the `except ...`. If you saw something wrong in the thermal / sts logic we should change it there too.

antarest/study/dao/database/database_reserve_symmetries_dao.py
```diff
@@ -115,3 +122,47 @@ def _save_reserve_symmetries(self, area_ids: set[str], table: Table, values: lis
...
+    def save_hydro_reserve_symmetries(self, data: HydroReserveSymmetriesMapping) -> None:
+        values = []
+        for area_id, symmetries in data.items():
+            if not (any(symmetry for symmetry in symmetries)):
+                continue
+            values.append(
+                {
+                    "study_data_id": self._study_data_id,
+                    "area_id": area_id,
+                    "symmetries": json.dumps([symmetry for symmetry in symmetries if symmetry]),
```
MartinBelthle, line 150: Perhaps worth to put in common as we use it here and inside `common.py`

antarest/study/dao/database/database_reserve_symmetries_dao.py
MartinBelthle: If I understand correctly, the only case where this check is useful is when a user sends a fake area but with empty symmetries.
IMO we don't really care about this case but if we decide it's needed we should also do this for thermal and sts

antarest/study/dao/database/common.py
```diff
@@ -113,7 +113,7 @@ def serialize_frequency_filters(encoded_value: set[FrequencyFilter]) -> str:
 """
 
 
-def _convert_row_to_symmetries(row: Row[Any]) -> ReserveSymmetries:
+def convert_row_to_symmetries(row: Row[Any]) -> ReserveSymmetries:
```
MartinBelthle, line 116: Did not see where you used it outside of this file. If there is no such place, i think we should keep it private

antarest/study/dao/database/database_reserve_certification_dao.py
```diff
@@ -34,6 +34,28 @@
 from antarest.study.dao.common import AreaId, ThermalId
 from antarest.study.dao.database.common import ReserveObjectType, validate_areas_exist
 from antarest.study.dao.database.dao_context import DatabaseDaoBase
+from antarest.study.dao.database.models.hydro_reserve_certification import HYDRO_RESERVE_CERTIFICATION_TABLE
+
+_HYDRO_TABLE = HYDRO_RESERVE_CERTIFICATION_TABLE
+
+
+def _convert_hydro_row_to_model(row: Any) -> StorageReserveCertification:
+    values = row._mapping
```
MartinBelthle, line 43: We should use `get_row_representation_as_dict` instead to avoid accessing private members everywhere in case there's an API change

antarest/study/dao/database/database_reserve_certification_dao.py
```diff
@@ -152,6 +174,78 @@ def _save_certifications(
...
+        try:
+            area_ids = set(new_certifications)
+            stmt = delete(_HYDRO_TABLE).where(
+                (_HYDRO_TABLE.c.study_data_id == self._study_data_id) & (_HYDRO_TABLE.c.area_id.in_(area_ids))
+            )
+            self._db_session.execute(stmt)
+            if values:
+                self._db_session.execute(insert(_HYDRO_TABLE), values)
+            else:
+                validate_areas_exist(self._db_session, self._study_data_id, set(new_certifications))
```
MartinBelthle, line 219: Same remark here

tests/variantstudy/model/command/test_replace_hydro_reserve_symmetries.py
```diff
@@ -0,0 +1,165 @@
...
+
+    cmd = ReplaceHydroReserveSymmetries(
+        area_id="fr",
+        symmetries=[["r1", "r3"]],
+        command_context=command_context,
+        study_version=STUDY_VERSION_10_2,
+    )
+    output = cmd.apply(dao_10_2)
+    assert not output.status
+    assert "r3" in output.message
```
MartinBelthle, line 165: Minor but could we ensure the message is a bit clearer than just this ?

tests/variantstudy/model/command/test_replace_hydro_reserve_certifications.py
```diff
@@ -0,0 +1,169 @@
...
+        certifications={"fake_reserve": StorageReserveCertification()},
+        command_context=command_context,
+        study_version=STUDY_VERSION_10_2,
+    )
+    output = cmd.apply(dao_10_2)
+    assert not output.status
+    assert "Reserve definitions not found: {'fr': {'fake_reserve'}}" in output.message
+
+
+def test_command_dto_round_trip(command_context: CommandContext) -> None:
```
MartinBelthle, line 148: Once you've done the UT in `test_command_factory` you'll be able to remove this test

MartinBelthle (changes requested):
Few remarks:

1) We're missing 4 UTs inside `test_command_factory` to ensure we're able to do the round-trip around a command. 2 for STS reserves and also 2 for hydro reserves

2) Seems we should remove the folder `input/hydro/reserves` in `file_study_area_dao` when removing an area

antarest/study/dao/database/database_reserve_certification_dao.py
```diff
@@ -152,6 +181,80 @@ def _save_certifications(
                 else:
                     self.get_impl().delete_orphan_st_storage_symmetries(area_id, missing_reserves)
 
+    @override
+    def get_all_hydro_reserve_certifications(self) -> dict[AreaId, HydroReserveCertificationMapping]:
```
olivierPigeon-RTE, line 184: Why don't you use the ReserveObjectType class instead like for thermal and st-storage? Is it because of the absence of a "_db_key()" return value for hydro?

antarest/study/dao/database/database_reserve_definition_dao.py
```diff
@@ -133,21 +136,27 @@ def delete_reserve_definitions(self, area_id: AreaId, reserve_ids: Sequence[Rese
         reserve_ids_to_delete = set(reserve_ids)
         self.delete_orphan_thermal_symmetries(area_id, reserve_ids_to_delete)
         self.delete_orphan_st_storage_symmetries(area_id, reserve_ids_to_delete)
+        self.delete_orphan_hydro_symmetries(area_id, reserve_ids_to_delete)
         self._db_session.commit()
 
+    def delete_orphan_hydro_symmetries(self, area_id: str, reserves: set[ReserveDefinitionId]) -> None:
+        symmetries = self.get_impl().get_hydro_reserve_symmetries(area_id)
+        if remove_reserves_from_symmetries(symmetries, reserves):
+            self.get_impl().save_hydro_reserve_symmetries({area_id: symmetries})
```
olivierPigeon-RTE: we should update symmetries with the result of `remove_reserves_from_symmetries(symmetries, reserves)` rather than setting the new value to `symmetries`, shouldn't we? It woudl do something like

```python
        if new_summetries:=remove_reserves_from_symmetries(symmetries, reserves):
            self.get_impl().save_hydro_reserve_symmetries({area_id: new_symmetries})
```

antarest/study/dao/file/common.py
olivierPigeon-RTE: I also think so

antarest/study/dao/file/file_study_area_dao.py
```diff
@@ -401,6 +402,11 @@ def _delete_area_files(self, area_id: str, study_data: Any) -> None:
             study_data.tree.delete(["input", "st-storage", "clusters", area_id])
             study_data.tree.delete(["input", "st-storage", "series", area_id])
 
+        if study_version >= STUDY_VERSION_10_2:
+            # The folder is only created when reserve participations are saved, so it may not exist.
+            with contextlib.suppress(ChildNotFoundError):
+                study_data.tree.delete(["input", "hydro", "reserves", area_id])
```
olivierPigeon-RTE, line 408: I didn't know this part of the code, souldn't we do the same thing for thermal clusters and st-storages?

antarest/study/dao/file/file_study_reserve_certification_dao.py
```diff
@@ -147,3 +152,35 @@ def get_all_st_storage_reserve_certifications(self) -> dict[AreaId, StorageReser
...
         return parse_st_storage_reserves_certifications(data)
+
+    @override
+    def get_all_hydro_reserve_certifications(self) -> dict[AreaId, HydroReserveCertificationMapping]:
+        result = {}
+        for area in self.get_file_study().config.areas:
+            certifications = self.get_hydro_reserve_certifications(area)
+            if certifications:
+                result[area] = certifications
+        return result
```
olivierPigeon-RTE, line 163: It may need some refactor to avoid lines duplication with thermal and st-storage, only this line`certifications = self.get_hydro_reserve_certifications(area)` changes from one asset type to another. Not sure this needs to be done in this PR though

antarest/study/storage/rawstudy/model/filesystem/root/input/hydro/reserves/area.py
```diff
@@ -0,0 +1,23 @@
...
+from antarest.study.storage.rawstudy.model.filesystem.folder_node import FolderNode
+from antarest.study.storage.rawstudy.model.filesystem.inode import TREE
+from antarest.study.storage.rawstudy.model.filesystem.yaml_file_node import YAMLFileNode
+
+
+class InputHydroReservesArea(FolderNode):
+    @override
+    def build(self) -> TREE:
+        # Reserve participations of the long-term storage owned by this area.
+        return {"reserve-participations": YAMLFileNode(self.config.next_file("reserve-participations.yml"))}
```
olivierPigeon-RTE, line 23: You should add a check on the version as for st storage and thermal

olivierPigeon-RTE (changes requested):
Nice PR, just a few comments before merging

antarest/study/dao/file/file_study_area_dao.py
smailio: no no I thought the same but when whe delete thermal and st-storage the whole folder gets deleted including the reserve stuff

antarest/study/dao/file/file_study_reserve_certification_dao.py
smailio: IMO, I'd rather have some duplication than abstract too early, since that can couple domains that may evolve independently. And I can imagine area assets and object assets being treated differently in the future.

antarest/study/dao/database/database_reserve_certification_dao.py
smailio: Yes :( I was hesitant to use a fake key, but I gave up on that. Then I thought about creating an abstract class for Area assets but I thought that would be too much abstraction overhead.

antarest/study/dao/file/common.py
smailio: It's because empty participations is a list for object level reserves (sst and thermal) while it's empty object for area level reserves (hydro). I'll parametrize _get_participations, let me know if you like it better this way after I push the commit.

antarest/study/storage/rawstudy/model/filesystem/root/input/hydro/reserves/area.py
smailio: Actually this is done one level higher, so this code is never reached when version <= 10.2. We can add it here too, that would be informative to reader but it would be redundant and potentially irrelevant in the case of a refactoring.

antarest/study/dao/database/common.py
```diff
@@ -57,18 +56,31 @@ def validate_areas_exist(session: Session, study_data_id: int, area_ids: set[str
         raise AreaNotFound(*invalid_areas)
 
 
-def validate_areas_without_rows_exist(
-    session: Session, study_data_id: int, area_ids: set[str], values: Sequence[Mapping[str, Any]]
+def delete_by_area_id(
+    session: Session, study_data_id: int, table: Table, area_ids: set[str], inserted_area_ids: set[str]
```
olivierPigeon-RTE, line 60: Not sure to understand why a delete function has an `inserted_area_ids`

smailio: It's because otherwise the validate_area exist will be triggered every time we create reserves in the area for the first time

antarest/study/storage/rawstudy/model/filesystem/root/input/hydro/reserves/area.py
olivierPigeon-RTE: let's keep like this then to make the code simpler

antarest/study/dao/database/common.py
smailio: I mean this code will trigger an unecessary validate_area_exists, because the delete would return 0 rows deleted, as they weren't inserted yet 

```
save_area(dao, "fr")
dao.save_reserve_definitions({"fr": [ReserveDefinition(name="r1", type=ReserveType.UP)]})
dao.save_hydro_reserve_certifications({"fr": {}}) 
```

olivierPigeon-RTE (approved):

PR #3374, opened 2026-08-31
feat(ui-reserves): add reserves symmetries

hdinia:


webapp/src/queries/reserves/keys.ts
```diff
@@ -63,4 +64,18 @@ export const reserveKeys = {
       "updateReservesCertifications",
     ];
   },
+  symmetries: (
+    studyId: Study["id"],
+    areaId: AreaWithId["id"],
+    productionType: SymmetryProductionType,
+  ) => {
+    return [...reserveKeys.all(), "symmetries", { studyId, areaId, productionType }];
```
skamril: 💡 "reserveSymmetries"

webapp/src/queries/reserves/queries.ts
```diff
@@ -67,4 +69,17 @@ export const reserveQueries = {
...
+    studyId: Study["id"],
+    areaId: AreaWithId["id"],
+    productionType: SymmetryProductionType,
+  ) => {
+    return queryOptions({
+      queryKey: reserveKeys.symmetries(studyId, areaId, productionType),
+      queryFn: () => getReservesSymmetries({ studyId, areaId, productionType }),
+      // Deleting a cluster or a reserve elsewhere cascades to its symmetries
+      // server-side without invalidating this cache.
+      ...EXTERNALLY_MUTATED,
```
skamril, line 82: I think it would be better to invalidate the cache manually when a cluster or reserve is deleted, rather than relying on `EXTERNALLY_MUTATED` to refetch on every mount. That way we avoid unnecessary refetches for cases where nothing actually changed.

webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/areas/$areaId/reserves/-components/SymmetriesTable/index.tsx
```diff
@@ -0,0 +1,373 @@
...
+  onUndo,
+  onRedo,
+  onSave,
+}: Props) {
+  const { t } = useTranslation();
+  const { isDarkMode } = useThemeColorScheme();
+  const [rowSelection, setRowSelection] = useState<MRT_RowSelectionState>({});
+  // Raw input string, clamping on each keystroke would fight typing;
+  // `symmetryCount` is the normalized value actually used.
+  const [symmetryCountInput, setSymmetryCountInput] = useState("1");
```
skamril, line 111: ❓ Shouldn't the user be able to see the actual clamped symmetry count, instead of just their raw input? Right now the displayed value stays as symmetryCountInput even after clamping, so the UI never reflects the value that's actually used (`symmetryCount`). It might be worth syncing the input back to the clamped value once it's computed, so the user isn't left seeing a number that doesn't match what's actually applied.

webapp/src/services/api/studies/areas/reserves/adapters.ts
```diff
@@ -0,0 +1,100 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * SPDX-License-Identifier: MPL-2.0
+ *
+ * This file is part of the Antares project.
+ */
+
+import type { ClusterGroup, ReservesSymmetries, SymmetryRow } from "./types";
+
```
skamril: ❌ Adapters should only be functions that transform backend data (DTO) into UI data. Some of the functions here are actually view-specific utils rather than adapters.

skamril (changes requested):

webapp/src/services/api/studies/areas/reserves/types.ts
```diff
@@ -33,6 +35,27 @@ export type ReserveGlobalParameters = z.infer<typeof reserveGlobalParametersSche
 export type CertificationProductionType = z.infer<typeof certificationProductionTypeSchema>;
 export type ReserveCertification = z.infer<typeof reserveCertificationSchema>;
 export type ReservesCertifications = z.infer<typeof reservesCertificationsSchema>;
+export type SymmetryProductionType = z.infer<typeof symmetryProductionTypeSchema>;
+export type ReservesSymmetries = z.infer<typeof reservesSymmetriesSchema>;
+
+// UI row for one symmetry, adapted from a `ReservesSymmetries` entry.
+// `uiId` and `index` have no API equivalent: they exist for row identity
+// (React keys, undo/redo) and display order in the table.
+export interface SymmetryRow {
```
skamril: ❌ This type is view-specific and doesn't belong in the api folder.

skamril (changes requested):

skamril (approved):

PR #3376, opened 2026-09-02
feat(ui-reserves): hide obsolete reserve options for studies >= 10.2

hdinia:


webapp/src/routes/_authenticated/studies/$studyId/explore/modeling/areas/$areaId/route.tsx
```diff
@@ -109,14 +109,24 @@ function AreaLayout() {
...
+                linkOptions: linkOptions({
+                  to: "/studies/$studyId/explore/modeling/areas/$areaId/reserves",
+                  params,
+                }),
+              }
+            : {
+                id: "reservesLegacy",
+                label: t("study.modeling.reserves"),
+                linkOptions: linkOptions({
+                  to: "/studies/$studyId/explore/modeling/areas/$areaId/reserves-legacy",
```
skamril, line 126: ❓ Why not keep a single route to the right version internally, instead of exposing two separate routes?

```ts
// in reserves/route.tsx
if (semver.lt(study.version, "10.2.0")) {
  return <ReservesLegacy />;
}
return <Reserves />;
```

This also means that if we have another major change in the future, we won't need to create yet another route,  we can just extend the version check here.

skamril (approved):

PR #3383, opened 2026-09-14
refactor(config): type export_mps as ExportMPS enum instead of str

smailio:


antarest/study/dao/database/database_study_settings_dao.py
```diff
@@ -112,16 +108,6 @@ def get_optimization_preferences(self) -> OptimizationPreferences:
...
-        # Handle `export_mps` differently as it is stored as String in DB, but it can either be a string or a boolean.
-        mps: bool | str
-        if row.export_mps.lower() == "true":
-            mps = True
-        elif row.export_mps.lower() == "false":
-            mps = False
-        else:
-            mps = row.export_mps
-        data["export_mps"] = mps
-
```
smailio, line 124: Now pydantic parses this correctly

antarest/study/dao/database/database_study_settings_dao.py
```diff
@@ -89,12 +89,8 @@ def get_general_config(self) -> GeneralConfig:
     @override
     def save_optimization_preferences(self, config: OptimizationPreferences) -> None:
         values = dict(study_data_id=self._study_data_id, **config.model_dump(exclude={"export_mps"}))
-        # Handle `export_mps` differently as it can either be a string or a boolean but will be stored as String in DB.
-        if isinstance(config.export_mps, bool):
-            mps = str(config.export_mps)
-        else:
-            mps = config.export_mps
-        values["export_mps"] = mps
+        # Handle `export_mps` differently as it can either be an enum or a boolean but will be stored as String in DB.
```
MartinBelthle, line 92: Seems this message is not needed anymore as you removed the if / else

MartinBelthle (commented):
Minor remark. Also, we should perhaps ensure that we can no longer give a fake string inside a test

MartinBelthle (approved):

PR #3384, opened 2026-09-15
feat(gems): handle library inside the back-end

MartinBelthle:


alembic/versions/0974bca4078d_create_gems_library_tables.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+
+    op.create_table(
+        "gems_port_types",
+        study_data_id_col(),
+        Column("id", String(255), primary_key=True),
+        Column("description", String(), nullable=True),
+        Column("fields", String(), nullable=False),
+        Column("area_connection", String(), nullable=True),
+        Column("thermal_capacity_connection", String(), nullable=True),
```
olivierPigeon-RTE: What do "area_connection" and "thermal_capacity_connection" refer to ? I can't find them in the whole documentation about the library file

antarest/study/dao/database/gems/database_library_dao.py
```diff
@@ -0,0 +1,224 @@
...
+# See AUTHORS.txt
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+from typing_extensions import override
```
olivierPigeon-RTE, line 16: Minor but there are 2 copyrights headers in this file

alembic/versions/0974bca4078d_create_gems_library_tables.py
```diff
@@ -0,0 +1,87 @@
...
+down_revision = 'c7e21b9f4a83'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "gems_library_metadata",
+        study_data_id_col(),
+        Column("id", String(255), nullable=False),
```
smailio, line 24: This isn't a primary key ?

antarest/study/business/model/gems/library.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class _GemsPortTypeField(AntaresBaseModel):
+    model_config = ConfigDict(populate_by_name=True, extra="forbid")
+
+    id: str
+
+
+class _GemsAreaConnection(AntaresBaseModel):
+    model_config = ConfigDict(populate_by_name=True, extra="forbid", alias_generator=to_kebab_case)
```
smailio, line 27: why all these classes are private ?

alembic/versions/0974bca4078d_create_gems_library_tables.py
smailio: Also we may add a unique index on study_data_id_col as later in the DAO you assume there's at most one library per study.

antarest/study/business/model/gems/library.py
```diff
@@ -0,0 +1,98 @@
...
+    description: str | None = None
+    fields: list[_GemsPortTypeField] = Field(default_factory=list)
+    area_connection: _GemsAreaConnection | None = None
+    thermal_capacity_connection: _GemsThermalCapacityConnection | None = None
+
+
+class _GemsModelsProperties(AntaresBaseModel):
+    model_config = ConfigDict(populate_by_name=True, extra="forbid")
+
+    id: str
```
smailio, line 53: that's it 🤨 ?

antarest/study/dao/database/gems/database_library_dao.py
```diff
@@ -0,0 +1,224 @@
...
+)
+
+METADATA_TABLE = GEMS_LIBRARY_METADATA_TABLE
+
+
+class DatabaseGemsLibraryDao(GemsLibraryDao, DatabaseDaoBase):
+    """Database implementation of GemsLibraryDao"""
+
+    @override
+    def get_library(self) -> GemsLibrary | None:
```
smailio, line 37: maybe we can move the parsing logic in the model layer, that function contains a lot of logic.

antarest/study/dao/file/file_study_gems_library_dao.py
```diff
@@ -0,0 +1,56 @@
...
+    return study_path / "input" / "model-libraries"
+
+
+class FileStudyGemsLibraryDao(GemsLibraryDao, ABC):
+    @abstractmethod
+    def get_file_study(self) -> FileStudy:
+        pass
+
+    @override
+    def get_library(self) -> GemsLibrary | None:
```
smailio, line 34: I thought that FS dao wouldn't support GEMS ?

antarest/study/storage/rawstudy/model/filesystem/yaml_file_node.py
```diff
@@ -63,8 +63,8 @@ class YAMLWriter(IniWriter):
 
     @override
     def write(self, data: JSON, path: Path) -> None:
-        with open(path, "wb") as fh:
-            fh.write(yaml.safe_dump(data).encode("utf-8"))
+        with open(path, "w", encoding="utf-8") as fh:
+            yaml.safe_dump(data, fh)
```
smailio, line 67: nice but why ?

smailio: Ah I guess so the JSON string isn't encoded all at once and so it's more efficient memory wise ?

antarest/study/business/model/gems/library.py
MartinBelthle: Yep, see: https://gems-energy.readthedocs.io/en/stable/user-guide/input-files/library/#properties

antarest/study/business/model/gems/library.py
MartinBelthle: Just because I thought we shouldn't really expose them directly, more a implem detail of the library IMO. But I don't have a strong opinion on the matter

alembic/versions/0974bca4078d_create_gems_library_tables.py
MartinBelthle: For now we don't want to make "id" a primary key as we want to have only 1 library per study. We will change this if one day we want to support multi libraries for a study and we'll add the index at the same time

antarest/study/dao/file/file_study_gems_library_dao.py
MartinBelthle: We won't support unitary operation but we should be able to import and export studies with GEMS inside it so we must have the GET / SET operations for the whole content

antarest/study/storage/rawstudy/model/filesystem/yaml_file_node.py
MartinBelthle: Yes I think it's more efficient. Not really important to be fair

antarest/study/dao/database/gems/database_library_dao.py
MartinBelthle: Not really sure to understand. What do you want to move and where ?

alembic/versions/0974bca4078d_create_gems_library_tables.py
smailio: Without the unique index on study_data_id this table allows to have several libraries per study

MartinBelthle: As discussed via Teams, not an issue as the study_data_id key is the PK

PR #3386, opened 2026-09-16
feat(user-resources): create user resources view for database study

skamril:
ANT-5909


webapp/src/services/api/studies/userResources/index.ts
hdinia, line 1: suggestion: you can add doc strings like in other modules

.gitignore
```diff
@@ -155,4 +155,5 @@ watcher
 *.lock
 !uv.lock
 
-**/bucket
\ No newline at end of file
+**/bucket
+blobstore/
```
hdinia: Is the missing trailing slash intended ? I think it should be anchored to the root:

`/blobstore/`

webapp/src/components/router/RouterIconButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+import { IconButton } from "@mui/material";
+import { createLink } from "@tanstack/react-router";
+
+// Tanstack Router documentation:
+// https://tanstack.com/router/latest/docs/framework/react/guide/custom-link#button
+// https://tanstack.com/router/latest/docs/framework/react/how-to/integrate-material-ui#step-2-create-typed-mui-button-component
+
+const RouterIconButton = createLink(IconButton);
```
hdinia: Seems not used anywhere

hdinia (dismissed):

.gitignore
skamril: @copilot `/blobstore/` or `/blobstore`?

hdinia (approved):

PR #3391, opened 2026-09-16
fix(scenario-builder): handle binding constraints for studies before v8.6

olivierPigeon-RTE:


MartinBelthle (changes requested):
We should add an integration test (or modify an existing one) to test the full case the user saw.
You can simply create a v8.6 study, add a constraint and try to fetch the scenario builder with any scenario type and it should fail prior to your fix

tests/integration/studies_blueprint/test_scenario_builder.py
```diff
@@ -0,0 +1,26 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+from starlette.testclient import TestClient
+
+
+def test_get_scenario_builder_for_study_with_binding_constraint_without_group(admin_client: TestClient):
```
MartinBelthle: Move this test inside the already existing `test_scenario_builder.py` file and we'll be good to go

MartinBelthle (changes requested):

MartinBelthle (approved):

PR #3393, opened 2026-09-17
feat(ramp): add ramp boolean in optimization_preferences

smailio:


alembic/versions/e12d85a77641_add_thermal_cluster_ramping_to_.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision = "e12d85a77641"
+down_revision = "0974bca4078d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    default_boolean = sa.text("0") if op.get_context().dialect.name == "sqlite" else "f"
+    with op.batch_alter_table("optimization_preferences") as batch_op:
```
MartinBelthle, line 21: This is not what we should do at all. The parameter should be nullable as it only appears in v10.2

antarest/study/business/model/config/optimization_config_model.py
```diff
@@ -65,6 +65,7 @@ class OptimizationPreferences(AntaresBaseModel):
     transmission_capacities: bool | LegacyTransmissionCapacities | TransmissionCapacities = True
     thermal_clusters_min_stable_power: bool = True
     thermal_clusters_min_ud_time: bool = True
+    include_thermal_cluster_ramping: bool = False
```
MartinBelthle: Same remark here, should be optional as it only appears in v10.2

MartinBelthle (changes requested):
Didn't read everything as you'll have to change lots of things

TheoPascoli (dismissed):

